PR #3, opened 2015-10-07
Added QOS Meter adjust attribute.Modified the descriptions.

AshokDaparthi:
Attribute of maximum childs as mandatory for scheudler group creation.

### DIFF
--- a/inc/saischedulergroup.h
+++ b/inc/saischedulergroup.h
@@ -35,9 +35,6 @@ typedef enum _sai_scheduler_group_attr_t
 {
     /** READ-ONLY */
 
-    /** Maximum Number of childs on group [uint32_t] */
-    SAI_SCHEDULER_GROUP_ATTR_MAX_SUPPORTED_CHILDS = 0x00000000,
-
     /** Number of queues/groups childs added to
      * scheduler group [uint32_t] */
     SAI_SCHEDULER_GROUP_ATTR_CHILD_COUNT = 0x00000001,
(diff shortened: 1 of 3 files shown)


inc/saiswitch.h
```diff
@@ -412,6 +412,12 @@ typedef enum _sai_switch_attr_t
        Default no map */
     SAI_SWITCH_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP,
 
+    /** Number of preamble and Inter frame gap bytes to be added in the 
+     *  Meters(Policer, shaper and meter counters) [uint32_t]. 
+     *  Default to 20 Bytes. 
+     */
+    SAI_SWITCH_ATTR_QOS_METER_ADJUST,
```
lguohan: can you explain a little bit more about this attribute? why 20 bytes, why this can be different on different platforms?

inc/saischedulergroup.h
```diff
@@ -55,6 +52,10 @@ typedef enum _sai_scheduler_group_attr_t
        MANDATORY_ON_CREATE,  CREATE_ONLY */
     SAI_SCHEDULER_GROUP_ATTR_LEVEL = 0x00000004,
 
+    /** Maximum Number of childs on group [uint8_t]
+     * MANDATORY_ON_CREATE,  CREATE_ONLY */
+    SAI_SCHEDULER_GROUP_ATTR_MAX_CHILDS = 0x00000000,
```
lguohan: the enum value should be adjusted accordingly.

inc/saiswitch.h
AshokDaparthi: Based on MEF explanation, CIR includes the Preamble and and Inter frame gap in calculations, But some NPU's may/may not include this bytes in meter calculations by default. It will be controlled by external registers. Apps needs to have control to add/remove bytes from meter calculations.  
SAI model we can define instead of 20 bytes,  default behavior NPU should include these in  meter calculation. 
Preamble: 7 bytes Preamble
Start Frame Delimiter (SFD): 1 byte
Inter-frame Gap (IFG): 12 Byte(But it can be variable)

inc/saischedulergroup.h
AshokDaparthi: Intention is app to set the maximum child links as part of the SG creation. NPU has global node limits might be verified through error once limits exceeds.

lguohan:
merged, thanks.

PR #6, opened 2015-10-08
Added a new switch read-only attribute to synchronize counter statistics

Rajasekhar-Pulluru:
with the HW when SW based counters collection is enabled.

 Changes to be committed:
    modified:   inc/saiswitch.h


lguohan:
no longer needed

PR #8, opened 2015-10-09
Add ON_LINK next hop for directly reachable route

lguohan:
Directly reachable route will point to ON_LINK next hop.
Packet Destination IP is looked up in the neighbor table.
Attribute SAI_NEXT_HOP_NEIGHBOR_MISS_PACKET_ACTION for
ON_LINK next hop determines the packet action for neighbor
table lookup miss.

### DIFF
--- a/inc/sairouterintf.h
+++ b/inc/sairouterintf.h
@@ -89,6 +89,10 @@ typedef enum _sai_router_interface_attr_t
     /** MTU [uint32_t] (CREATE_AND_SET) (default to 1514 bytes) */
     SAI_ROUTER_INTERFACE_ATTR_MTU,
 
+    /** Packet action when neighbor table lookup miss for this router interface [sai_packet_action_t]
+     * (CREATE_AND_SET) (default to SAI_PACKET_ACTION_TRAP) */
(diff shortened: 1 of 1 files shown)


SathishKumarTR-Dell:
SAI_ROUTE_ATTR_NEXT_HOP_ID can be set with RIF Id for the directly reachable route, as mentioned in the comments. So, do we need the new ON_LINK next-hop type which requires the software to create a dummy next-hop id - as this does not get programmed in hardware. Can the NEIGHBOR_MISS_PACKET_ACTION be added as an attribute like ON_LINK_ROUTE_NEIGHBOR_MISS_PACKET_ACTION in the route entry itself?

lguohan:
NEIGHBOR_MISS_PACKET_ACTION is an attribute that controls the behavior in the neighbor lookup phase, it does not happen in the route lookup phase. So, better not to put that attribute in route entry itself. 

However, the another option is to still keep current header but add the neighbor_miss_packet_action in the route interface itself. 

So, the route entry still points itself next_hop_id to a router interface in the direct reachable route case, then we have an attribute in the router interface to control the missing neighbor action. Does this sound ok?

SathishKumarTR-Dell:
Agreed. This looks fine.

PR #9, opened 2015-10-12
SAI L2 Google UT

dgsudharsan:
Google unit test changes for L2 module


johnarnold:
Are these tests up to date with the latest SAI?

PR #13, opened 2015-10-20
Added ACL action attribute for egress block port mask.

SathishKumarTR-Dell:
Packets matching the ACL entry and egressing out the ports in the given port list will be dropped.


lguohan:
merged, thanks

PR #14, opened 2015-10-23
Added few more counter types and API for clear counters

dgsudharsan:
There are no counters defined in SAI to get total number of packets in a VLAN. Added both ingress and egress counter ids for total number of packets.
Added clear counters similar to the one present in port statistics


lguohan:
merged, thanks.

PR #17, opened 2015-10-28
Fix comment for SAI_PORT_ATTR_QOS_WRED_PROFILE_ID

itaibaz:
Fix comment - SAI_PORT_ATTR_QOS_DROP_TYPE is deprecated, instead attach
null object id to disable wred


lguohan:
merged, thanks.

PR #19, opened 2015-11-05
Modify Priority_group_to_pfc_priority -> pfc_priority_to_priority_group

itaibaz:
Support mapping of more than one pfc prio to a single PG


AshokDaparthi:
i am fine with change.

PR #21, opened 2015-11-12
ACL attributes to match UDF

SathishKumarTR-Dell:
Added attributes in SAI ACL objects to match User Defined Field data.
Defined type for passing the ACL UDF attribute values.


lguohan:
merged, thanks

PR #28, opened 2015-12-16
Recreated!! Extended SAI callbacks mechanism & extended host intf attribute type with ability to retrieve VLAN pkt belongs to.

rivasiv:
1) Extended SAI callbacks to allow the callback creator to provide callback with some data callback creator is needed.
2) Extended host intf attribute type with ability to retrieve VLAN pkt belongs to.

### DIFF
--- a/inc/saihostintf.h
+++ b/inc/saihostintf.h
@@ -627,6 +627,8 @@ typedef sai_status_t(*sai_send_hostif_packet_fn)(
  *    @param[in] buffer_size - actual packet size in bytes
  *    @param[in] attr_count - number of attributes
  *    @param[in] attr_list - array of attributes
+ *    @param[in] cb_data - Pointer to application data passed to
+ *          callback
  *
  * Return Values:
  *		None
@@ -635,7 +637,8 @@ typedef void(*sai_packet_event_notification_fn)(
     _In_ const void *buffer,
     _In_ sai_size_t buffer_size,
     _In_ uint32_t attr_count,
-    _In_ const sai_attribute_t *attr_list
+    _In_ const sai_attribute_t *attr_list,
+    void *cb_data
     );
 
 /**
--- a/inc/saihostintf.h
+++ b/inc/saihostintf.h
@@ -556,6 +556,9 @@ typedef enum _sai_hostif_packet_attr
     /** Ingress LAG [sai_object_id_t] (for receive-only) */
     SAI_HOSTIF_PACKET_INGRESS_LAG,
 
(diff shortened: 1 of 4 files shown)


inc/saiport.h
```diff
@@ -686,13 +689,16 @@ typedef void (*sai_port_state_change_notification_fn)(
...
+ *          callback
+ * 
  * Return Values:
  *    None
  */
 typedef void (*sai_port_event_notification_fn)(
     _In_ uint32_t count,
-    _In_ sai_port_event_notification_t *data
+    _In_ sai_port_event_notification_t *data,
+    void *cb_data
```
lguohan: Based on the comments, it is the application data passed to the callback. can you give an use case for this cb_data?

rivasiv: Sure, 

  for example for OvS it would be pointer to ofproto/netdev structure instance, in FBOSS it would be pointer to HwSwitch class instance, that's something callback should refer to once it called, some "context" from "original" cb environment. 

  Do we need some peace of code as example of usage?

-Rostyslav

inc/saihostintf.h
```diff
@@ -556,6 +556,9 @@ typedef enum _sai_hostif_packet_attr
     /** Ingress LAG [sai_object_id_t] (for receive-only) */
     SAI_HOSTIF_PACKET_INGRESS_LAG,
 
+    /** Ingress VLAN [sai_vlan_id_t] (for receive-only) */
+    SAI_HOSTIF_PACKET_INGRESS_VLAN,
```
lguohan: can you clarify the ingress vlan? What is this ingress vlan? Is it internal vlan tag added by the ASIC? What if the packet is untagged packet, or already tagged packet?

rivasiv: That's VLAN pkt is forwarded over ASIC. It may be either VLAN pkt is received with or in case pkt is not tagged - PVID.

lguohan: If it is the VLAN tag in the received packet, then the tag should be in the packet raw data. The application should be able to parse the packet to get the vlan tag. 

If it is PVID for untagged packet, then because the APP knows which incoming port the packet is coming from so it should know the PVID of that port.

It looks like the information can either be derived from the packet data or ASIC configuration. Is this true?

rivasiv: Yes you are right, we added this just like easier way to retrieve it.

kcudnik:
Closing, over 1 year old, moved to sai 1.0

PR #31, opened 2015-12-17
 Added ability to push egress physical port ID to neighbour and nexthop functionalites.

rivasiv:
```
  In case of VLAN routing there are no possibility to resolve physical egress port pkt should pass over.
  It must be resolved by control plane and pushed to SDK during neighbour or nexthop configuration.
```

### DIFF
--- a/inc/saineighbor.h
+++ b/inc/saineighbor.h
@@ -74,6 +74,7 @@ typedef enum _sai_neighbor_attr_t
 typedef struct _sai_neighbor_entry_t
 {
     sai_object_id_t rif_id;
+    sai_object_id_t egress_port_id; /*Must be set to SAI_NULL_OBJECT_ID in case rif_id points to router port(not a VLAN).*/
     sai_ip_address_t ip_address;
 
 } sai_neighbor_entry_t;
--- a/inc/sainexthop.h
+++ b/inc/sainexthop.h
@@ -70,6 +70,13 @@ typedef enum _sai_next_hop_attr_t
     /** Next hop entry router interface id [sai_object_id_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
     SAI_NEXT_HOP_ATTR_ROUTER_INTERFACE_ID,
 
+    /** Next hop entry egress port id [sai_object_id_t]
+     *  (MANDATORY_ON_CREATE|CREATE_ONLY).
+     *  Must be set to SAI_NULL_OBJECT_ID in case
+     *  SAI_NEXT_HOP_ATTR_ROUTER_INTERFACE_ID points to router
+     *  port(not a VLAN). */
+    SAI_NEXT_HOP_ATTR_EGRESS_PORT_ID,
+
     /* -- */
 
     /** Custom range base value */


lguohan:
the neighbor table only contains the MAC, it's up to the software or hardware to look the FDB table to find out the physical egress port. 

In the logical SAI pipeline, after L3 lookup, there is additional L2 lookup to resolve the port from the mac. If hardware does not support it, then SAI needs to lookup the L2 table. 

Thus, port in the neighbor table is not needed.

rivasiv:
Generally speaking you are right, but described scenario with L2 look up doesn't cover all cases.   There are:
- list of router applications which do not support FDB learning(FBOSS for example, and, perhaps, Quagga) or support software FDB learning which is running outside(above) SAI layer.
 - In ARP pkt the SRC/DST MAC of L2 Ethernet header not necessary are equal to those inside ARP header(as in ARP proxy).
 - there is possibility to set static neighbour entry from "some configuration software".

It those cases there are no FDB entry to be looked for port.  So there should be possibility to create neighbour without additional L2 lookup.

P.S.  I think we may think on creating some unified "Detailed SAI documentation" which suppose to explain all such corner cases.

lguohan:
In case there is no FDB entry, then L2 forward defines the the packet forwarding behavior of unknown unicast mac address, it can be either flood to the vlan or drop. 

You are right, we should have a documentation to describe such corner cases. It should describe the SAI packet logical pipeline. I am thinking to put the SAI spec into the wiki page and more contents into it, and maybe that's easier for people to find it.

garyachy:
@lguohan I ask you to reconsider this change request.
From my point of view we usually have an egress port when we are installing neighbor entry.
For example, in SONiC we translate ARP entries received from Linux into neighbors.

So why should we lose this information and then make pipeline or SAI look into FDB and make complex decisions to broadcast or unicast the packets?

Please advise what is bad into introducing the discussed parameter?

lguohan:
hi @garyachy , some asic have separate neighbor and fdb tables while some asics have folded view where neighbor table also contains output port information. The solution we found to unified these two approaches is to provide separated view as SAI interface and let sai library to translate them into the asic tables. Meanwhile, it's difficult to separate those information into multiple table when SAI provides the folded interface.

The separated view also matches the l3/l2 protocol layering principle as arp is Layer 2.5 protocol and fdb at Layer 2. In fact in linux, the interface in the ARP table should be mapped router interface, not the output port. For example, if you create a bridge in linux networking stack and assign a subnet to that bridge interface, then the arp entries will all show the bridge name as the interface and we do not really see the output port in the arp table.

lgh@acs-trusty6:/data/saibcm$ cat /proc/net/arp 
IP address       HW type     Flags       HW address            Mask     Device
192.168.5.3      0x1         0x2         02:42:c0:a8:05:03     *        bridge0
10.145.3.209     0x1         0x2         12:34:56:78:9a:bc     *        eth0
10.145.3.212     0x1         0x2         12:34:56:78:9a:bc     *        eth0
192.168.5.2      0x1         0x2         02:42:c0:a8:05:02     *        bridge0
lgh@acs-trusty6:/data/$ brctl showmacs bridge0
port no mac addr               is local?        ageing timer
  2     6a:7f:63:d2:b9:0b      yes                 0.00
  1     6e:44:9f:59:58:47      yes                 0.00
lgh@acs-trusty6:/data/$

kcudnik:
Closing, moved to sai 1.0

PR #32, opened 2015-12-18
    Extended Learning mode with per VLAN learn mode configuration.

rivasiv:
```
Proposed behaviour is like the following:

|--------------------------------------------------|
| Port learning  | VLAN learning | Result Learning |
|--------------------------------------------------|
| Enable         | Disable       | Port Enable     |
|--------------------------------------------------|
| Enable         | Enable        | Port Enable     |
|--------------------------------------------------|
| Disable        | Disable       | Disable         |
|--------------------------------------------------|
| Disable        | Enable        | VLAN Enable     |
|--------------------------------------------------|
```

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -121,28 +121,6 @@ typedef enum _sai_port_internal_loopback_mode_t
 } sai_port_internal_loopback_mode_t;
 
 /**
- * @brief Attribute data for SAI_PORT_ATTR_FDB_LEARNING
- */
-typedef enum _sai_port_fdb_learning_mode_t
-{
-    /** Drop packets with unknown source MAC. Do not learn. Do not forward */
-    SAI_PORT_LEARN_MODE_DROP,
-
-    /** Do not learn unknown source MAC. Forward based on destination MAC */
-    SAI_PORT_LEARN_MODE_DISABLE,
-
-    /** Hardware learning. Learn source MAC. Forward based on destination MAC */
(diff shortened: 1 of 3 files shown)


lguohan:
not sure I understand the column "Result learning". For the second row, what does the "Port enable" here. Say port 1 is enabled and vlan100 is disabled. Suppose port 1 is also in vlan100, for vlan100 packet comes in port 1, does the ASIC do mac learning or not? Which mac learning setting should we apply to this packet, the setting belongs to the vlan or the port?

rivasiv:
Here we are talking about configuration priorities. Usually more specific configuration takes precedence over more general one. In added design port configuration will have higher priority than configuration over VLAN, let me rephrase the table:

|--------------------------------------------------|-----------------|
| Port learning  | VLAN learning | Learning state  | Applied from    |
|--------------------------------------------------|-----------------|
| Enable         | Disable       | Enable          | Port            |
|--------------------------------------------------|-----------------|
| Enable         | Enable        | Enable          | Port            |
|--------------------------------------------------|-----------------|
| Disable        | Disable       | Disable         | ~               |
|--------------------------------------------------|-----------------|
| Disable        | Enable        | Enable          | VLAN            |
|--------------------------------------------------|-----------------|

For case you described behaviour suppose to be:  The ASIC does MAC learning and uses the port settings.

rivasiv:
Hi @lguohan, What do you think would it have sense?

kcudnik:
Closing, moved to sai 1.0

PR #33, opened 2015-12-21
   Added function prototype that hides initialization of service method table.

rivasiv:
This function just universalizes service method table initialization in host adapter
   and moves platform specific initialization outside host adapter(to bsp/platform adapter).

### DIFF
--- a/inc/sai.h
+++ b/inc/sai.h
@@ -130,7 +130,25 @@ typedef struct _service_method_table_t
 
 } service_method_table_t;
 
-
+/**
+ * Routine Description:
+ *     @brief Service method table initialization call.
+ *
+ * Arguments:
+ *     @param[in/out] services - methods table with services to
+ *           be initialised by host adapter
+ *
+ * Return Values:
+ *    @return  SAI_STATUS_SUCCESS on success
+ *             Failure status code on error
+ *  
+ * @note This function supposes to hide platform specific 
+ *       service method table initialization.
+ */
+sai_status_t
+sai_service_method_table_initialize(
+    _Inout_ service_method_table_t* services
+    );
 
 /**
  * Routine Description:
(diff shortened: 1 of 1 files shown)


lguohan:
can you provide example here? 

Method table that contains function pointers for services exposed by adapter host for adapter. The adapter host supply the functions so that SAI library can use the functions to get initial KV pairs. 

What does this sai_service_method_table_initialize do?

rivasiv:
Let's have a host adapter running over a single ASIC_X SAI adapter but two different boards called PlatformA and PlatformB. 

Each board uses the same ASIC_X but has its own profile which must be chosen when host runs over board A or B respectively. The profile functions for the boards could look as follows:
-PlatformA: 
 - profileAGetValue(), profileAGetNextValue()
 - profileBGetValue(), profileBGetNextValue()

Now when the host initializes service method table itself it would need to assign profileA... functions in case it works over boardA and profileB... ones over boardB to sai_service_method_table_t instance. This would mean that the whole universality of SAI interface is lost here since host adapter has to call platform(i.e. customer) specific code for service method table initialization. Thus the universality SAI provides will be limited.

If we add sai_service_method_table_initialize() function we move platform specific service method table initialization into it. Then sai_service_method_table_initialize() will be implemented for platforms A and B and will do the service method table initialization described above for each of them so it will be hidden from the host agent code. It can be placed to a small platform specific lib which will implement profileX... functions together with sai_service_method_table_initialize(). This lib can have common name and linked to a host agent.

With such an approach host agent will just create an instance of service_method_table_t and call sai_service_method_table_initialize() with a pointer to that instance. This will work for any platform host agent will work over without calls of any platform specific code since we will have a unified interface function provided by SAI.

rivasiv:
Hello, could we refresh this discussion, as it looks this approach will be very useful for any SAI client PTF/SoNIC/FBOSS etc... and will allow to provide binary with platform configuration instead of inventing work around each time.

rivasiv:
Ok, looks this is not relevant anymore. Closed

PR #37, opened 2015-12-23
Add mirror session CFI attribute

itaibaz:
VLAN_PRI is 3 bit PCP
VLAN_CFI is 1 bit CFI/DEI
Follow the concept used in saiacl.h


lguohan:
merged, thanks.

PR #38, opened 2015-12-23
Buffers - Add RX stats for PG, Add occupancy stats for Queue, PG, pool

itaibaz:
…G + Add occupancy stats for Queue, PG, pool

1) Port has ingress buffer profile list, can allow profile per ingress
pool
Same for egress
2) Added PG RX bytes / packets stats
3) Added PG, Queue, Pool current occupany and watermark stats


lguohan:
merged, thanks.

PR #42, opened 2015-12-28
tunnel proposal header files

itaibaz:


inc/saitunnel.h
```diff
@@ -0,0 +1,444 @@
...
+{
+    /** tunnel table point to point entry match on dst & src ip & tunnel type  **/
+    SAI_TUNNEL_TABLE_ENTRY_P2P,
+
+    /** tunnel table point to multi point entry match on dst ip  & tunnel type  **/
+    SAI_TUNNEL_TABLE_ENTRY_P2MP,
+
+} sai_tunnel_table_entry_type_t;
+
+typedef enum _sai_tunnel_table_entry_attr_t
```
lguohan: This table is used to match the tunnel termination packets. 

is _sai_tunnel_term_table_attr_t a better name?

inc/sairouterintf.h
```diff
@@ -46,7 +46,10 @@ typedef enum _sai_router_interface_type_t
     SAI_ROUTER_INTERFACE_TYPE_PORT,
 
     /** VLAN Router Interface Type */
-    SAI_ROUTER_INTERFACE_TYPE_VLAN
+    SAI_ROUTER_INTERFACE_TYPE_VLAN,
+
+    /** Loopback Router Interface Type */
+    SAI_ROUTER_INTERFACE_TYPE_LOOPBACK
```
lguohan: In the sai_router_interface_attr_t, what are the valid attributes for this LOOPBACK router interface?

inc/saitunnel.h
itaibaz: Thanks for the suggestion
Yes, tunnel term table is a better name
I will modify the header file accordingly

inc/saitunnel.h
```diff
@@ -0,0 +1,444 @@
...
+    SAI_TUNNEL_ATTR_UNDERLAY_INTERFACE,
+
+    /** tunnel overlay interafce [sai_object_id_t] */
+    SAI_TUNNEL_ATTR_OVERLAY_INTERFACE,
+
+    /** tunnel encap attribute*/
+
+    /** tunnel ip verssion ipv4/ipv6 [sai_ip_addr_family_t] 
+     *  (MANDATORY_ON_CREATE when (CREATE_ONLY) */
+    SAI_TUNNEL_ATTR_IP_VER,
```
lguohan: SRC_IP has ip family type built into it, is this redundant?

inc/saitunnel.h
```diff
@@ -0,0 +1,444 @@
...
+    /** READ-WRITE */
+
+    /** tunnel virtual router id [sai_object_id_t] (MANDATORY_ON_CREATE|CREATE_ONLY)*/
+    SAI_TUNNEL_TABLE_ENTRY_ATTR_VR_ID,
+
+    /** tunnel entry type [sai_tunnel_table_entry_type_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
+    SAI_TUNNEL_TABLE_ENTRY_ATTR_TYPE,
+
+    /** tunnel termination ip address [sai_ip_address_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
+    SAI_TUNNEL_TABLE_ENTRY_ATTR_DST_IP,
```
lguohan: is this the same as SAI_TUNNEL_ATTR_SRC_IP? If yes, can we remove this one here?

inc/saitunnel.h
```diff
@@ -0,0 +1,444 @@
...
+    /** tunnel encap attribute*/
+
+    /** tunnel ip verssion ipv4/ipv6 [sai_ip_addr_family_t] 
+     *  (MANDATORY_ON_CREATE when (CREATE_ONLY) */
+    SAI_TUNNEL_ATTR_IP_VER,
+
+    /** tunnel src ip [sai_ip_address_t] (MANDATORY_ON_CREATE when 
+     *  SAI_TUNNEL_ATTR_TYPE=SAI_TUNNEL_IPINIP,SAI_TUNNEL_IPINIP_GRE,
+     *  SAI_TUNNEL_IPINIP_GRE_AND_KEY) (CREATE_ONLY) */
+    SAI_TUNNEL_ATTR_SRC_IP,
```
lguohan: Since sai_tunnel_attr_t is used for both encap and decap, SAI_TUNNEL_ATTR_SRC_IP might not be a good name as the IP is the src IP for the encap, but the dest IP for the decap. Would it be better to rename it to SAI_TUNNEL_ATTR_IP?

inc/saitunnel.h
```diff
@@ -0,0 +1,444 @@
...
+     *  SAI_TUNNEL_ATTR_TYPE=SAI_TUNNEL_IPINIP_GRE_AND_KEY) (CREATE_ONLY) */
+    SAI_TUNNEL_ATTR_ENCAP_GRE_KEY,
+    
+    /** tunnel mapper [sai_object_id_t] */
+    SAI_TUNNEL_ATTR_ENCAP_MAPPER,
+
+    /** tunnel decap attribute **/
+
+    /**  tunnel decap ECN mapping [sai_object_id_t] */
+    SAI_TUNNEL_ATTR_DECAP_ECN_TABLE,
```
lguohan: What if the ASIC only support the hardcoded ECN mapping mode. Can we define another attribute SAI_TUNNEL_DECAP_ECN_MODE? Where the sai_tunnel_attr_decap_ecn_mode_t can be defined as
{
   SAI_TUNNEL_DECAP_ECN_MODE_COPY_FROM_OUTER,

   SAI_TUNNEL_DECAP_ECN_MODE_KEEP_INNER,

   SAI_TUNNEL_DECAP_ECN_MODE_USER_DEFINED
}

when it's USER_DEFINED, then you can supply with an ECN_MAPPER.

inc/saitunnel.h
```diff
@@ -0,0 +1,444 @@
...
+
+} sai_tunnel_type_t;
+
+typedef enum _sai_tunnel_ttl_mode_t
+{
+    SAI_TUNNEL_TTL_COPY_FROM_INNER,
+
+    SAI_TUNNEL_TTL_USER_DEFINE
+
+} sai_tunnel_ttl_mode_t;
```
lguohan: the sai_tunnel_ttl_mode_t is current define for the encap, we probably need separate it into sai_tunnel_encap_ttl_mode and sai_tunnel_decap_ttl_mode

inc/saitunnel.h
```diff
@@ -0,0 +1,444 @@
...
+
+typedef enum _sai_tunnel_dscp_mode_t
+{
+    SAI_TUNNEL_DSCP_COPY_FROM_INNER,
+
+    SAI_TUNNEL_DSCP_USER_DEFINE
+
+} sai_tunnel_dscp_mode_t;
```
lguohan: this is define only for the encap, we probably need another one for the decap.

inc/saitunnel.h
```diff
@@ -0,0 +1,444 @@
...
+    _Inout_ sai_attribute_t *attr_list
+    );
+
+typedef enum _sai_tunnel_type_t
+{
+    SAI_TUNNEL_IPINIP,
+
+    SAI_TUNNEL_IPINIP_GRE,
+
+    SAI_TUNNEL_IPINIP_GRE_AND_KEY,
```
lguohan: is this NVGRE? if yes, why not use the name NVGRE?

inc/saitunnel.h
```diff
@@ -0,0 +1,444 @@
...
+/**
+ * @brief Enum defining tunnel map types.
+ */
+typedef enum _sai_tunnel_map_type_t
+{
+    /** TUNNEL Map overlay ECN to underlay ECN   */
+    SAI_TUNNEL_MAP_OECN_TO_UECN = 0x00000001,
+
+    /** TUNNEL Map underlay ECN to overlay ECN   */
+    SAI_TUNNEL_MAP_UECN_TO_OECN = 0x00000002,
```
lguohan: the decap ECN mapper is slightly more complex than currently defined, in RFC6040, Section 4.2, the mapping maps (inner ECN, outer ECN) into final ECN.

inc/sairouterintf.h
mattyk: SAI_ROUTER_INTERFACE_ATTR_VIRTUAL_ROUTER_ID,
 SAI_ROUTER_INTERFACE_ATTR_TYPE,
 SAI_ROUTER_INTERFACE_ATTR_ADMIN_V4_STATE,
 SAI_ROUTER_INTERFACE_ATTR_ADMIN_V6_STATE,
 SAI_ROUTER_INTERFACE_ATTR_MTU,

inc/saitunnel.h
mattyk: yes it is you are right this is not needed

inc/saitunnel.h
mattyk: you are right . it is bad name 
this attribute is only used for the encap should be changed ot 
SAI_TUNNEL_DECAP_ATTR_SRC_IP

inc/saitunnel.h
mattyk: yes we can it is a good idea

inc/saitunnel.h
mattyk: you are right wev shoud have smiler attributes for decap as well

inc/saitunnel.h
mattyk: you are right we  should have similar attributes for decap as well

inc/saitunnel.h
mattyk: not it isn't NVGRE it is L3 IP tunnel  with GRE

inc/saitunnel.h
mattyk: you are right will  be fix

inc/saitunnel.h
lguohan: what's the difference between the GRE and GRE_AND_KEY?

inc/saitunnel.h
SathishKumarTR-Dell: Is the encap SRC_IP attribute redundant? Tunnel Encap Next Hop IP Address can be just used?

inc/saitunnel.h
```diff
@@ -0,0 +1,461 @@
...
+     *  (MANDATORY_ON_CREATE when SAI_TUNNEL_ENCAP_DSCP_MODE = SAI_TUNNEL_DSCP_USER_DEFINE) */
+    SAI_TUNNEL_ATTR_ENCAP_DSCP_VAL,
+
+    /** tunnel GRE key [sai_uint32_t] (MANDATORY_ON_CREATE when 
+     *  SAI_TUNNEL_ATTR_TYPE=SAI_TUNNEL_IPINIP_GRE_AND_KEY) (CREATE_ONLY) */
+    SAI_TUNNEL_ATTR_ENCAP_GRE_KEY,
+    
+    /** tunnel mapper [sai_object_id_t] */
+    SAI_TUNNEL_ATTR_ENCAP_MAPPER,
+
```
SathishKumarTR-Dell: Is it mandatory to pass tunnel encap attribute always? Should we allow to create a receive-only tunnel end point with just decap?

inc/saitunnel.h
```diff
@@ -0,0 +1,461 @@
...
+typedef enum _sai_tunnel_term_table_entry_attr_t
+{
+    /** READ-ONLY */
+
+    /** READ-WRITE */
+
+    /** tunnel virtual router id [sai_object_id_t] (MANDATORY_ON_CREATE|CREATE_ONLY)*/
+    SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_VR_ID,
+
```
SathishKumarTR-Dell: Do we need an attribute to match VNI for the VXLAN type?

inc/saitunnel.h
```diff
@@ -0,0 +1,461 @@
...
+typedef enum _sai_tunnel_decap_ecn_mode_t
+{
+    SAI_TUNNEL_DECAP_ECN_MODE_COPY_FROM_OUTER,
+
+    SAI_TUNNEL_DECAP_ECN_MODE_KEEP_INNER,
+
+    SAI_TUNNEL_DECAP_ECN_MODE_USER_DEFINED
+} sai_tunnel_decap_ecn_mode_t;
+
```
SathishKumarTR-Dell: Will there be a separate mode for the encap? rfc6040 specifies some functionality at tunnel encapsulation as well.

inc/saitunnel.h
mattyk: the next hop ip address -> tunnel destination ,  
the SAI_TUNNEL_DECAP_ATTR_SRC_IP  is for the source ip

inc/saitunnel.h
mattyk: in the GRE header the key file is optional , 
So one can build GRE header with or without key

inc/saitunnel.h
mattyk: you are right , should be change to - create only

inc/saitunnel.h
mattyk: not sure , the approach is to have a single tunnel for multiple VNIs

lguohan:
merged, thanks.

PR #43, opened 2016-01-03
Port has ingress and egress buffer profile lists

itaibaz:
Port has ingress buffer profile list, can allow profile per ingress pool
For example, suppose we have ingress lossy and lossless pools. Define
one profile of usage from the lossy pool, and one profile of usage from
the lossless pool
Same for egress


lguohan:
merged,thanks.

PR #44, opened 2016-01-03
update host-if doc 

mattyk:
new host interface  table was added
user define trap-id object was removed


lguohan:
merged, thanks

PR #45, opened 2016-01-05
Warmboot

AnbalaganNatchimuthu:


inc/saiswitch.h
```diff
@@ -296,6 +315,18 @@ typedef enum _sai_switch_attr_t
...
+
+    /** Minimum interval of time required by SAI for planned restart [sai_uint32_t]
+     *  in milliseconds. Will be 0 for SAI_RESTART_TYPE_NONE.
+     *  The Host Adapter will have to wait for this minimum interval of time before it decides
+     *  to bring down SAI due to init failure. */
+    SAI_SWITCH_ATTR_MIN_PLANNED_RESTART_INTERVAL,
+
+    /** Nonvolatile storage required by both SAI and NPU [sai_uint64_t] 
+     * Will be 0 for SAI_RESTART_TYPE_NONE */
+    SAI_SWITCH_ATTR_NV_STORAGE_SIZE,
```
tushar-ty: Need units specified for storage size. Could be KB or MB.

inc/saiswitch.h
```diff
@@ -140,6 +140,25 @@ typedef enum _sai_hash_algorithm_t
...
 /**
+ * @brief Attribute data for SAI_SWITCH_ATTR_RESTART_TYPE
+ */
+typedef enum _sai_switch_restart_type_t
+{
+    /** Planned restart only */
+    SAI_RESTART_TYPE_PLANNED = 0,
+
+    /** Unplanned restart only */
+    SAI_RESTART_TYPE_UNPLANNED = 1,
```
tushar-ty: I really doubt if someone would have 'unplanned only'. 
More practical options would be: none, planned, any (means planned and unplanned).

inc/saiswitch.h
lguohan: added KB

inc/saiswitch.h
lguohan: changed

PR #47, opened 2016-01-08
PFC stats and Shared current and watermark bytes

dgsudharsan:
Included port level counters for PFC TX and RX bytes per priority.
Included counters for shared current and watermark bytes per ingress priority group and per queue. Included current occupancy and watermark bytes per port.


lguohan:
merged, thanks.

PR #51, opened 2016-01-13
Add l2 multicast and ip multicast proposal

chenyq2:
L2 multicast is used for IGMP snooping entry, and IP Multicast is used for IPMC protocol, please review it and feedback to me.


lguohan:
close the pull request as we have new proposals

PR #52, opened 2016-01-13
IEEE 802.1BR SAI proposal

rsivasankar:
SAI proposal to implement IEEE 802.1BR functionality


chenyq2:
hi 
I have 2 problems.
1, how to add PE port to VLAN, it should be easy to extend "sai_add_ports_to_vlan_fn". but my main concern is the "remote replication";
2, considering chip performance and chip capability, can you add some FLAG to get "remote replication " capability on SAI.
and then, when user want to "add port to vlan", user should consider the chip capability and performance. user can add PE port to VLAN for CB replication or add Cascade port to VLAN for PE replication.

another concern is IPMC, if user use CB replication, it is easy to support it. but if user want to use PE replication, it is a issue. currently, SAI don't support IPMC, but i think we need consider it later.

rsivasankar:
Hi Chen,
1. The addition of the PE ports to the VLAN will be taken care by the communication between the control planes in CB and PE. The CB control plane will send control message to the control plane in PE to add the PE ports to the vlan. The applications running in the PE will then invoke the SAI vlan api to add the ports to the vlan in the PE. 
2. The replication/flooding will be handled using Multicast ECID. If replication is desired in CB, then CB will send multiple copies on the Cascading port, with each copy carrying the unicast ECID of the PE port. If replication is desired in PE, then CB will send one copy per Cascading port with the Vlan's Multicast ECID. The control plane will program the PE data plane, to replicate the frames based on the Multicast ECID.

For IPMC, separate ECID must be assigned to each Multicast group.

The PE ports will be instantiated in the CB as extended ports. There is no need to program any port attributes for these extended ports in CB data plane.

chenyq2:
Hi Rsivasankar
1,but if the replication is desired in CB, then extend port(ECID) should be added to VLAN on CB. here, the current SAI don't support add extend port(ECID) to vlan. 
or can you give me an example to do that?
2, if the replication is desired in PE, yes you are right that Control plane can handle it on PE. but on CB, how can we add Cascade port to VLAN. maybe we should consider that using 1br_port_object_id with Cascade type to "add_to_vlan".
3, for IPMC, it is also related with "remote replication" option. if IPMC replication is performed in CB, there are no problem. if IPMC replication is performed in PE, a Multicast ECID should be assigned to that IPMC group as you said. but in this case, we must consider that PE should copied the IPMC packet VLAN by VLAN. it is means the IPMC packets is flooded in multiple VLAN.
4, if you can add a switch attribute to describe the ASIC "remote replication" capability, it is convenient for my customer to design system.

rsivasankar:
Hi Chen,

Please see my responses inlined...

1,but if the replication is desired in CB, then extend port(ECID) should be added to VLAN on CB. here, the current SAI don't support add extend port(ECID) to vlan. 
or can you give me an example to do that?
[Ravi] I missed this part. The Extended port must also be added in the Vlan in the CB. The current   
          SAI add vlan port API takes the list of ports (object). The 1br_port_object_id can also be 
          allowed. So we can bring this support without changing the vlan api.  

2, if the replication is desired in PE, yes you are right that Control plane can handle it on PE. but on CB, how can we add Cascade port to VLAN. maybe we should consider that using 1br_port_object_id with Cascade type to "add_to_vlan".
[Ravi] SAI can implicitly add the Cascade Port to the Vlan (or) we can mandate that the application 
          must do this explicitly. For replication in PE, CB must transmit the frame on the cascading 
          port with the Vlan's Multicast ECID. So we need to associate the Multicast ECID to the Vlan. 
          This is missed in my proposal. I will add an ECID attribute in the vlan object.

3, for IPMC, it is also related with "remote replication" option. if IPMC replication is performed in CB, there are no problem. if IPMC replication is performed in PE, a Multicast ECID should be assigned to that IPMC group as you said. but in this case, we must consider that PE should copied the IPMC packet VLAN by VLAN. it is means the IPMC packets is flooded in multiple VLAN.
[Ravi] I have to correct my previous statement on this. We should assign one Multicast Ecid for 
          each vlan in a Multicast group. This is because, the PE will forward the packets based on 
          ECID only. It will not look into the vlan id present in the packet. 
          
          Let us say there are 3 PE ports ExP1V1, ExP2V1 and ExP3V2. 
          ExP1V1 and ExP2V1 are in vlan V1, while ExP3V2 is in V2. 
          Let the Multicast ECIDs assigned to the Vlans in this Multicast Group be ECID-Mc-V1 and 
          ECID-Mc-V2 (Note ECID-Mc-V1 and ECID-Mc-V2 are different than the ECIDs assigned to
          Vlans V1 and V2 for flooding purpose). 
          Let the cascading port through which the PE is reachable be CP.

          CB should transmit 2 packets on CP as follows:
          Packet 1: Vlan = V1, ECID = ECID-Mc-V1
          Packet 2: Vlan = V2, ECID = ECID-Mc-V2
           
          PE must be programmed such that,
              Packet with ECID-Mc-V1 is sent on Ports ExP1V1 and ExP2V1
              Packet with ECID-Mc-V2 is sent on Port ExP1V2

4, if you can add a switch attribute to describe the ASIC "remote replication" capability, it is convenient for my customer to design system.
[Ravi] This can be added. We can discuss this in detail during the meeting.

chenyq2:
hi Ravi,

good, it's cleared. specially, I accept your idea for IPMC replication.

PR #63, opened 2016-02-10
SAI api proposal for INT

pbhide:
Document for review


tushar-ty:
I think there are several typos in the docs examples.
Where is this inst_list on the LHS initialized ? 
attr[2].val.int_session_inst_list.inst_list = inst_list;

PR #65, opened 2016-02-12
Added threshold mode at buffer profile level

dgsudharsan:
Added threshold mode at buffer profile level to provide a finer granularity. If this is not set, it default the pool level threshold mode else overrides it.


lguohan:
merged, thanks.

PR #72, opened 2016-03-03
replace sai_add_ports_to_vlan_fn and sai_remove_ports_to_vlan_fn

lguohan:
with sai_(create|remove)_vlan_member_fn API.

### DIFF
--- a/inc/saivlan.h
+++ b/inc/saivlan.h
@@ -53,17 +53,6 @@ typedef enum _sai_vlan_tagging_mode_t
 } sai_vlan_tagging_mode_t;
 
 /**
- *   @brief Port/vlan membership structure
- */
-typedef struct _sai_vlan_port_t
-{
-    sai_object_id_t port_id;
-
-    sai_vlan_tagging_mode_t tagging_mode;
-
-} sai_vlan_port_t;
-
-/**
  *  @brief Attribute Id in sai_set_vlan_attribute() and
  *  sai_get_vlan_attribute() calls
  */
@@ -71,8 +60,8 @@ typedef enum _sai_vlan_attr_t
 {
     /** READ-ONLY */
 
-    /** List of ports in a VLAN [sai_vlan_port_list_t]*/
-    SAI_VLAN_ATTR_PORT_LIST,
+    /** List of vlan members in a VLAN [sai_object_list_t]*/
+    SAI_VLAN_ATTR_MEMBER_LIST,
 
     /** READ-WRITE */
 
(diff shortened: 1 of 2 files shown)


rivasiv:
Hi ,

  We already have all necessary information to operate with VLAN/port pair in SAI adapter client. The VLAN member is a bit synthetic instance which adding some extra complexity, isn't it? Isn't better just to add function which checks if port belongs to VLAN?

inc/saivlan.h
```diff
@@ -192,57 +216,54 @@ typedef sai_status_t (*sai_get_vlan_attribute_fn)(
...
+    \param[out] vlan_member_id VLAN member ID
+    \param[in] attr_count number of attributes
+    \param[in] attr_list array of attributes
+    \return Success: SAI_STATUS_SUCCESS
+            Failure: failure status code on error
+*/
+typedef sai_status_t (*sai_create_vlan_member_fn)(
+    _Out_ sai_object_id_t* vlan_member_id,
+    _In_ uint32_t attr_count,
+    _In_ const sai_attribute_t *attr_list
```
tushar-ty: From a usage point of view it appears a bit un-optimized compared to the previous routine where one could add multiple ports to a vlan in a single call. Now does the vlan_member object represent a single vlan-port pair ? and does one have to call this routine as many times as the number of ports in a vlan ?
It would help to see some example usage of these new objects.

lguohan:
@rivasiv The design rationale is explained in the https://github.com/lguohan/SAI/blob/uom/doc/SAI-Proposal-7-UOID.md 

@tushar-ty to optimize the performance, in the future, we'd like to have an API to do batch SAI object creation which allows user to create multiple sai object in one function call.

rivasiv:
Hi @lguohan , 

  I understand UOID conception, but 

  LAG_member and VLAN member sounds synthetic for me :

-  for example LAG is logical interface(it is complete, self managed instance) all it properties are consistent and suppose to be the same for all LAG members, so do we really need to have LAG_member object, operate with it as with single instance? Are there any real use case for this?  
Why we can not operate with single LAG object? That's more natural way, isn't it ?

 - we do not have a way to get VLAN from VLAN_member as UOID API conception doesn't provide such possibility(as UOID creation algorithm suppose to be vendor specific). So to operate with VLAN_member    
we still need to keep VLAN/VLAN_member map in client application. Are there any benefits of one more extra object(VLAN_member) ?  

Thanks,
   Rostyslav

lguohan:
A correction, please take a look at this proposal. https://github.com/lguohan/SAI/blob/uom/doc/SAI-Proposal-15-UOM.md

@rivasiv,  In LAG_member, there are EGRESS/INGRESS_DISABLE attribute for individual members. In VLAN_member, there is also tagged mode for each individual member.

In both cases, the member objects do have member specific attribute we need to configure. Then, somewhat we need to find way to expose it. the previous approach is to design another structure, e.g. sai_vlan_port_t, which does not fit well with the SAI API.

PR #73, opened 2016-03-04
IEEE 802.1BR SAI proposal.

rsivasankar:


inc/saidot1brfdb.h
```diff
@@ -0,0 +1,130 @@
...
+
+/** \defgroup SAIDOT1BRFDBENTRY SAI - 802.1BR ECID based forwarding specific public APIs and datastructures in PE.
+ *
+ *  \{
+ */
+
+/**
+ * @brief SAI attributes for SAI_OBJECT_TYPE_DOT1BR_FDB_ENTRY 
+ */
+typedef enum _sai_dot1br_fdb_entry_attr_t
```
atitjain: Cant this be merged with the existing fdb object. Can have an attribute to mark egress as BR port/list or if the BR port list is non null, the FDB points to BR list.

rsivasankar: Dell - Internal Use - Confidential
Hi Atit Jain,

There are few differences between existing FDB object and the proposed dot1brFdb object:

1.      The key (sai_fdb_entry_t) to existing FDB object is <Vlan Id, Mac Addr> tuple, whereas the key to dot1brFdb object is <Ecid>. If we want to merge these 2 objects then sai_fdb_entry_t must be converted to a union.

2.      The dot1brfdb entries cannot be learnt (and hence cannot be aged out too). They can only be configured.

3.      As a result of point 2, there is no need for fdb event notification and flushing of entries

So the object key and most of the attributes of existing FDB are not relevant to dot1brfdb object. I feel it will be better not to merge the 2 objects.

Regards,
Ravi

atitjain: You mean this is Port Extender Forwarding Table.

rsivasankar: Dell - Internal Use - Confidential
Hi Atit Jain,

Yes. This is Port Extender Forwarding Table.

Regards,
Ravi

atitjain: Could you please explain, in the SAI pipeline where this will be positioned.

rsivasankar: Dell - Internal Use - Confidential
Hi Atit Jain,

The SAI pipeline model does not describe the packet flow in the Port Extender device.

I have explained in section 3.2 of my proposal document, how this might fit in the pipeline model in a PE.

Regards,
Ravi

atitjain: You mean this table is not in the controller bridge but is in port extender?

inc/saidot1brport.h
```diff
@@ -0,0 +1,145 @@
...
+    SAI_DOT1BR_PORT_TYPE_NONE,
+    SAI_DOT1BR_PORT_TYPE_UPSTREAM,
+    SAI_DOT1BR_PORT_TYPE_CASCADE,
+    SAI_DOT1BR_PORT_TYPE_ACCESS,
+} sai_dot1br_port_type_t;
+
+/**
+ * @brief SAI attributes for SAI_OBJECT_TYPE_DOT1BR_PORT */
+typedef enum _sai_dot1br_port_attr_t
+{
```
atitjain: How do we map the dot1br_port to actual/physical port.

inc/saidot1brfdb.h
rsivasankar: Dell - Internal Use - Confidential
Hi Atit Jain,

Yes. This table is only in the Port Extender.

Regards,
Ravi

inc/saidot1brport.h
rsivasankar: Dell - Internal Use - Confidential
Hi Atit Jain,

I missed it. In the initial versions of the proposal, these dot1br port attributes were part of the port data attributes. One of the review comments is to create a separate dot1br port object containing only the dot1br attributes. While addressing this comment, I missed adding the physical port to which the dot1br port is mapped to.

I will update the proposal. Following will be the change (highlighted in blue) in the file inc/saidot1brport.h


typedef enum _sai_dot1br_port_attr_t

{

    /** READ-WRITE */



   /** The Port to which the 802.1BR Port is mapped to [sai_object_id_t]  (MANDATORY_ON_CREATE|CREATE_ONLY).

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_PORT,



    /** 802.1BR Port Type [sai_dot1br_port_type_t]

     * (MANDATORY_ON_CREATE|CREATE_ONLY).

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_TYPE,



    /** 802.1BR Port default ECID [sai_uint32_t] (MANDATORY_ON_CREATE|CREATE_AND_SET).

     * This attribute is valid only if the attribute SAI_DOT1BR_PORT_ATTR_TYPE

     * is set to SAI_DOT1BR_PORT_TYPE_ACCESS.

     * ECID to be added on receiving dot1br untagged frames.

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_ECID,



    /** 802.1BR Port default PCP [sai_uint8_t] (CREATE_AND_SET) (default to 0).

     * This attribute is valid only if the attribute SAI_DOT1BR_PORT_ATTR_TYPE

     * is set to SAI_DOT1BR_PORT_TYPE_ACCESS.

     * PCP to be added on receiving dot1br untagged frames.

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_PCP,



    /** 802.1BR Port default DEI [sai_uint8_t] (CREATE_AND_SET) (default to 0).

     * This attribute is valid only if the attribute SAI_DOT1BR_PORT_ATTR_TYPE

     * is set to SAI_DOT1BR_PORT_TYPE_ACCESS.

     * DEI to be added on receiving dot1br untagged frames.

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_DEI,



    /* -- */



    /* Custom range base value */

    SAI_DOT1BR_PORT_ATTR_CUSTOM_RANGE_BASE  = 0x10000000



} sai_dot1br_port_attr_t;

rsivasankar: Dell - Internal Use - Confidential
Hi Atit Jain,

I have updated the document.

https://github.com/opencomputeproject/SAI/pull/52

Regards,
Ravi

From: Sivasankar, Ravikumar
Sent: Monday, June 20, 2016 10:44 AM
To: 'opencomputeproject/SAI' <reply@reply.github.com>; opencomputeproject/SAI <SAI@noreply.github.com>
Cc: Author <author@noreply.github.com>
Subject: RE: [opencomputeproject/SAI] IEEE 802.1BR SAI proposal. (#73)


Dell - Internal Use - Confidential
Hi Atit Jain,

I missed it. In the initial versions of the proposal, these dot1br port attributes were part of the port data attributes. One of the review comments is to create a separate dot1br port object containing only the dot1br attributes. While addressing this comment, I missed adding the physical port to which the dot1br port is mapped to.

I will update the proposal. Following will be the change (highlighted in blue) in the file inc/saidot1brport.h


typedef enum _sai_dot1br_port_attr_t

{

    /** READ-WRITE */



   /** The Port to which the 802.1BR Port is mapped to [sai_object_id_t]  (MANDATORY_ON_CREATE|CREATE_ONLY).

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_PORT,



    /** 802.1BR Port Type [sai_dot1br_port_type_t]

     * (MANDATORY_ON_CREATE|CREATE_ONLY).

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_TYPE,



    /** 802.1BR Port default ECID [sai_uint32_t] (MANDATORY_ON_CREATE|CREATE_AND_SET).

     * This attribute is valid only if the attribute SAI_DOT1BR_PORT_ATTR_TYPE

     * is set to SAI_DOT1BR_PORT_TYPE_ACCESS.

     * ECID to be added on receiving dot1br untagged frames.

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_ECID,



    /** 802.1BR Port default PCP [sai_uint8_t] (CREATE_AND_SET) (default to 0).

     * This attribute is valid only if the attribute SAI_DOT1BR_PORT_ATTR_TYPE

     * is set to SAI_DOT1BR_PORT_TYPE_ACCESS.

     * PCP to be added on receiving dot1br untagged frames.

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_PCP,



    /** 802.1BR Port default DEI [sai_uint8_t] (CREATE_AND_SET) (default to 0).

     * This attribute is valid only if the attribute SAI_DOT1BR_PORT_ATTR_TYPE

     * is set to SAI_DOT1BR_PORT_TYPE_ACCESS.

     * DEI to be added on receiving dot1br untagged frames.

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_DEI,



    /* -- */



    /* Custom range base value */

    SAI_DOT1BR_PORT_ATTR_CUSTOM_RANGE_BASE  = 0x10000000



} sai_dot1br_port_attr_t;

rsivasankar: Dell - Internal Use - Confidential
Hi Atit Jain,

The include file is also updated.

https://github.com/opencomputeproject/SAI/pull/73

Regards,
Ravi

From: Sivasankar, Ravikumar
Sent: Monday, June 20, 2016 11:43 AM
To: 'opencomputeproject/SAI' <reply@reply.github.com>; 'opencomputeproject/SAI' <SAI@noreply.github.com>
Cc: 'Author' <author@noreply.github.com>
Subject: RE: [opencomputeproject/SAI] IEEE 802.1BR SAI proposal. (#73)


Dell - Internal Use - Confidential
Hi Atit Jain,

I have updated the document.

https://github.com/opencomputeproject/SAI/pull/52

Regards,
Ravi

From: Sivasankar, Ravikumar
Sent: Monday, June 20, 2016 10:44 AM
To: 'opencomputeproject/SAI' <reply@reply.github.com>; opencomputeproject/SAI <SAI@noreply.github.com<mailto:SAI@noreply.github.com>>
Cc: Author <author@noreply.github.com<mailto:author@noreply.github.com>>
Subject: RE: [opencomputeproject/SAI] IEEE 802.1BR SAI proposal. (#73)


Dell - Internal Use - Confidential
Hi Atit Jain,

I missed it. In the initial versions of the proposal, these dot1br port attributes were part of the port data attributes. One of the review comments is to create a separate dot1br port object containing only the dot1br attributes. While addressing this comment, I missed adding the physical port to which the dot1br port is mapped to.

I will update the proposal. Following will be the change (highlighted in blue) in the file inc/saidot1brport.h


typedef enum _sai_dot1br_port_attr_t

{

    /** READ-WRITE */



   /** The Port to which the 802.1BR Port is mapped to [sai_object_id_t]  (MANDATORY_ON_CREATE|CREATE_ONLY).

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_PORT,



    /** 802.1BR Port Type [sai_dot1br_port_type_t]

     * (MANDATORY_ON_CREATE|CREATE_ONLY).

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_TYPE,



    /** 802.1BR Port default ECID [sai_uint32_t] (MANDATORY_ON_CREATE|CREATE_AND_SET).

     * This attribute is valid only if the attribute SAI_DOT1BR_PORT_ATTR_TYPE

     * is set to SAI_DOT1BR_PORT_TYPE_ACCESS.

     * ECID to be added on receiving dot1br untagged frames.

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_ECID,



    /** 802.1BR Port default PCP [sai_uint8_t] (CREATE_AND_SET) (default to 0).

     * This attribute is valid only if the attribute SAI_DOT1BR_PORT_ATTR_TYPE

     * is set to SAI_DOT1BR_PORT_TYPE_ACCESS.

     * PCP to be added on receiving dot1br untagged frames.

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_PCP,



    /** 802.1BR Port default DEI [sai_uint8_t] (CREATE_AND_SET) (default to 0).

     * This attribute is valid only if the attribute SAI_DOT1BR_PORT_ATTR_TYPE

     * is set to SAI_DOT1BR_PORT_TYPE_ACCESS.

     * DEI to be added on receiving dot1br untagged frames.

     * Applicable only to Physical ports. */

    SAI_DOT1BR_PORT_ATTR_DEI,



    /* -- */



    /* Custom range base value */

    SAI_DOT1BR_PORT_ATTR_CUSTOM_RANGE_BASE  = 0x10000000



} sai_dot1br_port_attr_t;

inc/saiport.h
```diff
@@ -431,6 +433,14 @@ typedef enum _sai_port_attr_t
      * given port list will be dropped. */
     SAI_PORT_ATTR_EGRESS_BLOCK_PORT_LIST,
 
+    /** Dropping of 802.1BR untagged frames on ingress [bool] (default to FALSE).
+     * Applicable only to Physical ports. */
+    SAI_PORT_ATTR_DOT1BR_DROP_UNTAGGED,
```
atitjain: How these attributes interact with dot1br and dot1br_extended port objects. If a given port is set as physical port for one of the br types (cascade, upstream, access) , we can use the br type as a hint to derive drop etag tagged/untagged. Is there some other use case I am missing?

rsivasankar: Dell - Internal Use - Confidential
Hi Atit,

Yes. It is possible to derive the ‘drop etag tagged/untagged’ parameter from the br type (cascade, upstream, access). But there could be some violation scenarios, where receiving a particular type of frame on a port is invalid. For example, receiving dot1br Tagged frames might be invalid on dot1br Access port. Similarly receiving dot1br Untagged frames might be invalid on cascading port. It all depends on the SAI application’s need. We have some proprietary SAI applications in our software with these requirements.

Also on a non-dot1br port (port that is neither dot1br Access/Cascading/Upstream type), dot1br tagged frames should be dropped. Receiving dot1br frames on a non-dot1br port indicates a mis-configuration in the network.

Regards,
Ravi

atitjain:
Hi Ravi,

I was just thinking on the similar lines that you have mentioned:
1. Receiving dot1br Tagged frames might be invalid on dot1br Access port. 
2. Similarly receiving dot1br Untagged frames might be invalid on cascading port. 
3. Also on a non-dot1br port (port that is neither dot1br Access/Cascading/Upstream type), dot1br tagged frames should be dropped.

So atleast with above usecases in mind we have enough information to implicitly configure. I understand you have some application, so may be for that its not sufficient. I am just thinking about how the port type and drop will interact. For example on the access port if I configute accept etagged then what happens etc. 

- Atit

rsivasankar:
Dell - Internal Use - Confidential
Hi Atit,

IMHO the port type and the drop can be 2 independent parameters.

Normally there will not be a need to accept Etagged frames on Access port. But we never know that will be true always. In future, if there are hosts (connected to the PEs) which are Etag aware, they could transmit frames with just the PCP field set in the ETAG (aka to 802.1Q Priority tagged frames). The PE could then add the ECID and preserve the PCP field. This way the traffic from the hosts can be prioritized. For such cases we might have to allow Etagged frames on access frames.

Regards,
Ravi

lguohan:
We are working on the SAI logical pipeline model which define how packet are processed (logically) for a SAI-compliant NPU. Since 802.1BR defines a type of packet processing function, can you try to update the current [SAI pipeline model](https://github.com/opencomputeproject/SAI/blob/master/doc/behavioral%20model/)?

inc/saidot1brfdb.h
kcudnik: there is plan to convert fdb_entry_t to actual sai_object_id_t and mac and vlan could mandatory attributes, Type attribute could be introduced as mandatory attribute to determine wheter its fdb entry or fdb_dot1br entry and based on type only specific attributes would apply to specific fdb entry type.
IMO we dont need more unions, they make things more complicated and next to union we still need a type to know which union member is used. it would be better to use attributes instead

rsivasankar: Dell - Internal Use - Confidential
Hi Kamil Cudnik,

This FDB table is the forwarding table in Port Extender (PE). There are quite a number differences between the native L2 Fdb table and this one and hence it is not combined with the native L2 Fdb table.


1.      There is no learning and all entries are static

2.      Flush table requests are not applicable

3.      This table contains both unicast and multicast entries. In case of multicast entries, it will contain a list of ports.


This pull request is an old pull request. With respect to the SAI pipeline model and Bridge port proposals, the dot1br proposal required quite a number of changes. These changes were reflected in the proposal document. The changes in the dot1br header files cannot be incorporated until the bridge port header file changes come in. So waiting for those changes to update the dot1br pull request.

So the I have to come up with an updated dot1br proposal.

I did not want to delete the pull request, in order to preserve the comments provided on the pull request.

Regards,
Ravi

inc/saivlan.h
```diff
@@ -98,6 +98,12 @@ typedef enum _sai_vlan_attr_t
       * Value Range SAI_SWITCH_ATTR_VLAN_USER_META_DATA_RANGE */
     SAI_VLAN_ATTR_META_DATA,
 
+    /**
+      * The Multicast 802.1BR E-Channel Id (ECID), to be used in the flooding traffic
+      * sent by the CB to Port Extenders (PE). [sai_uint32_t] (CREATE_AND_SET) (default to 0)
+      */
+    SAI_VLAN_ATTR_FLOODING_ECID,
```
sonicyao1985: @rsivasankar ECID is not a global concept, it is allocated within the scope of an Internal External Bridge. There could be multiple Internal External Bridges within a CB. So ECID can not be used globally. And I don't understand the purpose of this FLOODING_ECID. If it's for PE replication (multicast ECID), then it's wrong. A packet could be flooded to normal access ports and multiple multicast E-Channels at the same time. For CB replication, this attribute is not needed.

PS; Suggest adding new switch level attribute to indicate that the switch is a CB or PE.

rsivasankar: Dell - Internal Use - Confidential
Hi Min Yao,

The ECID is global in a dot1br domain that contains a single CB and a set of associated PEs. I am not clear on what you mean by “Internal External Bridge”. Do you mean to refer to the CB domain?

A Vlan can have both non dot1br ports (native ports) and Extended ports (remote PE port). So unknown L2 traffic must be sent to both types of ports. Unknown unicast/flooding in a vlan can be achieved by any of the following mechanisms:


1.      CB Replication

CB can replicate a copy for each of the Extended ports as unicast frames (the ECID field will be set to the E-channel id of the extended port. It must ensure that source suppression is achieved here

2.      CB and PE Replication

CB can replicate a copy for each PE. PE would then replicate copies to its dot1br access ports. If there are 4 extended ports, 2 connected to PE1 and other 2 connected to PE2, then CB will send 1 copy to PE1 and another copy to PE2. The Ecid will be the same in both these copies. The PEs will then replicate the packet.

In this scenario the frame must carry the multicast ECID. This ECID is nothing bug SAI_VLAN_ATTR_FLOODING_ECID. If this attribute is not present, how will CB determine the multicast ECID that must be filled in the flooding traffic?

I am not clear with the following statement,

“A packet could be flooded to normal access ports and multiple multicast E-Channels at the same time.”

Can you please elaborate on the scenario, where there is a need for Vlan flooding (unknown L2 traffic on the vlan), to be sent on multiple multicast E-Channels? As per my understanding the unknown L2 traffic on the vlan has to be sent only on one multicast E-Channel. Only L3 Multicast would require multiple multicast E-Channels? Are you referring to L3 Multicast traffic?

Regards,
Ravi

sonicyao1985: @rsivasankar Sorry, mistype. “Internal External Bridge” -->> "Internal Bridge Port Extender".

An ECID is allocated within the scope the an "Internal Bridge Port Extender". Check IEEE 802.1br page 40, Figure 8-3,

rsivasankar: Dell - Internal Use - Confidential
Hi Min Yao,

You are right, the Standard says that the ECID must be unique with an Internal Bridge Port Extender.

For each Cascading Port in a CB, there is an associated, Internal Bridge Port Extender. Each Extended Port (remote port on the external PE) has an associated Internal Bridge Port. As per the standard, an ECID assigned to the External Extended Port can also be associated to another Extended Port, as long as these External Extended Ports are not associated to the same Internal Bridge Port Extender (or in other words, these external extended ports are not reachable via same Cascading Port).

But when a ECID is assigned to a Vlan for flooding, in a global scope, it will still not violate the Standards’ requirement of having unique ECID with in an Internal Bridge Port Extender. A CB is a system, containing the Internal Bridge Port Extenders as its subsystem. If the ECID is unique in CB, it would still be unique within each of its subsystems (Internal Bridge Port Extenders).

If we were to do otherwise, then each Vlan must be associated with a list of {Cascading Port, ECID} tuple. This would really complicate the SAI functionality.
Regards,
Ravi

sonicyao1985: @rsivasankar Now we're on the same page. : )   
"each Vlan must be associated with a list of {Cascading Port, ECID} tuple" --  this is exactly what I mean. My opinion is that we should allow the possibility that ControlStack allocates same ECID for two E-Channel in different Internal Bridge Port Extenders within one CB, to keep conformance to the IEEE standard. 
I remember that DELL has mentioned that only CB replication is required for now. In this case, this SAI_VLAN_ATTR_FLOODING_ECID attribute is not needed. Multicast ECID is only need when doing PE replication. I'm OK that DELL allocates different ECID for every ExtendedPort but I'm not OK with this SAI_VLAN_ATTR_FLOODING_ECID attribute.

rsivasankar: Dell - Internal Use - Confidential
Hi Min Yao,

I will address this.

The replication in CB can be achieved using the Multicast object that is being discussed in OCP. Will address this after the Multicast Pull request gets merged.

The dot1br header file pull request is an old one that needs update. Will take care of this then.

Regards,
Ravi

PR #77, opened 2016-03-21
SAI ACL attribute to match Route table Hit Meta Data.

SathishKumarTR-Dell:
Added new attribute to qualify Route table hit meta data.
Renamed the existing attributes in SAI ACL Table to be symmetric in the naming convention.
Added boolean data in sai_acl_field_data union.


lguohan:
merged, thanks.

PR #78, opened 2016-03-21
Detach switch_sai_thrift_rpc_thread after creation

pavel-shirshov:
We don't need to wait until this nonterminating thread finishes.


lguohan:
merged, thanks.

PR #79, opened 2016-03-22
fix stub autogen

stcheng:
fix autogen.sh to check libtoolize
add all SAI header files as dependencies for stub


lguohan:
merged, thanks

PR #81, opened 2016-03-24
Add sample packet mode.

MELLANOX-zhichenr:
Add sample packet mode which defines 2 behaviors: profile and netflow.

### DIFF
--- a/inc/saisamplepacket.h
+++ b/inc/saisamplepacket.h
@@ -52,15 +52,17 @@ typedef enum _sai_samplepacket_type_t
 typedef enum _sai_samplepacket_mode_t
 {
     /** A profile defining sample packet attributes. 
-      * The profile can be configured on a port, in which case, 
-      * an actual session is instantiated which is private for that port. 
-      * In case multiple ports are configured with the same profile, 
-      * sampling is independently done per port. */
+      * The profile can be configured on a port or as ACL entry action,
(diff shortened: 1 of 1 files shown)


inc/saisamplepacket.h
```diff
@@ -47,6 +47,25 @@ typedef enum _sai_samplepacket_type_t
...
+{
+    /** A profile defining sample packet attributes. 
+      * The profile can be configured on a port, in which case, 
+      * an actual session is instantiated which is private for that port. 
+      * In case multiple ports are configured with the same profile, 
+      * sampling is independently done per port. */
+    SAI_SAMPLEPACKET_MODE_EXCLUSIVE,
+    
+    /** An instance of sample packet session. 
+      * This session can be attached to multiple ACL entries. Traffic matching
```
lguohan: By its definition, the mode is not bound to acl rule, it can also be applied to the port. the same is true for the exclusive mode. Can you make this clear in the comments?

MELLANOX-zhichenr: Comment for sample packet mode has been updated. Thanks!

MELLANOX-zhichenr: Hi Guohan,

The comment for sample packet mode has been updated.

https://github.com/opencomputeproject/SAI/pull/81/commits

Thanks,

Zhichen

PR #82, opened 2016-03-28
Add tunnel gre key for decap

MELLANOX-zhichenr:
Tunnel encap and decap should be independent on each other. The case
that different gre keys for encap and decap should be allowed.

Without this change, tunnel decap needs to use encap gre key to decap
the packet, which might not be intuitive for the SAI user.


MELLANOX-zhichenr:
Discussion: decap key is used for validation, thus is not part of the general functionality of tunnel decap.

PR #86, opened 2016-03-29
Proposal specifying which attributes are applicable for LAG and for p…

itaibaz:
…ort when it's a LAG member


vgurevich:
Hello Itai,

A general suggestion: please, add a section to describe what happens when
the port is removed from a LAG.

For example, initially a port was not a member of a LAG and had an
attribute (e.g. SAI_PORT_ATTR_PORT_VLAN_ID) set to the value 5. The
LAG has SAI_PORT_ATTR_PORT_VLAN_ID
value as 7. The port was added to the LAG and later removed from it. Should
SAI restore the value of SAI_PORT_ATTR_PORT_VLAN_ID to 5?

I would prefer this to be a common, rather than case-by-case policy.

Another general suggestion is to describe the basic principle how you
decided whether to allow LAG ports to have individual values or not.

In the mirroring section you might have missed an attribute:
SAI_MIRROR_SESSION_ATTR_MONITOR_PORT

Thanks,
Vladimir

itaibaz:
Hi Vladimir,
Sorry for the delay in response

I will add the following details to our proposal :

·         Pure bridge attributes (.1q) are applicable only for a LAG, and not for a port which is a member of a LAG

·         When an attribute is not a pure bridge one, it is applicable both for a LAG and a member port.



·         When a port joins a LAG, it inherits from the LAG the pure bridge attributes configuration, meaning we are aligning these settings from the LAG to the new member port

·         When a port leaves a LAG :

1.       Pure bridge attributes will be set to the default values in SAI spec

2.       The rest of the attributes will remain as is


·         SAI_MIRROR_SESSION_ATTR_MONITOR_PORT added for the mirroring section

Thanks

vgurevich:
Hello Itai,

I think you'll have to start by defining what is a "pure bridge attribute",
since this is a totally new term, at least to me. :)

Thanks,
Vladimir

AshokDaparthi:
Hi Itai, 

Please let me know my understanding is correct or not. 

1# Admin/Operational state	Yes	Yes
if LAG is Admin/Oper state is down, are you expected member port admin state can be set to up by application?

2#Buffering	Yes 	Yes
if lag set with 100KB buffer, my understanding is aggregate of all ports buffer should not exceed the 100KB, if it is true, What in case member port buffer is set to 200KB. are you expecting this will be cached and used once port is out of LAG?

Still i am not clear about the approach. could you please help us to understand the proposal. 

Regards
Ashok D

AnbalaganNatchimuthu:
@itaibaz Adding to queries of @AshokDaparthi,

Policer/WRED applied to port and lag - What should be the behavior. Should the aggregate of all ports in LAG take effect or that of the individual members take effect. please explain.

PR #87, opened 2016-03-29
Remove SAI_TUNNEL_IPINIP_GRE

MELLANOX-zhichenr:
SAI_TUNNEL_IPINIP_GRE has the same functionality of
SAI_TUNNEL_ATTR_ENCAP_GRE_KEY_VALID in sai_tunnel_attribute_t which
indicates whether GRE encapsulation is needed or not.


MELLANOX-zhichenr:
Discussion: There can be a packet which has GRE header without GRE key. 
So this type is needed.

PR #96, opened 2016-04-19
Changes to add the action attributes as part of table id creation

AnbalaganNatchimuthu:


itaibaz:
Hi
I think all the action fields as part of table creation are of type boolean - meaning may be present in entries or won't be present
can you please fix the type of the new fields

AnbalaganNatchimuthu:
Hi @itaibaz 

I have addressed your comments.

Thanks
Anbalagan

PR #105, opened 2016-05-10
Add new traps : IP2ME, SSH and SNMP

itaibaz:


inc/saihostintf.h
```diff
@@ -230,6 +230,18 @@ typedef enum _sai_hostif_trap_id_t
     /** default packet action is forward */
     SAI_HOSTIF_TRAP_ID_MLD_V2_REPORT = 0x0000200f,
 
+    /** IP packets to local router IP address (routes with SAI_ROUTE_ATTR_NEXT_HOP_ID=SAI_SWITCH_ATTR_CPU_PORT)
```
atitjain: 1. My IP should be hit in the route/host table and then should go the CPU as pkt cmd. There is no other possible logical action. Worst case some user might want to drop. 
2. Why we need specific TRAP IDs. Also do we expect the HW to do deeper packet inspection to even figure out application.

itaibaz: Hi,
The motivation for having specific traps is to be able to control the Qos settings – cpu queue and policer rate limit, on these specific types of traffic.
There is actual use case for this in Sonic.
We don't expect deep packet inspection, basically just match ip is ip2me, and the standard ports for the protocol.

atitjain: Hi Itai,

I could guess the use case being QoS and in my previous implementations there is a queue for my IP. When we trap from the host/route table for the my IP how do you expect to account for the type. Or the Idea is to use the ACLs for my IP as well.

Regards,
Atit

itaibaz: The idea is to use the new trap SAI_HOSTIF_TRAP_ID_IP2ME
Classify it to a trap group, attach a policer to the trap group (SAI_HOSTIF_TRAP_GROUP_ATTR_POLICER) and define the queue (SAI_HOSTIF_TRAP_GROUP_ATTR_QUEUE).
Internally, each vendor can implement with his SDK host IFC API, or with ACL API.
The traffic that matches this new trap, has next hop id = cpu port, as explained in
https://github.com/opencomputeproject/SAI/pull/106

inc/saihostintf.h
```diff
@@ -230,6 +230,18 @@ typedef enum _sai_hostif_trap_id_t
...
+     * (default packet action is trap) */
+    SAI_HOSTIF_TRAP_ID_IP2ME = 0x00002010,
+
+    /** SSH traffic to local router IP address
```
lguohan: can you add dest port = 22 criteria to the comment to make this trap more clear?

inc/saihostintf.h
```diff
@@ -230,6 +230,18 @@ typedef enum _sai_hostif_trap_id_t
...
+     * (default packet action is trap) */
+    SAI_HOSTIF_TRAP_ID_IP2ME = 0x00002010,
+
+    /** SSH traffic to local router IP address 
+     * (default packet action is trap) */
+    SAI_HOSTIF_TRAP_ID_SSH = 0x00002011,
+
+    /** SNMP traffic to local router IP address
+     * (default packet action is trap) */
+    SAI_HOSTIF_TRAP_ID_SNMP = 0x00002012,
```
lguohan: can you add udp port = 161 criteria to the comments to make this trap more clear?

inc/saihostintf.h
itaibaz: Ports info added

PR #106, opened 2016-05-10
Add IP2ME route

itaibaz:


SathishKumarTR-Dell:
Is this intended to trap packets without binding the route to a local RIF? 
I guess the packet has to be mapped to a RIF and classified to a VRF and so the SAI model requires to bind the route to a local RIF (Port/VLAN/Loopback) to trap the self destined packets to CPU. So, this looks like adding a redundant case.

lguohan:
this is to provide meta data (class) for certain route which can be used in later stage for classification. In this case, for the BGP trap we'd like only to match the bgp traffic with dest ip equal to one of our interface ip or loopback ip while src/dst port = 179. 

The IP2ME route will trap those packets to CPU while add those meta data, so that later stage can use the meta data information to do the match.

We also thought about using the meta data approach in SAI. But that approach seems to be too generic for this specific problem and asic platform may various in terms of this IP2ME implementation. One specific problem that this meta does not really propogate to acl on certain platforms. Therefore, we prefer to use this way to solve this problem.

PR #112, opened 2016-05-18
allow enabling of ecn without wred in wred profile, change enable fro…

itaibaz:
…m bool to enum

### DIFF
--- a/inc/saiwred.h
+++ b/inc/saiwred.h
@@ -29,13 +29,27 @@
  *  \{
  */
 
+/**
+*  @brief wred ecn enable modes
+*/
+typedef enum _sai_wred_enable_mode_t
+{
+    SAI_WRED_ENABLE_MODE_NONE,
+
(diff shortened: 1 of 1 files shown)


inc/saiwred.h
```diff
@@ -92,9 +106,6 @@ typedef enum _sai_wred_attr_t
     /** 0 ~ 15 [sai_uint8_t], Default 0*/
     SAI_WRED_ATTR_WEIGHT = 0x0000000c,
```
AshokDaparthi: Hi Itai, 

Please let me know, how to enable ECN in case ECN enable/disable is not supported by colour?

Regards
Ashok D

itaibaz: Hi
While the SAI model allows setting ECN and Wred enable per color,
If specific HW only support global control for ECN,
I can suggest implementation wise this logic :
If at least one color is enabled, enable global ECN, and if the user tries explicitly sets different values (for example, green ecn enable=true, yellow ecn enable=false), return not supported

PR #113, opened 2016-05-19
Fix types for virtual functions.

paulnice:
Otherwise they don't match to the generated from thrift abstract class. thrift doesn't have unsigned integer types.

### DIFF
--- a/test/saithrift/src/switch_sai_rpc_server.cpp
+++ b/test/saithrift/src/switch_sai_rpc_server.cpp
@@ -1691,10 +1691,10 @@ class switch_sai_rpcHandler : virtual public switch_sai_rpcIf {
   }
 
   void sai_thrift_get_port_stats(
-          std::vector<uint64_t> & thrift_counters,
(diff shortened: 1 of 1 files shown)


test/saithrift/src/switch_sai_rpc_server.cpp
```diff
@@ -1691,10 +1691,10 @@ class switch_sai_rpcHandler : virtual public switch_sai_rpcIf {
   }
 
   void sai_thrift_get_port_stats(
-          std::vector<uint64_t> & thrift_counters,
+          std::vector<int64_t> & thrift_counters,
```
lguohan: can we try to use thrift 0.9.3?

paulnice: Yes. We can try. But why? I don't see thrift-0.9.3 supports unsigned integers.
https://github.com/apache/thrift/blob/master/CHANGES

PR #119, opened 2016-05-27
Updated HostIF packet attr comments

SathishKumarTR-Dell:


inc/saihostintf.h
```diff
@@ -558,8 +558,10 @@ typedef enum _sai_hostif_packet_attr_t
     /** packet transmit type [sai_hostif_tx_type_t]. (MANDATORY_ON_SEND) */
     SAI_HOSTIF_PACKET_ATTR_TX_TYPE,
 
-    /** Egress port or LAG [sai_object_id_t] (for send-only).
-     * (MANDATORY_ON_SEND when SAI_HOSTIF_PACKET_TX_TYPE == SAI_HOSTIF_TX_TYPE_PIPELINE_BYPASS) */
+    /** Egress port or LAG [sai_object_id_t].
+     * (MANDATORY_ON_SEND when SAI_HOSTIF_PACKET_TX_TYPE == SAI_HOSTIF_TX_TYPE_PIPELINE_BYPASS) 
+     * For the receive case, filled with the egress destination port for unicast packets.
```
zubinshah: Specifically on the RX case, for the example mentioned in the comment: for an ingress sampled packet, are we expecting the destination port to be filled in the packet's meta data ?

SathishKumarTR-Dell: Yes. It can be filled by a SAI implementation for the unicast sampled packets when destination port is known and provided by the switch element.

inc/saihostintf.h
```diff
@@ -558,8 +558,10 @@ typedef enum _sai_hostif_packet_attr_t
     /** packet transmit type [sai_hostif_tx_type_t]. (MANDATORY_ON_SEND) */
     SAI_HOSTIF_PACKET_ATTR_TX_TYPE,
 
-    /** Egress port or LAG [sai_object_id_t] (for send-only).
-     * (MANDATORY_ON_SEND when SAI_HOSTIF_PACKET_TX_TYPE == SAI_HOSTIF_TX_TYPE_PIPELINE_BYPASS) */
+    /** Egress port or LAG [sai_object_id_t].
```
lguohan: is this port or LAG? which one specifically?

inc/saihostintf.h
zubinshah: What happens in case when the packet is trapped at the ingress of a pipe, and the destination port is not resolved ?

inc/saihostintf.h
SathishKumarTR-Dell: It should be filled with port and application can lookup the LAG Id from port when needed. If a specific SDK/ASIC provides only the LAG for some reason, that can be filled in the attribute in that case. I have just removed the comment that defines this attribute "for send-only" case.

inc/saihostintf.h
SathishKumarTR-Dell: This attribute can be filled only for the known unicast packets that is resolved a destination port in the pipeline and the meta data is available to SAI.

PR #122, opened 2016-06-01
Port can require different Hardware configuration based on the attach…

Rajasekhar-Pulluru:
…ed media type,

attached cable length

Example for an hardware configuration could be (but not limited to) interface type setting,
tuning port pre-emphasis settings, modifying phy registers etc.

Handling SAI_PORT_ATTR_HW_PROFILE_ID configuration is implementation defined.
If the port hardware configuration doesn't vary depending on the attached media type/cable,
then the SAI implementation may chose to ignore this parameter.

 Changes to be committed:
	modified:   saiport.h

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -179,24 +179,6 @@ typedef enum _sai_port_media_type_t
     /** Media type not known */
     SAI_PORT_MEDIA_TYPE_UNKNONWN,
 
-    /** Media type QSFP fiber optic */
-    SAI_PORT_MEDIA_TYPE_QSFP_FIBER,
-
-    /** Media type QSFP copper optic */
-    SAI_PORT_MEDIA_TYPE_QSFP_COPPER,
-
-    /** Media type SFP fiber optic */
-    SAI_PORT_MEDIA_TYPE_SFP_FIBER,
-
-    /** Media type SFP copper optic */
-    SAI_PORT_MEDIA_TYPE_SFP_COPPER,
-
-    /** Media type QSFP28 fiber optic */
-    SAI_PORT_MEDIA_TYPE_QSFP28_FIBER,
-
-    /** Media type QSFP28 copper optic */
-    SAI_PORT_MEDIA_TYPE_QSFP28_COPPER,
-
     /** Media type fiber. Remote advertise medium information as fiber */
     SAI_PORT_MEDIA_TYPE_FIBER,
 
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -431,6 +431,15 @@ typedef enum _sai_port_attr_t
      * given port list will be dropped. */
     SAI_PORT_ATTR_EGRESS_BLOCK_PORT_LIST,
 
+    /** Port Hardware Configuration Profile ID [sai_uint64_t]
+     * Port can require different hardware configuration based on the attached
+     * media type, cable length etc. A Profile ID maps to a Port Hardware
+     * configuration settings that needs to be applied on the Port.
+     * This attribute need not be implemented and can be ignored if the port
+     * doesn't require any specific hardware settings based on media type/cable.
+     */
+    SAI_PORT_ATTR_HW_PROFILE_ID,
+
     /** -- */
 
     /* Custom range base value */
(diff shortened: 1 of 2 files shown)


inc/saiport.h
```diff
@@ -431,6 +431,15 @@ typedef enum _sai_port_attr_t
...
     SAI_PORT_ATTR_EGRESS_BLOCK_PORT_LIST,
 
+    /** Port Hardware Configuration Profile ID [sai_uint64_t]
+     * Port can require different hardware configuration based on the attached
+     * media type, cable length etc. A Profile ID maps to a Port Hardware
+     * configuration settings that needs to be applied on the Port.
+     * This attribute need not be implemented and can be ignored if the port
+     * doesn't require any specific hardware settings based on media type/cable.
+     */
+    SAI_PORT_ATTR_HW_PROFILE_ID,
```
atitjain: For my understanding. 
1. A profile is collection of certain switch configurations needed to support, tune given port media etc.
2. Switch/NPU vendors will define profile ID per supported media etc 
3. The user of SAI APIs will select the profile from the adapter host depending on their plugable/cable etc.

Rajasekhar-Pulluru: Hi @atitjain 
Your understanding is correct except that it's not intended to mean a collection of certain switch configurations (as you mentioned in point 1), instead it's a collection of certain port specific hardware configuration.
Example could be to configure the port interface type to SR4, speed to 40G and any special phy tuning paramters when QSFP SR4 type optic is inserted.

mikelazar: This is supposed to work the following way:

The SAI implementation defines a vendor specific configuration file, which maps a "HW Port profile IDs" to a set of parameters. 
- The name of the configuration file is specified using the SAI_KEY_HW_PORT_PROFILE_ID_CONFIG_FILE (as a key/value pair, see saiswitch.h)
- The contents of the configuration file is vendor and hardware specific, but it is assumed that it can map HW Port Profiles IDs to specific hardware parameters.

For readability, consider the following format :
<ProfileID>:<ParameterName>=<Value>,ParameterName>=<Value>,...

For instance:
1: Pre-emphasis=valuePreEmphasis1,SerDesPolarity=valueSerDesPolarity1
2: Pre-emphasis=valuePreEmphasis2,SerDesPolarity=valueSerDesPolarity2
3: Pre-emphasis=valuePreEmphasis3,SerDesPolarity=valueSerDesPolarity3
...


When the application (host adapter) sets the SAI_PORT_ATTR_HW_PROFILE_ID (say, SAI_PORT_ATTR_HW_PROFILE_ID=2), the SAI simply uses the corresponding values from the mapping table and sets the appropriate vendor specific values. Typically, those values correspond to a specific transceiver type (SFP, QSPF), but we do not need to specify transceiver types in the SAI API (and keep adding these types as new SFP's are supported).

This approach is useful because: 
- we do not need to tell the SAI what is the type of transceiver (we only need to provide a generic port profile ID). Furthermore, given that the profile ID to NPU HW register mapping is contained in a configuration file, the code does not need to be modified to add new transceiver types (just add new entries to the configuration file)
- the information can be vendor specific, but this is transparent to the SAI API

- the SAI user application ('host-adapter') does not need to "hard-code" transceiver types to "HW Port Profile IDs". This mapping can be provided as a separate configuration file as well (although this is outside the scope of the SAI API). Also, it should be noted that the same HW port profile ID may correspond to multiple transceiver types.
For instance:

QSFP_TypeA: 1
QSFP_TypeB: 1
SFP_TypeC: 2
SFP_TypeD: 2
...
When a transceiver is plugged in into a port, the host adapter determines the SAI_PORT_ATTR_HW_PROFILE_ID value that corresponds to the transceiver, and calls the appropriate SAI API for the port.

As new transceiver types are supported, the host-adapter-configuration file can be updated.

In summary, with this approach:
- the SAI does not need to be aware of transceiver types
- the SAI API is NPU agnostic/independent, yet specific NPU registers can be supported (inside an NPU specific SAI implementation)
- both the 'host-adapter' and SAI implementation can be configuration/data driven (no hard-coded mappings)

PR #124, opened 2016-06-08
Changes to add action list as part of table creation

AnbalaganNatchimuthu:


inc/saiacl.h
```diff
@@ -339,8 +437,12 @@ typedef enum _sai_acl_table_attr_t
     /** Range type defined in sai_acl_range_type_t */
     SAI_ACL_TABLE_ATTR_FIELD_RANGE,
 
+    /** List of actions in sai_acl_table_action_list_t [sai_u32_list_t] 
+     * Default action list will be present which can be modified*/
+    SAI_ACL_TABLE_ATTR_ACTION_LIST,
```
atitjain: When you say the default action list, will it be defined by SAI proposal or the provider of the adapter will init this to what he can support. Will a adapter host select a subset from this. I thing just being able to query the action list (read only parameter) should be sufficient for the application to adapt its logic.

atitjain: There are two lists in the switch api, mandatory and the supported. which is assumed to be default. The supported looks like the superset of all. So if switch can support all then what would be different here.

inc/saiswitch.h
```diff
@@ -349,7 +349,14 @@ typedef enum _sai_switch_attr_t
...
     SAI_SWITCH_ATTR_NV_STORAGE_SIZE,
-
+    
+    /** Whether action list is mandatory for table creation [bool] 
+     * Default [false]*/
+    SAI_SWITCH_ATTR_ACL_TABLE_ACTION_LIST_MANDATORY,
+
+    /** List of actions supported [sai_u32_list_t] 
+     * List of all supported actions will be returned */
+    SAI_SWITCH_ATTR_ACL_TABLE_ACTION_LIST,
```
atitjain: What is use of SAI_SWITCH_ATTR_ACL_TABLE_ACTION_LIST_MANDATORY is this something that cant be knokced off while creating the tables. Also all ACL tables (in all stages) might not support same set of actions, so how will this global be useful. Do we assume mandatory to be same for all stages etc. 
Typically Keys can be programmed to save on the TCAM width. What is the use of restricting actions, could ASICs get advantage on the width of action memory (say SRAM etc)

AnbalaganNatchimuthu: Dell - Internal Use - Confidential
HI Atit,

Updated the comments in saiacl.h.

The SAI_SWITCH_ATTR_ACL_TABLE_ACTION_LIST_MANDATORY is a Boolean variable which will indicate whether action list is a mandatory attribute needed during table creation.

The SAI_SWITCH_ATTR_ACL_TABLE_ACTION_LIST will be the list of all the actions supported by the NPU defined in SAI. This action list can be queried and SAI_ACL_TABLE_ATTR_ACTION_LIST can be set.

If the SAI_SWITCH_ATTR_ACL_TABLE_ACTION_LIST_MANDATORY is true then application has to explicitly send the action list.



Thanks
Anbalagan

atitjain: in case when its not mandatory, could application still get the supported list and change the logic.

lguohan: @atitjain @AnbalaganNatchimuthu "Do we assume mandatory to be same for all stages etc", looks like it does. However, this may not be a valid assumption here. In this case, for each stage the SAI supports, we need to know whether the action list is mandatory or not.

inc/saiacl.h
```diff
@@ -339,8 +437,13 @@ typedef enum _sai_acl_table_attr_t
     /** Range type defined in sai_acl_range_type_t */
     SAI_ACL_TABLE_ATTR_FIELD_RANGE,
 
+    /** List of actions in sai_acl_table_action_list_t [sai_u32_list_t]
```
lguohan: sai_s32_list_t

AnbalaganNatchimuthu:
@atitjain @lguohan  Addressed the issues and modified the approach based on the discussion. Please review.

inc/saiacl.h
```diff
@@ -734,6 +841,36 @@ typedef enum _sai_acl_range_attr_t
...
+    /* Type of acl stage */
+    sai_acl_stage_t stage;
+
+    /* boolean indicating whether action list is mandatory for table creation */
+    bool            is_action_list_mandatory;
+
+    /* List of actions supported per stage from the sai_acl_table_action_list_t.
+     * Count can be obtained from the switch attribute SAI_SWITCH_ATTR_MAX_ACL_ACTION_COUNT*/
+    sai_s32_list_t  action_list;
+}sai_acl_attr_per_stage_t;
```
lguohan: rename to sai_acl_attr_stage_info_t

lguohan: need to put this in sai_attribute_value_t

inc/saiacl.h
```diff
@@ -734,6 +841,36 @@ typedef enum _sai_acl_range_attr_t
...
+
+/**
+ * @brief Structure to get the acl capabilities 
+ */
+typedef struct _sai_acl_capabality_t
+{
+    /* Number of stages */
+    int no_of_stages;
```
kcudnik: should this be named count ? to match current list's format?

inc/saiacl.h
```diff
@@ -734,6 +841,36 @@ typedef enum _sai_acl_range_attr_t
...
+
+    /* Pointer to structure having the attributes per stage */
+    sai_acl_attr_per_stage_t *acl_attrs;
```
kcudnik: should this be named acl_attrs_list? to match current list's format?

inc/saiacl.h
```diff
@@ -734,6 +841,36 @@ typedef enum _sai_acl_range_attr_t
...
+ *  action_list alone is added now. Qualifier list can also be added 
+ *  when needed.
+ */
+typedef struct _sai_acl_attr_per_stage_t
+{
+    /* Type of acl stage */
+    sai_acl_stage_t stage;
+
+    /* boolean indicating whether action list is mandatory for table creation */
+    bool            is_action_list_mandatory;
```
kcudnik: some extra spaces between bool and is_action_list_mandatory

inc/saiacl.h
```diff
@@ -734,6 +841,36 @@ typedef enum _sai_acl_range_attr_t
...
+
+    /* List of actions supported per stage from the sai_acl_table_action_list_t.
+     * Count can be obtained from the switch attribute SAI_SWITCH_ATTR_MAX_ACL_ACTION_COUNT*/
+    sai_s32_list_t  action_list;
```
kcudnik: if list is used then this structure should contain count entry as well

inc/saiacl.h
lguohan: if we have stage as a SAI object, then it will be easier to solve query its attributes. but I do not know whether it makes sense to create ACL stage or remove ACL stage.

inc/saiswitch.h
```diff
@@ -496,6 +505,8 @@ typedef enum _sai_switch_attr_t
     /** Port Breakout mode [sai_port_breakout_t] */
     SAI_SWITCH_ATTR_PORT_BREAKOUT,
 
+
+
```
stcheng: extra lines?

inc/saiacl.h
AnbalaganNatchimuthu: @kcudnik Modified as count.

inc/saiacl.h
AnbalaganNatchimuthu: @kcudnik Modified

inc/saiacl.h
AnbalaganNatchimuthu: Removed extra spaces

inc/saiacl.h
AnbalaganNatchimuthu: sai_s32_list_t has a count and pointer for the list.

inc/saiswitch.h
AnbalaganNatchimuthu: Removed the extra lines

inc/saiacl.h
AnbalaganNatchimuthu: @lguohan  Moved the structures to saitypes.h and added as part of sai_attribute_value_t.

inc/saitypes.h
```diff
@@ -506,6 +506,35 @@ typedef struct _sai_tunnel_map_list_t
...
+
+    /* List of actions supported per stage from the sai_acl_table_action_list_t.
+     * Count can be obtained from the switch attribute SAI_SWITCH_ATTR_MAX_ACL_ACTION_COUNT*/
+    sai_s32_list_t  action_list;
+}sai_acl_attr_stage_info_t;
+
+/**
+ * @brief Structure to get the acl capabilities 
+ */
+typedef struct _sai_acl_capabality_t
```
mikelazar: Just noticed: typo in "sai_acl_capabality_t" - should be ...capability_t

inc/saitypes.h
```diff
@@ -506,6 +506,35 @@ typedef struct _sai_tunnel_map_list_t
...
+{
+    /* Number of stages */
+    int count;
```
mikelazar: unsigned int ...

inc/saitypes.h
```diff
@@ -506,6 +506,28 @@ typedef struct _sai_tunnel_map_list_t
...
 /**
+ *  @brief Structure for acl attributes supported at each stage.
+ *  action_list alone is added now. Qualifier list can also be added 
+ *  when needed.
+ */
+typedef struct _sai_acl_capability_t
+{
+    /* Type of acl stage. Input to get the action list
+     * Failure to pass the stage as input will be treated as error */
+    uint32_t stage;
```
lguohan: In the comment, note it is using enum. sai_acl_stage_t

sai_int32_t stage

AnbalaganNatchimuthu:
@lguohan addressed the comment. Please review.

PR #127, opened 2016-06-21
Remove decap ttl val and decap dscp val

MELLANOX-zhichenr:
For uniform mode, ttl val and dscp val are copied from outer header in
decap.
For pipe mode, ttl val and dscp val are copied from innner header in
decap.
Thus ttl val and dscp val are not used for either mode.


MELLANOX-zhichenr:
Merged this change to pull request #128 

https://github.com/opencomputeproject/SAI/pull/128

PR #128, opened 2016-06-21
Tunnel ttl and dscp pipe model comment change

MELLANOX-zhichenr:
Make the comment be more clear on the behavior for encapsulation and
decapsulation.

### DIFF
--- a/inc/saitunnel.h
+++ b/inc/saitunnel.h
@@ -156,8 +156,8 @@ typedef enum _sai_tunnel_ttl_mode_t
      *  where the outer header is independent of that in the inner header so
      *  it hides the TTL field of the inner header from any interaction
      *  with nodes along the tunnel.
-     *  TTL field is user-defined for outer header on encapsulation and copied
-     *  from inner header on decapsulation. */
+     *  TTL field is user-defined for outer header on encapsulation. TTL
+     *  field of inner header remains the same on decapsulation. */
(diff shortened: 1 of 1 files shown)


inc/saitunnel.h
```diff
@@ -171,7 +173,9 @@ typedef enum _sai_tunnel_dscp_mode_t
     /** The pipe model:
      *  where the outer header is independent of that in the inner header so
      *  it hides the DSCP field of the inner header from any interaction
-     *  with nodes along the tunnel. */
+     *  with nodes along the tunnel.
+     *  DSCP field is user-defined for outer header on encapsulation and copied
+     *  from inner header on decapsulation. */
```
lguohan: on decapsulation, the DSCP field is kept the same as before.

inc/saitunnel.h
```diff
@@ -155,7 +155,9 @@ typedef enum _sai_tunnel_ttl_mode_t
     /** The pipe model:
      *  where the outer header is independent of that in the inner header so
      *  it hides the TTL field of the inner header from any interaction
-     *  with nodes along the tunnel. */
+     *  with nodes along the tunnel.
+     *  TTL field is user-defined for outer header on encapsulation and copied
+     *  from inner header on decapsulation. */
```
lguohan: on decapsulation, the DSCP field is kept the same as before.

PR #129, opened 2016-06-21
Allow color aware control of ECN marking in wred ecn profile

itaibaz:
In case ASIC is color unaware, implementation can support mode_none and
mode_all, and return not supported for the rest of the modes


AshokDaparthi:
i am fine with this

PR #133, opened 2016-06-29
add fast boot option

lguohan:


inc/saiswitch.h
```diff
@@ -532,8 +532,10 @@ typedef enum _sai_switch_attr_t
 #define SAI_KEY_NUM_QUEUES                        "SAI_NUM_QUEUES"
 #define SAI_KEY_NUM_CPU_QUEUES                    "SAI_NUM_CPU_QUEUES"
 #define SAI_KEY_INIT_CONFIG_FILE                  "SAI_INIT_CONFIG_FILE"
-/** 0: means cold boot, and 1: means warm boot */
-#define SAI_KEY_WARM_BOOT                         "SAI_WARM_BOOT"
+/** 0: cold boot. Initialize NPU and external phys.
+ *  1: warm boot. Do not re-initialize NPU or external phys, reconstruct SAI/SDK state from stored state.
+ *  2: fast boot. Only initilize NPU */
```
zubinshah: @lguohan  just to clarify this mode of option requires *ALL* NPU tables/state/configs to be reset like it was a COLD BOOT , only avoid resetting the external PHYs. So that means the port speed configuration will not be changed across this type of reboot?

lguohan: * fast boot == Cold boot for NPU , meaning SAI/SDK state should not be persisted except for those related to physical port attributes such as SPEED, AUTONEG mode, admin state, oper status. 

Once we separate physical port attribute with logical port attribute, we'll have more clear definition what are persisted and what are not persisted.

PR #134, opened 2016-06-29
add clarification on the definition of UDF based offset

lguohan:


tushar-ty:
The changes make this very clear.

PR #141, opened 2016-07-12
Update saithrift infrastructure

ysegev:
Add profilefile handler
Add front_port_to_hw_lanes mapping handler
Note that saiserver must get now two files :
1. profile file
2. front_port_to_hw_lanes map
Add support in all test environment, evantually port_list in saithrift
will contain sai_ports in order of the user interfaces


test/saithrift/tests/sai_base_test.py
```diff
@@ -39,6 +42,25 @@ def setUp(self):
...
+        elif self.test_params.has_key("port_map_file"):
+            user_input = self.test_params['port_map_file']
+            f = open(user_input, 'r')
+            for line in f:
+                if (len(line) > 0 and (line[0] == '#' or line[0] == ';' or line[0]=='/')):
+                    continue;
+                interface_front_pair = line.split("@")
+                interface_to_front_mapping[interface_front_pair[0]] = interface_front_pair[1].strip()
+        else:
+            exit("No interface<->front port mapping, please specify as parameter or external file")
```
lguohan: can you generate default binding, i.e, 0@0, 1@1, ... for all ptf host interfaces?

test/saithrift/tests/switch.py
```diff
@@ -74,6 +76,15 @@ def switch_init(client):
...
     # wait till the port are up
     time.sleep(10)
+
+    thrift_attr = client.sai_thrift_get_port_list_by_front_port()
+    if thrift_attr.id == SAI_SWITCH_ATTR_PORT_LIST:
+        for port_id in thrift_attr.value.objlist.object_id_list:
+            front_port_list.append(port_id)
+
+    for interface,front in interface_to_front_mapping.iteritems():
+        port_list[int(interface)]=front_port_list[int(front)-1]
```
lguohan: why not both ptf interface and front panel port start from 0?

test/saithrift/src/saiserver.cpp
```diff
@@ -128,20 +165,162 @@ void sai_diag_shell()
...
+                options.profileMapFile = std::string(optarg);
+                break;
+
+            case 'f':
+                printf("port map file: %s", optarg);
+                options.portMapFile = std::string(optarg);
+                break;
+
+            case '?':
+                printf("unknown get opti option %c", optopt);
```
lguohan: unknown option

test/saithrift/tests/sai_base_test.py
```diff
@@ -39,6 +42,30 @@ def setUp(self):
...
+                    continue;
+                interface_front_pair = line.split("@")
+                interface_to_front_mapping[interface_front_pair[0]] = interface_front_pair[1].strip()
+        else:
+            f = open('./tests/default_interface_to_front_map.ini', 'r')
+            for line in f:
+                if (len(line) > 0 and (line[0] == '#' or line[0] == ';' or line[0]=='/')):
+                    continue;
+                interface_front_pair = line.split("@")
+                interface_to_front_mapping[interface_front_pair[0]] = interface_front_pair[1].strip()
```
lguohan: given that we change the interface map file format, I think it's better to make the map file as a mandatory input file.

test/saithrift/src/switch_sai_rpc_server.cpp
```diff
@@ -1060,6 +1060,153 @@ class switch_sai_rpcHandler : virtual public switch_sai_rpcIf {
...
+      sai_object_list_t *port_list_object;
+      int max_ports = 0;
+      sai_thrift_object_id_t port_id;
+      extern std::map<std::set<int>, std::string> gPortMap;
+      std::map<std::set<int>, std::string>::iterator gPortMapIt;
+
+      status = sai_api_query(SAI_API_SWITCH, (void **) &switch_api);
+      if (status != SAI_STATUS_SUCCESS) {
+          printf("sai_api_query failed!!!\n");
+          return 0;
```
lguohan: >0 [](start = 17, length = 1)

SAI_NULL_OBJECT_ID

rspolyak:
Could you please elaborate more on changes done?
What is the purpose of adding front_port_to_hw_lanes mapping?

ysegev:
The purpose is to create a generic test environment that will work on all platforms and with all cable connectivity setups.

rspolyak:
thanks. I am wondering because now PTF client does not longer work with following message:
_SystemExit: No ptf interface<-> switch front port mapping, please specify as parameter or in external file_

That's because PTF client now expects either port_map or port_map_file.
Could you please provide an example of PTF client command line?

ysegev:
This is Correct. You can see an example for this map in msn_2700 subfolder /saitests/src/msn_2700.

As for the command line: you have to options:

1. -t "port_map='0@Ethernet0,1@Ethernet1'" according to your cable connectivity.  This will map the ptf interface to a front port of the switch.

2. -t "port_map_file='<FILE LOCATION>'". As I mentioned earlier you have an example for this file in msn subfolder.

This way, and with the front_port to hw_lanes map that the saiserver receives on startup, the port_list is always sorted with the ptf interface as an index ( port_list[<PTF INTERFACE>]  )

And I also added a list called front_port_list that is sorted according to the switch front ports.

rspolyak:
got it. Thanks for details.

PR #143, opened 2016-07-12
update host interface trap API

lguohan:
Add host interface trap create/remove API

### DIFF
--- a/inc/saihostintf.h
+++ b/inc/saihostintf.h
@@ -139,7 +139,7 @@ typedef sai_status_t(*sai_get_hostif_trap_group_attribute_fn)(
     _Inout_ sai_attribute_t *attr_list
     );
 
-typedef enum _sai_hostif_trap_id_t
+typedef enum _sai_hostif_trap_type_t
 {
     /** control plane protocol*/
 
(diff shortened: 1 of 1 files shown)


inc/saihostintf.h
```diff
@@ -276,6 +276,10 @@ typedef enum _sai_hostif_trap_channel_t
 
 typedef enum _sai_hostif_trap_attr_t
 {
+    /** Host interface trap ID [sai_hostif_trap_id_t]
+     * (CREATE_ONLY|MANDATORY_ON_CREATE|KEY) */
+    SAI_HOSTIF_TRAP_ATTR_TRAP_ID,
```
atitjain: Calling this and the trap handle (hostif_trap_id) same might cause confusion. Please think about renaming the Enum as sai_hostif_trap_t/sai_hostif_trap_type_t and for eg SAI_HOSTIF_TRAP_ATTR_ID_VRRP ->SAI_HOSTIF_TRAP_ATTR_VRRP/SAI_HOSTIF_TRAP_ATTR_TYPE_VRRP.

lguohan: changed as suggested.

PR #144, opened 2016-07-12
add port/queue create/remove API

lguohan:
remove SAI_SWITCH_ATTR_PORT_BREAKOUT. Port breakout will be replaced
via port create/remove API.


lguohan:
merged.

PR #145, opened 2016-07-12
add next hop group member API

lguohan:
remove SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_LIST attribute
remove sai_add_next_hop_to_group_fn and sai_remove_next_hop_from_group_fn API
add Read Only SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_MEMBER_LIST attribute

### DIFF
--- a/inc/sainexthopgroup.h
+++ b/inc/sainexthopgroup.h
@@ -59,21 +59,39 @@ typedef enum _sai_next_hop_group_attr_t
     /** Number of next hops in the group [uint32_t] */
     SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_COUNT,
 
+    /** Next hop list [sai_object_list_t] */
+    SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_MEMBER_LIST,
+
     /** READ-WRITE */
 
     /** Next hop group type [sai_next_hop_group_type_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
     SAI_NEXT_HOP_GROUP_ATTR_TYPE,
 
-    /** Next hop list [sai_object_list_t] (MAXDATORY_ON_CREATE) 
-     * The next hop group must have at least one next hop member at the creation time */
-    SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_LIST,
-
-    /* -- */
+     /* -- */
 
     /* Custom range base value */
     SAI_NEXT_HOP_GROUP_ATTR_CUSTOM_RANGE_BASE  = 0x10000000
 
 } sai_next_hop_group_attr_t;
+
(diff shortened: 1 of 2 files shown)


inc/sainexthopgroup.h
```diff
@@ -59,22 +59,40 @@ typedef enum _sai_next_hop_group_attr_t
     /** Number of next hops in the group [uint32_t] */
     SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_COUNT,
 
+    /** Next hop list [sai_object_list_t] */
```
lguohan: next hop member list

inc/sainexthopgroup.h
```diff
@@ -59,22 +59,40 @@ typedef enum _sai_next_hop_group_attr_t
...
+    SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_MEMBER_LIST,
+
     /** READ-WRITE */
 
     /** Next hop group type [sai_next_hop_group_type_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
     SAI_NEXT_HOP_GROUP_ATTR_TYPE,
 
-    /** Next hop list [sai_object_list_t] (MAXDATORY_ON_CREATE) 
-     * The next hop group must have at least one next hop member at the creation time */
-    SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_LIST,
```
tushar-ty: We discussed about this but I am NOT convinced that leaving this as optional attribute creates a complicated behavior (in conjunction with a readonly member list retrieval attribute). 
But in the absence of the bulk create mechanism in 1.0 I do see this having two benefits: 
a. We don't break any existing test and application code. 
b. We have a simple and optimized way to create the nexthop group with a known list of nexthops up front instead of calling the new create member API multiple times.

Once the bulk create interface is defined post 1.0 this mechanism could be deprecated and the existing code could then be easily modified to use the new mechanism.

lguohan: @itaibaz, @kcudnik, one thing here is that we'll have create_switch API, where a few defaults objects are created along with the switch object. In this sense, creating next hop members in a next hop group call is not different from creating default hash object during the switch creation.

sonicyao1985:
@lguohan hi guohan, will this PR be merged soon?

inc/sainexthopgroup.h
lguohan: the tushar's suggestion looks ok to me. As about the remove_nexthopgroup api, I assume we need to remove all nexthop members first before removing the nexthop group. @itaibaz , @kcudnik , are you ok with it?

itaibaz: sounds ok to me

lguohan:
regenerate pull request in #248

PR #146, opened 2016-07-12
add Read-Only SAI_VLAN_ATTR_PORT_LIST to retrive port list for a vlan

lguohan:


itaibaz:
Hi
Can you explain why do we need both SAI_VLAN_ATTR_MEMBER_LIST and SAI_VLAN_ATTR_PORT_LIST.
If you want the port list, you can read the member list, and then the SAI_VLAN_MEMBER_ATTR_PORT_ID from each member. isn't one of them redundant ?

lguohan:
@atitjain , seems we can get the port list using the member_list. Is that ok if we drop this commit?

stcheng:
agree to have only member list here.

atitjain:
@lguohan I missed the member list part. I am ok with dropping this commit.

lguohan:
unnecessary per discussion

PR #149, opened 2016-07-20
add Broadcom switch shell attribute

lguohan:


inc/saiswitch.h
```diff
@@ -511,6 +511,9 @@ typedef enum _sai_switch_attr_t
     /* Custom range base value */
     SAI_SWITCH_ATTR_CUSTOM_RANGE_START = 0x10000000,
 
+    /* Enable Broadcom switch shell [bool] (default to 0) */
+    SAI_SWITCH_ATTR_BRCM_SWITCH_SHELL_ENABLE,
```
zubinshah: @lguohan  can this be made generic and not BRCM specific ? 

@cavium we support similar private extension to enable shell, making this generic would be  better.

lguohan: the thread model for shell is different on different platforms. some starts a thread, some starts a process. If you take a look at our syncd code. https://github.com/Azure/sonic-sairedis/blob/master/syncd/syncd.cpp#L1194

Does the calling pattern work for you?

zubinshah: yes! XPliant's shell is "initialize"/"enable" in a very similar fashion. Internal to the SAI adapter we invoke a thread. The user interface to the shell definitely is different, but should be agnostic to SAI.

inc/saiswitch.h
```diff
@@ -500,6 +500,9 @@ typedef enum _sai_switch_attr_t
        Default no map */
     SAI_SWITCH_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP,
 
+    /* Enable switch shell [bool] (default to false) */
+    SAI_SWITCH_ATTR_SWITCH_SHELL_ENABLE,
```
atitjain: How about clubbing all this as part of a saidebug module. This may have all logger/debug functionality and everything else needed for debugging. Also are the shell commands etc supposed to be vendor specific and more like HW debug?

lguohan: this switch shell is vendor specific and is more like hw debug, maybe we can rename as SWITCH HW SHELL. Will add more into the comments.

PR #155, opened 2016-07-27
fixed issue with incorrect socket number during saithrift server init

rspolyak:
Fixed issue which is observed during saithrift server initialization
_TTransportException: Could not connect to localhost:XXXX

The root cause is in start_sai_thrift_rpc_server() which passes local variable pointer to pthread_create(). This location is sometimes overwritten and socket is being bind to incorrect port.


rspolyak:
@lguohan Please review

lguohan:
the reason this that start_sai_thrift_rpc_server() might exit early than the thread creation, so the stack buffer is reallocated so that you'll get a wrong value. 

Thanks for find this bug and fix it.

PR #158, opened 2016-08-01
update pipeline 

mattyk:
add MC flows


sonicyao1985:
Q1: In the 1Q_flow/1Q_MC/1Q_MC router diagrams, the lookup key "BR_ID" in tables LearningFDB/L3interface/MCFDB actually means VID, right?

sonicyao1985:
The 1Q_MC pipeline looks fine with me. 
Regrading to the 1Q MC router pipeline, I have one question:
Q2: A routerinterface can be retrieved after MC FDB lookup, which means the outif_list(nexthop group) includes a sai routerinterface object. My question is that will the control stack be required to add the routerinterface to the nexthop group explicitly or ASIC vendors could add that internally?

lguohan:
@sonicyao1985 , when the router interface has IPV4 MC enabled, then the bridge port that connect the router interface to the bridge should be added the outif_list correspondingly.

sonicyao1985:
@lguohan Thanks for the clarification. That is the behavior that I'm expecting.

PR #159, opened 2016-08-02
Add switch attributes for default objects

kcudnik:
Motivation for this attributes is to provide knowledge to end user what objects are created by default after SAI initialization, for example: all port should be on that list, default virtual router, all vlan members objects for vlan 1. All objects with valid object ID should be on that list, this list should be updated every time new object is created or removed.

In my understandind some default objects are created. Extra object without ID is VLAN 1 which is created by default and all ports are added to that vlan createing vlanmember object.
No fdb_entry, route_entry, neighbor_entry should be create by default.

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -500,6 +500,12 @@ typedef enum _sai_switch_attr_t
        Default no map */
     SAI_SWITCH_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP,
 
+    /** The number of default objects created on the switch [sai_uint32_t] */
+    SAI_SWITCH_ATTR_DEFAULT_OBJ_NUMBER,
+
+    /** Get the list of default objects created on the switch [sai_object_list_t] */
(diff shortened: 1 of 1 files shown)


inc/saiswitch.h
```diff
@@ -500,6 +500,12 @@ typedef enum _sai_switch_attr_t
        Default no map */
     SAI_SWITCH_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP,
 
+    /** The number of default objects created on the switch [sai_uint32_t] */
```
zubinshah: @kcudnik @lguohan what is the motivation for the default objects created by the switch? can you provide a use case ? 

Typically there are a lot of attributes/resources available on the ASIC, they are not really created or deleted. Are you referring to OBJECT IDs only or hardware resources.

kcudnik: i updated description for this pull request

kcudnik:
Closing this PR since it was already added

PR #163, opened 2016-08-03
Cavm port cfg

rspolyak:
- Introduced Cavium port configuration files.

- Reduced number of times SAI adapter is queried from saithrift server in order to speed up switch initialization. Without this change, 128 port switch initialization  was very slow.

- Substituted hard coded lanes number (4) with actual value returned from sai adapter. This allows having other port configurations rather than 32 ports only.


test/saithrift/src/switch_sai_rpc_server.cpp
```diff
@@ -1204,36 +1207,37 @@ class switch_sai_rpcHandler : virtual public switch_sai_rpcIf {
...
 
-          std::set<int> port_lanes;
-          for (int j=0 ; j<4 ; j++){
-              port_lanes.insert(port_lane_list_attribute.value.u32list.list[j]);
-          }
+
+      port_lane_list_attribute.id = SAI_PORT_ATTR_HW_LANE_LIST;
+      port_lane_list_attribute.value.u32list.list = (uint32_t *) malloc(sizeof(uint32_t) * 4);
+      port_lane_list_attribute.value.u32list.count = 4;
+      port_api->get_port_attribute(port_list_object_attribute.value.objlist.list[portNum], 1, &port_lane_list_attribute);
```
lguohan: the portNum here is the front_port_number. But the index for port_list_object_attribute.value.objlist.list is not front panel port. Maybe on cavium platform it is arranged in this way, but it is not guaranteed for other platform.

I can see why the performance of original code is not good as it needs to query the HW_LANE_LIST for every port in the function. To solve the problem, I propose to save the hw_lane_list as a cache so that you do not need query it for every port in this function.

lguohan:
:clock1:

zubinshah:
@rspolyak - please add a new directory as7512-32x to locate these port mapping files.

PR #170, opened 2016-08-10
Add SAI_BUFFER_POOL_ATTR_XOFF_SIZE attribute for allocating buffer pool headroom

gmirek:


inc/saibuffer.h
```diff
@@ -328,14 +328,23 @@ typedef enum _sai_buffer_profile_attr_t
...
-     * Generate XON when the total buffer usage of this PG
-     * is less this threshold and available buffer in the PG buffer
-     * is larger than the XOFF threahold.
+     * Ensures to generate XON, and ignore hysteresis, when the total 
+     * buffer usage of this PG is less than this threshold and available 
+     * buffer in the PG buffer is larger than the XOFF threshold.
      * default to 0. */
     SAI_BUFFER_PROFILE_ATTR_XON_TH,
 
+    /** set the buffer profile XON hysteresis threshold in byte [sai_uint32_t]
```
atitjain: Please could you give an example for this attribute. I think already there is a hysteresis in picture due to use of XOFF and XON.

lguohan: agree with you. This is another mechanism to generate XON. The existing one compares the available PG buffer usage with a threshold to generate xon. The new one in below is an OFFSET, when the available PG buffer is less than XOFF threshold minus the OFFSET, it will generate the xon.

When a user sets both of them, the XON trigger threshold is max(XON_TH, XOFF_TH-XON_OFFSET)

It is probably more clear to rename the new attribute as  XON_OFFSET_TH.

@gmirek , can you add the equation into the comments so that user understand the relationship between the two XON threshold.

inc/saibuffer.h
```diff
@@ -190,6 +190,9 @@ typedef enum _sai_buffer_pool_attr_t
 
     /** buffer pool size in bytes [sai_uint32_t] (MANDATORY_ON_CREATE|CREATE_AND_SET) */
     SAI_BUFFER_POOL_ATTR_SIZE,
+    
+    /** For the ingress buffer pool, we allocate a shared XOFF (GLOBAL HEADROOM) room for lossless traffic */
```
lguohan: add comment: default value 0

inc/saibuffer.h
```diff
@@ -325,14 +329,29 @@ typedef enum _sai_buffer_profile_attr_t
...
+    /** set the buffer profile XON hysteresis threshold in byte [sai_uint32_t]
+     * Valid only for ingress PG (CREATE_AND_SET)
+     * Ensures to generate XON, taking into account hysteresis, when the total 
+     * buffer usage of this PG plus this threshold is less than the XOFF 
+     * threshold and available buffer in the PG buffer is larger than the 
+     * XON threshold.
+     * When the user sets both XON and XON_OFFSET thresholds, the XON
+     * trigger condition is governed by:
+     *          max(XOFF_TH - XON_OFFSET_TH, XON_TH)
+     * default to 0. */
```
lguohan: default set to 0xffffffff, by default the xon_offset_th is very large so it never gets chance to be triggered.

PR #171, opened 2016-08-15
ACL model changes for SAI 1.0.0

zubinshah:
Created this preliminary version of ACL model based on the community discussions and changes discussed during the 8/8-8/9 workshop. 

Primary changes includes binding model. Some more changes and examples to follow.


doc/SAI-Proposal-ACL-1.md
```diff
@@ -0,0 +1,89 @@
...
+ Authors     | Cavium Inc.
+ Status      | In review
+ Type        | Standards track
+ Created     | 08/09/2016
+ SAI-Version | 1.0.0
+
+-------------------------------------------------------------------------------
+
+## Overview ##
+SAI Access Control Lists (ACL) object implements ACL management functions. In SAI 0.9.1 through 0.9.5 versions, SAI ACL contained three types of objects, ACL table, ACL entry and ACL counter. The ACL table contains a number of ACL entries. Each ACL table defines a set of unique matching fields for all its ACL entries. A packet can match rules in different ACL tables and take non-conflicting actions from all the matched rules. However, within an ACL table, if a packet matches multiple rules, only the actions from the rule of highest priority are executed. ACL counters can also be created and attached to an ACL entry in order to counter the number of packets or bytes that match the ACL entry. The initial version of ACL table object has several ambiguities and limitations. We propose the following enhancements to address some of those and propose a generic and simple ACL model for operators:
```
tushar-ty: Would have helped if some examples of ambiguities with the existing spec had been provided. 
Would have also helped if the benefits of the change in the existing model be provided with some example.
So that it helps a user understand - How should the existing application code be modified to start using the benefits of the new model and overcome the limitations of the existing model.

inc/saiacl.h
```diff
@@ -1154,7 +1154,8 @@ typedef struct _sai_acl_api_t
     sai_remove_acl_range_fn             remove_acl_range;
     sai_set_acl_range_attribute_fn      set_acl_range_attribute;
     sai_get_acl_range_attribute_fn      get_acl_range_attribute;
-
+    sai_create_acl_group_fn             create_acl_group;
```
tushar-ty: Whats the signature of this new function ?

doc/SAI-Proposal-ACL-1.md
itaibaz: Hi,
I have a question re multi group binding to the same binding point.
With the old API, we had the definition :
     * Within one ACL group, only one ACL entry will be hit which is based on the
     * table priority as well as ACL entry priority within the table.
     * Across the ACL groups, one hit ACL entry from each group will be selected
     * and all non-conflicting actions from them will be performed.

With the new API, I can only see how to bind one group to each binding point, since each binding point has just one object ID, either a group or a table.
Within the group, tables will be evaluated according to table priority, and so there can only be one match per binding point.

So not sure if i'm missing something - how can we support multiple matches, one from each group, with the new API ?
Should we change the binding point to accept object list, thus allowing multiple groups to be bounded ?

Thanks

zubinshah: Від: Shah, Zubin [mailto:Zubin.Shah@cavium.com] 
Надіслано: Wednesday, October 5, 2016 12:30 AM
Кому: Itai Baz <itaib@mellanox.com>
Копія: Shah, Zubin <Zubin.Shah@cavium.com>; Guohan Lu <gulv@microsoft.com>; Matty Kadosh <mattyk@mellanox.com>; Yakiv Huryk <yakivh@mellanox.com>; Jain, Atit <Atit.Jain@cavium.com>
Тема: Re: question re multi group binding
 
HI Itai,
Extremely sorry for the delay in responding. It seems this message got lost while i was on vacation. Please see my answers inline.
 
Thanks,
Zubin
 
On Sep 21, 2016, at 7:35 PM, Itai Baz <itaib@mellanox.com> wrote:
 
Hi Zubin,

I have a question re multi group binding to the same binding point (also posted it on the github).
With the old API, we had the definition :
* Within one ACL group, only one ACL entry will be hit which is based on the
* table priority as well as ACL entry priority within the table.
* Across the ACL groups, one hit ACL entry from each group will be selected
* and all non-conflicting actions from them will be performed.

Zubin>> Just to clarify that during the community review , it was clear that there were folks who had grave objections to support “resolution” of non-conflicting actions across ACL tables of a specific group, hence to simplify the behavior was explicitly discussed to be what you wrote below (highlighted in green). I know your questions relates to non-conflicting actions across ACL group(s).
 
With the new API, I can only see how to bind one group to each binding point, since each binding point has just one object ID, either a group or a table.
Within the group, tables will be evaluated according to table priority, and so there can only be one match per binding point.

So not sure if i'm missing something - how can we support multiple matches, one from each group, with the new API ?
Should we change the binding point to accept object list, thus allowing multiple groups to be bounded ?

 
Zubin>> I am completely okay and would be able to accommodate this change of accepting the attribute list at every binding point. However, I’d suggest keeping it optional, because there are folks who would disagree to this. What do you guys in the CC list think?

zubinshah: The idea to enable programming any ACL Group or ACL Table be applied to filter traffic INGRESSING or EGRESSING the switch (all ports). It would be filtering all traffic. The real reason to provide such a configuration is to provide backward compatibility to the Pre SAI1.0 ACL model. (Which was assuming such a behavior).

inc/saiacl.h
zubinshah: @tushar-ty - I have updated the spec/document with review comments as well. Please find the updated signature in the PR, that lists the following signatures for the two functions: 

typedef sai_status_t (*sai_create_acl_table_group_fn)(                              
        _Out_ sai_object_id_t *acl_table_group_id,                                  
        _In_ uint32_t attr_count,                                                   
        _In_ const sai_attribute_t *attr_list);                                     

typedef sai_status_t (*sai_remove_acl_table_group_fn)(                              
        _In_ sai_object_id_t acl_table_group_id);                                   


The attributes used for sai_create_acl_table_group_fn : SHOULD be updated ENUM defined by : sai_acl_table_group_attr_t.

doc/SAI-Proposal-ACL-1.md
zubinshah: @tushar-ty : I will update the spec with the examples that you have asked for. Please review and let me know if you have further queries. 

The benefits of the new ACL model for SAI 1.0.0, and the motivation to enhance/update the current ACL model (pre SAI 1.0) is very clearly mentioned in the beginning of the section: Again listing for reference in this comment:

This proposal introduced for SAI 1.0.0, proposes the following enhancements:       
1. Introduce well defined binding point for an ACL table (or ACL Group) to be applied
2. Clarify the behavior and usage of ACL group ID and metadata fields              
3. Simplify ACL table stages that were relevant in absense of a binding point   
4. Address scaling concerns in absense of a well defined binding point             
5. Introduce behavioral model for ACLs                                             
6. Adding tunnel and bridge interface specific ACL behavior

zubinshah: @itaibaz @lguohan @AshokDaparthi @atitjain : currently here is what I have listed in the updated specification and documentation. Lets discuss these details before the review so that we have clear agenda and asks if there needs to be a change or provision put in. But the changes for not addressing non-conflicting behavior are very specifically coming from the previous ACL review we had in AUG 2016, and listing them here: 

**

> ### ACL Table and ACL Group Match Behavior                                      
> Within one ACL GROUP TABLE only one ACL entry will be hit which is based on the table priority as well as ACL entry priority within the table. Since only on ACL Group (or ACL Table) can be associated with a bind point, it is clear that only one entry (if hit) has corresponding actions to be taken. SAI 1.0.0 does not support resolution of non-conflicting actions across various ACL Groups. However, considering the behavioral pipeline (model) for SAI 1.0.0, multiple bind points can be valid for a specific flow. There are use cases when multiple ACL Tables and/or ACL Groups are assigned at different bind points, but the very first bind point with a valid ACL entry (if hit - irrespective of an ACL Table or ACL Group) takes precedence over rest of the entries. This avoids resolving non-conflicting actions across various ACL Groups.

**

tushar-ty:
Its not very clear whats the functional outcome of binding to a lag or intf or tunnel ?
It seems its eventually going to act as a filter or qualifier just like for a port and vlan.
If not then whats the difference between binding to a port vs Lag ?
If yes then things become unclear for router interface and tunnel – as those objects are constructed out of several params/attributes – is the expectation to physically match all those attributes ?
Pls clarify.

-thanks

zubinshah:
To answer your question in brief "Whats the difference say between binding to a port vs a lag ?"
- when an ACL is applied to a port all traffic igress-ing into that port : goes through the filter checks (and corresponding actions)
- when an ACL is applied to a LAG all traffic ingressing on the LAG/BOND :  goes through the filter checks (and corresponding actions)
- same goes through other binding points : namely RIFs/TUNNELS/Etc.

Also, to clarify the abstract idea behind any bind point is to map them to the SAI pipeline’s logical stage. In the SAI logical pipeline (or even in a switch silicon’s pipeline) there is a point in the various stages where a packet can be clearly identified for services to be applied for :
- ports
- LAGs (for ingress lag properties or egress)
- RIFs (layer 3 interface properties , like v4/v6 routing enable or disable, etc..)
— and so on other interfaces like ingress tunnels, etc.

The idea that applications can simply create ACL tables (like in iptables, or openswitch CLI or cisco like NOSes)..
- access-lists 1 , with specific rules.
These access lists created can be bound to logical interfaces independent of any rules or match/key/mask they are programmed with . HENCE the packets ingressing (or egressing) through these logical interfaces would be subject to rules and match/action criteria specified in "access-list 1".

Something on the lines of cisco like configuration also mentioned in the openswitch forums as well :
- http://lists.openswitch.net/pipermail/ops-dev/2015-December/000659.html
- http://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus5000/sw/configuration/guide/cli/CLIConfigurationGuide/sec_ipacls.html


On Oct 13, 2016, at 10:11 AM, Tushar Tyagi <notifications@github.com<mailto:notifications@github.com>> wrote:


Its not very clear whats the functional outcome of binding to a lag or intf or tunnel ?
It seems its eventually going to act as a filter or qualifier just like for a port and vlan.
If not then whats the difference between binding to a port vs Lag ?
If yes then things become unclear for router interface and tunnel – as those objects are constructed out of several params/attributes – is the expectation to physically match all those attributes ?
Pls clarify.

-thanks

tushar-ty:
So you do agree that the lag binding point also creates a Lag value match|filter criteria, just like port - which would mean that compared to the existing spec this is a new filter - right.

I understand that you are describing things from the CLI/NOS point of view. But i'm trying to understand the details from the SAI/SDK/hw implementation point of view. And I need to know that when a ACL group is bound to another complex SAI object like a router intf or Tunnel what are the physical match criteria that get created. I think the spec does not clearly indicate this.

zubinshah:
Its not very clear whats the functional outcome of binding to a lag or intf or tunnel ?

>> Zubin : the functional or logical outcome of applying an ACL table to a lag or rif  or tunnel -> is to subject any flow or traffic specifically classified to be ingress on a LAG or RIF or TUNNEL be applied or subject to the ACL table (or GROUP) created by the user distinctly.

It seems its eventually going to act as a filter or qualifier just like for a port and vlan.

>> the difference between that we can enable applications and other ASIC architecture agnostic vendors to move the qualifiers out of the filter rules.

If not then whats the difference between binding to a port vs Lag ?

>> Biggest difference is to avoid the logistics of managing the port vs lag : states and awareness in the NOS. NOS can abstract security services using “interfaces” and let the SAI implementation handle these interface to ACL rule management. - SCALABILITY AND FLEXIBILITY IN SAI abstraction.

If yes then things become unclear for router interface and tunnel – as those objects are constructed out of several params/attributes – is the expectation to physically match all those attributes ?

>> To clarify : the packets classified to be ingressing on a RIF or TUNNELS are well defined in SAI as well as mostly all switch silicon. Since these objects may be constructed out of several parameters and attributes — in one silicon they can be completely be handled in a different fashion in another switch silicon. hence, it makes sense to abstractly define them in SAI abstract definition and let the users of SAI manage them flexibly .

tushar-ty:
I think this is a big problem if the behavior is ambiguous and the app does not know the outcome of a specific SAI API call. 

>If yes then things become unclear for router interface and tunnel – as those objects are constructed out of several params/attributes – is the expectation to physically match all those attributes ?
…

>> To clarify : the packets classified to be ingressing on a RIF or TUNNELS are well defined in SAI as well as mostly all switch silicon. Since these objects may be constructed out of several parameters and attributes — in one silicon they can be completely be handled in a different fashion in another switch silicon. hence, it makes sense to abstractly define them in SAI abstract definition and let the users of SAI manage them flexibly .

zubinshah:
Completely agree on the first part. “ lag binding point also creates a Lag value match|filter criteria, just like port” - But you see thats very hardware or implementation specific to your switch silicon. There are other switch silicon that may implement the same qualifiers in completely different fashion. And to let the SAI users expose to these level of details is not reasonable. Hence it is better to abstract general switch concepts - (interfaces) ports, lags, rifs, tunnels, bridge interface, etc. and handle them within SAI implementations. I will clarify those details in the specification, but i believe these abstract concepts are already part of SAI object starting from SAI 0.9.1 releases and some even before.

Looking at the spec before pre-1.0.0 as i mentioned in my presentation we never had any qualifiers. For eg. creating a port qualifier was never mandatory nor expected. So from that perspective all the attributes I added inside port, lag, rif, switch, tunnel, bridge interfaces are new.


When an ACL group or ACL table is bound to any complex (but well abstracted) SAI object — it definitely creates several physical match criteria - completely relevant for Broadcom ASIC and will be completely different in Cavium or other ASICs.

zubinshah:
@tushar : the behavior is very crisp and clear from a SAI modeling perspective. I would disagree that any of the below binding points are not well defined: 
	- Ports, LAGs, VLANs, Router Interfaces , Tunnel Termination and Tunnel Originations
	- I agree with the definition of bridge interfaces (but that spec is not yet discussed nor the code changes are in the pull request).

tushar-ty:
Ok, so the Lag binding point is very clear now.

I'm still struggling to get a clear picture from the new proposal about the complex objects like Router intf and Tunnel - I think we need to explicitly mention if 'all' the attributes of the particular object will be matched upon or only 'some' could be matched upon - depending on the ASIC. 
May be one of those is your assumption but not documenting this and leaving the behavior open ended is going to create confusion for the app writer and SAI implementer alike.

zubinshah:
I'm still struggling to get a clear picture from the new proposal about the complex objects like Router intf and Tunnel - I think we need to explicitly mention if 'all' the attributes of the particular object will be matched upon or only 'some' could be matched upon - depending on the ASIC.
May be one of those is your assumption but not documenting this and leaving the behavior open ended is going to create confusion for the app writer and SAI implementer alike.

This proposal does not introduce any new terminology or concepts within SAI. IMO RIFs and Tunnels are well defined objects that can be classified by the attributes these objects are defined - sairouterintf.h and saitunnel.h.

If identifying the router interface (and the flow) in a asic is well defined , i believe applying filters to those interfaces should , and applying filters to that object should be straight forward. I was assuming this approach for binding model - to rely on other specifications/and documents in SAI would have clarified those concepts - RIF and Tunnel. However, i will document this in the spec update.

inc/sailag.h
```diff
@@ -52,6 +52,26 @@ typedef enum _sai_lag_attr_t
...
+    /** READ_WRITE */
+
+    /** Bind/Unbind an ingress acl or acl-group on a lag [sai_object_id_t].
+     *      Enable (or bind) ingress ACL filtering by assigning the ACL table 
+     *      object id or ACL group object id as attribute value. 
+     *      Disable (or unbind) ingress ACL filtering by assigning 
+     *      SAI_NULL_OBJECT_ID in the attribute value.
+     *      Default Value: SAI_NULL_OBJECT_ID
+     */
+    SAI_LAG_ATTR_INGRESS_ACL_ID,
```
atitjain: No. This is the acl table-id to which the LAG table is a binding point..

inc/saiport.h
```diff
@@ -695,6 +695,28 @@ typedef enum _sai_port_attr_t
...
     SAI_PORT_ATTR_FDB_LEARNING_LIMIT_VIOLATION,
 
+    /** Bind/Unbind an ingress acl or acl-group on a port [sai_object_id_t].
+     *      Enable (or bind) ingress ACL filtering by assigning the ACL table 
+     *      object id or ACL group object id as attribute value. 
+     *      Disable (or unbind) ingress filtering by assigning 
+     *      SAI_NULL_OBJECT_ID in the attribute value.
+     *      Default Value: SAI_NULL_OBJECT_ID
+     */
+    SAI_PORT_ATTR_INGRESS_ACL_ID,
```
AshokDaparthi: As per review, if you are updating this as list of oid, please update document with 
- is oid list can have both ACL group OID and ACL table OID's? 
if yes, is there how the priority will work.

zubinshah: The OID list will contain both the ACL group OIDs and ACL table OIDs. The priorities of the ACL tables - outside of the group as well as part of the group should be resolved. considering that binding point. This may seem complex but there is no prioirty for aCL group, so ACL tables is all we need. And ofcourse resolving non-conflicting actions across different OID objects.

zubinshah: @AshokDaparthi - updated the document and review comments. see if your question gets clarified.

lguohan:
probably to simplify the binding points, we can remove the tunnel related binding because after tunnel decap or encap, the packet will go to either a router or a bridge for further processing in the logical pipeline. Thus, we can bind the acl to the ingress bridge port or router port after the tunnel to match those packets. See SAI_proposal_Bridge_port_v0.2.docx, figure 4.3.

zubinshah:
@lguohan updated the spec with 2 open action items from today's meeting. 
@xinliu-seattle @atitjain @itaibaz @mattyk @tushar-ty @AshokDaparthi - please review.

inc/saiacl.h
```diff
@@ -227,17 +241,47 @@ typedef enum _sai_acl_table_group_attr_t
...
+     */
+    SAI_ACL_TABLE_GROUP_ATTR_STAGE = SAI_ACL_TABLE_GROUP_ATTR_START,
+
+    /**
+     * @brief List of ACL bind points where this group will be applied
+     *
+     * @type sai_acl_bind_point_t 
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_ACL_TABLE_GROUP_ATTR_BIND_POINT,
```
AnbalaganNatchimuthu: @zubinshah  Please modify MANDATORY_ON_CREATE. This should be optional

inc/saiacl.h
```diff
@@ -227,17 +241,47 @@ typedef enum _sai_acl_table_group_attr_t
...
-     * @flags READ_ONLY
+     * Value must be in the range defined in
+     * [SAI_SWITCH_ATTR_ACL_TABLE_GROUP_MINIMUM_PRIORITY,
+     * SAI_SWITCH_ATTR_ACL_TABLE_GROUP_MAXIMUM_PRIORITY]
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      */
-    SAI_ACL_TABLE_GROUP_ATTR_DUMMY = SAI_ACL_TABLE_GROUP_ATTR_START,
+    SAI_ACL_TABLE_GROUP_ATTR_PRIORITY,
```
lguohan: why do we need this group attribute priority as all groups are parallel lookups?

inc/saiacl.h
```diff
@@ -1997,7 +2073,8 @@ typedef struct _sai_acl_api_t
     sai_remove_acl_range_fn             remove_acl_range;
     sai_set_acl_range_attribute_fn      set_acl_range_attribute;
     sai_get_acl_range_attribute_fn      get_acl_range_attribute;
-
+    sai_create_acl_table_group_fn       create_acl_table_group;
+    sai_remove_acl_table_group_fn       remove_acl_table_group;
```
AnbalaganNatchimuthu: @zubinshah @lguohan 

Consider an application which sends multiple rules dynamically and below is a combination we want to achieve. 

 MAC1  with priority P1 on port 1, port 2  
 MAC2 with priority P2 on port 2, port 3

Scenario 1 :

Create ACL table T1 – Qualifier Mac1, Priority 1 – Bind to ports P1 and P2

Create ACL table T2 – Qualifier Mac2, Priority 2 - Bind to ports P2 and P3

Scenario 2:

Create ACL table T1 – Qualifier Mac1, Priority 1 – Bind to port P1

Create ACL table T2 – Qualifier Mac2, Priority 2 - Bind to port P3

Create ACL table T3 with 2 entries – Qualifier Mac1, Priority 1 – Bind to port P2  ---- Here priority will become entry priority.
                                                                  Qualifier Mac2, Priority 2 

Scenario 3:

Create ACL table T3  with 2 entries  ---- Bind to switch bind point 
            Qualifier Mac1, Priority 1 –   Inports list will have ports P1 and P2                          
            Qualifier Mac2, Priority 2 – Inports list will have ports P2 and P3.

What is the recommendation from you for the above scenario? 
In the community meeting on Thursday Scenario #3 was recommended.   Please provide your thoughts on this.

AnbalaganNatchimuthu: SAI_ACL_TABLE_ATTR_SIZE  Is this the guaranteed size or the maximum size? Is it possible to have both  guaranteed and maximum size attribute for each table.

doc/SAI-Proposal-ACL-1.md
```diff
@@ -0,0 +1,385 @@
...
+    if (saistatus != SAI_STATUS_SUCCESS) {
+        return saistatus;
+    }
+    
+    // Create an ACL table *acl_table_id3* , to be part of this group *acl_grp_id1*
+    sai_object_id_t acl_table_id3 = 0ULL;
+    acl_attr_list[0].id = SAI_ACL_TABLE_ATTR_STAGE;
+    acl_attr_list[0].value.s32 = SAI_ACL_STAGE_INGRESS;
+
+    acl_attr_list[1].id = SAI_ACL_TABLE_GROUP_ATTR_BIND_POINT;
```
tushar-ty: Is this a typo - should be SAI_ACL_TABLE_ATTR_BIND_POINT ?

inc/saiacl.h
zubinshah: @AnbalaganNatchimuthu - done in the updated pull request.

inc/saiacl.h
```diff
@@ -265,6 +309,14 @@ typedef enum _sai_acl_table_attr_t
     SAI_ACL_TABLE_ATTR_STAGE = SAI_ACL_TABLE_ATTR_START,
 
     /**
+     * @brief List of ACL bind point where this ACL can be applied
+     *
+     * @type sai_acl_bind_point_t 
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
```
zubinshah: @AnbalaganNatchimuthu - done in the updated pull request. 
@itaibaz - done in the updated pull request.

inc/saiacl.h
zubinshah: @lguohan based on DELL's feedback the idea was to use this GROUP priority to resolve parallel lookups in some cases that ASIC may not support parallel lookups based on various rules. And in that case the list of GROUPs that are bound may need some priority. 

Also, i made it optional and not mandatory on create. What do you think?

inc/saiacl.h
```diff
@@ -205,10 +219,10 @@ typedef enum _sai_acl_action_type_t
     /** Set Meta Data to carry forward to next ACL Stage */
     SAI_ACL_ACTION_TYPE_SET_ACL_META_DATA,
 
-    /** Egress block port list */
+    /** egress block port list */
```
zubinshah: @lguohan - done in the updated pull request.

inc/saiacl.h
zubinshah: @AnbalaganNatchimuthu IMHO Scenario #1 and Scenarios #3 both should be supported by the ACL Model. I wouldn't recommend Scenario#2, however the ACL model does support that and as a community we should consider that scenario. 

From your examples of 2 rules bound to ports where i can see you really need port list, and you can use both #1 and #3. However, the value addition of the new acl model is when you scale the number of rules beyond ports when there are ACL tables you create for routing interfaces and you will realize port list is not sufficient when you have both port based routing interfaces and VLAN based router interfaces -- where Scenario#3 is not scalable for you. And Scenario #1 is really really useful to scale your application and handle the complexity in the SAI implementation.

doc/SAI-Proposal-ACL-1.md
zubinshah: @tushar-ty based on the updated pull request, it should be "SAI_ACL_TABLE_GROUP_ATTR_BIND_POINT".

inc/saiacl.h
AshokDaparthi: @zubinshah - We discussed internally, if there is not use case of others, you can remove priority. We can resolve the priority based on TABLES always

inc/saiacl.h
AnbalaganNatchimuthu: @zubinshah Few applications would need a dedicated size for them to work. If the size is treated as maximum then such applications would have issues. Can we treat this size as guaranteed?

zubinshah:
Thanks. We will revisit this when we relook at the group concept.
Seems no one needs the group concept (which was existing in the previous spec, we can completely move it out).

On Nov 9, 2016, at 10:41 PM, Ashok Daparthi <notifications@github.com<mailto:notifications@github.com>> wrote:


@AshokDaparthi commented on this pull request.

In inc/saiacl.h<https://github.com/opencomputeproject/SAI/pull/171>:

>       */
-    SAI_ACL_TABLE_GROUP_ATTR_DUMMY = SAI_ACL_TABLE_GROUP_ATTR_START,
+    SAI_ACL_TABLE_GROUP_ATTR_PRIORITY,


@zubinshah<https://github.com/zubinshah> - We discussed internally, if there is not use case of others, you can remove priority. We can resolve the priority based on TABLES always

inc/saiacl.h
```diff
@@ -227,17 +241,36 @@ typedef enum _sai_acl_table_group_attr_t
     SAI_ACL_TABLE_GROUP_ATTR_START,
 
     /**
-     * @brief Dummy attribute
+     * @brief ACL stage
+     *
+     * @type sai_acl_stage_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_ACL_TABLE_GROUP_ATTR_ACL_STAGE = SAI_ACL_TABLE_GROUP_ATTR_START,
```
lguohan: can we add SAI_ACL_TABLE_GROUP_ATTR_TYPE attribute?

The group type can be an enum 
{

acl_table_group_type_parallel 

acl_table_group_type_sequential
}

In paralled, all acl tables in the group are looked up in parallel, and non conflicting actions are taken from multiple matched acl entries.

In sequential mode, priority is defined between the acl tables, and only one acl entry is matched.

inc/saiacl.h
```diff
@@ -1985,27 +2027,80 @@ typedef sai_status_t (*sai_get_acl_range_attribute_fn)(
...
+    sai_set_acl_counter_attribute_fn        set_acl_counter_attribute;
+    sai_get_acl_counter_attribute_fn        get_acl_counter_attribute;
+    sai_create_acl_range_fn                 create_acl_range;
+    sai_remove_acl_range_fn                 remove_acl_range;
+    sai_set_acl_range_attribute_fn          set_acl_range_attribute;
+    sai_get_acl_range_attribute_fn          get_acl_range_attribute;
+    sai_create_acl_table_group_fn           create_acl_table_group;
+    sai_remove_acl_table_group_fn           remove_acl_table_group;
+    sai_set_acl_table_group_attribute_fn    set_acl_table_group_attribute;
+    sai_get_acl_table_group_attribute_fn    get_acl_table_group_attribute;
```
lguohan: since the relation between acl_table_group and acl_table is similar to nexthopgroup and nexthop, we also need similiar API as nexthgroup.

In this case, we also need bring acl_table_group_member concept. 

We need create/remove/set/get APi for acl_table_group_member so that we can add/remove the ACL table to ACL table group.

lguohan:
> 

I am thinking to introduce an acl table group member object, and move these two attributes into the member object.

here is how the API works:

1. create_acl_table_group.
2. create_acl_table
3. create_acl_table_group_member

In acl_table_group_member, you specify the acl_table_id, and acl_table_group_id.

The group_member has an attribute PRIORITY which is to define the priority of acl table within the group, and it is only valid when the group type is sequential.



---
Refers to: inc/saiacl.h:389 in f9f5f31. [](commit_id = f9f5f31dc82c407bf9e1fd936794000917fcbed9, deletion_comment = False)

inc/saiacl.h
zubinshah: @lguohan - done. pls see the most updated commit to the PR.

zubinshah:
@lguohan -- appreciate the feedback on group member apis. i've incorporated these changes with your suggestions, please see the updated pull request and specifically the following commit : *https://github.com/zubinshah/SAI/commit/59c8537cc3cc2c6b2af26a44518b331583177804

doc/SAI-Proposal-ACL-1.md
```diff
@@ -0,0 +1,362 @@
...
+
+Figure 1 shows the relationship between an ACL table and various bind points. Examples 1 and 2 in the subsequent sections show how to bind/unbind an ACL table.<br>
+
+![SAI acl design](figures/sai_aclobjs.png "Figure 1: Relationship between ACL table ID and various binding points.")
+__Figure 1: Relationship between ACL table ID and various binding points.__<br>
+
+## ACL group management 
+Grouping of ACL tables was supported in the previous version. This specification enhances the ACL table grouping model and primarily the management of ACL groups - for scalability and flexibility. Two new APIs are introduced in saiacl.h object to manage group ID creation and removal. create_acl_table_group - allocates a unique object ID based ACL group table ID and remove_acl_table_group API removes the unique object ID for recycle. This proposal introduces a new ACL group object. ACL groupid is bound to all the previously discussed bind points that are valid for ACL tables.<br>
```
lguohan: can you update the description for ACL group management? an ACL group is a group of ACL tables that can be applied to the bind points. There are various types of ACL groups which defines the packet matching behavior across the ACL tables within one group. 

Current two group types are defined, sequential and parallel. In sequential mode, each ACL table is assigned with a unique priority within that group. With a packet matches multiple ACL entries within a group, only the one with highest table priority within the group and highest entry priority within the table wins. In parallel mode, all ACL tables are matches and non-conflicting actions are executed.

inc/saiacl.h
```diff
@@ -44,15 +44,29 @@ typedef enum _sai_acl_stage_t
     /** Egress Stage */
     SAI_ACL_STAGE_EGRESS,
 
-    /** Ingress Sub Stages */
+} sai_acl_stage_t;
+
+/**
+ * @brief Attribute data for SAI_ACL_TABLE_ATTR_BIND_POINT
+ */
+typedef enum _sai_acl_bind_point_t
```
lguohan: is it better to rename it as _sai_acl_bind_point_type_t, bind_points are the instances such as rif id, port id, bind_point_type is the type of the bind points.

doc/SAI-Proposal-ACL-1.md
```diff
@@ -0,0 +1,362 @@
...
+        SAI_ACL_BIND_POINT_VLAN,
+    
+        /** RIF Bind Point */
+        SAI_ACL_BIND_POINT_ROUTER_INTF,
+    
+        /** Port Bind Point */
+        SAI_ACL_BIND_POINT_SWITCH
+    } sai_acl_bind_point_t;
+
+## Metadata Usage Model
```
lguohan: not sure if we need this section here. this seems to be an independent topic and I cannot find it mentioned in the introduction.

doc/SAI-Proposal-ACL-1.md
```diff
@@ -0,0 +1,362 @@
...
+    acl_attr_list[1].value.objlist.count = 1;
+    acl_attr_list[1].value.objlist.list[0] = SAI_ACL_BIND_POINT_PORT;
+
+    acl_attr_list[2].id = SAI_ACL_TABLE_ATTR_PRIORITY;
+    acl_attr_list[2].value.s32 = 101;
+    acl_attr_list[3].id = SAI_ACL_TABLE_ATTR_FIELD_SRC_MAC;
+    acl_attr_list[3].value.booldata = True;
+    
+    acl_attr_list[4].id = SAI_ACL_TABLE_ATTR_GROUP_ID;
+    acl_attr_list[4].value.oid = acl_grp_id1;
```
lguohan: can you update the example based on the new header files?

inc/saiacl.h
```diff
@@ -227,21 +254,149 @@ typedef enum _sai_acl_table_group_attr_t
...
+
+    /**
+     * @brief List of ACL bind points where this group will be applied
+     *
+     * @type sai_s32_list_t sai_acl_bind_point_t
+     * @flags CREATE_ONLY
+     * @default empty
      */
-    SAI_ACL_TABLE_GROUP_ATTR_DUMMY = SAI_ACL_TABLE_GROUP_ATTR_START,
+    SAI_ACL_TABLE_GROUP_ATTR_ACL_BIND_POINT_LIST,
```
lguohan: I do not know if we need this or not since ACL_TABLE has the bind points. Assume you want to bind an ACL table group to a bind point, then all the acl tables should include that bind point as their types.

doc/SAI-Proposal-ACL-1.md
```diff
@@ -0,0 +1,362 @@
...
+
+This specification proposes the following enhancements and changes:<br>
+1. Introduce the concept of binding points for ACLs.<br>
```
zubinshah: done.

doc/SAI-Proposal-ACL-1.md
zubinshah: Will do so in the updated pull request.

inc/saiacl.h
zubinshah: @lguohan as we discussed - this will simplify the implementation. ACL groups act as a container where we can hold various acl tables. The user's intention to simply specifiy what bind points any specific acl group is going to be associated -- at creation time gives a better error checking/handling when new acl tables (via acl group members) are added to the groups. It makes the sai implementation management simple.

zubinshah:
@lguohan @atitjain - please see the updated pull request . all Guohan's comments and other open comments have been addressed. 

Please verify the final approval/acceptance in the 12/1 weekly syncup/call. 

Next steps for Cavium would be to add PTF test cases that are in the Examples section of the proposal, as well as few more that exercise and provides a compliance for the ACL model.

lguohan:
:shipit:

inc/saiswitch.h
```diff
@@ -471,6 +487,37 @@ typedef enum _sai_switch_attr_t
...
+     * ingress ACL table or ACL group filtering by assigning the list of valid 
+     * object id . Disable ingress filtering by assigning SAI_NULL_OBJECT_ID 
+     * in the attribute value. 
+     *
+     * @type sai_object_id_t
+     * @objects SAI_OBJECT_TYPE_ACL_TABLE, SAI_OBJECT_TYPE_ACL_TABLE_GROUP
+     * @flags CREATE_AND_SET
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_SWITCH_ATTR_DEFAULT_INGRESS_ACL,
```
AshokDaparthi: SAI_SWITCH_ATTR_DEFAULT_INGRESS_ACL --> SAI_SWITCH_ATTR_INGRESS_ACL can make it similar to other objects

inc/saiacl.h
```diff
@@ -227,21 +254,155 @@ typedef enum _sai_acl_table_group_attr_t
...
+     *
+     * Without grouping, all non-conflicting actions from different ACL entry
+     * hits across the tables will be performed.
+     *
+     * @type sai_object_id_t
+     * @objects SAI_OBJECT_TYPE_ACL_TABLE_GROUP
+     * @flags CREATE_ONLY
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_ACL_TABLE_GROUP_MEMBER_ATTR_ACL_TABLE_GROUP_ID = SAI_ACL_TABLE_GROUP_MEMBER_ATTR_START,
```
AshokDaparthi: Why it is different from other member objects example

 @flags MANDATORY_ON_CREATE | CREATE_ONLY
     */
    SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_GROUP_I

 @flags MANDATORY_ON_CREATE | CREATE_ONLY
     */
    SAI_VLAN_MEMBER_ATTR_VLAN_ID

 it should be MANDATORY_ON_CREATE since object we are creating here is group member.

inc/saiswitch.h
zubinshah: fixed. thank you

inc/saiacl.h
zubinshah: done, thanks.

PR #173, opened 2016-08-17
L3VlanNeighborMacUpdate test added

engdorm:
added unknown unicast test in sail2.py
and added counter constants to switch.py


test/saithrift/tests/sail2.py
```diff
@@ -20,6 +20,27 @@
 import sai_base_test
 
 @group('l2')
+class UnknownUnicastTest(sai_base_test.ThriftInterfaceDataPlane):
+    def runTest(self):
```
stcheng: this test need to be applied on vlan. i've just checked the previous BcastUcastTest and i think this test can be appended at the previous test.
try:
send packet with the dst mac not in the FDB table
verify no ports received that packet
check the counters_egress_drop == 1

test/saithrift/tests/sail2.py
```diff
@@ -20,6 +20,27 @@
...
+        print
+        print "Sending L2 packet to port 1 with unknown-unicast"
+        switch_init(self.client)
+        sai_thrift_clear_all_counters(self.client)
+        in_port = 0
+        pkt = simple_tcp_packet(eth_dst='00:11:11:11:11:11',
+                                eth_src='00:22:22:22:22:22',
```
lguohan: >00:22:22:22:22:22 [](start = 41, length = 17)

can you call remove_fdb() API to remove the fdb entry first, to make sure there is no such fdb entry in the asic?

test/saithrift/tests/sail2.py
```diff
@@ -20,6 +20,27 @@
...
+                                ip_dst='10.0.0.1',
+                                ip_id=101,
+                                ip_ttl=64)
+
+        send_packet(self, in_port, str(pkt))
+        verify_no_other_packets(self)
+        for i in range(len(port_list)):
+            if (i ==  in_port): continue
+            port_counters, queue_counters = sai_thrift_read_port_counters(self.client, port_list[i])
+            assert(port_counters[COUNTERS_EGRESS_DROP] == 1)
```
lguohan: > assert(port_counters[COUNTERS_EGRESS_DROP] == 1) [](start = 11, length = 49)

this will be tricky, not sure if all platforms treat such drop as egress drop. I think it's ok not to read counters in the tests.

test/saithrift/tests/sail2.py
lguohan: I agree this should be applied in the vlan context, but I suggest to keep it as a separate test since the verification is different.

Another thing to do is to set SAI_SWITCH_ATTR_FDB_UNICAST_MISS_ACTION = DROP.

---
In reply to: [75367140](https://github.com/opencomputeproject/SAI/pull/173#discussion_r75367140) [](ancestors = 75367140)

test/saithrift/tests/switch.py
```diff
@@ -51,6 +51,21 @@
...
+COUNTERS_PFC_PRIO_1 = 3
+COUNTERS_PFC_PRIO_2 = 4
+COUNTERS_PFC_PRIO_3 = 5
+COUNTERS_PFC_PRIO_4 = 6
+COUNTERS_PFC_PRIO_5 = 7
+COUNTERS_PFC_PRIO_6 = 8
+COUNTERS_PFC_PRIO_7 = 9
+COUNTERS_TRANSMITTED_OCTETS = 10
+COUNTERS_TRANSMITTED_PKTS = 11
+COUNTERS_RECEIVED_PKTS = 12
```
lguohan: are these counters used? can we remove them if not used.

engdorm: I can delete it and it still gonna work
but if in the future someone will want to use the read counter function 
(sai_thrift_read_port_counters)
it's return an array, and the use of const will make it easy to understand the array fields

*when I run the test the first time 
I wanted to see in my eyes the actual numbers of the counters

test/saithrift/tests/sail3.py
```diff
@@ -1214,3 +1214,116 @@ def runTest(self):
...
+                                eth_src=router_mac,
+                                ip_dst=ip_port1,
+                                ip_src='11.11.11.1',
+                                dl_vlan_enable=True,
+                                vlan_vid=vlan_id,
+                                ip_id=105,
+                                ip_ttl=63)
+            send_packet(self, 2, str(L3_pkt))
+            verify_packets(self, exp_pkt, [0])
+            verify_no_other_packets(self)
```
lguohan: No need for verify_no_other_packets

verify_packets :     """
    Check that a packet is received on each of the specified port numbers for a
    given device (default device number is 0).
    Also verifies that the packet is not received on any other ports for this
    device, and that no other packets are received on the device (unless --relax
    is in effect).

test/saithrift/tests/sail3.py
```diff
@@ -1214,3 +1214,116 @@ def runTest(self):
...
+                                ip_src=ip_port1,
+                                ip_id=102,
+                                ip_ttl=64)
+        
+
+        try:
+            #sending unkown UC for learning the ports mac and expecting flooding only on the vlan
+            send_packet(self, 0, str(local_pkt))
+            verify_packets(self, local_pkt, [1])
+            verify_no_other_packets(self)
```
lguohan: no need for verify_no_other_packets

test/saithrift/tests/sail3.py
```diff
@@ -1214,3 +1214,114 @@ def runTest(self):
...
+        vr_id = sai_thrift_create_virtual_router(self.client, v4_enabled, v6_enabled)
+        rif_vlan_id = sai_thrift_create_router_interface(self.client, vr_id, 0, 0, vlan_id, v4_enabled, v6_enabled, mac1)
+        rif_port_id = sai_thrift_create_router_interface(self.client, vr_id, 1, port3, 0, v4_enabled, v6_enabled, mac2)
+
+        attr_value2 = sai_thrift_attribute_value_t(s32=SAI_PORT_LEARN_MODE_HW)
+        attr2 = sai_thrift_attribute_t(id=SAI_PORT_ATTR_FDB_LEARNING, value=attr_value2)
+        self.client.sai_thrift_set_port_attribute(port1, attr2)
+        self.client.sai_thrift_set_port_attribute(port2, attr2)
+        
+        sai_thrift_create_neighbor(self.client, addr_family, rif_vlan_id, ip_port1, mac_port1)
```
nazarii-gnydyn: I see a difference between the testcase summary and code implementation. Why the neighbor and route are created before sending packet to update the FDB entry? Could you provide some explanations?

PR #177, opened 2016-08-23
Clarify several tunnel attributes

lguohan:


inc/saitunnel.h
```diff
@@ -222,10 +224,12 @@ typedef enum _sai_tunnel_attr_t
     /** tunnel type [sai_tunnel_type_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
     SAI_TUNNEL_ATTR_TYPE = SAI_TUNNEL_ATTR_START,
 
-    /** tunnel underlay interface [sai_object_id_t] */
+    /** tunnel underlay interface [sai_object_id_t] (MANDATORY_ON_CREATE|CREATE_ONLY)
```
erohsik, line 227: @lguohan If the tunnel is only used for Encap or Decap, does it make sense to relax the Mandatory on create requirement for Overlay and Underlay RIFs?

kcudnik: why enum specifies MANDATORY_ON_CREATE|CREATE_ONLY which is only available on attributes

erohsik: The current attribute looks like this.. [(https://github.com/opencomputeproject/SAI/blob/master/inc/saitunnel.h#L449)](https://github.com/opencomputeproject/SAI/blob/master/inc/saitunnel.h#L449)

    /**
     * @brief Tunnel underlay interface
     *
     * Underlay interface is router interface when conditions are met.
     *
     * @type sai_object_id_t
     * @condition SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_GRE or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_VXLAN or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_SRV6 or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_NVGRE or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_ESP or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_UDP_ESP or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_VXLAN_UDP_ESP
     */
    SAI_TUNNEL_ATTR_UNDERLAY_INTERFACE,

I believe this PR#177 introduced the flags, hence left a comment here..

kcudnik: ahh, missted that ,this is from 2016 :D

PR #179, opened 2016-08-23
Added switch ID to all Objects

AshokDaparthi:


inc/saiswitch.h
```diff
@@ -188,8 +188,68 @@ typedef enum _sai_switch_attr_t
...
 
+    /** Handle for the switch profile.
+     * sai_switch_profile_id_t of size _In_reads_z_(SAI_MAX_HARDWARE_ID_LEN) 
+     * (MANDATORY_ON_CREATE|CREATE_ONLY) */
+    SAI_SWITCH_ATTR_PROFILE_ID = SAI_SWITCH_ATTR_START,
+
+    /** Switch hardware ID to open 
+     * sai_string_t of size _In_reads_z_(SAI_MAX_HARDWARE_ID_LEN)
+     * (MANDATORY_ON_CREATE|CREATE_ONLY) */
+    SAI_SWITCH_ATTR_HARDWARE_INFO,
```
lguohan: add more description for the HARDWARE_INFO format

inc/saiswitch.h
```diff
@@ -188,8 +188,68 @@ typedef enum _sai_switch_attr_t
...
+
+    /** Vendor specific path name of the firmware to load
+     * sai_string_t of size _In_reads_z_(SAI_MAX_FIRMWARE_PATH_NAME_LEN)
+     * (MANDATORY_ON_CREATE|CREATE_ONLY) */
+    SAI_SWITCH_ATTR_FIRMWARE_PATH_NAME,
```
lguohan: is this mandatory what if the npu does not need a firmware path?

inc/saiswitch.h
```diff
@@ -188,8 +188,68 @@ typedef enum _sai_switch_attr_t
...
+     * After this call the capability attributes should be ready for retrieval
+     * via sai_get_switch_attribute()
+     * Defaults to True */
+    SAI_SWITCH_ATTR_INIT_SWITCH,
+
+    /** Switch oper status change notification callback function passed to the adapter.
+     *  (CREATE_AND_SET) 
+     *  [sai_switch_state_change_notification_fn as sai_pointer_t]
+     **/
+    SAI_SWITCH_ATTR_OPER_STATUS_CHANGE_NOTIFY,
```
lguohan: default is null.

inc/saiacl.h
```diff
@@ -221,11 +221,15 @@ typedef enum _sai_acl_table_attr_t
     
     /** READ-ONLY */
 
+    /** Switch Object ID [sai_object_id_t] (CREATE_ONLY),
+     * Default SAI_NULL_OBJECT_ID */
```
lguohan: per discussion, it is mandatory on create

kcudnik: Maybe its more convinient to use SAI_NULL_OBJECT_ID as default switch id, but there should be api that will get default switch ID on sai_switch_api->get_switch_attribute(0, 1, &attr) where attr.id = SAI_SWITCH_DEFAULT_SWITCH_ID; since when multiple switches are present there should be way to get their id, and query type of that object should be SAI_OBJECT_TYPE_SWITCH; 

IMHO SAI_NULL_OBJECT_ID should only be used to disable some functionality that uses object id, instead of pointing to "default" instance of some object. A good example here is DEFAULT_VIRTUAL_ROUTER, when creating ROUTE we don't set vr_id to NULL but to actual object id, this helps to track object dependencies without any special corner cases logic

So for all objects Default should be to SAI_SWITCH_DEFAULT_SWITCH_ID instead of SAI_NULL_OBJECT_ID 

only inconvinience here is that 
 sai_switch_api->get_switch_attribute(0, 1, &attr)  uses "zero" to get default_switch_id

also IMHO since we are moving to sai_switch_create() api, there should be no default switch, user should be obligated to create it and then get proper switch id, then all that "default" switch id would be gone

lguohan: agree that use SAI_NULL_OBJECT_ID to refer the default switch id is not a good idea. The proposal is to make this attribute as mandatory_on_create, the use must supply a value, which is returned by sai_create_switch() API.

inc/saiswitch.h
AshokDaparthi: Changed  as optional attribute.

inc/saiacl.h
AshokDaparthi: Modified as Mandatory.

mikelazar:
As discussed during the last meeting (Thur Sep 1st), we should have a 'switch ID' function argument for all create function. Here, the switch id is the SAI object ID returned by the switch object create function. 

This avoids the need to add new attributes, and makes the switch ID mandatory in an explicit and safer manner, since it is verifiable at compile-time vs. run-time.

For instance:

typedef sai_status_t (*sai_create_port_fn)(
    **_In_ sai_object_id_t switch_id,**
    _Out_ sai_object_id_t* port_id,
    _In_ uint32_t attr_count,
    _In_ const sai_attribute_t *attr_list
    );

Note that the returned OID ("port_id", in this case) must include the identity of the switch ID to which the port_id belongs to, thus only create functions need the separate switch ID parameter.

kcudnik:
What about create_switch function? Will this lack the switch_id ?

mikelazar:
The decision of OCP SAI meeting on Thursday Sep 22 was to:

1. An explicit function parameter “switch_id” is added to all ‘create’ functions – except ‘sai_create_switch_fn‘. The ‘switch_id’ is the value returned by ‘sai_create_switch_fn‘. 

The OID returned by ‘create’ functions must embed ‘switch_id’ information.

Example:
typedef sai_status_t (*sai_create_port_fn)(
    _Out_ sai_object_id_t* port_id, // port_id must embed 'switch_id' information
    _In_ uint32_t attr_count,
    _In_ const sai_attribute_t *attr_list
    );

Note. There is no need to add extra 'SWITCH_ID' attributes to SAI objects (e.g. SAI_PORT_ATTR_SWITCH_ID is not needed)

2. ‘set’, ‘get’, ’delete’ functions are unmodified.
All SAI OID’s embed switch ID information

3. A function ‘sai_get_switch_id’ allows callers to retrieve the switch_id that corresponds to any given OID

sai_status_t sai_get_switch_id(
        _Out_ sai_object_id_t* switch_id,
        _In_ sai_object_id_t  obj_id,
        );


The API changes are going to be updated accordingly.

PR #184, opened 2016-08-25
Change buffer profile threshold mode to enum

kcudnik:
I'n not sure but SAI_BUFFER_PROFILE_ATTR_TH_MODE should probably be CREATE_ONLY


lguohan:
@kcudnik , can you update this PR?

inc/saibuffer.h
```diff
@@ -349,9 +365,9 @@ typedef enum _sai_buffer_profile_attr_t
      * If not set, use #SAI_BUFFER_POOL_ATTR_TH_MODE.
      * (default to value set in #SAI_BUFFER_POOL_ATTR_TH_MODE)
```
lguohan: remove this line as it conflicts with the below value.

kcudnik: i see no conflicts ?

lguohan: I mean the default value should be SAI_BUFFER_PROFILE_THRESHOLD_MODE_INHERIT_BUFFER_POOL_MODE

kcudnik: but it is, line 370

kcudnik: ahh, in comments, reoved

kcudnik:
rebuild

PR #185, opened 2016-08-26
Unify fdb, route and neighbor to quad API

kcudnik:


AshokDaparthi:
As part of this please make saibuffer.h ingress priority groups and  saivlan.h also as object model.

atitjain:
Some thoughts about this: 
For forwarding entries (flow entries) this looks like an overhead from the application/stack perspective. Typically IP and MAC addresses are directly provided by CLI or the protocol, now if we have oid for them in SAI, then application needs to maintain states to map MAC/IP etc back to oid. 
What is the motivation here, I understand quad API is one but its already quad. Retrieval of objects for Warmboot etc could be another but I think that should be manageable w/o this change.

kcudnik:
@AshokDaparthi : I though vlan was already converted to object_id_t, but if not i can add it here also, i think @lguohan was working on that, i can't find it right now, but its not a problem. I will sync with him.
As for buffer it's already converted to sai_object_id_t on master branch.

@atitjain: yes you are correct, and this change for those object is a little overhead, and yes sai would need to keep mapping inside to find corresponding fdb. But the same thing exist in case of vlan_member or lag_member, sai also needs to keep mapping for them as well, and those objects are not referenced any where else since they are leafs (or roots, depends how you look at the tree) in object dependency tree.
Idea behind this is to unify quad API to have the same signature for all corresponging 4 functions across all objects used in SAI. This will make a verry simple interface for users to produce any generic code like for validation, object reference counting, creation, remove, set, get, and many others. I'm working on sairedis code and i always need to keep special cases for those objects.
I maybe don't see whole picture but from software perspective this is very simple use of hash/map/dictionaly for lookup for entry.

atitjain:
@kcudnik, some of my thoughts. 
For VLAN and LAG the scaling is not as high. VLAN is 4K space, lag may be even smaller. FDB and routes are bigger tables and may run into 100s of K. OID model requires this state to be replicated in both Host Adapter (kye -> oid) and SAI adapter (oid->).  SAI implementation can always have a static formula to map a vlan to a oid so doesn't need a state, in case of lag/port/tunnel etc, the id itself is used as the key to refer or program entries. And from the control plane perspective, it uses the id for such objects. But for MAC and IP typically control plane use the key directly.

kcudnik:
@atitjain, yes agree on your explanation, that this mapping on vlan, port and lag can be encoded directly inside oid itself, but from SAI user perspective it does not matter how this is encoded, user just uses oid, i have only little background on ASIC stuff, but it still seems to me that you can have this mapping inside SAI, mapping will be only on sdk side. Also for other object you need to have "state" inside SAI anyway (i suspect) since probably there is no 1:1 mapping between SAI objects and asic hw.
Good argument here is that routes, fdb and neighbors will be probably largest sai objects groups.

Counting roughly 10k routes * (16 bytes ipv6 + 16 bytes maskv6 + 8 bytes virtual router id + some hash ovehead) = ~ 500kb, yea maybe that's too much for map only? But it would made sai interface so much beautifull having all the same functions signatures across all objects :)

kcudnik:
Im closing this PR after this discussion.

PR #192, opened 2016-08-30
Change acl table group attributes to list

kcudnik:
Also a clarification will be needed what kind of object type can appear on this list.
I'm not sure if this change will be valid if those objects on list can be different types.


lguohan:
can this be dropped?

kcudnik:
For now i think yes, we can discuss those user attributes

PR #214, opened 2016-09-14
Update saihostintf.h

tushar-ty:
Add outbound cos queue number selection attribute.

### DIFF
--- a/inc/saihostintf.h
+++ b/inc/saihostintf.h
@@ -527,7 +527,8 @@ typedef enum _sai_hostif_attr_t
     /** Set the operational status for this host interface [bool] (default to false) */
     SAI_HOSTIF_ATTR_OPER_STATUS,
     
-    /** Set the cos queue number to be used for packets going out through this interface */
+    /** Set the cos queue number to be used for packets going out through this interface [uint32_t] 
+    * (default 0) This is an index. */
     SAI_HOSTIF_ATTR_COS_QUEUE,
 
     /* --*/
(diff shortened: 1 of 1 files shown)


inc/saihostintf.h
```diff
@@ -526,6 +526,9 @@ typedef enum _sai_hostif_attr_t
 
     /** Set the operational status for this host interface [bool] (default to false) */
     SAI_HOSTIF_ATTR_OPER_STATUS,
+    
+    /** Set the cos queue number to be used for packets going out through this interface */
+    SAI_HOSTIF_ATTR_COS_QUEUE,
```
kcudnik: Should that be queue index ? or actual QUEUE object ID? Please provide value type.

tushar-ty: Yes, its an index not an object. Will update.

inc/saihostintf.h
```diff
@@ -526,6 +526,10 @@ typedef enum _sai_hostif_attr_t
 
     /** Set the operational status for this host interface [bool] (default to false) */
     SAI_HOSTIF_ATTR_OPER_STATUS,
+    
+    /** Set the cos queue number to be used for packets going out through this interface [uint32_t] 
+    * (default 0) This is an index. */
+    SAI_HOSTIF_ATTR_COS_QUEUE,
```
SathishKumarTR-Dell: Will this mean the same cos queue is used for all packets? There is a new attribute SAI_PACKET_EGRESS_QUEUE proposed in the Host IF enhancements proposal (SAI-Proposal-4-host-if-ver7.docx) which would solve this requirement?

lguohan:
test

lguohan:
@tushar-ty , the build fails due to missing meta data for this attribute. Can you help add the meta data for this attribute?

PR #219, opened 2016-09-15
Adding New action to block learning per flow

AshokDaparthi:


AshokDaparthi:
@itaibaz  @tushar-ty @atitjain Please review

inc/saiacl.h
```diff
@@ -211,6 +211,9 @@ typedef enum _sai_acl_action_type_t
     /** Set User Defined Trap ID */
     SAI_ACL_ACTION_TYPE_SET_USER_TRAP_ID,
 
+    /** Set Do Not Learn Action */
+    SAI_ACL_ACTION_TYPE_SET_DO_NOT_LEARN,
```
atitjain: What happens to the packet, does it get flooded to the vlan if the rule is hit with this action?

PR #226, opened 2016-09-20
Proposal/pipeline/headerfile for multicast

sonicyao1985:
Proposal/pipeline/headerfile for multicast


sonicyao1985:
Conflict with PR 208, will resolve it and make a new PR.

PR #230, opened 2016-09-22
New test added 

engdorm:
new test added to test lag with different port numbers (16 to 1) through L3


test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,159 @@ def runTest(self):
...
+        """
+       
+          
+        print
+        print "L3MultipleLagTest"
+        #general configuration 
+        random.seed(1)
+        switch_init(self.client)
+        self.src_port = port_list[16]
+        if (len(port_list) < 17 ) : return
```
lguohan: print message like "skip this test as it requires 17 ports"

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,159 @@ def runTest(self):
...
+        pkt_counter = [0] * 16
+        destanation_ports = range(16)
+        sport = 0x1234
+        dport = 0x50
+        IP_LAST_WORD_RANGE = 254
+        IP_2ND_LAST_WORD_RANGE = 16
+        for i in xrange(IP_LAST_WORD_RANGE):
+                for j in xrange(IP_2ND_LAST_WORD_RANGE):
+                    ip_src = '10.0.' + str(j) + '.' + str(i)
+                    src_mac = self.dataplane.get_mac(0, 16)
```
lguohan: this get mac only needs to be run once, no need to run every time you send a packet.

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,159 @@ def runTest(self):
...
+               '00:11:37:33:44:65',
+               '00:11:38:33:44:66']
+    
+    
+    def config_for_lags(self, num_of_lags, port_list):
+        print port_list
+        mac =''
+        for i in xrange(num_of_lags):
+            self.lags.append(self.client.sai_thrift_create_lag([]))
+        for i in xrange(16):
```
lguohan: pass 16 as parameter for setup_lags(self, number_of_lags, total_port_num, port_list)

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,159 @@ def runTest(self):
...
+            self.lag_members.append(sai_thrift_create_lag_member(self.client, self.lags[i % num_of_lags], port_list[i]))
+        for i in xrange(num_of_lags):
+            self.lags_rifs.append(sai_thrift_create_router_interface(self.client, self.vr_id, 1, self.lags[i], 0, self.v4_enabled, self.v6_enabled, mac))
+        for i in xrange(16):
+            sai_thrift_create_neighbor(self.client, self.addr_family, self.lags_rifs[i%len(self.lags_rifs)], "10.10.%s.1" % str(i+1), self.mac_pool[i])
+            sai_thrift_create_route(self.client, self.vr_id, self.addr_family, "10.10.%s.0" % str(i+1), self.ip_mask, self.lags_rifs[i% len(self.lags_rifs)])
```
lguohan: there is one neighbor for one rif, and one directly reachable route for one neighbor. the range should be number_of_lags, not 16.

lguohan: the whole process can be like following
member_num = total_port/num_of_lags
for i in range(num_of_lags):
   create_lag()
   for j in range(i*member_num, (i+1)*member_num)
   create_lag_member()
   create_rif()
   create_neighbor()
   create_route()

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,159 @@ def runTest(self):
             attr = sai_thrift_attribute_t(id=SAI_PORT_ATTR_PORT_VLAN_ID, value=attr_value)
             self.client.sai_thrift_set_port_attribute(port1, attr)
             self.client.sai_thrift_set_port_attribute(port2, attr)
-
+@group('lag')            
+@group('l3')
+class L3MultipleLagTest(sai_base_test.ThriftInterfaceDataPlane):
+v4_enabled = 1
```
lguohan: space here?

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,159 @@ def runTest(self):
...
+    v6_enabled = 1
+    ip_mask = '255.255.255.0'
+    addr_family = SAI_IP_ADDR_FAMILY_IPV4
+    lag_members = []
+    lags = []
+    lags_rifs = []
+    neighbors = []
+    routes = []
+    vr_id = 0
```
lguohan: can we add total_lag_port = 16, and use total_lag_port in the reset of the program instead of hardcoded number 16?

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,159 @@ def runTest(self):
...
+    def runTest(self):
+        """
+        For sai server, testing different lags with router
+        ---- Test for 17 ports minimun ----
+        Steps
+        1. Create virtual router
+        2. Reserve one port for sending packets
+        3. Create router interfaces 1-for all the lags, 2-for the source port 
+        3. Create sixteen LAGs with each hash one member
+        4. Config neighbors,and FDB entries
```
lguohan: there is no fdb entries used, can you update the description to match the test?

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,159 @@ def runTest(self):
...
+            self.lag_members.append(sai_thrift_create_lag_member(self.client, self.lags[i % num_of_lags], port_list[i]))
+        for i in xrange(num_of_lags):
+            self.lags_rifs.append(sai_thrift_create_router_interface(self.client, self.vr_id, 1, self.lags[i], 0, self.v4_enabled, self.v6_enabled, mac))
+        for i in xrange(16):
+            sai_thrift_create_neighbor(self.client, self.addr_family, self.lags_rifs[i%len(self.lags_rifs)], "10.10.%s.1" % str(i+1), self.mac_pool[i])
+            sai_thrift_create_route(self.client, self.vr_id, self.addr_family, "10.10.%s.0" % str(i+1), self.ip_mask, self.lags_rifs[i% len(self.lags_rifs)])
+
+    def cleaning_lags(self, num_of_lags, port_list):
+        if (num_of_lags == 0 ): return
+        for i in xrange(16):
```
lguohan: as in the setup_lags, this 16 can be changed to number_of_lags.

test/saithrift/tests/sail3.py
engdorm: did a global const "total_lag_port = 16"
because this const affect all the test parameters 
and this way, I dont need to pass it

test/saithrift/tests/sail3.py
engdorm: Rif can have more than one neighbor connect
and I did it this way for 3 main reasons :
1. To make to code more generic, this way the send and verify packets function is working with all the different combinations of lags, without changing its logic  
2. All of the hash parameters are constant besides the TCP ports and  the src IP, and this way I'm changing to the DST IP as well, it's not much because it's from a 16 preconfigured different IPs.
but it still preferred from const DST IP that will happen in lag with one neighbor and one DST port
3. This test is more comprehensive, it also checks rif with one neighbor connected to it and rif with 16 neighbors connected to it

test/saithrift/tests/sail3.py
engdorm: done

engdorm: added

test/saithrift/tests/sail3.py
```diff
@@ -1325,3 +1326,160 @@ def runTest(self):
...
+               '00:11:32:33:44:60',
+               '00:11:33:33:44:61',
+               '00:11:34:33:44:62',
+               '00:11:35:33:44:63',
+               '00:11:36:33:44:64',
+               '00:11:37:33:44:65',
+               '00:11:38:33:44:66']
+    
+    
+    def config_for_lags(self, num_of_lags, port_list):
```
engdorm: fixed

test/saithrift/tests/sail3.py
```diff
@@ -1325,3 +1326,160 @@ def runTest(self):
...
+        for i in xrange(num_of_lags):
+            self.lags.append(self.client.sai_thrift_create_lag([]))
+        for port_num in xrange(16):
+            self.lag_members.append(sai_thrift_create_lag_member(self.client, self.lags[port_num % num_of_lags], port_list[port_num]))
+        for i in xrange(16): 
+            sai_thrift_create_fdb(self.client, 1, self.mac_pool[i], self.lags[i % len(self.lags)], self.mac_action)
+            
+    def cleaning_lags(self, num_of_lags, port_list):
```
engdorm: fixed

test/saithrift/tests/sail3.py
engdorm: fixed

test/saithrift/tests/sail3.py
```diff
@@ -1325,3 +1326,160 @@ def runTest(self):
...
+            self.assertTrue((pkt_counter[stat_port] <= ((IP_LAST_WORD_RANGE ) * 1.2)),
+                    "Not all paths are equally balanced, %s" % pkt_counter[stat_port])
+                    
+    def runTest(self):
+        """
+        For sai server, testing different lags with router
+        ---- Test for 17 ports minimun ----
+        Steps
+        1. Create virtual router
+        2. Reserve one port for sending packets
```
engdorm: fixed

test/saithrift/tests/sail3.py
engdorm: fixed

test/saithrift/tests/sail3.py
engdorm: fixed

test/saithrift/tests/sail3.py
```diff
@@ -1325,3 +1326,160 @@ def runTest(self):
...
+        ip_addr = '10.10.0.0'
+        ip_mask = '255.255.0.0'
+        
+        print
+        print "L3MultipleLagTest"
+        #general configuration 
+        random.seed(1)
+        switch_init(self.client)
+        #not for git
+        sai_thrift_clear_all_counters(self.client)
```
engdorm: fixed

test/saithrift/tests/sail3.py
```diff
@@ -1325,3 +1326,160 @@ def runTest(self):
...
+        #not for git
+        self.src_port = port_list[16]
+        if (len(port_list) < 17 ) : return
```
engdorm: fixed

test/saithrift/tests/sail3.py
engdorm: fixed

test/saithrift/tests/sail3.py
```diff
@@ -1325,3 +1326,160 @@ def runTest(self):
...
+                    
+    def runTest(self):
+        """
+        For sai server, testing different lags with router
+        ---- Test for 17 ports minimun ----
+        Steps
+        1. Create virtual router
+        2. Reserve one port for sending packets
+        3. Create router interfaces 1-for all the lags, 2-for the source port 
+        3. Create sixteen LAGs with each hash one member
```
engdorm: fixed

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,158 @@ def runTest(self):
...
+    
+    
+    def setup_lags(self, num_of_lags, port_list):
+        for i in xrange(num_of_lags):
+            self.lags.append(self.client.sai_thrift_create_lag([]))
+        for i in xrange(self.total_lag_port):
+            self.lag_members.append(sai_thrift_create_lag_member(self.client, self.lags[i % num_of_lags], port_list[i]))
+        for i in xrange(num_of_lags):
+            self.lags_rifs.append(sai_thrift_create_router_interface(self.client, self.vr_id, 1, self.lags[i], 0, self.v4_enabled, self.v6_enabled, ''))
+        for i in xrange(self.total_lag_port):
```
lguohan: I am not sure why you need to create route and neighbor for every port, I think it should be every router interface, so it should be number_of_lags here.

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,158 @@ def runTest(self):
...
+            sai_thrift_create_neighbor(self.client, self.addr_family, self.lags_rifs[i%len(self.lags_rifs)], "10.10.%s.1" % str(i+1), self.mac_pool[i])
+            sai_thrift_create_route(self.client, self.vr_id, self.addr_family, "10.10.%s.0" % str(i+1), self.ip_mask, self.lags_rifs[i% len(self.lags_rifs)])
+
+    def teardown_lags(self, num_of_lags, port_list):
+        if (num_of_lags == 0 ): return
+        for i in xrange(self.total_lag_port):
```
lguohan: same as above.

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,158 @@ def runTest(self):
...
+        
+    def send_and_verify_packets(self, num_of_lags, port_list):
+        exp_pkts = [0]*self.total_lag_port
+        pkt_counter = [0] * self.total_lag_port
+        destanation_ports = range(self.total_lag_port)
+        sport = 0x1234
+        dport = 0x50
+        src_mac = self.dataplane.get_mac(0, 16)
+        IP_LAST_WORD_RANGE = 254
+        IP_2ND_LAST_WORD_RANGE = 16
```
lguohan: I think this can be num_of_lags here, it does not have to be 16.

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,158 @@ def runTest(self):
...
+                                            eth_dst=self.mac_pool[0],
+                                            eth_src=router_mac,
+                                            ip_src=ip_src,
+                                            ip_dst=ip_dst,
+                                            ip_id=i,
+                                            tcp_sport=sport,
+                                            tcp_dport=dport,
+                                            ip_ttl=63)
+                    masked_exp_pkt = Mask(exp_pkt)
+                    masked_exp_pkt.set_do_not_care_scapy(ptf.packet.Ether,"dst")
```
lguohan: there is no need to mask the dst, the dstmac will be mac_pool[j]

test/saithrift/tests/sail3.py
```diff
@@ -1324,4 +1325,158 @@ def runTest(self):
...
+                                            eth_dst=router_mac,
+                                            eth_src=src_mac,
+                                            ip_src=ip_src,
+                                            ip_dst=ip_dst,
+                                            ip_id=i,
+                                            tcp_sport=sport,
+                                            tcp_dport=dport,
+                                            ip_ttl=64)
+                    exp_pkt = simple_tcp_packet(
+                                            eth_dst=self.mac_pool[0],
```
lguohan: use as mac_pool[j]

test/saithrift/tests/sail3.py
lguohan: In the worst case, when you have 1 lag with 16 members, each port only get 254/16 packets, then you need to vary the source IP in a larger range. for example i = 65536. and src ip = 10.0.i/255.i%255

test/saithrift/tests/sail3.py
engdorm: Rif can have more than one neighbor connect
and I did it this way for 3 main reasons :
1. To make to code more generic, this way the send and verify packets function is working with all the different combinations of lags, without changing its logic

2. All of the hash parameters are constant besides the TCP ports and the src IP, and this way I'm changing to the DST IP as well, it's not much because it's from a 16 preconfigured different IPs.
but it still preferred from const DST IP that will happen in lag with one neighbor and one DST port
3. This test is more comprehensive, it also checks rif with one neighbor connected to it and rif with 16 neighbors connected to it

test/saithrift/tests/sail3.py
engdorm: answered in comment above

PR #231, opened 2016-09-22
update switch init

engdorm:
update switch init to check the ports status instead of time out in switch.py


test/saithrift/tests/switch.py
```diff
@@ -85,7 +84,19 @@ def switch_init(client):
     for interface,front in interface_to_front_mapping.iteritems():
         sai_port_id = client.sai_thrift_get_port_id_by_front_port(front);
         port_list[int(interface)]=sai_port_id
-           
+        
+    for num_of_tries in range(20):
+        # wait till the port are up
```
lguohan: we should wait till all port to be up. all the ports that have mapped in the interfaces.

test/saithrift/tests/switch.py
```diff
@@ -69,23 +69,36 @@ def switch_init(client):
...
+                if attribute.id == SAI_PORT_ATTR_OPER_STATUS:
+                    if attribute.value.s32 != SAI_PORT_OPER_STATUS_UP:
+                        all_ports_are_up = False
+                        print "port %h is down" % port
+        if all_ports_are_up:
+            break
+        else:
+            all_ports_are_up = True
+    if not all_ports_are_up:
+        raise NameError('Not all of the  ports are up')
```
lguohan: NameError -> RuntimeError

PR #233, opened 2016-09-24
add sai tunnel test cases

lguohan:


test/saithrift/tests/switch.py
```diff
@@ -209,7 +219,8 @@ def sai_thrift_remove_route(client, vr_id, addr_family, ip_addr, ip_mask, nhop):
     route = sai_thrift_unicast_route_entry_t(vr_id, ip_prefix)
     client.sai_thrift_remove_route(thrift_unicast_route_entry=route)
 
-def sai_thrift_create_nhop(client, addr_family, ip_addr, rif_id):
+def sai_thrift_create_nhop(client, addr_family, is_tunnel, ip_addr, rif_id):
```
lguohan: many existing tests are using this API, in order not break existing test, I suggest the API to be like

sai_thrift_create_nhop(client, addr_family, ip_addr, rif_id, is_tunnel=False):

srikanthre: Added the same in the newer request

PR #237, opened 2016-09-27
Hold these SAI objects for the sencond test with ACL applied

xdwu:
Route, next hop, router_interface and virtual router will be removed when the second test (with ACL applied) is completed

### DIFF
--- a/test/saithrift/tests/saiacl.py
+++ b/test/saithrift/tests/saiacl.py
@@ -63,8 +63,17 @@ class IPAclTest(sai_base_test.ThriftInterfaceDataPlane):
         try:
             send_packet(self, 2, str(pkt))
             verify_packets(self, exp_pkt, [1])
-        finally:
+        except IPSendRecvErr:
             # cleanup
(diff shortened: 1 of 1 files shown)


test/saithrift/tests/saiacl.py
```diff
@@ -65,15 +65,6 @@ def runTest(self):
             verify_packets(self, exp_pkt, [1])
         finally:
             # cleanup
```
lguohan: cleanup should not be removed, should change the finally to except:

test/saithrift/tests/saiacl.py
```diff
@@ -63,7 +63,7 @@ def runTest(self):
         try:
             send_packet(self, 2, str(pkt))
             verify_packets(self, exp_pkt, [1])
-        finally:
+        except IPSendRecvErr:
```
lguohan: catch all exceptions, can clean up the system. after clean up, should return.

xdwu:
Withdraw the Pull request

PR #248, opened 2016-10-03
add next hop group member API

lguohan:


lguohan:
bulk create api will come later.

lguohan:
test it

PR #250, opened 2016-10-03
Multi NPU Switch id addition to API's

AshokDaparthi:


lguohan:
@kcudnik , can you take a look at this PR?

inc/saiswitch.h
```diff
@@ -979,6 +993,119 @@ typedef enum _sai_switch_attr_t
...
+     * @flags CREATE_ONLY
+     * @allownull true
+     */
+    SAI_SWITCH_ATTR_SWITCH_PROFILE_ID,
+
+    /**
+     * @brief Device Information for switch initialization.
+     * 
+     * Hardware information format is based on SAI implementations by vendors.
+     *   Example: Like PCI location, I2C adddress etc.
```
AshokDaparthi: Add comment in case of NULL

kcudnik: Use sanity check to see if it compile form meta dir

inc/sairoute.h
```diff
@@ -142,35 +144,42 @@ typedef sai_status_t (*sai_create_route_fn)(
...
  *
+ * @param[in] switch_id - Switch id
  * @param[in] route_entry - route entry
  *
  * @return #SAI_STATUS_SUCCESS on success Failure status code on error
  */
 typedef sai_status_t (*sai_remove_route_fn)(
-        _In_ const sai_route_entry_t *route_entry);
+        _In_ sai_switch_id_t switch_id,
+        _In_ const sai_route_entry_t *route_entry,
```
kcudnik: Will this switch id will be in every create api ? Also route contains vr id inside which should identyfiy switch already seems like in tyis case this id is not needet at all

lguohan: while I agree the switch be every where, I think it is better to put the switch_id into the sai_route_entry_t, sai_neighbor_entry_t, and sai_fdb_entry_t. 

Because we only agree to add switch_id as a parameter in the create API, not in set,get,remove API.

AshokDaparthi: OK, I will make changes.

lguohan:
what about vlan_id as an object id? are you going to change that in this commit or a separate commit?

AshokDaparthi:
making vlan_id as object will crate software cache in SAI, This is same case and we discussed  making the  route, fdb as objects. We did not get to conclusion. After that we can think of modifying this also.

lguohan:
I do not like the fact that the vlan api is different from other API. Either we go with route approach to have a vlan struct that has vlan_id and switch_id, or we make vlan_id as a UOID. I prefer the second approach as we can just embed the uint16 vlan_id into the UOID vlan_id. It is simple.

inc/sairoute.h
lguohan: @kcudnik , the switch_create api is different than all other create_api as the switch_id is an out parameter.

lguohan:
sai_flush_fdb_entries_fn

lguohan:
closed by mistake

inc/saiswitch.h
AshokDaparthi: it is string with null terminated. Not sure what is missed here.

AshokDaparthi:
vlan we will close this in today's discussion. I am fine with any approach, But in 2nd approach we need to pass vlan id to create vlan object.

kcudnik:
Use sanity check to see if it compile form meta dir:
$ cd meta
$ make

lguohan:
test it

AshokDaparthi:
@lguohan @tushar-ty @itaibaz @atitjain - Please expedite review

inc/saiobject.h
```diff
@@ -58,37 +58,43 @@ typedef struct _sai_object_key_t
 /**
```
atitjain: As per my understanding of the current model, only create time we need the switch-id. Later we only need the OID and OID somehow maps to switch-id. If that is the model, for the calls in this file we dont need switch-id as parameter.

inc/saiport.h
```diff
@@ -1456,22 +1458,26 @@ typedef sai_status_t (*sai_clear_port_all_stats_fn)(
  *
  * Passed as a parameter into sai_initialize_switch()
  *
+ * @param[in] switch_id Switch id
  * @param[in] count Number of notifications
  * @param[in] data Array of port operational status
  */
 typedef void (*sai_port_state_change_notification_fn)(
```
atitjain: As per my understanding of the current model, only create time we need the switch-id. Later we only need the OID and OID somehow maps to switch-id. So if the sai_port_oper_status_notification_t  has port OID we should not need switch-id.

inc/saiswitch.h
```diff
@@ -1113,98 +1236,73 @@ typedef enum _sai_switch_attr_t
...
  *
+ * @param[in] switch_id Switch Id
  * @param[in] switch_oper_status New switch oper state
  */
 typedef void (*sai_switch_state_change_notification_fn)(
+        _In_ sai_object_id_t switch_id,
         _In_ sai_switch_oper_status_t switch_oper_status);
 
 /**
- * @brief Switch notification table passed to the adapter via sai_initialize_switch()
```
atitjain: So now each of these is a switch attribute?

inc/saiswitch.h
```diff
@@ -976,8 +990,117 @@ typedef enum _sai_switch_attr_t
...
+    /**
+     * @brief Set Switch oper status change notification callback 
+     * function passed to the adapter.
+     *
+     * Use sai_switch_state_change_notification_fn as notification function.
+     *
+     * @type sai_pointer_t
+     * @flags CREATE_AND_SET
+     */
+    SAI_SWITCH_ATTR_OPER_STATUS_CHANGE_NOTIFY,
```
atitjain: @type should besai_switch_state_change_notification_fn for clarity. Though its captured in comment.

inc/saiobject.h
AshokDaparthi: yes, your understanding is correct. This file is defined to get the objects from SAI. switch id for other than create derived internally in SAI, in this bluk get cases,  caller can not know switch id by looking object id,
since this is implementation specific. 
bulk api's get(defined for warmboot),  it is good to know from which NPU they are getting bulk objects.

inc/saiswitch.h
AshokDaparthi: yes

inc/saiswitch.h
AshokDaparthi: ok

inc/saifdb.h
```diff
@@ -282,22 +285,26 @@ typedef sai_status_t (*sai_get_fdb_entry_attribute_fn)(
...
 
 /**
  * @brief FDB notifications
  *
+ * @param[in] switch_id Switch object id
  * @param[in] count Number of notifications
  * @param[in] data Pointer to fdb event notification data array
  */
 typedef void (*sai_fdb_event_notification_fn)(
+        _In_ sai_object_id_t switch_id,
```
itaibaz: Why do we need switch_id for sai_fdb_event_notification
The event data has event notification data items
Each data item contains fdb_entry, which already contains switch_id

inc/saiport.h
```diff
@@ -1456,22 +1458,26 @@ typedef sai_status_t (*sai_clear_port_all_stats_fn)(
...
+        _In_ sai_object_id_t switch_id,
```
itaibaz: why do we need switch id in port notifications ?
each port event contains port object, which already has the switch id encoded in it

inc/saiswitch.h
itaibaz: should the default for all notification be null, meaning don't call the callback ?

inc/saiswitch.h
```diff
@@ -1113,98 +1236,73 @@ typedef enum _sai_switch_attr_t
...
  */
-typedef sai_status_t (*sai_connect_switch_fn)(
-        _In_ sai_switch_profile_id_t profile_id,
-        _In_reads_z_(SAI_MAX_HARDWARE_ID_LEN) char *switch_hardware_id,
-        _In_ sai_switch_notification_t *switch_notifications);
-
-/**
- * @brief Disconnect this SAI library from the SDK.
- */
-typedef void (*sai_disconnect_switch_fn)(void);
```
itaibaz: how can we now disconnect from a switch without removing the switch object ?
with the old API app 1 could init switch
App 2 could connect and then disconnect, to free connection resources
With new API, app 2 can connect, but i don't see a way for it to disconnect

inc/saiswitch.h
```diff
@@ -976,8 +990,117 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * TRUE - Initialize switch/SDK. 
+     * FALSE - Connect to SDK. This will connects library to the initialized SDK.
+     * After this call the capability attributes should be ready for retrieval
+     * via sai_get_switch_attribute()
+     *
+     * @type bool
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_SWITCH_ATTR_INIT_SWITCH,
```
itaibaz: not clear about connect flow
So I will call create switch ID with init switch=false, and the same switch hardware ID as as given to a previous call that created & initialized the switch
Is the expectation that this create will return the same object ID as the original create - in which case it is not really creating a new object and different than any other create
Or the expectation is to return a new object ID, in which case we will have 2 object IDs referring to the same switch object
Why is it necessary to remove connect/disconnect ? IMO, actually the old syntax was clearer

inc/saiswitch.h
```diff
@@ -979,6 +993,119 @@ typedef enum _sai_switch_attr_t
...
+     * FALSE - Connect to SDK. This will connects library to the initialized SDK.
+     * After this call the capability attributes should be ready for retrieval
+     * via sai_get_switch_attribute()
+     *
+     * @type bool
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @default true
+     */
+    SAI_SWITCH_ATTR_INIT_SWITCH,
+
```
vkochan-mlnx: Does it mean that to perform switch connect it is needed to pass existing switch id as pointer into create function ?
I really thing that we need to keep switch_connect/disconnect which will take switch id as value parameter.

inc/saiswitch.h
```diff
@@ -1213,10 +1311,8 @@ typedef sai_status_t (*sai_get_switch_attribute_fn)(
  */
 typedef struct _sai_switch_api_t
```
rspolyak: Is it worth to update applications that uses SAI (i.e., PTF) in this PR? After this PR is merged, will PTF compile?

inc/saiswitch.h
```diff
@@ -1113,98 +1236,73 @@ typedef enum _sai_switch_attr_t
...
+ * @param[in] attr_list Array of attributes
  *
- * @param[in] warm_restart_hint Hint that indicates controlled warm restart.
- * Since warm restart can be caused by crash (therefore there are no guarantees
- * for this call), this hint is really a performance optimization.
+ * @return #SAI_STATUS_SUCCESS on success Failure status code on error
  */
-typedef void (*sai_shutdown_switch_fn)(
-        _In_ bool warm_restart_hint);
+typedef sai_status_t(*sai_create_switch_fn)(
```
rspolyak: Before creating new switch instance using sai_create_switch_fn(), how application should get available NPU numbers for particular hardware?
For example, application wants to initialize two NPU's, but particular platform is single NPU? Can application retrieve NPU number prior to create new switch instances?

inc/saiswitch.h
AshokDaparthi: Yes, This is not mandatory attribute, if you are not included this attribute it is NULL inside the SAI.

inc/saiswitch.h
AshokDaparthi: it will crate the new switch object id for connect case also and will be used for disconnect.
As per our previous discussion with guohan, we thought of reusing API's and also there is no need for application to exchange the switch ID's since every connect will get the logical switch id for each connect.  if every one feels having separate is more cleaner  i can keep connect and disconnect API's. @lguohan - Please comment on this.

inc/saiport.h
AshokDaparthi: typedef struct _sai_port_oper_status_notification_t
{
    /** Port id */
    sai_object_id_t port_id;

    /** Port operational status */
    sai_port_oper_status_t port_state;
} sai_port_oper_status_notification_t;

This structure does not has switch id it has only port id. There is noway to get switch object id from the port object id in applications because switch object id embedding in port oid is SAI vendor specific implementation. Application will not know about format.
I am not sure is there any need to know switch id for port notifications from applications side. Since this will be uniform having switch object id for all notifications instead of retrieving from other OID's.

inc/saiobject.h
AshokDaparthi: @atitjain - is it clear or still have any open items.

inc/saifdb.h
AshokDaparthi: For API uniformity i added this. All notifications will have switch object id and notification list will have FDB entries belong to same Switch ID.

inc/saiswitch.h
lguohan: itai, your understanding is correct except that the object ID returned does not have to be the exact the same object although it is used to identify the same asic. 

Since create API can also be used for connect, so it is not neccessary to have connect/disconnect.

inc/saiswitch.h
lguohan: this is likely be in a separate CR. We need some example first, then we can change the ptf tests to use the new API.

inc/saiswitch.h
lguohan: this is likely be in a separate CR. We need some example first, then we can change the ptf tests to use the new API.

inc/saifdb.h
lguohan: in the fdb entry, there is already a switch_id field, it is a little redudent here.

inc/saiswitch.h
rspolyak: great, example with be helpful. 
@AshokDaparthi are you planning to provide some sample code that uses new API's?

inc/saifdb.h
AshokDaparthi: Removed

inc/saiswitch.h
AshokDaparthi: @lguohan , While looking at previous connect API, API will allows notification as argument, 
is that means multiple applications can register of same notifications example FDB/port notification need by app1 and app2 . 
In this scenario we need to generate  2 call back notification to 2 applications for same FDB event. 
correct me if i am wrong? 

in this case, each notification event will go with switch object created with init=true or should generate events with switch object id create for connect app1 and connect app2.

if this is real scenario, then it is better to go with connect and disconnect API's and use switch id as input. Then SAI can replicate events and send to multiple applications.

@itaibaz - Consider above scenario, app1 and app2 are used connect and registered for FDB notification. Now app2 alone is disconnected how SAI will aware it should trigger notifications to app1 alone, because present disconnect does not have any args.    

I am not aware of background of connect/disconnect API's. Please let me know expected behavior based on that we can converge on above solution.

inc/saiswitch.h
rspolyak: @lguohan , @AshokDaparthi 
Couple of questions:
1) for example, we want to initialize device with 2 NPU's. Thus we call sai_create_switch_fn() twice, each time with different SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO attribute value, i.e. NPU0-PCI0, NPU1-PCI1.
Right?

2) for each call, unique switch_id OID is returned. We can then use them to address specific NPU.
If SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO is NULL, then it it assumed as single NPU device.
Right?

3) On SAI adapter side, SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO is encoded into switch_id OID and returned to application. 
Right?

4) What is the purpose of SAI_SWITCH_ATTR_SWITCH_PROFILE_ID then? In current SAI implementation it is usually 0 (for example, PTF and SONIC pass 0).

inc/saiport.h
atitjain: What I mean is not the application side. I mean in the application once we call create with NPU, the adapter could maintain the state (it will anyway need it to do Read, Update and Delete calls for the NPU). So in your way of things how is the OID space, is it per NPU or its a per adapter instance (same OID space for all the NPUs managed by adapter).

inc/saiswitch.h
```diff
@@ -976,8 +990,117 @@ typedef enum _sai_switch_attr_t
...
+    /**
+     * @brief Set Switch oper status change notification callback 
+     * function passed to the adapter.
+     *
+     * Use sai_switch_state_change_notification_fn as notification function.
+     *
+     * @type sai_pointer_t
+     * @flags CREATE_AND_SET
+     */
+    SAI_SWITCH_ATTR_SWITCH_STATE_CHANGE_NOTIFY,
```
lguohan: If it is not a mandatory attribute, then it should have a default value, which in this case is NULL. Can you take a look at #287 PR, it requires the sai_pointer_t to also have a default value.

inc/saiswitch.h
lguohan: assume you have two host adapters running on top of SAI, one uses create_switch(INIT=true), one uses create_switch(INIT=false). both registers FDB notification events, then SAI would deliver FDB events to both host adapters. I think it is the same as the connect API. 

The difference between INIT is true and false is in the remove_switch(). When INIT=true, the remove_switch will deinit the switch, otherwise, the remove_switch will just disconnect the switch.

inc/saiswitch.h
lguohan: 1) yes.
2) yes
3) I do not quite understand what "encoded" means. The returned switch_id OID will be used to identify the NPU in the rest of SAI calls.
4) It was designed to address the scenario where the profile has many sub-profiles, and user can use profile_id to indicate which sub-profile to use for switch init. Since now usually the profile does not have sub-profiles, so it is always zero.

inc/saiswitch.h
AshokDaparthi: in this case then switch id is FDB event should be original switch id( created with init=true) right?

lguohan:
test sai_pointer_t default value

inc/saiswitch.h
lguohan: should be the switch_id that it got from switch_init(). I see your point, in case the two host adapters need to communicate two different switch_id representing the same asic will be confusing. Then, I think we can still keep this switch_init(INIT) and let them return the same switch_id. So, we can still use all the switch attribute defined. Meanwhile, we add disconnect_switch(switch_id) function.

AshokDaparthi: By return the same switch id  to all adapters(for init and connect) and using disconnect_switch(switch id) should have problem while disconnect SAI does not know which notifications that adaptor registered, to stop notifications. this problem exits in old API also.

lguohan: since the host adapters specify the hardware_info in create_switch, even it receives two different switch_id, it know they are representing the same asic. so, looks like returning different switch_id is ok.

inc/saiport.h
AshokDaparthi: Removed

inc/saiport.h
AshokDaparthi: Removed

inc/saiobject.h
zubinshah: @AshokDaparthi considering the proposed approach it seems that the current model assumes the host adapter needs to maintain some sort of object id storage per switch-id, to be reconciled post warm restart. is there a way to query the number of switch-ids supported post warm restart?

zubinshah: @AshokDaparthi :  as a follow up on @atitjain's question just for my understanding what is the motivation that there the following will be different for various switch-ids in the system : 
1. max attributes per object type per switchid (heterogeneous ASIC supported ? )
2. number of objects per object type per switchid (asymmetric type of programming done across multiple ASICs?)

AshokDaparthi: 1# Yes 
2# Yee

inc/saiswitch.h
rspolyak: So to sum this up:
1) switch_init() will return different switch_id's for create and connect cases
2) disconnect_switch(switch id) is not necessary in this case 
3) host adapter should be capable to store notification callbacks for each call of switch_init(). This will allow to trigger different callbacks for each registered app.

Please confirm.

AshokDaparthi: 1# it should return the same switch object id for both connect and disconnect.
https://github.com/opencomputeproject/SAI/pull/250/commits/39e44e6270af4ce2b7a1deadbe8097d5afc0dd63
2# disconnect switch is needed even if we return the same object id, each app can register(exampe app1 with FDB and app2 with port event) with different call back notifications.  when we disconnect from app1, FDB notification should be removed  since it is registered by app1.

rspolyak:
Ok, but instead of adding new function, can we pass flag True/false to shutdown_switch() in order to disconnect or remove switch (the same as is done for create_switch()?

7 ????. 2016 ?. ? 10:48 Ashok Daparthi <notifications@github.com<mailto:notifications@github.com>> ???????(??):


@AshokDaparthi commented on this pull request.

In inc/saiswitch.h<https://github.com/opencomputeproject/SAI/pull/250>:

> +     * @default empty
+     */
+    SAI_SWITCH_ATTR_FIRMWARE_PATH_NAME,
+
+    /**
+     * @brief Set to switch initialization or connect to NPU/SDK.
+     *
+     * TRUE - Initialize switch/SDK.
+     * FALSE - Connect to SDK. This will connects library to the initialized SDK.
+     * After this call the capability attributes should be ready for retrieval
+     * via sai_get_switch_attribute()
+     *
+     * @type bool
+     */
+    SAI_SWITCH_ATTR_INIT_SWITCH,


1# it should return the same switch object id for both connect and disconnect.
39e44e6<https://github.com/opencomputeproject/SAI/pull/250/commits/39e44e6270af4ce2b7a1deadbe8097d5afc0dd63>
2# disconnect switch is needed even if we return the same object id, each app can register(exampe app1 with FDB and app2 with port event) with different call back notifications. when we disconnect from app1, FDB notification should be removed since it is registered by app1.

-

inc/saiswitch.h
rspolyak: Ok, but instead of adding new function, can we pass flag True/false to shutdown_switch() in order to disconnect or remove switch (the same as is done for create_switch()?

PR #251, opened 2016-10-06
remove add/remove child node from scheduler group, add parent node attribute

lguohan:
remove add/remove child node from scheduler group, add parent node attribute in scheduler group and queue object

### DIFF
--- a/inc/saiqueue.h
+++ b/inc/saiqueue.h
@@ -73,12 +73,32 @@ typedef enum _sai_queue_attr_t
     SAI_QUEUE_ATTR_TYPE = SAI_QUEUE_ATTR_START,
 
     /**
+     * @brief Pord id
+     *
+     * @type sai_object_id_t
+     * @objects SAI_OBJECT_TYPE_PORT
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY | KEY
+     */
+    SAI_QUEUE_ATTR_PORT = 0x00000001,
+
+    /**
      * @brief Queue index
      *
      * @type sai_uint8_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY | KEY
      */
-    SAI_QUEUE_ATTR_INDEX = 0x00000001,
+    SAI_QUEUE_ATTR_INDEX = 0x00000002,
+
+    /**
+     * @brief Parent scheduler node
+     *
+     * In case of Hierarchical Qos not supported, the parent node is the port
+     *
+     * @type sai_object_id_t
+     * @objects SAI_OBJECT_TYPE_SCHEDULER, SAI_OBJECT_TYPE_PORT
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
+     */
+    SAI_QUEUE_ATTR_PARENT_SCHEDULER_NODE = 0x00000003,
 
     /* READ-WRITE */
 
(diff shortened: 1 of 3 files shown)


lguohan:
@AshokDaparthi , please review

inc/saiqueue.h
```diff
@@ -80,6 +80,15 @@ typedef enum _sai_queue_attr_t
...
 
+    /**
+     * @brief Parent scheduler node
+     *
+     * @type sai_object_id_t
+     * @objects SAI_OBJECT_TYPE_SCHEDULER
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
+     */
+    SAI_QUEUE_ATTR_PARENT_SCHEDULER_NODE = 0x00000002,
+
```
AshokDaparthi: This is mandatory in case of hierarchy qos supported alone.

AshokDaparthi: In case of Hierarchical Qos not supported in some NPU's, queues are created directly on port. Then queue creation has attributes queue_index and port  oid. In you made it mandatory in those NPU's it again we have to give port oid as part it. which is redundant. So this attribute is mandatory in case hierarchy qos supported NPU's alone. else we can remove port oid and keep this one alone, So both case it will work.

AshokDaparthi: Sorry, i did not realize we do not have port oid attribute in queue attribute list. 
SAI_QUEUE_ATTR_PARENT_SCHEDULER_NODE can take 2 objects, 
SAI_OBJECT_TYPE_SCHEDULER_GROUP  or SAI_OBJECT_TYPE_PORT. 
Please change comment.

lguohan: I realized that, so added the port id as an attribute. I think it's better to add it because the queue type, port id, queue index then can be the key for this object. then scheduler node is additional attribute you need to set. 

Also parent node is a CREATE_AND_SET attribute, whereas the PORT is CREATE_ONLY attribute.

inc/saiqueue.h
```diff
@@ -73,12 +73,32 @@ typedef enum _sai_queue_attr_t
...
-    SAI_QUEUE_ATTR_INDEX = 0x00000001,
+    SAI_QUEUE_ATTR_INDEX = 0x00000002,
+
+    /**
+     * @brief Parent scheduler node
+     *
+     * In case of Hierarchical Qos not supported, the parent node is the port
+     *
+     * @type sai_object_id_t
+     * @objects SAI_OBJECT_TYPE_SCHEDULER, SAI_OBJECT_TYPE_PORT
```
AshokDaparthi: SAI_OBJECT_TYPE_SCHEDULER -> SAI_OBJECT_TYPE_SCHEDULER_GROUP

lguohan:
test this.

inc/saiqueue.h
```diff
@@ -73,12 +73,32 @@ typedef enum _sai_queue_attr_t
...
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
```
lguohan: updated. since HQoS is not supported, so let's keep the flags as MANDATORY_ON_CREATE, once the HQoS is introduced we can remove the flag.

inc/saischedulergroup.h
```diff
@@ -95,6 +95,17 @@ typedef enum _sai_scheduler_group_attr_t
     SAI_SCHEDULER_GROUP_ATTR_SCHEDULER_PROFILE_ID = 0x00000005,
 
     /**
+     * @brief Scheduler group parent node
+     *
+     * This is conditional when the level > 0, when level == 0, the parent is the port.
+     *
+     * @type sai_object_id_t
+     * @objects SAI_OBJECT_TYPE_SCHEDULER
```
lguohan: updated

PR #252, opened 2016-10-07
Changes meta

AshokDaparthi:
- Added the string and pointer 
- Using "empty" as default to represent NULL for sai_string_t and sai_pointer_t types.

### DIFF
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -522,7 +522,7 @@ typedef union {
     sai_int32_t s32;
     sai_uint64_t u64;
     sai_int64_t s64;
-    sai_string_t string; 
+    sai_string_t str; 
     sai_pointer_t ptr; 
     sai_mac_t mac;
     sai_ip4_t ip4;
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -114,7 +114,8 @@ typedef uint8_t sai_uint8_t;
 typedef int8_t sai_int8_t;
 typedef size_t sai_size_t;
 typedef uint64_t sai_object_id_t;
-
+typedef char *sai_string_t;
+typedef void *sai_pointer_t;
 /**
  * @def SAI_NULL_OBJECT_ID
  * Sai NULL object ID
@@ -521,6 +522,8 @@ typedef union {
     sai_int32_t s32;
     sai_uint64_t u64;
     sai_int64_t s64;
+    sai_string_t string; 
+    sai_pointer_t ptr; 
     sai_mac_t mac;
     sai_ip4_t ip4;
     sai_ip6_t ip6;
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -114,7 +114,6 @@ typedef uint8_t sai_uint8_t;
 typedef int8_t sai_int8_t;
 typedef size_t sai_size_t;
 typedef uint64_t sai_object_id_t;
-typedef char *sai_string_t;
 typedef void *sai_pointer_t;
 /**
  * @def SAI_NULL_OBJECT_ID
@@ -524,7 +523,6 @@ typedef union {
     sai_int32_t s32;
     sai_uint64_t u64;
     sai_int64_t s64;
-    sai_string_t str; 
     sai_pointer_t ptr; 
     sai_mac_t mac;
     sai_ip4_t ip4;
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -211,6 +211,10 @@ typedef struct _sai_u8_list_t {
     uint8_t *list;
 } sai_u8_list_t;
 
+/**
+ * @brief Defines a s8 list or string.
+ * string should be null terminated and count should include '\0'    
+ */
 typedef struct _sai_s8_list_t {
     uint32_t count;
     int8_t *list;
--- a/meta/parse.pl
+++ b/meta/parse.pl
@@ -208,7 +208,6 @@ sai_uint32_t            u32
 sai_int32_t             s32
 sai_uint64_t            u64
 sai_int64_t             s64
-sai_string_t            str
 sai_pointer_t           ptr
 sai_mac_t               mac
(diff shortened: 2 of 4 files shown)


AshokDaparthi:
@kcudnik - Please review, Please edit if needed.

lguohan:
test this

meta/parse.pl
```diff
@@ -208,6 +208,8 @@ sub ExtractDescription
 sai_int32_t             s32
 sai_uint64_t            u64
 sai_int64_t             s64
+sai_string_t            string
+sai_pointer_t           pointer
```
kcudnik: this should be same as in saitypes value so "str" and "ptr"

meta/saisanitycheck.c
```diff
@@ -548,15 +549,14 @@ void check_attr_default_required(
...
-        case SAI_ATTR_VALUE_TYPE_UINT32_LIST:
+             break;
+        case SAI_ATTR_VALUE_TYPE_UINT32_LIST:        
         case SAI_ATTR_VALUE_TYPE_INT32_LIST:
         case SAI_ATTR_VALUE_TYPE_OBJECT_LIST:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_OBJECT_LIST:
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_LIST:
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT8_LIST:
-
+
```
kcudnik: remove whitespace

meta/saisanitycheck.c
```diff
@@ -719,7 +721,9 @@ void check_attr_default_value_type(
 
             switch (md->attrvaluetype)
             {
-                case SAI_ATTR_VALUE_TYPE_UINT32_LIST:
+                case SAI_ATTR_VALUE_TYPE_STRING:
+                case SAI_ATTR_VALUE_TYPE_POINTER:
```
kcudnik: those are not list types, you can remove them from here

meta/saisanitycheck.c
```diff
@@ -567,6 +567,8 @@ void check_attr_default_required(
             break;
 
         case SAI_ATTR_VALUE_TYPE_UINT8_LIST:
+        case SAI_ATTR_VALUE_TYPE_STRING:
+        case SAI_ATTR_VALUE_TYPE_POINTER:
```
kcudnik: those are not list types, please move them to separate case statement and allow default value type as CONST or NONE only

AshokDaparthi: in case of pointer and string, default value type NONE is used.

meta/saisanitycheck.c
AshokDaparthi: Removed

meta/saisanitycheck.c
AshokDaparthi: Done

meta/parse.pl
AshokDaparthi: Modified to str and ptr

kcudnik:
Ack

AshokDaparthi:
We can document saying size not more than size x or we need to define string as below structure 
sai_string_t  { int len;  char *str}.

kcudnik:
We already have s8list struct which could be reused for that purpose which also have abbility to request how long the string is setting initial count to zero, i will ping Guohan to take a look

meta/saisanitycheck.c
```diff
@@ -719,7 +739,7 @@ void check_attr_default_value_type(
 
             switch (md->attrvaluetype)
             {
-                case SAI_ATTR_VALUE_TYPE_UINT32_LIST:
+                case SAI_ATTR_VALUE_TYPE_UINT32_LIST:
```
lguohan: this line is not changed, please remove the trailing whitespaces.

lguohan:
@AshokDaparthi , can we use s8list as suggested by kcudnik? I suggest not including '\0' to get the count for the string, that is how strlen() is doing.

AshokDaparthi:
@lguohan @kcudnik I will remove the sai_string_t and use the sai_s8list_t and update the pull request,

inc/saitypes.h
```diff
@@ -211,6 +211,10 @@ typedef struct _sai_u8_list_t {
     uint8_t *list;
 } sai_u8_list_t;
 
+/**
+ * @brief Defines a s8 list or string.
+ * string should be null terminated and count should include '\0'
```
kcudnik: This seems to be different then common string usage, usually string length don't contain \0 i that would be intuitive behaviour

AshokDaparthi: @lguohan - Based on our discussion we thought of adding the \0 also in count right. Otherwise we can leave it to strlen behavior.

kcudnik: ok

meta/saisanitycheck.c
```diff
@@ -596,15 +612,14 @@ void check_attr_default_required(
         case SAI_ATTR_VALUE_TYPE_UINT64:
         case SAI_ATTR_VALUE_TYPE_MAC:
         case SAI_ATTR_VALUE_TYPE_IP_ADDRESS:
-            break;
-
+             break;
```
lguohan: extra space here?

meta/saisanitycheck.c
```diff
@@ -630,6 +645,9 @@ void check_attr_default_required(
 
             break;
 
+        case SAI_ATTR_VALUE_TYPE_POINTER:
+             break;
```
lguohan: >break; [](start = 13, length = 6)

extra space here?

AshokDaparthi: Removed

meta/saisanitycheck.c
AshokDaparthi: Removed

PR #253, opened 2016-10-07
allow SAI_ROUTE_ATTR_NEXT_HOP_ID point to SAI_NULL_OBJECT_ID by default

lguohan:
When SAI_ROUTE_ATTR_NEXT_HOP_ID = SAI_NULL_OBJECT_ID, packet is dropped.
This allows user to change packet action from DROP to FORWARD using set API.
Otherwise, the behavior transition from DROP to FORWARD is undefined.

### DIFF
--- a/inc/sairoute.h
+++ b/inc/sairoute.h
@@ -81,8 +81,8 @@ typedef enum _sai_route_entry_attr_t
      *
      * @type sai_object_id_t
      * @objects SAI_OBJECT_TYPE_NEXT_HOP, SAI_OBJECT_TYPE_NEXT_HOP_GROUP, SAI_OBJECT_TYPE_ROUTER_INTERFACE, SAI_OBJECT_TYPE_PORT
-     * @flags CREATE_AND_SET
      * @default SAI_NULL_OBJECT_ID
+     * @flags CREATE_AND_SET
      * @allownull true
      */
     SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID,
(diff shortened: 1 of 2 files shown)


lguohan:
@itaibaz , @AshokDaparthi , @atitjain , @tushar-ty , @krambn, please review.

lguohan:
@kcudnik , this PR give following error. 

Processing saiwred_8h.xml
invalid default value 'SAI_NULL_OBJECT_ID' on SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID (sai_object_id_t)
make: *** [saimetadata.c] Error 1
lgh@acs-trusty6:/data/sai/meta$ 

Can you help?

inc/sairoute.h
```diff
@@ -67,16 +67,23 @@ typedef enum _sai_route_entry_attr_t
...
+     *  FORWARD, COPY, LOG, TRANSIT
+     *
      * The next hop id can be a generic next hop object, such as next hop, next
      * hop group. Directly reachable routes are the IP subnets that are
      * directly attached to the router. For such routes, fill the router
      * interface id to which the subnet is attached. IP2ME route adds a local
      * router IP address. For such routes, fill the CPU port
      * (#SAI_SWITCH_ATTR_CPU_PORT).
      *
+     * When it is SAI_NULL_OBJECT_ID, then packet will be dropped.
```
SathishKumarTR-Dell: Just to confirm the understanding that NULL_OBJECT_ID can be set when packet action is set to TRAP before and the packet action will take precedence in that case.  In other words, the packet will be dropped when packet action is one of the following: FORWARD, COPY, LOG, TRANSIT as mentioned in the comments.

johnarnold:
Jenkins, test this please

PR #254, opened 2016-10-07
Added new port learn mode

dgsudharsan:
SAI_PORT_FDB_LEARNING_MODE_FDB_NOTIFICATION mode is proposed so that when a packet from unknown source MAC comes, it is sent as a notification via FDB callback as new MAC learn. This mode will not forward traffic(until application re-installs the FDB entry. This mode will generate only one notification per unknown source MAC to FDB callback.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -136,8 +136,8 @@ typedef enum _sai_port_fdb_learning_mode_t
     /** Trap packets with unknown source MAC to CPU. Do not learn. Forward based on destination MAC */
     SAI_PORT_FDB_LEARNING_MODE_CPU_LOG,
 
-    /** Send MAC notification via FDB callback. Do not learn. Do not forward */
-    SAI_PORT_FDB_LEARNING_MODE_SW_TRAP,
+    /** Notify unknown source MAC using FDB callback. Do not learn. Do not forward */
+    SAI_PORT_FDB_LEARNING_MODE_FDB_NOTIFICATION,
 
 } sai_port_fdb_learning_mode_t;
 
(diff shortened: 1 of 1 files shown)


inc/saiport.h
```diff
@@ -136,6 +136,9 @@ typedef enum _sai_port_fdb_learning_mode_t
     /** Trap packets with unknown source MAC to CPU. Do not learn. Forward based on destination MAC */
     SAI_PORT_FDB_LEARNING_MODE_CPU_LOG,
 
+    /** Send MAC notification via FDB callback. Do not learn. Do not forward */
```
AshokDaparthi: generate FDB Event notifications for unknown source MAC

inc/saiport.h
```diff
@@ -136,6 +136,9 @@ typedef enum _sai_port_fdb_learning_mode_t
...
+    SAI_PORT_FDB_LEARNING_MODE_SW_TRAP,
```
lguohan: is it better to change the name to SAI_PORT_FDB_LEARNING_MODE_FDB_NOTIFICATION

lguohan: generate notification only for unknown source MAC? Does it generate notification everytime when receiving a same unknown source mac packet repeatedly?

atitjain: @lguohan, for this issue to be sorted, one needs to have a FDB entry added as "pending". So on the first packet add that entry to the HW and then let the stack do verification etc and then make the pending entry active.

dgsudharsan: This mode will generate only one per unknown source MAC. This is similar to pending entries in hardware where only one notification goes per unknown source MAC

AshokDaparthi:
@itaibaz , @zubinshah @atitjain @tushar-ty @lguohan  - Please review

lguohan:
ok to test

AshokDaparthi:
@itai @mattyk - Could you please let us know support of this behavior.

PR #255, opened 2016-10-09
Fixed issue with hardcoded HW lane value

rspolyak:
For 10G port speed lanes number is 1 and not 4, so use lane count retrieved from HW instead of hard-coded one.

### DIFF
--- a/test/saithrift/src/switch_sai_rpc_server.cpp
+++ b/test/saithrift/src/switch_sai_rpc_server.cpp
@@ -1102,7 +1102,8 @@ class switch_sai_rpcHandler : virtual public switch_sai_rpcIf {
           port_api->get_port_attribute(port_list_object_attribute.value.objlist.list[i], 1, &port_lane_list_attribute);
 
           std::set<int> port_lanes;
-          for (int j=0 ; j<4 ; j++){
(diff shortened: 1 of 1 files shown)


lguohan:
merged #256 , this one seems to be duplicated. should we close this one?

rspolyak:
We need this fix for both master and sonic branches. If you are planning to merge master to sonic branch than it is ok to skip second PR.

Regards,
Rostyslav

Sent from my phone

10 ????. 2016 ?. ? 10:43 lguohan <notifications@github.com<mailto:notifications@github.com>> ???????(??):


merged #256<https://github.com/opencomputeproject/SAI/pull/256> , this one seems to be duplicated. should we close this one?

-

rspolyak:
@lguohan Could you please merge this fix to SONIC branch too?

lguohan:
current all tests assume to work with sonic branch headers although they are checked in under the master branch. I would like to separate the test repo with the sai header repo, and in the test repo to reference which sai header repo to use, so that there is no ambiguity here.

rspolyak:
PTF tests does not run on 10G ports without this change.
We use SAI SONIC branch (not SAI master) for PTF and thus need the same fix here.

PR #266, opened 2016-10-20
host if spec changes for 1.0

itaibaz:
Add host interface table entry, for demux of traps to host if object
Remove host if user defined traps, and add them to regular trap list


lguohan:
@tushar-ty , @AshokDaparthi , @atitjain ,  please review.

inc/saihostintf.h
```diff
@@ -594,28 +497,33 @@ typedef enum _sai_hostif_attr_t
...
-     * Valid only when #SAI_HOSTIF_ATTR_TYPE == #SAI_HOSTIF_TYPE_NETDEV
+     * Valid only when #SAI_HOSTIF_ATTR_TYPE == #SAI_HOSTIF_TYPE_PORT 
+     * || #SAI_HOSTIF_TYPE_LAG || #SAI_HOSTIF_TYPE_VLAN
+     * should be port object when type is SAI_HOSTIF_TYPE_PORT
+     * should be lag object when type is SAI_HOSTIF_TYPE_LAG
+     * should be VLAN rif object when type is SAI_HOSTIF_TYPE_VLAN
      *
      * @type sai_object_id_t
-     * @objects SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_ROUTER_INTERFACE
+     * @objects SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_LAG, SAI_OBJECT_TYPE_ROUTER_INTERFACE
```
lguohan: can we still use SAI_HOSTIF_TYPE_NETDEV here? Since we can use the object id to know whether it is port-based, lag-based, or router interface based host interface type.

inc/saihostintf.h
```diff
@@ -186,7 +191,7 @@ typedef enum _sai_hostif_trap_type_t
     /** default action is forward */
     SAI_HOSTIF_TRAP_TYPE_IGMP_TYPE_V3_REPORT = 0x000000009,
 
-    /** default action is trap */
```
atitjain: I think for this, there is a prerequisite that the sampling is enabled. If not enabled this trap has no significance. Once the sampling is enabled, the default should be trap.

inc/saihostintf.h
```diff
@@ -353,35 +372,14 @@ typedef enum _sai_hostif_trap_attr_t
...
-     * @brief Enable trap for a list of SAI ports
+     * @brief List of SAI ports to be excluded (disabled) from the trap generation
      *
      * @type sai_object_list_t
      * @objects SAI_OBJECT_TYPE_PORT
      * @flags CREATE_AND_SET
      * @default empty
      */
-    SAI_HOSTIF_TRAP_ATTR_PORT_LIST,
+    SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST,
```
atitjain: what is the motivation here?

inc/saihostintf.h
```diff
@@ -692,6 +587,179 @@ typedef sai_status_t(*sai_get_hostif_attribute_fn)(
...
+    SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_CB,
+
+    /** receive packets via file descriptor */
+    SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_FD,
+
+    /** receive packets via Linux netdev type port */
+    SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_PHYSICAL_PORT,
+
+    /** receive packets via Linux netdev logical port (LAG or port) */
+    SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_LOGICAL_PORT,
```
atitjain: ** receive packets via Linux netdev logical port (LAG or port) - is this a superset of *_PHYSICAL_PORT.

inc/saihostintf.h
```diff
@@ -186,7 +191,7 @@ typedef enum _sai_hostif_trap_type_t
     /** default action is forward */
     SAI_HOSTIF_TRAP_TYPE_IGMP_TYPE_V3_REPORT = 0x000000009,
```
atitjain: put SAI_HOSTIF_TRAP_TYPE_START and SAI_HOSTIF_TRAP_TYPE_END Macros, like other objects.

inc/saihostintf.h
```diff
@@ -692,6 +587,179 @@ typedef sai_status_t(*sai_get_hostif_attribute_fn)(
...
+    *
+    * Valid only when #SAI_HOSTIF_TABLE_ENTRY_ATTR_TYPE == #SAI_HOST_INTERFACE_TABLE_ENTRY_TYPE_PORT || 
+    * #SAI_HOST_INTERFACE_TABLE_ENTRY_TYPE_LAG || #SAI_HOST_INTERFACE_TABLE_ENTRY_TYPE_VLAN ||
+    * #SAI_HOST_INTERFACE_TABLE_ENTRY_TYPE_TRAP_ID
+    * @type sai_object_id_t
+    * @objects SAI_OBJECT_TYPE_HOSTIF_TRAP
+    * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+    * @condition SAI_HOSTIF_TABLE_ENTRY_ATTR_TYPE == SAI_HOST_INTERFACE_TABLE_ENTRY_TYPE_PORT or SAI_HOSTIF_TABLE_ENTRY_ATTR_TYPE == SAI_HOST_INTERFACE_TABLE_ENTRY_TYPE_VLAN or SAI_HOSTIF_TABLE_ENTRY_ATTR_TYPE == SAI_HOST_INTERFACE_TABLE_ENTRY_TYPE_LAG or SAI_HOSTIF_TABLE_ENTRY_ATTR_TYPE == SAI_HOST_INTERFACE_TABLE_ENTRY_TYPE_TRAP_ID
+    */
+    SAI_HOSTIF_TABLE_ENTRY_ATTR_TRAP_ID,
```
atitjain: trap id has a exclude port list. Hostif table entry has a port (SAI_HOSTIF_TABLE_ENTRY_ATTR_OBJ_ID), how do they interact? If I have a port here that is excluded in the trap what is expected.

inc/saihostintf.h
```diff
@@ -692,6 +587,179 @@ typedef sai_status_t(*sai_get_hostif_attribute_fn)(
...
+    */
+    SAI_HOSTIF_TABLE_ENTRY_ATTR_START,
+
+    /**
+    * @brief Host interface table entry type
+    *
+    * @type sai_host_interface_table_entry_type_t
+    * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+    */
+    SAI_HOSTIF_TABLE_ENTRY_ATTR_TYPE = SAI_HOSTIF_ATTR_START,
```
atitjain: why is the type needed, the type part should be decoded from SAI_HOSTIF_TABLE_ENTRY_ATTR_OBJ_ID

inc/saihostintf.h
```diff
@@ -330,8 +350,7 @@ typedef enum _sai_hostif_trap_attr_t
      * @brief Trap action
      *
      * @type sai_packet_action_t
-     * @flags CREATE_AND_SET
-     * @default SAI_PACKET_ACTION_FORWARD
+     * @flags CREATE_AND_SET | MANDATORY_ON_CREATE
      */
     SAI_HOSTIF_TRAP_ATTR_PACKET_ACTION,
```
atitjain: I think this attribute should move. The action should be definable based on per ingress port etc. So if same trap is applied on different ports, the action could be different.

inc/saihostintf.h
```diff
@@ -353,35 +372,14 @@ typedef enum _sai_hostif_trap_attr_t
     SAI_HOSTIF_TRAP_ATTR_TRAP_PRIORITY,
```
atitjain: should be per trap, per ingress source.

inc/saihostintf.h
```diff
@@ -692,6 +587,179 @@ typedef sai_status_t(*sai_get_hostif_attribute_fn)(
...
+    /** receive packets via Linux netdev logical port (LAG or port) */
+    SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_LOGICAL_PORT,
+
+    /** receive packets via Linux netdev L3 interface */
+    SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_L3
+
+} sai_host_interface_table_entry_channel_type_t;
+
+/**
+* @brief Host interface table entry attribute IDs
```
atitjain: Please add some explanation for "Host interface table entry"

inc/saihostintf.h
```diff
@@ -692,6 +587,179 @@ typedef sai_status_t(*sai_get_hostif_attribute_fn)(
...
+/**
+* @brief Create host interface table entry
+*
+* @param[out] hif_table_entry Host interface table entry
+* @param[in] attr_count Number of attributes
+* @param[in] attr_list Aarray of attributes
+*
+* @return #SAI_STATUS_SUCCESS on success Failure status code on error
+*/
+typedef sai_status_t(*sai_create_hostif_table_entry_fn)(
```
tushar-ty: Is the host intf document updated to describe this change ?
Would need an example to understand the new flow/usage.

inc/saihostintf.h
```diff
@@ -212,101 +217,116 @@ typedef enum _sai_hostif_trap_type_t
...
+    /** neighbor table traps (default packet action is drop) */
+    SAI_HOSTIF_TRAP_TYPE_NEIGH_MIN = SAI_HOSTIF_TRAP_TYPE_ROUTER_MAX + 1,
+
+    /** neighbor table traps max */
+    SAI_HOSTIF_TRAP_TYPE_NEIGH_MAX = SAI_HOSTIF_TRAP_TYPE_NEIGH_MIN + SAI_HOSTIF_TRAP_TYPE_TABLE_RANGE,
 
-    /** receive packets via OS net device */
-    SAI_HOSTIF_TRAP_CHANNEL_NETDEV,
+    /** ACL traps (default packet action is drop) */
+    SAI_HOSTIF_TRAP_TYPE_ACL_MIN = SAI_HOSTIF_TRAP_TYPE_NEIGH_MAX + 1,
```
tushar-ty: Do you need SAI_HOSTIF_TRAP_TYPE_TABLE_RANGE here as well ?

itaibaz: For each type (Router, neigh, ACL, FDB)
We allow SAI_HOSTIF_TRAP_TYPE_TABLE_RANGE
So the MAX trap ID for the type is MIN+SAI_HOSTIF_TRAP_TYPE_TABLE_RANGE
Then the next type is MAX+1

inc/saihostintf.h
itaibaz: There is explanation of the host interface table entry and flows in proposal document v9 which is already committed in github
https://github.com/opencomputeproject/SAI/commit/8f277b732c4f0c369a56ce125c1304fff6400b1a

inc/saihostintf.h
itaibaz: The motivation is that the most common use case is to have the trap enabled for all the ports
With the old attribute, port list, the default is empty list. So the application had to create a list of all the ports and provide it for this use case. In addition, whenever breaking out a port or creating a LAG, the application would have had to update this list.
With the new attribute, exclude port list, it is very easy to address the global use case. Just provide an empty list of ports to exclude, and the trap will be enabled for all ports. 
And still, with the new attribute, it is possible to have the trap enabled only for a subset of ports.

inc/saihostintf.h
itaibaz: Not exactly
When the channel is physical port, if the packet ingressed from port or LAG, it will be muxed to port netdev
When the channel is logical port, if the packet ingressed from LAG, it will be muxed to LAG netdev. If the packet ingressed from port, it will be muxed to port netdev.
Example for choosing physical port channel is LACP protocol, where you probably want to receive the packet on the port netdev and not LAG netdev.
Mux flow is explained also in the proposal doc

inc/saihostintf.h
itaibaz: In the new spec, you have to explicitly create traps, to match the unified object model.
Before you create a trap object, the behavior chosen is 
Drop L2 PDUs, IP2ME, pipeline exceptions
Forward L3 protocols
Drop user defined traps CPU copy

This choice is done for clean behavior.
If application wants to receive traps, it needs to explicitly create them. And so is the case for sample packet - application just needs to create the trap for it

inc/saihostintf.h
```diff
@@ -692,6 +587,179 @@ typedef sai_status_t(*sai_get_hostif_attribute_fn)(
...
+    /**
+    * @brief Host interface table entry action target host interface object
+    *
+    * Valid only when #SAI_HOSTIF_TABLE_ENTRY_ATTR_CHANNEL = #SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_FD
+    * @type sai_object_id_t
+    * @objects SAI_OBJECT_TYPE_HOSTIF
+    * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+    * @condition SAI_HOSTIF_TABLE_ENTRY_ATTR_CHANNEL == SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_FD
+    */
+    SAI_HOSTIF_TABLE_ENTRY_ATTR_HOST_IF,
```
itaibaz: You can create multiple host interface objects of type file descriptor
You can then mux the packet to these objects, according to hitting match criteria in the host table entry, and executing the channel + object mux action 
The packet doesn't have file descriptor info in it

inc/saihostintf.h
tushar-ty: 1. It seems there is a discrepancy in the doc and this header as the sai_create_hostif_trap_group_fn's are renamed to sai_create_hostif_group_fn's ?
The examples have that issue too.

2. There are several typos in the example for the host table. The attrib variable name should be changed and the eventual function name is wrong too.

3. Is the host table object not associated with any other objects in this module ?

itaibaz: I will go over the doc, and fix some of the typos
However please note the intention of the doc is to explain the idea and flows, and the examples aren't fully compiling.
The header file is fully accurate.

Re association of host table entries with other objects :
1) When filling host table entry action channel SAI_HOSTIF_TABLE_ENTRY_ATTR_CHANNEL=SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_FD, you are also filling SAI_HOSTIF_TABLE_ENTRY_ATTR_HOST_IF, which references a host IF object of type SAI_HOSTIF_TYPE_FD, and muxes the packet to that object.
2) When filling SAI_HOSTIF_TABLE_ENTRY_ATTR_TRAP_ID, you are filling a trap object ID created by sai_create_hostif_trap_fn.

itaibaz: I updated the doc in
https://github.com/opencomputeproject/SAI/pull/281

inc/saihostintf.h
atitjain: so you mean the exclude list takes priority?

inc/saihostintf.h
```diff
@@ -692,6 +597,179 @@ typedef sai_status_t(*sai_get_hostif_attribute_fn)(
...
+    */
+    SAI_HOSTIF_TABLE_ENTRY_ATTR_END,
+
+    /** Custom range base value */
+    SAI_HOSTIF_TABLE_ENTRY_ATTR_CUSTOM_RANGE_START = 0x10000000,
+
+    /** End of custom range base */
+    SAI_HOSTIF_TABLE_ENTRY_ATTR_CUSTOM_RANGE_END
+
+} sai_hostif_table_entry_attr_t;
```
lguohan: what is the priority among those host entries? Do we need a priority attribute for the hostif table entry?

itaibaz: Match is according to this order : 
{Port,trap_id}, {LAG,trap_id},{vlan,trap_id}, {wildcard,trap_id},{wildcard, wildcard}
Process is described in proposal doc

inc/saihostintf.h
```diff
@@ -212,101 +222,121 @@ typedef enum _sai_hostif_trap_type_t
...
+    /** Router traps custom range start */
+    SAI_HOSTIF_TRAP_TYPE_ROUTER_CUSTOM_RANGE_BASE = 0x0003000,
+
+    /* local IP traps */
 
     /**
      * @brief IP packets to local router IP address (routes with
      * #SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID = #SAI_SWITCH_ATTR_CPU_PORT)
-     * (default packet action is trap)
+     * (default packet action is drop)
```
atitjain: what is the difference between this and the following ones which are also IP2ME like:
SAI_HOSTIF_TRAP_TYPE_SSH 
SAI_HOSTIF_TRAP_TYPE_SNMP 
SAI_HOSTIF_TRAP_TYPE_BGP etc.

This looks like a superset. Do you mean a better match gets more priority.

itaibaz: Correct. SSH, SNMP, BGP are more specific cases of IP2ME with higher prio
Please also note this hasn't changed at all from the previous spec version definition

inc/saihostintf.h
atitjain: @itaibaz ideally it should be towards the better match, in the above the left is highest priority or rightmost.

itaibaz: Port, trap id is highest prio
wildcard, wildcard the lowest prio

inc/saihostintf.h
```diff
@@ -148,16 +148,26 @@ typedef sai_status_t(*sai_get_hostif_trap_group_attribute_fn)(
         _Inout_ sai_attribute_t *attr_list);
 
 /**
+* @brief Host interface user defined trap type table range
+*/
+#define SAI_HOSTIF_TRAP_TYPE_TABLE_RANGE 0x1
```
aarthygnanamurthy: sai_hostif_trap_type_t is being used as the key and needs to be unique for each trap entry created. Having this SAI_HOSTIF_TRAP_TYPE_TABLE_RANGE value set to 1 limits the no.of traps to 2 for a particular type. How do we need to handle in case of table size more than 2 i.e 255 or 1024 entries ?

lguohan:
Looks like we are good here? Plan to check-in by the end of this week.

inc/saihostintf.h
AshokDaparthi: @lguohan @kcudnik @itaibaz - 
 /**
     * @brief Host interface trap type []
     *
     * @type sai_hostif_trap_type_t
     */
    SAI_HOSTIF_TRAP_ATTR_TRAP_TYPE = SAI_HOSTIF_TRAP_ATTR_START,

What was meaning of KEY here,  will SAI allow creating only object with this TYPE or it just indication to application like this is input for creating object.

itaibaz:
The meaning is there can be up to one object instance of a given type
For example, there can be exactly zero or one trap object of type BGP. It is not allowed to create 2 objects of trap type BGP, as both would be managing the behavior for the same trap.

AshokDaparthi:
@itaibaz - We understand that for regular traps. BGP/STP will have only one trap. 
in case of ACL trap
/** ACL traps (default packet action is drop) */
 +    SAI_HOSTIF_TRAP_TYPE_ACL_MIN = SAI_HOSTIF_TRAP_TYPE_NEIGH_MAX + 1,
       SAI_HOSTIF_TRAP_TYPE_ACL_MAX = SAI_HOSTIF_TRAP_TYPE_ACL_MIN + SAI_HOSTIF_TRAP_TYPE_TABLE_RANGE,

That means i can not create more than 1 ACL trap Id's right? if no,  can you add example to configure multiple ACL traps and use it by application.

itaibaz:
There is just a single enum set :

/**
* @brief Attribute data for SAI_HOSTIF_TABLE_ENTRY_ATTR_CHANNEL_TYPE
*/
typedef enum _sai_hostif_table_entry_channel_type_t
{
    /** receive packets via callback */
    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_CB,

    /** receive packets via file descriptor */
    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_FD,

    /** receive packets via Linux netdev type port */
    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_PHYSICAL_PORT,

    /** receive packets via Linux netdev logical port (LAG or port) */
    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_LOGICAL_PORT,

    /** receive packets via Linux netdev L3 interface */
    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_L3

} sai_hostif_table_entry_channel_type_t;

You might be seeing in the history the older enum set, that was renamed

tushar-ty:
No,

I’m seeing this with history turned off in the specification section.



    /*host intarfce table entry action channel  [
sai_host_interface_table_entry_channel_type_t]

     Valid only when
SAI_HOST_INTERFACE_TABLE_ENTRY_ACTION_USER_CHANNEL_TYPE ==
SAI_HOST_INTERFACE_ENTRY_USER_CHANNEL_FD

     * (MANDATORY_ON_CREATE | CREATE_ONLY)  */

     SAI_HOST_INTERFACE_TABLE_ENTRY_ACTION_CHANNEL,





    /*host intarfce table entry action user channel type [
sai_host_interface_table_user_channel_t]

     * (MANDATORY_ON_CREATE | CREATE_ONLY)  */

     SAI_HOST_INTERFACE_TABLE_ENTRY_ACTION_USER_CHANNEL_TYPE,





And the following in the examples:





sai_if_channel_attr[2].id= SAI_HOSTIF_TABLE_ENTRY_ATTR_CHANNEL;

sai_if_channel_attr[2].value=SAI_HOST_INTERFACE_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_LOGICAL_PORT;



sai_create_hostif_table_entry_fn(&host_table_entry[0], 3,
sai_if_channel_attr);





It’s a bit confusing to understand.



-thanks





*From:* Itai Baz [mailto:notifications@github.com]
*Sent:* Thursday, November 10, 2016 12:58 PM
*To:* opencomputeproject/SAI
*Cc:* Tushar Tyagi; Mention
*Subject:* Re: [opencomputeproject/SAI] host if spec changes for 1.0 (#266)



There is just a single enum set :

/**
* @brief Attribute data for SAI_HOSTIF_TABLE_ENTRY_ATTR_CHANNEL_TYPE
*/
typedef enum _sai_hostif_table_entry_channel_type_t
{
/** receive packets via callback */
SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_CB,

/** receive packets via file descriptor */
SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_FD,

/** receive packets via Linux netdev type port */
SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_PHYSICAL_PORT,

/** receive packets via Linux netdev logical port (LAG or port) */
SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_LOGICAL_PORT,

/** receive packets via Linux netdev L3 interface */
SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_L3

} sai_hostif_table_entry_channel_type_t;

You might be seeing in the history the older enum set, that was renamed

itaibaz:
The headers version is the most updated
There was a slight rename enforced by the meta data tool
The document uses the old name, but I would hope it is still understandable

inc/saihostintf.h
```diff
@@ -598,16 +509,19 @@ typedef enum _sai_hostif_attr_t
     SAI_HOSTIF_ATTR_TYPE = SAI_HOSTIF_ATTR_START,
 
     /**
-     * @brief Assosiated port or router interface
+     * @brief Host interface object ID
      *
-     * Valid only when #SAI_HOSTIF_ATTR_TYPE == #SAI_HOSTIF_TYPE_NETDEV
+     * Valid only when #SAI_HOSTIF_ATTR_TYPE == #SAI_HOSTIF_TYPE_NETDEV
```
andriykokhan: Shouldn't it be valid for SAI_HOSTIF_TYPE_PORT/LAG/VLAN as well?

inc/saihostintf.h
```diff
@@ -214,101 +224,121 @@ typedef enum _sai_hostif_trap_type_t
...
+
+    /** Custom range base */
+    SAI_HOSTIF_TRAP_TYPE_CUSTOM_RANGE_BASE = 0x00009000,
+
+    /**
+    * @brief End of trap types
+    */
+    SAI_HOSTIF_TRAP_TYPE_END,
+
+} sai_hostif_trap_type_t;
```
andriykokhan: Does KEY flag in SAI_HOSTIF_TRAP_ATTR_TRAP_TYPE description mean that its allowed to created only one trap of a specific type?

inc/saihostintf.h
```diff
@@ -461,120 +469,23 @@ typedef sai_status_t(*sai_get_hostif_trap_attribute_fn)(
...
+    SAI_HOSTIF_TYPE_VLAN,
 
-/**
- * @brief Attribute data for SAI_HOSTIF_ATTR_TYPE
- */
-typedef enum _sai_hostif_type_t
-{
-    /** Netdev-based Host Interface Type */
+    /** Netdevice */
     SAI_HOSTIF_TYPE_NETDEV,
```
tushar-ty: Pls clarify the purpose of the above three types - port, lag, vlan vs the Netdev. How will the other three types be used if the netdev itself can be of the same three types. An example will help.

inc/saihostintf.h
```diff
@@ -698,6 +612,181 @@ typedef sai_status_t(*sai_get_hostif_attribute_fn)(
...
+    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_CB,
+
+    /** receive packets via file descriptor */
+    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_FD,
+
+    /** receive packets via Linux netdev type port */
+    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_PHYSICAL_PORT,
+
+    /** receive packets via Linux netdev logical port (LAG or port) */
+    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_LOGICAL_PORT,
```
tushar-ty: The purpose of physical port type is not clear when the type logical port itself can cover both lag and port. We are not exposing physical port object anywhere else in SAI. An example will help.

inc/saihostintf.h
lguohan: @tushar-ty , I cannot find the sai_host_interface_table_user_channel_t in the latest header file.

inc/saihostintf.h
lguohan: correct

inc/saihostintf.h
```diff
@@ -461,120 +469,23 @@ typedef sai_status_t(*sai_get_hostif_trap_attribute_fn)(
...
  */
-typedef enum _sai_hostif_user_defined_trap_attr_t
+typedef enum _sai_hostif_type_t
 {
-    /**
-     * @brief Start of attributes
-     */
-    SAI_HOSTIF_USER_DEFINED_TRAP_ATTR_START,
+    /** Port-based Host Interface Type */
+    SAI_HOSTIF_TYPE_PORT,
```
lguohan: @itaibaz , please remove SAI_HOSTIF_TYPE_PORT/LAG/VLAN in the latest header, it is not in redundant.

inc/saihostintf.h
lguohan: it should be removed, only SAI_HOSTIF_TYPE_NETDEV/FD is needed.

inc/saihostintf.h
lguohan: @tushar-ty , SAI_HOSTIF_TYPE_PORT/LAG/VLAN should be removed according to the latest doc.

inc/saihostintf.h
lguohan: the logical_port is different then physical port in the host table match in terms of directing the packets to the correct netdev.

look at the figure on page 6, when the action is phy_port, the packet will be directing the port-based net dev based on its incoming port.

when the action is logical port, it will first check if the incoming port belongs to a LAG or not, if it is then it will send to the LAG net dev, otherwise, it will send the port net dev.

for example, imagine you have two port-base netdev and one lag-base netdev. 
For lldp packet, you will send them to net_phy_port.
For stp, you will send them to net_logical_port.

inc/saihostintf.h
```diff
@@ -321,19 +349,38 @@ typedef enum _sai_hostif_trap_attr_t
     SAI_HOSTIF_TRAP_ATTR_START,
 
     /**
+    * @brief Is user defined trap
+    *
+    * @type bool
+    * @flags CREATE_ONLY
+    * @default false
+    */
+    SAI_HOSTIF_TRAP_ATTR_IS_USER_DEFINED_TRAP = SAI_HOSTIF_TRAP_ATTR_START,
```
andriykokhan: What was the motivation to move user defined traps into the separate enum? Does it limits somehow the list of supported SAI_HOSTIF_TRAP_ATTRs by user defined traps? E.g., is it valid to use EXCLUDE_PORT_LIST or GROUP for user defined traps?

inc/saihostintf.h
```diff
@@ -861,8 +956,6 @@ typedef struct _sai_hostif_api_t
     sai_remove_hostif_trap_fn                      remove_trap;
     sai_set_hostif_trap_attribute_fn               set_trap_attribute;
     sai_get_hostif_trap_attribute_fn               get_trap_attribute;
-    sai_set_hostif_user_defined_trap_attribute_fn  set_user_defined_trap_attribute;
-    sai_get_hostif_user_defined_trap_attribute_fn  get_user_defined_trap_attribute;
     sai_recv_hostif_packet_fn                      recv_packet;
     sai_send_hostif_packet_fn                      send_packet;
```
andriykokhan: How to get Switch ID in CB mode ( hif_id == SAI_NULL_OBJECT_ID )?

/**
 * @brief Hostif send function
 *
 * @param[in] hif_id Host interface id. only valid for send through FD channel. Use #SAI_NULL_OBJECT_ID for send through CB channel.
 * @param[in] buffer Packet buffer
 * @param[in] buffer size Packet size in bytes
 * @param[in] attr_count Number of attributes
 * @param[in] attr_list Array of attributes
 *
 * @return #SAI_STATUS_SUCCESS on success Failure status code on error
 */
typedef sai_status_t(*sai_send_hostif_packet_fn)(
        _In_ sai_object_id_t hif_id,
        _In_ void *buffer,
        _In_ sai_size_t buffer_size,
        _In_ uint32_t attr_count,
        _In_ sai_attribute_t *attr_list);

itaibaz:
Thanks
I submitted a fix for this
When you want to send a packet through CB channel, pass the switch ID object

inc/saihostintf.h
```diff
@@ -214,101 +219,124 @@ typedef enum _sai_hostif_trap_type_t
...
+    SAI_HOSTIF_TRAP_TYPE_TTL_ERROR = 0x00006001,
 
     /** Exception traps custom range start */
-    SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE = 0x00005000,
+    SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE = 0x00007000,
+
+    /**
+     * @brief End of trap types
+     */
+    SAI_HOSTIF_TRAP_TYPE_END,
```
andriykokhan: Shouldn't it be (SAI_HOSTIF_TRAP_USER_DEFINED_TYPE_END = SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE + 0x1000) ?

lguohan:
there is a conflict in inc/saihostintf.h

inc/saihostintf.h
lguohan: looks like we need to add switch_id as a parameter

inc/saihostintf.h
itaibaz: the user defined type and trap type are in 2 separated enums

andriykokhan: Sorry, meant SAI_HOSTIF_TRAP_TYPE_END. Corrected.

meta/saisanitycheck.c
```diff
@@ -849,14 +849,13 @@ void check_attr_conditions(
...
-                    md->objecttype != SAI_OBJECT_TYPE_HOSTIF_TRAP &&
-                    md->objecttype != SAI_OBJECT_TYPE_HOSTIF_USER_DEFINED_TRAP)
+                    md->objecttype != SAI_OBJECT_TYPE_HOSTIF_TRAP)
             {
                 META_ASSERT_FAIL(md, "marked as conditional on non mirror session /hostif_trap/hostif_user_defined_trap, but invalid creation flags: 0x%u", md->flags);
             }
 
             break;
-
+        case SAI_ATTR_FLAGS_MANDATORY_ON_CREATE | SAI_ATTR_FLAGS_CREATE_ONLY | SAI_ATTR_FLAGS_KEY:
```
kcudnik: SAI_ATTR_FLAGS_KEY attributes should not be conditional, please remove and revisit your change

itaibaz: Here is the motivation
Trap object can either be regular trap or user defined trap (controlled by SAI_HOSTIF_TRAP_ATTR_IS_USER_DEFINED_TRAP)
When it is regular trap, SAI_HOSTIF_TRAP_ATTR_TRAP_TYPE field is a key, meaning there can be exactly one instance of each type.
When it is user defined trap, we allow creating multiple instances of each SAI_HOSTIF_TRAP_ATTR_USER_DEFINED_TYPE.
So the key field is conditional if it's trap or user defined trap
Please explain whether I should remove the key field, but then we won't be able to specify we allow just one of each type

kcudnik: In previous design user defined trap was different object, making kay conditional will make this hard to maintain

itaibaz: Based on discussion Guohan and Matty had, and additional inputs, it was decided to unify trap and user defined trap to one object
Perhaps we should have more notions of keys ? Primary key (unique non null, meaning mandatory) and unique key (conditional, allow nulls, unique when value is used). In this case, we will use the unique key, vs in other cases we will use primary key when the attribute is mandatory

lguohan:
@itaibaz , after I merged the ACL, there is now a conflict on the header files.

itaibaz:
Done

inc/saihostintf.h
wendani: In SAI behavioral/pipeline model, all tables use the match-action primitive. Does this primitive have the TCAM-like behavior that a key can match multiple entries. If multiple entires hit, only the highest-priority (lowest-index) one wins? In the latest host-if proposal (ver9, https://github.com/opencomputeproject/SAI/pull/281) 2.2.1 Switch trap -> user channel port example, if an L2 control protocol packet comes from a physical port, which is part of a LAG, its metadata will be IN port=N, and IN l2_if=M, it can match two entries in the Host Interface table---{l2_if, trap_id} and {IN port, trap_id}.

PR #267, opened 2016-10-20
add mandatory attributes for STP PORT

lguohan:


lguohan:
@krambn please review

PR #274, opened 2016-10-26
saithrift: No need to parse LAG attributes on create

vkochan-mlnx:
As SAI_LAG_ATTR_PORT_LIST became read only there are noe more
LAG attributes for create operation, thus lets remove not needed
attributes parsing.


stcheng:
ack

PR #282, opened 2016-10-28
BST Feature API Proposal

bandaru-viswanath:
Updated based on community feedback


sonicyao1985:
In case transporting the statistics to a remote IP, how will the statistics data be organized in a data packet? I think this part should be done by SAI control plane, not in SAI implementation.

bandaru-viswanath:
Much of these questions have been answered in the updated document which has been renamed as TAM. Closing this request and opening a new pull request for the updated proposal.

PR #283, opened 2016-10-29
saithrift: tests: Remove default vlan before add ports to the LAG

vkochan-mlnx:
Regarding the following discussed pull request:

```
https://github.com/opencomputeproject/SAI/pull/86
```

Port must be cleaned from .1q bridge attributes (VLANs) before
adding to the LAG.

Added separate function to remove all ports from the specified vlan
which is used from few places in sail2.py test module.

Signed-off-by: Vadim Kochan vadymk@mellanox.com

### DIFF
--- a/test/saithrift/tests/sail2.py
+++ b/test/saithrift/tests/sail2.py
@@ -19,6 +19,19 @@ import socket
 from switch import *
 import sai_base_test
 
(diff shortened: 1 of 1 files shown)


vkochan-mlnx:
will rework in new pull request ...

PR #284, opened 2016-10-29
saithrift: tests: Fix pvid resetting in sail2.py tests

vkochan-mlnx:
Use correct port1 for resetting pvid on finally stage.

Signed-off-by: Vadim Kochan vadymk@mellanox.com


test/saithrift/tests/sail2.py
```diff
@@ -210,7 +210,7 @@ def runTest(self):
 
             attr_value = sai_thrift_attribute_value_t(u16=1)
             attr = sai_thrift_attribute_t(id=SAI_PORT_ATTR_PORT_VLAN_ID, value=attr_value)
-            self.client.sai_thrift_set_port_attribute(port2, attr)
+            self.client.sai_thrift_set_port_attribute(port1, attr)
```
lguohan: in this test, since port2 is untagged, i think the line 186 should be changed to port2 instead of here?

PR #286, opened 2016-11-01
Modify SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE to accept a list

yakivh:
The ability to pass only one value for SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE attribute
disallows the creation of SAI Entries with different range types within one ACL Table.

Signed-off-by: Yakiv Huryk <yakivh@mellanox.com>

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -816,7 +816,7 @@ typedef enum _sai_acl_table_attr_t
     /**
      * @brief Range type defined
      *
-     * @type sai_acl_range_type_t
+     * @type sai_s32_list_t sai_acl_range_type_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      */
     SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE,


inc/saiacl.h
```diff
@@ -816,7 +816,7 @@ typedef enum _sai_acl_table_attr_t
     /**
      * @brief Range type defined
      *
-     * @type sai_acl_range_type_t
+     * @type sai_s32_list_t sai_acl_range_type_t
```
lguohan: I think the original type is incorrect, it should be a bool.

lguohan: discussed with @itaibaz , it seems s32_list is needed by mellanox since they need to know specify range type at the acl table creation time. @AshokDaparthi , please review.

kcudnik: list suggest you can have multiple range types is that correct ?

itaibaz: Yes, you can put any item from sai_acl_range_type_t :

    /** L4 Source Port Range */
    SAI_ACL_RANGE_L4_SRC_PORT_RANGE,

    /** L4 Destination Port Range */
    SAI_ACL_RANGE_L4_DST_PORT_RANGE,

    /** Outer Vlan Range */
    SAI_ACL_RANGE_OUTER_VLAN,

    /** Inner Vlan Range */
    SAI_ACL_RANGE_INNER_VLAN,

    /** Packet Length Range in bytes */
    SAI_ACL_RANGE_PACKET_LENGTH

So currently there can be up to 5 items

kcudnik: so if you set for example L4 src port range and outer vlan, and then you add acl entry which is range packet length what then ?

itaibaz: It is no different than any other match field, which are all create only - 
CREATE_ONLY, match fields cannot be changed after the table is created
When you create the table, you need to specify all the possible match fields in the key
If you then try to create an entry with different keys, you should get an error

lguohan: looks like we are good here?

lguohan: Plan to check in by the end of this week.

PR #287, opened 2016-11-02
make sure sai_pointer_t also need default value

lguohan:


meta/saisanitycheck.c
```diff
@@ -527,12 +527,6 @@ void check_attr_default_required(
             return;
         }
 
-        if (md->attrvaluetype == SAI_ATTR_VALUE_TYPE_POINTER)
-        {
-            /* Pointer or String may not have default value */
-            return;
-        }
-
```
AshokDaparthi: do we have have keyword to represent "@default" as  NULL or null. Can we use the 0.

lguohan: update the check. Please use @default NULL for sai_pointer_t

PR #288, opened 2016-11-03
Port ADD/DELETE Notifcation removed

AshokDaparthi:
As SAI supports explicit port Create and Port delete API's, Port event notification is redundant.


AshokDaparthi:
@itaibaz @tushar-ty @lguohan @atitjain - Please review

PR #247, opened 2016-09-30
[copp] Added copp tests

hrachya-mellanox:
- List of tests:
- ARPTest
- DHCPTest
- LLDPTest
- BGPTest
- LACPTest
- SNMPTest
- SSHTest
- IP2METest
- DefaultTest
- The test closely replicates SONIC's setup for copp, for the trap ids mentioned
- above.

Signed-off-by: hrachya@mellanox.com

PR #290, opened 2016-11-03
saithrift: tests: Fix sail2/3 tests failing

vkochan-mlnx:
After 4042cf9 initializing of test_params was moved to
createRpcClient() but they are used earlier in loadPortMap()
and it caused sail2.py & sail3.py fail.

Fixed by move initializing test params earlier in setUp().

Fixes: 4042cf9 ("[copp] Added copp tests (#247)")


lguohan:
fixed in https://github.com/opencomputeproject/SAI/commit/f057001d9687568352e3fb24ec2d0a1c89cc8a09

Thanks for the patch!

PR #299, opened 2016-11-08
Add IPv6PrefixTest

taoyl-ms:


test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
...
+        ip_addr1 = '1234:5678:9abc:def0:4422:1133:5577:99aa'
+        ip_addr1_int = ip6_to_integer(ip_addr1)
+
+        dmac1 = '00:11:22:33:44:55'
+        sai_thrift_create_neighbor(self.client, addr_family, rif_id1, ip_addr1, dmac1)
+        nhop1 = sai_thrift_create_nhop(self.client, addr_family, ip_addr1, rif_id1)
+        
+        pkt = simple_tcpv6_packet( eth_dst=router_mac,
+                                eth_src='00:22:22:22:22:22',
+                                ipv6_dst='1234:5678:9abc:def0:4422:1133:5577:99aa',
```
qiluo-msft: Do you need a IPv6 unicast address here?

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
...
+                sai_thrift_remove_route(self.client, vr_id, addr_family, net, mask, rif_id1)
+            self.client.sai_thrift_remove_next_hop(nhop1)
+            sai_thrift_remove_neighbor(self.client, addr_family, rif_id1, ip_addr1, dmac1)
+
+            self.client.sai_thrift_remove_router_interface(rif_id1)
+            self.client.sai_thrift_remove_router_interface(rif_id2)
+
+            self.client.sai_thrift_remove_virtual_router(vr_id)
+
+
```
qiluo-msft: remove one empty line?

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
...
+        a, b = unpack(">QQ", ip6)
+        return (a << 64) | b
+
+def integer_to_ip6(ip6int):
+        a = (ip6int >> 64) & ((1 << 64) - 1)
+        b = ip6int & ((1 << 64) - 1)
+        return socket.inet_ntop(socket.AF_INET6, pack(">QQ", a, b))
+        
+@group('l3')
+class L3IPv6PrefixTest(sai_base_test.ThriftInterfaceDataPlane):
```
lguohan: can you add a description to describe this test, what does it test? something like test packet forwarding for all IPv6 prefix lenghs (from 0 to 128)

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
...
+                                ipv6_src='2000::1',
+                                ipv6_hlim=63)
+        try:
+            for i in range(128):
+                mask_int = ( ( 1 << (128-i) ) - 1 ) << i
+                net_int = ip_addr1_int & mask_int
+                mask = integer_to_ip6(mask_int)
+                net = integer_to_ip6(net_int)
+
+                print net + ' / ' + mask
```
lguohan: print "test packet forwarding for %s" % ipprefix_string

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
...
+        v4_enabled = 1
+        v6_enabled = 1
+        mac = ''
+
+        vr_id = sai_thrift_create_virtual_router(self.client, v4_enabled, v6_enabled)
+
+        rif_id1 = sai_thrift_create_router_interface(self.client, vr_id, 1, port1, 0, v4_enabled, v6_enabled, mac)
+        rif_id2 = sai_thrift_create_router_interface(self.client, vr_id, 1, port2, 0, v4_enabled, v6_enabled, mac)
+
+        addr_family = SAI_IP_ADDR_FAMILY_IPV6
```
lguohan: do you need to create a default route with drop to make the the packet is not accidental forward to port 1?

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
...
+def integer_to_ip6(ip6int):
+        a = (ip6int >> 64) & ((1 << 64) - 1)
+        b = ip6int & ((1 << 64) - 1)
+        return socket.inet_ntop(socket.AF_INET6, pack(">QQ", a, b))
+        
+@group('l3')
+class L3IPv6PrefixTest(sai_base_test.ThriftInterfaceDataPlane):
+    def runTest(self):
+        print
+        print "Sending packet port 1 -> port 2 (2000::1 -> 3000::1)"
```
lguohan: remove the message here

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
...
+        rif_id1 = sai_thrift_create_router_interface(self.client, vr_id, 1, port1, 0, v4_enabled, v6_enabled, mac)
+        rif_id2 = sai_thrift_create_router_interface(self.client, vr_id, 1, port2, 0, v4_enabled, v6_enabled, mac)
+
+        addr_family = SAI_IP_ADDR_FAMILY_IPV6
+        ip_addr1 = '1234:5678:9abc:def0:4422:1133:5577:99aa'
+        ip_addr1_int = ip6_to_integer(ip_addr1)
+
+        dmac1 = '00:11:22:33:44:55'
+        sai_thrift_create_neighbor(self.client, addr_family, rif_id1, ip_addr1, dmac1)
+        nhop1 = sai_thrift_create_nhop(self.client, addr_family, ip_addr1, rif_id1)
```
lguohan: you need to create a directly reachable route here to cover the next hop. check L160 in sail3.py

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
 
             self.client.sai_thrift_remove_virtual_router(vr_id)
 
+def ip6_to_integer(ip6):
+        ip6 = socket.inet_pton(socket.AF_INET6, ip6)
```
lguohan: tab space = 4

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
...
+                                eth_src=router_mac,
+                                ipv6_dst='1234:5678:9abc:def0:4422:1133:5577:99aa',
+                                ipv6_src='2000::1',
+                                ipv6_hlim=63)
+        try:
+            for i in range(128):
+                mask_int = ( ( 1 << (128-i) ) - 1 ) << i
+                net_int = ip_addr1_int & mask_int
+                mask = integer_to_ip6(mask_int)
+                net = integer_to_ip6(net_int)
```
lguohan: suggest to use https://pypi.python.org/pypi/netaddr to manipulate the ipv6 address instead of create our own.

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
...
+
+                print net + ' / ' + mask
+                sai_thrift_create_route(self.client, vr_id, addr_family, net, mask, rif_id1)
+                send_packet(self, 1, str(pkt))
+                verify_packets(self, exp_pkt, [0])
+                sai_thrift_remove_route(self.client, vr_id, addr_family, net, mask, rif_id1)
+                mask=""
+        finally:
+            if mask!="":
```
lguohan: the exception most like happen in verify_packet, the checking mask is not necessary. what if the exception happens inside sai_thrift_remove_route?

taoyl-ms: Though sai_thrift_remove_route might be called twice in this case, I don't see a problem here. What's your concern about that?

taoyl-ms: Though sai_thrift_remove_route might be called twice in this case, I don't see a problem here. What's your concern about that?

taoyl-ms: Though sai_thrift_remove_route might be called twice in this case, I don't see a problem here. What's your concern about that?

test/saithrift/tests/sail3.py
taoyl-ms: I don't want to make the testing environment complicated by introducing pip packages as the functionality we need from netaddr is minimum.

test/saithrift/tests/sail3.py
```diff
@@ -184,6 +187,77 @@ def runTest(self):
 
             self.client.sai_thrift_remove_virtual_router(vr_id)
 
+def ip6_to_integer(ip6):
```
lguohan: can we use the above api to manage the ip address?

taoyl-ms: To do that will need us to install that dependency in the ptf client docker. Will apply that change as an independent PR after finish the docker modification.

PR #302, opened 2016-11-11
Change in saifdb.h for CRUD. 

atitjain:
This is proposed as part of CRUD changes. The flush call is not one of the quad APIs. In stead, I remember during nexthop group and other CRUD change discussions we brought up the notion of bulk APIs. I feel
the bulk delete should be able to take care of the flush.

### DIFF
--- a/inc/saifdb.h
+++ b/inc/saifdb.h
@@ -142,76 +142,6 @@ typedef enum _sai_fdb_entry_attr_t
 
 } sai_fdb_entry_attr_t;
 
-/**
- * @brief FDB Flush entry type.
- */
-typedef enum _sai_fdb_flush_entry_type_t
-{
-    /** Flush dynamic FDB entries */
-    SAI_FDB_FLUSH_ENTRY_TYPE_DYNAMIC,
-
-    /** Flush static FDB entries */
-    SAI_FDB_FLUSH_ENTRY_TYPE_STATIC,
-
-} sai_fdb_flush_entry_type_t;
-
-/**
- * @brief Attribute for FDB flush API to specify the type of FDB entries being flushed.
- *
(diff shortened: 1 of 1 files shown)


lguohan:
@AshokDaparthi , can you comment?

kcudnik:
To make it compile remove FDB_FLUSH from object_type enum since you removed attribute

AshokDaparthi:
@atitjain @lguohan - Intention of these flush API's is to simplify the flush logic when STP state change etc. Instead of sending the FDB entries to removed as bulk or individual entry based on port, vlan , port+vlan or all, they can indicate just port oid, vlan or etc. Then underlying SDK's or hardware will delete those entries. this will be much faster than deleting entry as bulk/individual.

In case some ASIC's does not support this model, they can handle in their SAI.  SAI logical API model is to enforce support this model for future ASIC vendors.  

We have lot other non CRUD API's at present, that should not be reason for deleting this.

atitjain:
@AshokDaparthi 
Inline - 

Intention of these flush API's is to simplify the flush logic when STP state change etc. Instead of sending the FDB entries to removed as bulk or individual entry based on port, vlan , port+vlan or all, they can indicate just port oid, vlan or etc. Then underlying SDK's or hardware will delete those entries. this will be much faster than deleting entry as bulk/individual.
<Atit> I understand the usecase and ASIC implication. Only GAP I have probably is that I have not seen the bulk remove yet. In our SDK (HW assisted) we can do bulk remove and update etc on any set of key values with rest of those masked. 
In case some ASIC's does not support this model, they can handle in their SAI. SAI logical API model is to enforce support this model for future ASIC vendors.

We have lot other non CRUD API's at present, that should not be reason for deleting this.

lguohan:
the bulk remove api will be given a list of fdb entries and then remove. this flush api is to used to remove the fdb entries based on certain attribute of the fdb entries

PR #304, opened 2016-11-14
Commit header files for L2/L3 multicast

sonicyao1985:


lguohan:
@mattyk , please review

sonicyao1985:
@lguohan I've updated saitypes.h. but I'm getting ACL related errors:

saimetadata.c:22276:27: error: 'SAI_ACL_TABLE_GROUP_MEMBER_ATTR_START' was not declared in this scope
     .attridstart        = SAI_ACL_TABLE_GROUP_MEMBER_ATTR_START,
                           ^
saimetadata.c:22277:27: error: 'SAI_ACL_TABLE_GROUP_MEMBER_ATTR_END' was not declared in this scope
     .attridend          = SAI_ACL_TABLE_GROUP_MEMBER_ATTR_END,
                           ^
saimetadata.c:22278:28: error: 'metadata_enum_sai_acl_table_group_member_attr_t' was not declared in this scope
     .enummetadata       = &metadata_enum_sai_acl_table_group_member_attr_t,
                            ^
Makefile:61: recipe for target 'saisanitycheck' failed

kcudnik:
yes, you added table group member object but you didnt defined start and end attributes

kcudnik:
please convert +    SAI_OBJECT_TYPE_L2MC_ENTRY and SAI_OBJECT_TYPE_IPMC_ENTRY               to object_id and use fields as attributes mandatory on create and create_only

kcudnik:
mcast_fdb_entry has exactly the same fields as fdb_entry, how then they can be distinguised ?

kcudnik:
saiobject.h sai_object_key_t needs to be updated to support fdb_mcast_entry and ipmc and l2mc entries

sonicyao1985:
@kcudnik 
1. I'm not touching anything about ACL in this PR. So I'm confused about the error info.
2. Is this about the proposal of making all entries(fdb) as SAI objects? I think that is not merged. FDB is still a entry in currnet implementation. I can modify this once that proposal is merge. And I'm not sure if that proposal would be included in SAI 1.0.
3. By destination mac address in the keys. My original proposal is to re-use fdb entry for mcast fdb, but it is microsoft and mlnx's idea to seperate unicast/mcast fdb in code.
4. That I missed...will fix

lguohan:
@sonicyao1985, once you have resolve the conflict in saitypes.h, the acl error will go away, it is there due to saitype.h conflict.

lguohan:
@sonicyao1985 , about your 2nd question, please add those new fdb entries in sai_object_key_t.
@kcudnik, the mcast fdb entry use different dst mac than the ucast fdb entry.

kcudnik:
if mcast fdb_entry uses only different mac addres why do we need extra set of apis ?

lguohan:
because the attribute are different, for ucast fdb, there is entry type (dynamic or static) because the asic can do hw learning and create dynamic ucast fdb. For mcast fdb, it is always static, the asic cannot do any hw learning. 

If we merge ucast fdb and mcast fdb into one, then the entry type (dynamic or static) is conditional, we need to introduce another type to indicate whether the fdb is mcast or ucast.

PR #309, opened 2016-11-17
add bulk object creation and removal API for several sai object types

lguohan:
add bulk object creation and removal API for next_hop_member, stp_port, vlan_member, and lag_member

### DIFF
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -566,9 +566,11 @@ typedef struct _sai_attribute_t {
  *         of attribute for each object to create.
  *
  * @param[in] attrs List of attributes for every object.
+ * @param[out] object_statuses List of status for every object. Caller needs to allocate the buffer.
  *
- * @return #SAI_STATUS_SUCCESS on success when all objects are created. When
- * there is failure, none of the objects are created, and failure code is returned.
+ * @return #SAI_STATUS_SUCCESS on success when all objects are created or #SAI_STATUS_FAILURE when
+ * any of the objects fails to create. When there is failure, Caller is expected to go through the
+ * list of returned statuses to find out which fails and which succeeds.
  */
 
 typedef sai_status_t (*sai_bulk_object_create_fn)(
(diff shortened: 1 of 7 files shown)


lguohan:
@tushar-ty , @itaibaz , @atitjain , @krambn, @AshokDaparthi , @SathishKumarTR-Dell , @sonicyao1985 , please review.

inc/saitypes.h
```diff
@@ -557,6 +557,49 @@ typedef struct _sai_attribute_t {
...
+ *
+ * @return #SAI_STATUS_SUCCESS on success when all objects are created or #SAI_STATUS_FAILURE when
+ * any of the objects fails to create. When there is failure, Caller is expected to go through the
+ * list of returned statuses to find out which fails and which succeeds.
+ */
+
+typedef sai_status_t (*sai_bulk_object_create_fn)(
+        _In_ sai_object_id_t switch_id,
+        _In_ uint32_t object_count,
+        _Out_ sai_object_id_t *object_id,
```
tushar-ty: Would it look cleaner to have the two _Out_'s in the end of the param list. 
_Out_ sai_object_id_t *object_id, _Out_ sai_status_t *object_statuses ?

Everything else looks fine.

atitjain:
@lguohan Some thoughts:
Is this currently the infrastructure code and solves some of the earlier discussed issues during CRUD adaption. Do we have a plan to extend it to other objects. Rather the architecture we have for saiobject.h, which provides common getters and setters based on the object type could be considered for this. 
For example if some application with SW based learning does bulking of say FDB notifications and can do one shot FDB creations.

inc/saitypes.h
```diff
@@ -556,6 +556,63 @@ typedef struct _sai_attribute_t {
...
+ * @param[in] object_count Number of objects to create
+ * @param[in] attr_count List of attr_count. Caller passes the number
+ *         of attribute for each object to create.
+ * @param[in] attrs List of attributes for every object.
+ * @param[in] type bulk operation type.
+ *
+ * @param[out] object_id List of object ids returned
+ * @param[out] object_statuses List of status for every object. Caller needs to allocate the buffer.
+ *
+ * @return #SAI_STATUS_SUCCESS on success when all objects are created or #SAI_STATUS_FAILURE when
```
atitjain: I think you mean:
```..... #SAI_STATUS_SUCCESS on success when all objects are created or a failure code (sai_status_t) when any of the objects failes....```. 
The correct status could be used for further debug/analysis. Can we just clarify the comment so that implementation handles it properly.

File: inc/saitypes.h [hunk 3]
```diff
@@ -556,6 +556,63 @@ typedef struct _sai_attribute_t {
...
+ * @param[out] object_statuses List of status for every object. Caller needs to allocate the buffer.
+ *
+ * @return #SAI_STATUS_SUCCESS on success when all objects are removed or #SAI_STATUS_FAILURE when
+ * any of the objects fails to remove. When there is failure, Caller is expected to go through the
+ * list of returned statuses to find out which fails and which succeeds.
+ */
+
+typedef sai_status_t (*sai_bulk_object_remove_fn)(
+        _In_ uint32_t object_count,
+        _In_ sai_object_id_t *object_id,
```
atitjain: I think we could make use of:
```
typedef struct _sai_object_list_t {
    uint32_t count;
    sai_object_id_t *list;
} sai_object_list_t;
```

kcudnik:
Add bulk route_entry create/remove also ?

inc/saitypes.h
lguohan: leave it as it is to keep it in similar format as object_create

inc/saitypes.h
lguohan: I thin it is SAI_STATUS_FAILURE, the failure code for each individual object are encoded in the object_statuses return.

lguohan:
@atitjain , this function pointer is also universal to all UID based object. For other key based object, we can later introduce the bulk function for each type as the existing create object function.
@kcudnik , batch route will be added later.

PR #310, opened 2016-11-17
enforce default value check for all non mandatory_on_create attribute

lguohan:


lguohan:
@kcudnik , can you review?

PR #311, opened 2016-11-18
Flags for SAI_SWITCH_ATTR_*_HASH* should be CREATE_AND_SET

vveskera:
User should have the possibility to set created hash objects as an attribute for switch object, not only to read it.

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -558,11 +558,10 @@ typedef enum _sai_switch_attr_t
      * #SAI_NATIVE_HASH_FIELD_ETHERTYPE\]
      * #SAI_HASH_ATTR_UDF_GROUP_LIST empty list
      *
-     * The object id is read only, while the object attributes can be modified
      *
      * @type sai_object_id_t
      * @objects SAI_OBJECT_TYPE_HASH
-     * @flags READ_ONLY
+     * @flags CREATE_AND_SET
      */
     SAI_SWITCH_ATTR_ECMP_HASH,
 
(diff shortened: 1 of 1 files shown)


inc/saiswitch.h
```diff
@@ -558,11 +558,10 @@ typedef enum _sai_switch_attr_t
      * #SAI_NATIVE_HASH_FIELD_ETHERTYPE\]
      * #SAI_HASH_ATTR_UDF_GROUP_LIST empty list
      *
-     * The object id is read only, while the object attributes can be modified
      *
      * @type sai_object_id_t
      * @objects SAI_OBJECT_TYPE_HASH
-     * @flags READ_ONLY
+     * @flags CREATE_AND_SET
```
lguohan: as noted in the comment, the object id is read only, but user can still modify the attribute of the hash object, so user can still change the hash fields. The reason we do not allow user to change the attribute is that this is the root hash function and we do not allow user to set it to null and avoid undefined behavior. The same applies to LAG_HASH. 

For the rest of HASH attribute, it is ok to change to CREATE_AND_SET. Since is CREATE_AND_SET, we need then give default value. In this case, I suggest it to be SAI_NULL_OBJECT_ID, and we need to explain the behavior of this in the comment. In my mind, the behavior should be inherit the root hash object in the hash tree.

The meta data validation for this PR is failing, can you modify your meta according to above suggestion and make it pass?

kcudnik: hmm, could this attribute be create and set, and allow null, which null will mean 'use default' and it should be well defined, this way we will avoid more "default" objects created on the switch since i just noticed that i will also need to add those fields to "hidden" objects and during syncd restart i will have to "skip" those hash creations which are just readonly and it will require special logic

PR #315, opened 2016-11-27
Proposal for 'How To' add SAI API Extensions

mikelazar:


tushar-ty:
How do we resolve conflicts if someone uses extensions from more than one provider ?
Don’t we need to allocate vendor specific base enum offsets..

I think an example showing how will the extension headers be defined and included will help.
It seems to me that the enum values can conflict when two vendors define their own enums for different purposes in the same modules.
Also conflicts can occur for module ids, obj ids, API ids etc. at the top level when including extensions from more than one source.

mikelazar:
Added usage notes. Note that there are no conflicts between symbols:
- for attribute extensions, if two vendors provide the same functionality (same extended attributes, say PORT_LED), they should use the same extension attribute
- module extensions are completely independent, and if two extension modules provide the same functionality (say, MPLS) they are not used at the same time

PR #316, opened 2016-11-27
SAI bridge API headers

itaibaz:


inc/saitypes.h
```diff
@@ -463,6 +465,9 @@ typedef struct _sai_tunnel_map_params_t
     /** VNI id */
     sai_uint32_t vni_id;
 
+    /** Bridge IF */
+    sai_object_id_t bridge_if;
```
kcudnik: does this needs to be here ? it will make really hard to track this object id, there is already oid in attribute_t directly in the structure

inc/saiport.h
```diff
@@ -1038,6 +994,15 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_EEE_WAKE_TIME,
 
     /**
+    * @brief Query/Configure Port's Advertised flow control mode
```
lguohan: the comment is probably not correct.

inc/saiport.h
```diff
@@ -47,6 +47,19 @@ typedef enum _sai_port_type_t
...
+*/
+typedef enum _sai_port_bind_mode_t
+{
+    /** Port */
+    SAI_PORT_BIND_MODE_PORT,
+
+    /** Sub port */
+    SAI_PORT_BIND_MODE_SUB_PORT,
+
+} sai_port_bind_mode_t;
```
lguohan: I do not find any explanation for the bind mode in the v6 doc. what is mode_port, and what is sub_port?

itaibaz: Section 2 "sub port interface" explains about sub port and legacy port modes

Section 6.1 gives examples of using sub port
//configure port 3,4,5,6 to sub port mode  
sai_attribute_t port_attr;
port_attr.id=SAI_PORT_BIND_MODE;
port_attr.value= SAI_PORT_BIND_MODE_SUB_PORT;
sai_set_port_attribute_fn(port_3,port_attr);
sai_set_port_attribute_fn(port_4,port_attr);

Section 6.2 gives example of using port mode
sai_attribute_t port_attr;
port_attr.id=SAI_PORT_BIND_MODE;
port_attr.value= SAI_PORT_BIND_MODE_PORT;
sai_set_port_attribute_fn(port_3,port_attr);
sai_set_port_attribute_fn(port_4,port_attr);

inc/saitypes.h
itaibaz: The idea is to support 2 new types of maps - VNI to Bridge IF, and Bridge IF to VNI
This requires adding bridge_if to the map params
There can be many map elements
Not sure I understand the comment - "there is already oid in the strcut". Can you explain how you suggest to do it ?

kcudnik:
Adding objec type also make some issues in metadata since currently serialization type describes allowed object types that can be set/get on current attribute, and serialization type determines whether this type is object id or not, in here serialization type will be qos map, but it also will require allowed object types to be present which make some issues in current metadata logic

kcudnik:
Convert qos map struct to qos map object with type as mandatory on create and other attribute As conditional on type, and qos map list will become object id list, this will be consistent and metadata will be easy

lguohan:
@atitjain , @krambn, @AshokDaparthi , @tushar-ty , @sonicyao1985 , please help to review

inc/saibridge.h
```diff
@@ -0,0 +1,376 @@
...
+    */
+    SAI_BRIDGE_PORT_FDB_LEARNING_MODE_FDB_NOTIFICATION,
+
+} sai_bridge_port_fdb_learning_mode_t;
+
+/**
+ * @brief Attribute data for #SAI_BRIDGE_PORT_ATTR_TYPE
+ */
+typedef enum _sai_bridge_port_type_t
+{
```
rsivasankar: We might need the ADMIN_STATE attribute on the bridge port. For example, on a sub_port, say P1, we could have 2 dot1d ports created (P1.V1 and P1.V2). There might be a need to shut down one of the 2 ports.

Currently the ADMIN_STATE is present only in saiport.h (for physical and Lag ports) and hence we can only shutdown all the sub ports on a port, but not selectively.

inc/saitypes.h
kcudnik: it will be really hard to track this in metadata, it would be better if this would be an attribute

inc/saifdb.h
```diff
@@ -57,9 +70,15 @@ typedef struct _sai_fdb_entry_t
...
 
-    /** Vlan ID */
+    /** Bridge type */
+    sai_fdb_entry_bridge_type_t bridge_type;
+
+    /** Vlan ID. Valid for .1Q */
     sai_vlan_id_t vlan_id;
 
+    /** Bridge ID. Valid for .1D */
+    sai_object_id_t bridge_id;
```
kcudnik: this value is conditional based on bridge_type which makes this a lot harder to maintain in metadata

inc/saifdb.h
```diff
@@ -101,17 +120,13 @@ typedef enum _sai_fdb_entry_attr_t
     SAI_FDB_ENTRY_ATTR_TYPE = SAI_FDB_ENTRY_ATTR_START,
 
     /**
-     * @brief FDB entry port id
-     *
-     * The port id here can refer to a generic port object such as SAI port object id,
-     * SAI LAG object id and etc. or to a tunnel next hop object in case the entry is
-     * l2 tunnel
+     * @brief FDB entry bridge port id
```
jtardo: Why the complete change of destination, why not just consider the bridge port as another generic port object? Suggest leaving it unchanged.

inc/saibridge.h
```diff
@@ -0,0 +1,376 @@
...
+    _In_ uint32_t attr_count,
+    _Inout_ sai_attribute_t *attr_list);
+
+/**
+ * @brief Attribute data for #SAI_BRIDGE_ATTR_TYPE
+ */
+typedef enum _sai_bridge_type_t
+{
+    /** vlan aware bridge */
+    SAI_BRIDGE_TYPE_1Q,
```
atitjain: Isnt saivlan the SAI_BRIDGE_TYPE_1Q. I think we dont need this type until we unify saivlan and saibridge?

inc/saibridge.h
```diff
@@ -0,0 +1,376 @@
...
+     *
+     * @type sai_object_list_t
+     * @objects SAI_OBJECT_TYPE_BRIDGE_PORT
+     * @flags READ_ONLY
+     */
+    SAI_BRIDGE_ATTR_PORT_LIST,
+
+    /**
+     * @brief End of attributes
+     */
```
atitjain: Dont we need equivalent of following attributes here. I think we should have it.
SAI_VLAN_ATTR_MAX_LEARNED_ADDRESSES
SAI_VLAN_ATTR_STP_INSTANCE
SAI_VLAN_ATTR_LEARN_DISABLE
SAI_VLAN_ATTR_INGRESS/EGRESS_ACL

inc/saifdb.h
```diff
@@ -57,9 +70,15 @@ typedef struct _sai_fdb_entry_t
     /** Mac address */
     sai_mac_t mac_address;
 
-    /** Vlan ID */
+    /** Bridge type */
+    sai_fdb_entry_bridge_type_t bridge_type;
```
atitjain: Why not just use bridge_type_t.

inc/saifdb.h
atitjain: agree to this.

inc/saiswitch.h
```diff
@@ -472,6 +472,17 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_DEFAULT_VIRTUAL_ROUTER_ID,
 
     /**
+     * @brief Default .1Q Bridge ID
```
atitjain: What is this? Isn't 1Q bridge = vlan, in that case we have default vlan = 1.

inc/saivlan.h
```diff
@@ -160,13 +160,13 @@ typedef enum _sai_vlan_member_attr_t
...
-     * @brief Logical port ID
+     * @brief Bridge port ID. Valid only for .1Q Bridge ports
      *
      * @type sai_object_id_t
-     * @objects SAI_OBJECT_TYPE_PORT
+     * @objects SAI_OBJECT_TYPE_BRIDGE_PORT
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      */
-    SAI_VLAN_MEMBER_ATTR_PORT_ID,
+    SAI_VLAN_MEMBER_ATTR_BRIDGE_PORT_ID,
```
atitjain: To use the port here, do I need to create a bridge port with type SAI_BRIDGE_PORT_TYPE_PORT and use it here?

inc/saifdb.h
```diff
@@ -101,17 +120,13 @@ typedef enum _sai_fdb_entry_attr_t
...
-     * l2 tunnel
+     * @brief FDB entry bridge port id
      *
      * @type sai_object_id_t
-     * @objects SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_LAG, SAI_OBJECT_TYPE_TUNNEL
+     * @objects SAI_OBJECT_TYPE_BRIDGE_PORT
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      */
-    SAI_FDB_ENTRY_ATTR_PORT_ID,
+    SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID,
```
atitjain: To add a normal 1Q port, do I need to create a bridge port of type SAI_BRIDGE_PORT_TYPE_PORT?

inc/sairouterintf.h
```diff
@@ -45,7 +45,13 @@ typedef enum _sai_router_interface_type_t
...
 
     /** Loopback Router Interface Type */
-    SAI_ROUTER_INTERFACE_TYPE_LOOPBACK
+    SAI_ROUTER_INTERFACE_TYPE_LOOPBACK,
+
+    /** Sub port Router Interface Type */
+    SAI_ROUTER_INTERFACE_TYPE_SUB_PORT,
+
+    /** .1D Bridge Router Interface Type */
+    SAI_ROUTER_INTERFACE_TYPE_BRIDGE
```
atitjain: Is it possible to create a RIF on a sub port and then create a bridge port using the same and then create a RIF on the .1D bridge router interface?

inc/saiport.h
lguohan: ok with me.

inc/saibridge.h
jtardo: I tend to agree, it's kind of confusing now. Is it really necessary to have this kind of redundant bridge type configuration on both the bridge and bridge_port objects?

inc/saivlan.h
```diff
@@ -160,13 +160,13 @@ typedef enum _sai_vlan_member_attr_t
     SAI_VLAN_MEMBER_ATTR_VLAN_ID = SAI_VLAN_MEMBER_ATTR_START,
 
     /**
-     * @brief Logical port ID
+     * @brief Bridge port ID. Valid only for .1Q Bridge ports
      *
      * @type sai_object_id_t
-     * @objects SAI_OBJECT_TYPE_PORT
+     * @objects SAI_OBJECT_TYPE_BRIDGE_PORT
```
jtardo: This looks circular. For sub-ports don't you need to know the VLAN id from the packet to look up the bridge port? Shouldn't the PVID and/or VLAN filtering go with the front panel port based on the packet headers? What purpose does it serve to associate the VLAN id with a bridge port like this?

inc/saiport.h
mattyk: yes it is , 
thanks for the comment   we will update the diagram ti make it clearer

mattyk: currently we don't have admin state for all  logical interfaces E.g router interface 
I agree that we need to consider adding admin state to all logical interfaces including router interface and bridge interface . 
but i don't  think it is mandatory for 1.0 we can the Admin  state attribute leter do you agree ?

mattyk: yes it is , 
thanks for the comment   we will update the diagram ti make it clearer

inc/saibridge.h
mattyk: currently we don't have admin state for every logical  interface in SAI E.g router interface  
I think that bridge port interface is not different that router interface
i think we should have this discussion but it is not mandatory for 1.0 
we can always add Admin state attribute leter do you agree ?

inc/saitunnel.h
```diff
@@ -50,6 +50,12 @@ typedef enum _sai_tunnel_map_type_t
     /** TUNNEL Map VLAN ID to VNI */
     SAI_TUNNEL_MAP_TYPE_VLAN_ID_TO_VNI = 0x00000003,
 
+    /** TUNNEL Map VNI to Bridge IF */
+    SAI_TUNNEL_MAP_TYPE_VNI_TO_BRIDGE_IF = 0x00000004,
+
```
mattyk: it should be  type = =SAI_BRIDGE_PORT_TYPE_TUNNEL

inc/saibridge.h
mattyk: he bridge object look different (.1Q may have MSTP and vlan membership configuration ) 
the create abridge object according to the first added to the bridge look too flexible and undetermenistic to me  
iIthink we should have more strict and deterministic  model

inc/saibridge.h
mattyk: we still have those attributes on the vlan object .
will going to add :  
SAI_BRIDEG_ATTR_MAX_LEARNED_ADDRESSES
SAI_BRIDGE_ATTR_LEARN_DISABLE

inc/saifdb.h
mattyk: it is an FDB entry type don't you think we need the "sai_fdb_entry"  prefix

inc/saiswitch.h
mattyk: by default we create a .1Q bridge and add all the switch port to the that bridge 
we need this attribute in order to get the default .1Q bridge object

inc/saivlan.h
mattyk: yes Vlan is an attribute of a .1Q bridge port

inc/sairouterintf.h
mattyk: No , {port,Vlan} can point to a bridge interface Or router interface

inc/saivlan.h
mattyk: the Vlan object is valid only for .1Q bridge ,  and used vlan base bridge configuration (port vlan member, learning ...) 
.1D bridge are vlan unaware bridges

inc/saibridge.h
```diff
@@ -0,0 +1,376 @@
...
+     * @default SAI_BRIDGE_PORT_FDB_LEARNING_MODE_HW
+     */
+    SAI_BRIDGE_PORT_ATTR_FDB_LEARNING_MODE,
+
+    /**
+     * @brief Maximum number of learned MAC addresses
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 0
```
lguohan: 0 means no limit or cannot learn? it's better to explain.

inc/saivlan.h
atitjain: I dont think that answers the question. What I mean is can I add the port oid here or I need to first create a bridge port from the port oid and use it here.

inc/saibridge.h
atitjain: Not clear to me. MSTP is applied to a group of vlans. VLAN has membership configuration. Please see saivlan.h. Can you point me where is the vlan attribute in the 1Q bridge object.

inc/saifdb.h
atitjain: You have removed vlan from the fdb attribute, vlan is not part of the bridge object then how does the existing vlan object work.

inc/saiswitch.h
atitjain: This is change of behavior. In the current APIs all the ports are by default added to vlan 1. Please help me understand this.

inc/saifdb.h
itaibaz: vlan is not removed from fdb entry. It is valid for .1Q. Please see the complete change in the entry

    /** Bridge type */
    sai_fdb_entry_bridge_type_t bridge_type;

    /** Vlan ID. Valid for .1Q */
    sai_vlan_id_t vlan_id;

    /**
     * Bridge ID. Valid for .1D
     *
     * @objects SAI_OBJECT_TYPE_BRIDGE
     */
    sai_object_id_t bridge_id;

inc/saibridge.h
```diff
@@ -0,0 +1,376 @@
...
+{
+    sai_create_bridge_fn         		create_bridge;
+    sai_remove_bridge_fn 				remove_bridge;
+    sai_set_bridge_attribute_fn  		set_bridge_attribute;
+    sai_get_bridge_attribute_fn  		get_bridge_attribute;
+    sai_create_bridge_port_fn         	create_bridge_port;
+    sai_remove_bridge_port_fn 			remove_bridge_port;
+    sai_set_bridge_port_attribute_fn  	set_bridge_port_attribute;
+    sai_get_bridge_port_attribute_fn    get_bridge_port_attribute;
+} sai_bridge_api_t;
```
itaibaz: fixed

inc/sairouterintf.h
```diff
@@ -94,7 +100,7 @@ typedef enum _sai_router_interface_attr_t
      * @type sai_object_id_t
      * @objects SAI_OBJECT_TYPE_VLAN
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
-     * @condition SAI_ROUTER_INTERFACE_ATTR_TYPE == SAI_ROUTER_INTERFACE_TYPE_VLAN
+     * @condition SAI_ROUTER_INTERFACE_ATTR_TYPE == SAI_ROUTER_INTERFACE_TYPE_VLAN or SAI_ROUTER_INTERFACE_ATTR_TYPE == SAI_ROUTER_INTERFACE_TYPE_SUB_PORT
      */
     SAI_ROUTER_INTERFACE_ATTR_VLAN_ID,
```
jtardo: Shouldn't this be BRIDGE_ID for SAI_ROUTER_INTERFACE_TYPE_SUB_PORT?

inc/saibridge.h
lguohan: in the current proposal, there is only one default 1Q bridge, so all 4k vlan objects are in the 1Q bridge. The all the vlan share same attribute such as max learned mac, and etc on that are defined in the bridge attribute.

inc/saiswitch.h
lguohan: no. we still have the default vlan 1, and all ports joined the vlan 1 by default. The behavior is the same as before.

inc/saivlan.h
lguohan: you need to first create the bridge port and use it here. However, it only be .1Q bridge port, not .1D bridge port. vlan member will be created based on the vlan id and the .1Q bridge port.

inc/sairouterintf.h
jtardo: In the examples aren't the "vports" are all bridge ports? Are you indicating that a 1D rif can be used instead of a "vport" bridge port?

PR #320, opened 2016-12-05
Default vlan object id get

AshokDaparthi:
Present SAI1.0 modified vlan as sai_object_id_t. 
Added get attribute to retrieve default vlan object id from switch, by using this application can make use of this to get member list and set/get attributed for default vlan object id.

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -452,6 +452,16 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_ACL_USER_TRAP_ID_RANGE,
 
     /**
+     * @brief Default SAI VLAN ID
+     *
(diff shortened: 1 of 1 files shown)


AshokDaparthi:
@lguohan, @atitjain @itaibaz - Please review

lguohan:
@kcudnik, please review.

kcudnik:
saiport.h there is attribute SAI_PORT_ATTR_PORT_VLAN_ID which is uint16 should be probably also converted to object id

AshokDaparthi:
@lguohan @kcudnik - saiport.h , saiacl.h etc as per previous discussions we leave it as uint16.

atitjain:
Was looking for this during our development. Really helpful will approve.

PR #321, opened 2016-12-05
Access type changes for WRED and SCHEDULER GROUP attributes

aarthygnanamurthy:
WRED profiles can be updated even after creation. Hence the access flags for these attributes have been updated to CREATE and SET.

Similarly, Scheduler profiles can be updated on a scheduler group even after create. Hence updating the same to CREATE and SET

### DIFF
--- a/inc/saischedulergroup.h
+++ b/inc/saischedulergroup.h
@@ -86,11 +86,11 @@ typedef enum _sai_scheduler_group_attr_t
     SAI_SCHEDULER_GROUP_ATTR_MAX_CHILDS = 0x00000004,
 
     /**
-     * @brief Scheucler ID
+     * @brief Scheduler ID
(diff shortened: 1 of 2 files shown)


kcudnik:
if you change SAI_WRED_ATTR_RED_ENABLE  dynamically, then it may happen that those red threshold min/max is no longer required (if you sed enable to false) but attributes were passed already, those should be actually "validonlywhen" rather than conditions, currently this option is not available in metadata yet, i will work on it in near future, for now you can comment out those conditions with comments

also there will be a problem when you enable RED but thresholds are not set, so they should have default value specified

sheduler it should be MANDATORY_ON_CREATE | CREATE_AND_SET since this value probably can't be NULL as you set it now for CREATE_AND_SET

PR #325, opened 2016-12-09
Add missing VLAN links in dependency graph

kcudnik:


lguohan:
we can change SAI_ROUTER_INTERFACE_ATTR_VLAN_ID to use object id, will this solve the problem?

kcudnik:
We can, just wondering if thats required, since you probably can't set other switch vlan to current switch router interface so even single vlan number should directly identify from which switch vlan comes, but yea we can use object id here.

lguohan:
Usually, when object have dependency we make those dependency via the object id, it is easier to manage the dependency than making those options.

For example, an ACL field can match a vlan id (uint16_t), but it does mean the ACL depends on a vlan object, because the ACL is match a packet, and the packet can have that vlan id whether or not the switch has created the vlan.

kcudnik:
Im closing this one since uint16 value vlans are not treated as actual dependency

PR #326, opened 2016-12-09
Modified WRED threshold flags

AshokDaparthi:


AshokDaparthi:
@kcudnik 

"if you change SAI_WRED_ATTR_RED_ENABLE dynamically, then it may happen that those red threshold min/max is no longer required (if you sed enable to false) but attributes were passed already, those should be actually "validonlywhen" rather than conditions, currently this option is not available in metadata yet, i will work on it in near future, for now you can comment out those conditions with comments

also there will be a problem when you enable RED but thresholds are not set, so they should have default value specified"

Based on your comments, i added validonly flag, it is correct or you are going to add vlaidonlywhen flag.
for can i comment condition and vlaidonly ?

kcudnik:
https://sonic-jenkins.westus.cloudapp.azure.com/job/sai-meta-pr-build/234/console
your build failed, please correct

kcudnik:
rebuild test

AshokDaparthi:
Done

AshokDaparthi:
@kcudnik - Please review

PR #330, opened 2016-12-10
Add non object id test

kcudnik:


kcudnik:
Please install "grep" on test docker

PR #344, opened 2016-12-22
PTF test cases for SAI 1.0 ACL Model 

zubinshah:
Modified the one and only existing ACL PTF test cases. 
Added 5 new ACL test cases that are provided as an example in the proposal doc (doc/SAI-Proposal-ACL-1.md)

Thanks @IGordynskyi for uplifting this.


lguohan:
merged in #450

PR #345, opened 2016-12-27
Fixed SAI thrift compilation.

nazarii-gnydyn:
SAI thrift compilation fixed according to SAI headers update. Also fixed ACL test case.


nazarii-gnydyn:
@lguohan @xdwu @itaibaz Please review PULL-request.

PR #360, opened 2017-01-12
Tunnel map items as objects proposal

kcudnik:
Purpose for this change is to remove bridge id from sai_tunnel_map_params_t and replace it with new object type to keep same feature. Current solution will require very special logic to handle bridge_if id in sai_tunnel_map_params_t in metadata in evey place where it will be used, and its not uniform with all other objects in SAI

This change could be much simpler if there would be requirement that when bridge id is used, then same bridge must be used for all tunnel map, then this bridge_if id could be moved tunnel itself instead of tunnel map

### DIFF
--- a/inc/saitunnel.h
+++ b/inc/saitunnel.h
@@ -778,6 +778,58 @@ typedef sai_status_t (*sai_get_tunnel_term_table_entry_attribute_fn)(
         _Inout_ sai_attribute_t *attr_list);
 
 /**
+ * @brief Create tunnel map item
+ *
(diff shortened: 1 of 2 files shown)


kcudnik:
@itaibaz Please review this change since you added bridge header changes, purpose is to remove
sai_object_id_t bridge_if; from sai_tunnel_map_params_t

itaibaz:
Fine with Matty and me.
Since we might need to create a large number (several K's elements), please consider bulk API as well

PR #361, opened 2017-01-12
Convert acl stages to switch attributes

kcudnik:
Purpose of this change is to unify attribute values, since for acl capability for different input stage we will get different results in output attribute value. This behavior is not present in any other attribute so it make sense to unify this. If in the future more acl stages will be defined they can be added as extra switch attributes without problems.


lguohan:
@kcudnik , can you resolve conflict for this one?

kcudnik:
Yes, resolved

PR #381, opened 2017-01-25
Tunnel map items as objects proposal

kcudnik:
Purpose for this change is to remove bridge id from sai_tunnel_map_params_t and replace it with new object type to keep same feature. Current solution will require very special logic to handle bridge_if id in sai_tunnel_map_params_t in metadata in evey place where it will be used, and its not uniform with all other objects in SAI

This change could be much simpler if there would be requirement that when bridge id is used, then same bridge must be used for all tunnel map, then this bridge_if id could be moved tunnel itself instead of tunnel map

### DIFF
--- a/inc/saitunnel.h
+++ b/inc/saitunnel.h
@@ -61,12 +61,12 @@ typedef enum _sai_tunnel_map_type_t
 
 } sai_tunnel_map_type_t;
 
-typedef enum _sai_tunnel_map_item_attr_t
+typedef enum _sai_tunnel_map_entry_attr_t
 {
     /**
      * @brief Start of attributes
      */
-    SAI_TUNNEL_MAP_ITEM_ATTR_START = 0x00000000,
+    SAI_TUNNEL_MAP_ENTRY_ATTR_START = 0x00000000,
 
     /**
      * @brief Tunnel Map type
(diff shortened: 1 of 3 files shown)


kcudnik:
This is reissued commit since preveious was merged by mistake by @lguohan

inc/saitunnel.h
```diff
@@ -61,6 +61,137 @@ typedef enum _sai_tunnel_map_type_t
...
+     */
+    SAI_TUNNEL_MAP_ITEM_ATTR_END,
+
+    /** Custom range base value */
+    SAI_TUNNEL_MAP_ITEM_ATTR_CUSTOM_RANGE_START = 0x10000000,
+
+    /** End of custom range base */
+    SAI_TUNNEL_MAP_ITEM_ATTR_CUSTOM_RANGE_END
+
+} sai_tunnel_map_item_attr_t;
```
tushar-ty: Could you call it map_'entry' or something else instead of 'item' ?

kcudnik: yes, i will update that

tushar-ty: Thanks. 
Just FYI, my input was for the entire file not just this instance.

MELLANOX-zhichenr:
Is there a reason to keep the old tunnel map object? Can it be fully replaced by the new tunnel map entry object?

kcudnik:
Idea was to not cause too much trouble for current vendor implementations. We can remove that in future eventually.

MELLANOX-zhichenr:
I see. Thanks!

MELLANOX-zhichenr:
Hi I have several questions:

1. Tunnel has an attribute called encap tunnel map. Could this attribute be either 'tunnel map' type or 'tunnel map entry' type? If yes then the metadata might need to be updated. If no, then how to associate tunnel map entry with tunnel?

2. Tunnel map entry has an attribute called tunnel map. Does this attribute need to be mandatory? If yes, then does setting oecn value on tunnel map entry changes the attribute in tunnel map?

3. In tunnel map, comment on tunnel map list said 'if empty, tunnel map can be used to assign tunnel map entry objects'.  Is there an example of how to do the assignment?

4. For SAI_TUNNEL_MAP_ENTRY_ATTR_OECN_KEY, should the condition be only SAI_TUNNEL_MAP_ENTRY_ATTR_TUNNEL_MAP_TYPE == SAI_TUNNEL_MAP_TYPE_OECN_TO_UECN since it is always 'key to value'?

PR #389, opened 2017-01-26
Added MAX value for sai_api_t and sai_native_hash_field_t enums

vveskera:


kcudnik:
Why you need those MAX values ? none of other enums has those excet object_type where it make sense

vveskera:
@kcudnik 
This will make them more scalable and flexible. 
And we will not need to update code every time when new API or hash field is added.

atitjain:
Looks good.

kcudnik:
if you add a new api enum you need to add code to support that anyway, this enum is i think used right now to set loglevel and to get specific api structure, so adding MAX is not helps in this case, on swich handling api_t you add default tag and add log that this value is not supported, same as some would pass _MAX to get api 
for hash it maybe make sense if you want to pass all fields to be used every time

vveskera:
From resource allocation side it simplifies code support, because we will always know MAX sai_api_t index.
In this case, if you add new API emum value before SAI_API_MAX, you don't need to change the memory borders.

lguohan:
we do not have this in the master, which mean if you later move to master your sai is probably going to break, so we'll need this in master branch too.

the other question I have is that why only  hash and api enum, what about other enum?

zubinshah:
I agree, it should be done for other enums as well.

On Jan 31, 2017, at 12:33 PM, lguohan <notifications@github.com<mailto:notifications@github.com>> wrote:


we do not have this in the master, which mean if you later move to master your sai is probably going to break, so we'll need this in master branch too.

the other question I have is that why only hash and api enum, what about other enum?

vveskera:
I had double checked the headers.
And IMHO for now there are no any enums, where this changes are also needed.

kcudnik:
Test this please

PR #396, opened 2017-02-02
Add FEC support to saiport.h

shihjeff:
Signed-off-by: jeff chen <jechen@cavium.com>


inc/saiport.h
```diff
@@ -572,6 +587,15 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_INTERNAL_LOOPBACK_MODE,
 
     /**
+     * @brief Forward Error Correction (FEC) control
+     *
+     * @type sai_port_fec_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PORT_FEC_MODE_NONE
```
shihjeff: Are you thinking in terms of advertising FEC abilities with auto-negotiation?  If I go with a list, and someone configures in static mode and (mis-)configures both FEC options (like in 25G), which type takes precedence?

SathishKumarTR-Dell: Yes, I was thinking that the list will fit the case where there is a hw ability to perform any of the supported modes based on the peer's mode. We are leaving it to the HA to configure the valid FEC mode for a port speed. So, any mis-configuration would lead the port to remain down. Isnt it? We can have this as the static enum now if the case i mentioned is not really useful.

lguohan: Is SAI_PORT_ATTR_AUTO_NEG_MODE is true, then this FEC_MODE will no longer be applicable to the port? if yes, then we need add validonly meta data to specify that FEC_MODE is only valid when AN_MODE = false.

SathishKumarTR-Dell: I have the same understanding. When AN is enabled, FEC mode should be resolved based on the capability advertised by the PHY as defined in standard. This attribute should define the mode for cases wherever AN is disabled.

shihjeff: Committed the changes mentioned in today's meeting. Feel free to review

AshokDaparthi:
@lguohan , @shihjeff - SAI_PORT_ATTR_REMOTE_SUPPORTED* this is to get the remote port capability, remote port supported/capable and advertised can be different like what we have for port supported and advertised. After discussing internally with my team, remote port capability not useful, only we need the remote port advertised fields. Not sure is any NPU's are supporting this. We can remove the remote_supported attributes and keep SAI_PORT_ATTR_REMOTE_ADVERTISED* alone.

shihjeff:
@AshokDaparthi / @lguohan - taken @AshokDaparthi comment on REMOTE_SUPPORTED_* and removed them as part of this PR.

AshokDaparthi:
@shihjeff  - Please add below counters also 
In sai_port_stat_t,

/*
* Forward Error Correction corrected counters.[sai_uint64_t]
*/
SAI_PORT_STAT_FEC_CORRECTED_COUNTERS

/*
* Forward Error Correction uncorrected counters.[sai_uint64_t]
*/
SAI_PORT_STAT_FEC_UNCORRECTED_COUNTERS

AshokDaparthi:
Please raise same to 0.9.6 branch.

shihjeff:
@AshokDaparthi @lguohan - We can probably add the FEC counters later; just want to get version 1 in. Naming doesn't seem to be compliant with the current naming convention, which is a bit overlapping to begin with.  I can see:
SAI_PORT_STAT_IF_IN/OUT_*
SAI_PORT_STAT_ETHER_STATS_* (Is this sum of IN/OUT?)
SAI_PORT_STAT_IP(V6)_IN/OUT_*
SAI_PORT_STAT_* (For WRED)
SAI_PORT_STAT_ETHER_IN/OUT_*
SAI_PORT_STAT_IN/OUT_*
SAI_PORT_STAT_*

I think we should have effort to clean above up or at least document the proper naming convention (even down to mundane things like PACKETS vs PKTS)

Should it which of the above?:
SAI_PORT_STAT_FEC_CORRECTED_PKTS
SAI_PORT_STAT_IF_IN_FEC_CORRECTED_PKTS
SAI_PORT_STAT_IN_FEC_CORRECTED_PKTS
etc..

AshokDaparthi:
counters we can raise it as next PR

shihjeff:
@lguohan - Is there any checks/changes left we need to go through to get this and PR#429 merged?

inc/saiport.h
```diff
@@ -254,6 +269,14 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_SUPPORTED_SPEED,
 
     /**
+     * @brief Query list of supported port FEC mode
+     *
+     * @type sai_u32_list_t
+     * @flags READ_ONLY
```
kcudnik: This should bez s32 list with enum name following

kcudnik: Od this list od list od enum values

inc/saiport.h
```diff
@@ -294,60 +317,20 @@ typedef enum _sai_port_attr_t
...
-
-    /**
-     * @brief Query list of Advertised remote port speed (Full-Duplex) in Mbps
+     * @brief Query list of Advertised remote port FEC control
      *
      * @type sai_u32_list_t
      * @flags READ_ONLY
      */
-    SAI_PORT_ATTR_REMOTE_ADVERTISED_SPEED,
+    SAI_PORT_ATTR_REMOTE_ADVERTISED_FEC_MODE,
```
kcudnik: This should be s32 list with enum name following

inc/saiport.h
```diff
@@ -470,6 +453,15 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_ADVERTISED_SPEED,
 
     /**
+     * @brief Query/Configure list of Advertised port FEC Mode
+     *
+     * @type sai_u32_list_t
```
kcudnik: Also s32 list with following enum name

inc/saiport.h
lguohan: agree with kcudnik

shihjeff:
@lguohan - Made the change @kcudnik mentioned

PR #400, opened 2017-02-06
Add sai_ip_prefix_t to sai_attribute_value_t for future use

kcudnik:
Member of attribute value ipprefix will be used later on
for generic extraction of members from non object id
structs and now is only used inside route_entry

### DIFF
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -555,6 +555,7 @@ typedef union {
     sai_ip4_t ip4;
     sai_ip6_t ip6;
     sai_ip_address_t ipaddr;
+    sai_ip_prefix_t ipprefix;
     sai_object_id_t oid;
     sai_object_list_t objlist;
     sai_u8_list_t u8list;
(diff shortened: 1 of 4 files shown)


inc/saitypes.h
```diff
@@ -555,6 +555,7 @@ typedef union {
     sai_ip4_t ip4;
     sai_ip6_t ip6;
     sai_ip_address_t ipaddr;
+    sai_ip_prefix_t ipprefix;
```
lguohan: the concern is that this doubles the sai_attribute_value size. for ipv6, the attribute value size will be 16*4 = 64 bytes.

kcudnik: Nope:
 36      printf ("value size: %zu\n", sizeof (sai_attribute_value_t));
 37      printf ("ipprefix: %zu\n", sizeof (sai_ip_prefix_t));
value size: 40
ipprefix: 36

lguohan: you are right. 16*2+4=36

kcudnik: also depends of compiler align, but by default seems to align 4 bytes

PR #403, opened 2017-02-07
[v0.9.4] Add PFC based pause state

stcheng:
Signed-off-by: Shuotian Cheng <shuche@microsoft.com>

### DIFF
--- a/inc/saiqueue.h
+++ b/inc/saiqueue.h
@@ -70,6 +70,9 @@ typedef enum _sai_queue_attr_t
     /** Attach scheduler to Queue [sai_object_id_t]*/
     SAI_QUEUE_ATTR_SCHEDULER_PROFILE_ID = 0x00000003,
 
+    /** Queue pause state [bool] (READ-ONLY) */
+    SAI_QUEUE_ATTR_PAUSE_STATUS = 0x00000004,
+
     /* -- */
     /* Custom range base value */
     SAI_QUEUE_ATTR_CUSTOM_RANGE_BASE = 0x10000000


inc/saiqueue.h
```diff
@@ -70,6 +70,9 @@ typedef enum _sai_queue_attr_t
     /** Attach scheduler to Queue [sai_object_id_t]*/
     SAI_QUEUE_ATTR_SCHEDULER_PROFILE_ID = 0x00000003,
 
+    /** PFC based pause state [booldata] */
```
lguohan: should not restricted to pfc, just queue pause state.
default false. (true: means queue in pause state).

stcheng: updated

AshokDaparthi:
@lguohan - is this READ_ONLY Attribute to know the pause state on queue or to enable honoring on queue ?

tushar-ty:
@AshokDaparthi, yes this is RO for knowing the current state.

stcheng:
@tushar-ty I will update the PR to indicate that this is RO. But if it is RO, why is there a default value? Or does it need to have a default value? What will trigger this attribute to change?

tushar-ty:
Yes, the default value field should be removed. I did not remember if it wasn't required for RO attribs. This attrib represents the queues internal h/w state and is updated upon receiving PFC frames.

stcheng:
@tushar-ty So here if the value is true, it means that the queue is paused?

stcheng:
@tushar-ty Could you also take a look at the changes on master?
https://github.com/opencomputeproject/SAI/pull/405

inc/saiqueue.h
```diff
@@ -70,6 +70,10 @@ typedef enum _sai_queue_attr_t
     /** Attach scheduler to Queue [sai_object_id_t]*/
     SAI_QUEUE_ATTR_SCHEDULER_PROFILE_ID = 0x00000003,
 
+    /** Queue pause state [bool]
+      * (READ_ONLY default is FALSE) */
```
lguohan: if it is read only, then there is no default value.

lguohan:
agree with tushar, there should be no default value for RO attribute.

inc/saiqueue.h
stcheng: updated

stcheng:
🏓

stcheng:
👆

PR #405, opened 2017-02-07
Add queue pause enabled status attribute

stcheng:
Signed-off-by: Shuotian Cheng <shuche@microsoft.com>


kcudnik:
Enums must be sorted, thats why build fails

stcheng:
@kcudnik So If I will need to add some more READ_ONLY attributes, the rest enums need to be renumbered. This will fail the backward compatibility?

kcudnik:
No, just add them at the end, w don't need to stick to rule that all read only attributes must be at the top

stcheng:
So, the comment flag: /* READ_ONLY */ and /* READ & WRITE */ make no sense in the future.

kcudnik:
Yes, we can remove them

stcheng:
updated

inc/saiqueue.h
```diff
@@ -118,8 +118,7 @@ typedef enum _sai_queue_attr_t
     SAI_QUEUE_ATTR_WRED_PROFILE_ID = 0x00000004,
 
     /**
-     * @brief Attach buffer profile to Queue
-     * Default no profile
```
atitjain: is the default taken out?

PR #408, opened 2017-02-08
Added switch id in missed object create.

AshokDaparthi:


kcudnik:
missing also sai_create_rpf_group_member_fn  sai_create_l2mc_group_member_fn sai_create_next_hop_group_member_fn sai_create_rpf_group_member_fn sai_create_stp_port_fn
in 1.0 there is already automatic check for that

PR #409, opened 2017-02-08
Added Ipv6 Next Header as qualifier

AshokDaparthi:


inc/saiacl.h
```diff
@@ -927,9 +927,18 @@ typedef enum _sai_acl_table_attr_t
     SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE,
 
     /**
+     * @brief IPv6 Next Header
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_IPv6_NEXT_HEADER,
```
kcudnik: don't use small letters in enum members

lguohan:
in case the packet has many ipv6 next header fields, which one?

inc/saiacl.h
```diff
@@ -927,9 +927,18 @@ typedef enum _sai_acl_table_attr_t
     SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE,
 
     /**
+     * @brief IPv6 Next Header
```
lguohan: clarify on this "the next header in IPV6 header not the next header in the extension headers."

AshokDaparthi:
@guohan - Please check now

PR #410, opened 2017-02-08
WRED profile on buffer pools

AshokDaparthi:


AshokDaparthi:
@lguohan @atitjain @tushar-ty @itaibaz - Please review

lguohan:
can you explain why buffer pool needs to attach a wred profile?

inc/saibuffer.h
```diff
@@ -264,6 +264,19 @@ typedef enum _sai_buffer_pool_attr_t
     SAI_BUFFER_POOL_ATTR_XOFF_SIZE,
 
     /**
+     * @brief Attach WRED ID to pool
```
lguohan: can you add description on its relationship with port/queue level wred?

PR #413, opened 2017-02-10
Add generic quad api to metadata

kcudnik:
This generic quad api will help us later to call any api, without doing any
switch cases for calling different signature functions including non object
id structs. Also later we will generate automatic serialize and deserialize
methods for non object id which will deserialize data to object union in
sai_object_meta_key_t to right place.


lguohan:
can you resolve merge conflict?

kcudnik:
Resolved

PR #419, opened 2017-02-16
Updated TAM proposal based on community feedback

bandaru-viswanath:
Changes are -
- sai_tam_statistic_id_t: Allow the parent_type parameter to be a not-needed parameter for snapshot / tam object creation
- sai_tam_statistic_id_t: change the type for counter_id to uint32_t


kcudnik:
since sai_tam_statistic_id_t is using object id's please convert this to new wobject type for easy metadata tracking (and add those fields as attributes) and then instead of sai_tam_statistic_t please use object_list and list of objects for statistic_id object type

also SAI_TAM_THRESHOLD_ATTR_SNAPSHOT_ON_BREACH is bool, allownull dont apply
also please add @flags for each attribute

lguohan:
I'll merge this doc. @bandaru-viswanath , please follow kcudnik's feedback.

PR #426, opened 2017-02-20
Block BUM traffic ingress on given port to port list provided in the …

AshokDaparthi:
Use Case: Once of our use case is to block Flooding(Broadcast/unknown Unicast/unknown multicast/All three types) from given port to other ports in system.  example, FCoE and FC port can be same vlan. Traffic ingress on FCoE port and it is BUM traffic then block to other FC ports alone and flood it to other Ethernet or FCoE ports and vice versa.  Those port can be many vlan's, instead of vlan control port level control will help us.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1019,6 +1019,58 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_EGRESS_BLOCK_PORT_LIST,
 
     /**
+     * @brief Egress broadcast flood traffic block port list
+     *
+     * Broadcast flood Traffic ingressing on this port and egressing out of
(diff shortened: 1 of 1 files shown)


AshokDaparthi:
@atitjain @lguohan - Please check now. 

In v0.9.6 branch, we do not have bridge port concept, so all these are port level controls. In SAI1.0, i will try to  keep in bridge port model.

atitjain:
So this will be effective for all the vlan's with ingress port is part of?

AshokDaparthi:
Yes Atit

AshokDaparthi:
@lguohan - could you please review this, v0.9.6 is does not have bridge port. so raised it as port attributes.

PR #429, opened 2017-02-21
0.9.6 FEC support

shihjeff:


shihjeff:
@kcudnik - made the same change as master branch for s32 / enum name

PR #434, opened 2017-03-01
Add FDB entry tunnel endpoint IP

itaibaz:


kcudnik:
Add @default SAI_NULL_OBJECT_ID

lguohan:
but it is not an object id.

lguohan:
@kcudnik, can we condition on SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID/ SAI_BRIDGE_PORT_ATTR_TYPE=SAI_BRIDGE_PORT_TYPE_TUNNEL?

kcudnik:
no, since it's cross attribute condition, metadata is not able to do that

kcudnik:
ahh, it's ip address, then allownull is not required and needs to be removed

kcudnik:
also right now parser don't support default value to be set since this attribute is CREATE_AND_SET what is default value on that attribute ?

kcudnik:
take a look ad #435 and add @default 0.0.0.0 and remove allownull

kcudnik:
you changed that too fast, about 4 hours ago, and @lguohan merged my change 2 hours ago thats why it failed

itaibaz:
Yes, just need to rerun the jenkins job
I ran it on my switch, and it passes

PR #439, opened 2017-03-05
Force isvlan tag on uint16 types

kcudnik:


stcheng:
is this used to have logics for checking if a vlanid is valid or not?

kcudnik:
It will be used for that purpose

inc/saiacl.h
```diff
@@ -1192,6 +1194,7 @@ typedef enum _sai_acl_entry_attr_t
      *
      * @type sai_acl_field_data_t sai_uint16_t
      * @flags CREATE_AND_SET
+     * @isvlan false
```
lguohan: can we make isvlan=false by default so that we do not need to mark this for all uint16?

kcudnik: it is false by default, but i made it explicit on uint16 types for future adding attributes for users to not forget to set it to true if attribute is vlan

kcudnik:
Is there any other issue with this PR ?

PR #448, opened 2017-03-09
Force filenames to follow api name convention

kcudnik:


inc/sai.h
```diff
@@ -31,7 +31,7 @@
 #include "saibuffer.h"
 #include "saifdb.h"
 #include "saihash.h"
-#include "saihostintf.h"
+#include "saihostif.h"
```
stcheng: why 'routerinterface' but not 'hostinterface'?

kcudnik: thats how api_t are called, this change will have no impact on current vendor source code, but if we change name more complicated changes maybe required on vendor side. @lguohan can you check this change and suggest ?

PR #458, opened 2017-03-15
remove SAI_ACL_ENTRY_ATTR_ACTION_SET_CPU_QUEUE attribute

lguohan:
For ACL-based trap, the packet should go to a trap group and CPU queue
is defined in the trap. SAI_ACL_ENTRY_ATTR_ACTION_SET_CPU_QUEUE is
redudent and should not be used.


lguohan:
@itaibaz , @tushar-ty , @zubinshah , please review

PR #475, opened 2017-03-28
meta: Fix saisanitycheck when gcc treats enum as unsigned

vkochan-mlnx:
gcc throws warning:

    comparison of unsigned expression >=0 is always true

when saisanitycheck tries to check the following condition:

    if (ot >= SAI_OBJECT_TYPE_NULL && ot <= SAI_OBJECT_TYPE_MAX)

it happens because gcc treats enum's as unsigned if there only positive
values.


itaibaz:
The fix changes the meaning a little bit, making SAI_OBJECT_TYPE_NULL no longer acceptable
Did you want that ?
The other alternative to avoid the compilation issue is just checking 
if (ot <= SAI_OBJECT_TYPE_MAX)
or casting to int ?

kcudnik:
Where you get warning ? on what version of gcc ?
Also this check should be as is NULL and MAX should be also included

meta/saisanitycheck.c
```diff
@@ -2283,7 +2283,7 @@ void check_non_object_id_object_types()
                 {
                     sai_object_type_t ot = m->allowedobjecttypes[k];
 
-                    if (ot >= SAI_OBJECT_TYPE_NULL && ot <= SAI_OBJECT_TYPE_MAX)
```
kcudnik: change to (ot > SAI_OBJECT_TYPE_NULL && ot < SAI_OBJECT_TYPE_MAX)

vkochan-mlnx:
The warning was caused by gcc (Debian 4.7.2-5) 4.7.2

PR #478, opened 2017-03-29
Add vlan header valid for ERSPAN

itaibaz:
Packet is always encalpsulated with vlan header for remote L2
Packet will be encapsulated with vlan header for ERSPAN with vlan header
valid
The vlan ID in the header may be 0 for priority tagged packets
Add default values for vlan prio, CFI and TPID

### DIFF
--- a/inc/saimirror.h
+++ b/inc/saimirror.h
@@ -159,8 +159,9 @@ typedef enum _sai_mirror_session_attr_t
      * @brief Vlan header valid
      *
      * @type bool
-     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
-     * @condition SAI_MIRROR_SESSION_ATTR_TYPE == SAI_MIRROR_SESSION_TYPE_ENHANCED_REMOTE
+     * @flags CREATE_ONLY
+     * @default false
+     * @validonly SAI_MIRROR_SESSION_ATTR_TYPE == SAI_MIRROR_SESSION_TYPE_ENHANCED_REMOTE
      */
     SAI_MIRROR_SESSION_ATTR_VLAN_HEADER_VALID,
 
--- a/inc/saimirror.h
+++ b/inc/saimirror.h
@@ -110,42 +110,43 @@ typedef enum _sai_mirror_session_attr_t
     SAI_MIRROR_SESSION_ATTR_TC,
 
     /**
-     * @brief Valid for RSPAN and ERSPAN
-     *
-     * L2 header TPID if vlanId is not zero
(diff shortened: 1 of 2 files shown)


inc/saimirror.h
```diff
@@ -110,42 +110,43 @@ typedef enum _sai_mirror_session_attr_t
...
-     * L2 header TPID if vlanId is not zero
+     * @brief L2 header TPID. Valid for RSPAN or ERSPAN with valid Vlan header
      *
      * @type sai_uint16_t
-     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
+     * @flags CREATE_AND_SET
      * @isvlan false
-     * @condition SAI_MIRROR_SESSION_ATTR_TYPE == SAI_MIRROR_SESSION_TYPE_ENHANCED_REMOTE or SAI_MIRROR_SESSION_ATTR_TYPE == SAI_MIRROR_SESSION_TYPE_REMOTE
+     * @default 0x8100
+     * @validonly SAI_MIRROR_SESSION_ATTR_VLAN_HEADER_VALID == true or SAI_MIRROR_SESSION_ATTR_TYPE == SAI_MIRROR_SESSION_TYPE_REMOTE
```
kcudnik: SAI_MIRROR_SESSION_ATTR_VLAN_HEADER_VALID is also conditional attribute, this is not allowed to chain conditional attributes

kcudnik:
Are you able to use those limitations to describe mirror session in simpler way ? Possibly make 
vVlan_header_valid as create_only with default value set to false ? Then you don’t need to make it
conditional.

itaibaz:
I tried making vlan_header_valid as non mandatory with default false
However, it will still need to be described as      * @validonly SAI_MIRROR_SESSION_ATTR_TYPE == SAI_MIRROR_SESSION_TYPE_ENHANCED_REMOTE
Because it's only relevant for ERSPAN mode
So, now I get validonly chain 
 ASSERT FAIL(on line 1271) SAI_MIRROR_SESSION_ATTR_VLAN_TPID: validonly attibute is also validonly attribute, not allowed

How do we proceed ?

kcudnik:
Hey, i can relax valid only condition so there will be possible chaining validonly attributes.
Same i can do later with conditional attributes, but for this i need to rethink metadata consequences what maybe impacted.

For your case, please update saisanitycheck.c at line 1260, add similar to GRE_KEY exception for your attributes, and after i rethink possible consequences i remove this chaining restriction.

itaibaz:
I modified the code accordingly, to allow valid only chain on vlan header valid

lguohan:
retest this please

itaibaz:
retested and passed

PR #483, opened 2017-03-31
Add custom range start end test

kcudnik:
purpose of this test is to make sure all objects define custom range start and end attributes


lguohan:
I probably did something wrong on resolving the conflict and cause the pr build error. @kcudnik , can you check?

kcudnik:
yea, will check

PR #487, opened 2017-04-03
Modified attr name to get max ports supported on switch

AshokDaparthi:
Previous attribute SAI_SWITCH_ATTR_PORT_NUMBER is not clear. 
is this for active ports or max supported ports by NPU ?
Max supported ports are more relevant, since apps can query SAI_SWITCH_ATTR_PORT_LIST to get active list or adapter will create/delete ports as part of switch_init. 
They will have active port count in DB.

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -225,12 +225,12 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_START,
 
     /**
-     * @brief The number of ports on the switch
+     * @brief Maximum number of supported ports on the switch
      *
      * @type sai_uint32_t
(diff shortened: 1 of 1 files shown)


kcudnik:
port_number was number of ports on the port_list, so for maximum ports should we have another attribute ? also ports can be created removed and splitted, also should number of lanes determine how many ports we can create ?

lguohan:
SAI_SWITCH_ATTR_PORT_NUMBER is the active ports. MAX NUMBER OF PORTS should be another attribute.

AshokDaparthi:
@lguohan @atitjain - Updated with new attribute.

PR #488, opened 2017-04-04
Added an ALIASE for "isvlan" flag in DoxyFile.

NaiduMarineni:
An ALIASE for "isvaln" flag is missing in DoxyFile. This flag is used in saimirror.h attribute properties.


kcudnik:
Actually those aliases could be removed since metadata in meta directory has its own Doxygen file where this is already supported

NaiduMarineni:
Is there any specific reason for maintaining two Doxyfiles? As there is a confusion of which file to use.

kcudnik:
Im using the one in metadata as helper for headers validation

NaiduMarineni:
Is there any specific reason for maintaining two Doxyfiles? As there is a confusion of which file to use.

PR #489, opened 2017-04-04
Add PG port and index attributes

itaibaz:
For PG creation, it is needed to specify what is the Port and PG index
(similar to Queue API)

### DIFF
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -1502,7 +1502,7 @@ void check_attr_key(
 
             case SAI_ATTR_VALUE_TYPE_OBJECT_ID:
 
-                if (md->objecttype == SAI_OBJECT_TYPE_QUEUE && md->attrid == SAI_QUEUE_ATTR_PORT)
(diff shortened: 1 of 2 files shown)


meta/saisanitycheck.c
```diff
@@ -1502,7 +1502,7 @@ void check_attr_key(
 
             case SAI_ATTR_VALUE_TYPE_OBJECT_ID:
 
-                if (md->objecttype == SAI_OBJECT_TYPE_QUEUE && md->attrid == SAI_QUEUE_ATTR_PORT)
+                if ((md->objecttype == SAI_OBJECT_TYPE_QUEUE || md->objecttype == SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP) && md->attrid == SAI_QUEUE_ATTR_PORT)
```
kcudnik: this condition is wrong !

since you are matching queue attribute with ingress priority group, please revert

kcudnik:
wrong condition !

PR #492, opened 2017-04-05
Added P4 Behavioral Model

snifbh:
This pull request contains:
- A general switch P4 behavioral model.
- Implementation of the SAI API over the P4 BM. 
- unit test example for the sai api usage.
- Thirft sai server implementation and ptf tests.

currently only L2 (1D,1Q) flows are supported.


kcudnik:
Almost 600 files in one commit ? can you split that into couple PR ? Are all files go to BM directory ?

snifbh:
1) All files are in the BM directory.
2) this PR  needs the P4 environment (~60% of the files), found under BM/p4-softswtich/. it is required to compile the BM.
3) this pull request can be splitted: the PTF, unittest, and sai_thrift_server (BM/sai_adapter/tests/ ~100 files)  are done in another PR, but I think that the tests show usage examples and simplify the model understanding.

kcudnik:
Does this needs to be in SAI repo ? is that developed for SAI ? or maybe this can be in other repo and jus be git submodule here? I believe nobody will be able to do review of this code in this form

lguohan:
yes, this provides the official SAI compliant software switch.

lguohan:
@snifbh, can you rename the root directory to bm? we are using non capital letters.

YonatanPitz:
@lguohan  done.

lguohan:
did we modify any p4 switch behavior model? Is it possible to use submodule to link to a p4 switch behavior model here? The concern is that from time to time we probably want to upgrade to a new p4 switch behavior model. 

I prefer to only have following in the SAI repo.

- our p4 program, P4-SAI.
- Implementation of the SAI API over the P4 BM.
- unit test example for the sai api usage.
- Thirft sai server implementation and ptf tests.

Then, we can have a submodule link to bmv2 software, so that we can compile our P4-SAI model. 

This may be cleaner and more easy to understand our own contribution.

snifbh:
@lguohan
I agree that using the behavioral model as submodule here is much cleaner.
But we have implementation issue: 
Currently the behavioral model doesn't support external target (our P4-SAI MUST be in the behavioral-model/targets/ folder)
We might be able work around this, but it will be much less clean.
I'll try to open the external target feature with the P4 community, but for now, I suggest we keep it like this.

lguohan:
we can do this using our own build script, for example, first copy our target to the behaviroal-model/targets/folder and then compile.

snifbh:
@lguohan done.

PR #501, opened 2017-04-13
saithrift: tests: Support of bridge API for L2 tests

vkochan-mlnx:


vkochan:
Hi Guohan,
Thanks for merge this, but why these commits were squashed into 1 commit ?

lguohan:
we always to do squash merge, in case you want to them to be 2 commits, then you can send two pull requests.

PR #505, opened 2017-04-15
IPv6 Segment Route Proposal for 1.2

shihjeff:
Created this preliminary draft of the IPv6 Segment Route model.  Includes addition to behavioral model and APIs to configure SR origination


lguohan:
can you rebase to the current master?

lguohan:
also it is better to separate the proposal with the actual header file

kcudnik:
Please read my email response

inc/saisegmentroute.h
```diff
@@ -0,0 +1,168 @@
...
+     * @type sai_tlv_list_t
+     * @flags CREATE_AND_SET
+     */
+    SAI_SEGMENTROUTE_ENTRY_ATTR_TLV_LIST,
+
+    /**
+     * @brief List of Segments to Originate
+     *
+     * @type sai_segment_list_t
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
```
kcudnik: Sińce new 2 apibtypes are added parser.pl And saimrtadatsttpes attrvaluetypes needs to be updated aswell as saisanitycheck.c

inc/saitypes.h
```diff
@@ -530,6 +531,64 @@ typedef struct _sai_acl_capability_t
...
+ */
+typedef struct _sai_segment_list_t
+{
+    /** Number of IPv6 Segment Route entries */
+    uint32_t count;
+    /** Segment list */
+    sai_ip6_t *list;
+} sai_segment_list_t;
+
+/**
```
kcudnik: Those above 4 struct should be defined on saitypes.h

meta/parse.pl
```diff
@@ -263,6 +265,8 @@ sub ExtractDescription
 sai_qos_map_list_t      QOS_MAP_LIST
 sai_tunnel_map_list_t   TUNNEL_MAP_LIST
 sai_acl_capability_t    ACL_CAPABILITY
+sai_tlv_list_t          TLV_LIST
+sai_segment_list_t      SEGMENT_LIST
```
kcudnik: According attr value types needs to be included i side saimrtadatsttpes.h And saisanitycheck.c needs to be updated

inc/saitypes.h
kcudnik: Never Mondeo they are 😁

shihjeff: Thanks @kcudnik - I believe I have incorporated your comments as best as I understand them.  It should meta make correctly now.  Let me know if I haven't addressed anything

kcudnik:
Please asa HMAC Andrzej TLV to parser.pl for aspell capitalcase list

meta/saisanitycheck.c
```diff
@@ -332,6 +332,26 @@ bool sai_metadata_is_acl_field_or_action(
     return false;
 }
 
+bool sai_metadata_is_tlv_or_segment(
```
kcudnik: Please remove this merhod, no default value od only for acl, on Your proposal add default value as empty od internautów od list is not empty And thenyou font need thisncheck, please see other attribute how lista are using default value

shihjeff: @kcudnik - I was having trouble with this and used this as a workaround.  Since sai_tlv_list_t and sai_segment_list_t are structures defined in saitypes.h, I couldn't find an example of a default value of a struct.  Most examples are enums or pointers.  I tried 0, NULL without success.  Other examples like qos_map_list_t / tunnel_map_list_t doesn't have a default value defined.

If you can point me out an example, that would help me a lot.

kcudnik: Structss but as list like s32list etc add default value empty And modify saisanitycheck adding those new 2 values to accepting empty as default, there is a section there on chek

kcudnik: IT will be line 777 And 950

shihjeff: Done

shihjeff:
@kcudnik - could you help me decipher the meta error message?  I made some changes given feedback and I made the segmentroute_endpoint_entry style similar to how sairoute or saifdb implements it.  Its erroring in the autogenerated file and I'm not sure how to fix it.

kcudnik:
As error says union dont have member, you need to add member to object jest on saiobject.h

kcudnik:
Looks good, just question about prefix _ST_ wht this stand for? And second sai_segmentroute_st object is not referenced anywhere yet should IT be used in some attribute in sai_segmentroute_endpoint_entry?

shihjeff:
@kcudnik - ST stands for Source / Transmit.  It used to be called transmit but it was misleading since this table is used for both source and transmit policy application.  and SOURCE_TRANSMIT naming seemed too long, so I abbreviated it.  Its not for the endpoint entry table.  It's used in the separate table.

inc/saiobject.h
```diff
@@ -56,7 +56,7 @@ typedef struct _sai_object_key_t
         sai_mcast_fdb_entry_t     mcast_fdb_entry;
         sai_l2mc_entry_t          l2mc_entry;
         sai_ipmc_entry_t          ipmc_entry;
-        sai_segmentroute_endpoint_entry_t segmentroute_endpoint_entry;
+        sai_sr_pe_entry_t         sr_pe_entry;
```
kcudnik: should be sai_sr_ep_entry_t ?

ps. i don't mind with long name here

inc/saiobject.h
```diff
@@ -32,7 +32,7 @@
 #include <saiipmc.h>
 #include <saineighbor.h>
 #include <sairoute.h>
-#include <saisegmentroute.h>
```
kcudnik: this was better i think self explaining name

inc/saisr.h
```diff
@@ -0,0 +1,490 @@
...
+
+    /**
+     * @brief IPv6 Destination
+     */
+    sai_ip_address_t destination;
+
+    /**
+     * @brief Color to match on policy
+     */
+    sai_uint32_t policy_color;
```
kcudnik: in this PR you changed this struct 2 times already, it seems to me that this design needs a little bit more polishing before final version ? also policy color should that be an attribute ? does this needs to really be in the struct? we should keep struct members to minimum

inc/saisr.h
```diff
@@ -0,0 +1,490 @@
...
+ */
+typedef sai_status_t (*sai_get_sr_pe_entry_attribute_fn)(
+        _In_ const sai_sr_pe_entry_t *pe_entry,
+        _In_ uint32_t attr_count,
+        _Inout_ sai_attribute_t *attr_list);
+
+/**
+ * @brief Segment Route methods table retrieved with sai_api_query()
+ */
+typedef struct _sai_sr_api_t
```
kcudnik: segmentroute api was better self explaining name

inc/saisr.h
kcudnik: also should color be enum ? and not uint32?

shihjeff:
@kcudnik thanks for the feedback.  Few points:
1. Color is actually an ID, can't be an enum.  IETF spec says color number & DIP to derive the SR policy, thus, I used unit32 
2. Abbreviations are due to comments from sai meeting feedback
3. Since this is proposal, going through a lot of churn due to various offline comments.  Apologize on the overtime since I'm trying to adapt to all the feedback

inc/saisr.h
kcudnik: ok

kcudnik:
I added some more pedantic rules about style in headers, can you respin your change to see if your changes passes those checks ?

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+
+![SAI v6SR bm](figures/sai_v6SR_bm.png "Figure 1: Behavioral Model Addition. ")
+__Figure 1: Behavioral Model Addition.__
+
+![SAI v6SR bm1](figures/sai_v6SR_bm1.png "Figure 2: Source Behavior ")
+__Figure 2: Source Behavior.__
+
+![SAI v6SR bm2](figures/sai_v6SR_bm2.png "Figure 3: Transit Behavior ")
+__Figure 3: Transit Behavior.__
+
```
rodnymolina: As you know, there's no 'transit' specialized behavior per-se in regular SRv6: you're either a 'source' or an 'endpoint'. Real 'transit' nodes are those that handle an IPv6 packet with a DA/DIP different than themselves, so they are not expected to process the SRH, thus, they are not existing from a SRv6 standpoint. All the other nodes, excepting the 'source', are SRv6 'endpoints'. But i guess that your 'transit' semantic is different than mine in the sense that you are referring to nodes that are neither 'ingress' nor 'egress' from a traffic-flow perspective. I got it, no issues there. The reason I'm making this point is because in your 'transit' diagram, you are drawing a "set sidlist" action as part of the nh-table processing, which can be understood as if real srv6-transit nodes require a new sid-list to be provided by the user to allow the srh-pop operation. Please confirm that all 'srv6-transit' nodes are simply expected to execute the following instructions (further below) as per SRv6 drafts, and that there's no need to pass a new/shorter sid-list to SAI. If my understanding is correct, please adjust/remove the transit diagram if that's ok with you.

   1. IF SegmentsLeft > 0 THEN
   2.    decrement SL
   3.    update the IPv6 DA with SRH[SL]
   4.    FIB lookup on updated DA
   5.    forward accordingly to the matched entry
   6.  ELSE
   7.    drop the packet

[ After writing all the above, i now realize that by 'transit' you might be referring to the semantics exposed in SR-TE/policy draft --to allow insertions of sid-lists within an incoming sid-list. If that's the case, please ignore my comment above, and write a note to prevent misunderstandings like the one i may have incurred ;-) ]

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+3. Introduce SAI APIs to define SR properties including segment and TLV definitions.
+
+## Behavioral Model
+
+In order to add IPv6 Segment Routing, it requires three mechanisms:
+1. Way to specify which flows will be marked for SR source or transit
+2. Way to add SR header (source / transit) or modify / remove SR header (endpoint) before normal L2/L3 processing
+
+For the first mechanism, ACL is used to match on any flow characteristics to determine the policy "color".  The color and packet DIP is used in the ACL match lookup to determine the Policy's Binding Segment ID (BSID) which is enabled in the next-hop group table. Alternatively, one can use the route table to just use DIP alone to determine the BSID.  The BSID is used in the next-hop group table to group Segment ID Lists (SID Lists) and ECMP hash between the possible paths
+
```
shihjeff: It is already supported via: SAI_NEXT_HOP_GROUP_MEMBER_ATTR_WEIGHT

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
shihjeff: I do mean the special "transit" scenario where transit node can append packet with additional SR headers creating a stack of them.  Will add note

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+### APIs
+
+To start with, the basic create/remove entry and set/get attributes APIs are included
+
+    create_segmentroute_sidlist;
+    remove_segmentroute_sidlist;
+    set_segmentroute_sidlist_attribute;
+    get_segmentroute_sidlist_attribute;
+    create_segmentroute_sidlists;
+    remove_segmentroute_sidlists;
```
shihjeff: If you look at other modules, adding a "s" at the end seems to be the syntax.

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Examples ##
+### Example 1 - SR Source / Transit
+The following example
+1. Creates a Next-Hop Group (BSID)
+2. Creates an ACL entry to specify a specific flow to bind to a Next-Hop Group (BSID)
+3. Writes SID List Object to add 3 Segments and an Ingress Node TLV
```
shihjeff: This is just an example.  I was using 3 as a concrete example.

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+    CONVERT_STR_TO_IPV6(v6sr_entry_attrs[0].value.objlist.list[1], "2001:db8:85a3::8a2e:370:2345");
+    CONVERT_STR_TO_IPV6(v6sr_entry_attrs[0].value.objlist.list[2], "2001:db8:85a3::8a2e:370:3456");
+    v6sr_entry_attrs[1].id = SAI_SEGMENTROUTE_SIDLIST_ATTR_TLV;
+    v6sr_entry_attrs[1].value.objlist.count = 1;
+    v6sr_entry_attrs[1].value.objlist.list[0].tlv_type = SAI_TLV_TYPE_INGRESS;
+    CONVERT_STR_TO_IPV6(v6sr_entry_attrs[1].value.objlist.list[0].ingress_node, "2001:db8:85a3::8a2e:370:9876");
+    v6sr_entry_attrs[2].id = SAI_SEGMENTROUTE_SIDLIST_ATTR_TYPE;
+    v6sr_entry_attrs[2].value = SAI_SEGMENTROUTE_SIDLIST_TYPE_ENCAPS_ORIGINATION; 
+    v6sr_entry_attrs[3].id = SAI_SEGMENTROUTE_SIDLIST_ATTR_SEGMENTROUTE_BSID;
+    v6sr_entry_attrs[3].value = bsid_id;
```
shihjeff: This is a typo.  It was leftover from the previous model.  Will remove

inc/saiswitch.h
```diff
@@ -1269,6 +1269,30 @@ typedef enum _sai_switch_attr_t
...
+    SAI_SWITCH_ATTR_SEGMENTROUTE_MAX_SIDLISTS,
+
+    /**
+     * @brief List of Type Length Value types supported for source
+     *
+     * @type sai_s32_list_t sai_tlv_type_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_SEGMENTROUTE_TLV_TYPE,
+
```
shihjeff: Its not maximum, its the actual individual TLV types a vendor can support.  For example if a vendor only supports Ingress Node TLV type but not all others, it would pass Ingress Node TLV enum in

inc/saiswitch.h
```diff
@@ -1269,6 +1269,30 @@ typedef enum _sai_switch_attr_t
...
+    SAI_SWITCH_ATTR_SEGMENTROUTE_MAX_SEGMENTS,
+
+    /**
+     * @brief Max number of SID Lists supported
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_SEGMENTROUTE_MAX_SIDLISTS,
+
```
shihjeff: I'll remove this for now.  My original idea was for the transit case where multiple SR headers can be appended and this would be the MAX # of headers a vendor could parse.  But due to confusion and folks not ready for SR header stack, I'll just remove this.  We can add if there is demand.

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+    saistatus = sai_acl_api->create_acl_entry(&acl_entry, 4, acl_entry_attrs);
+    if (saistatus != SAI_STATUS_SUCCESS) {
+        return saistatus;
+    }
+
+    v6sr_entry_attrs[0].id = SAI_SEGMENTROUTE_SIDLIST_ATTR_SEGMENT_LIST
+    v6sr_entry_attrs[0].value.objlist.count = 3;
+    CONVERT_STR_TO_IPV6(v6sr_entry_attrs[0].value.objlist.list[0], "2001:db8:85a3::8a2e:370:1234");
+    CONVERT_STR_TO_IPV6(v6sr_entry_attrs[0].value.objlist.list[1], "2001:db8:85a3::8a2e:370:2345");
+    CONVERT_STR_TO_IPV6(v6sr_entry_attrs[0].value.objlist.list[2], "2001:db8:85a3::8a2e:370:3456");
```
shihjeff: Fixed

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+
+### SID List Object APIs
+#### Vendor Support Advertisement
+
+Included is also a way for vendors to advertise device support include the number of segments and TLV types that can be originated
+
+    SAI_SWITCH_ATTR_SEGMENTROUTE_MAX_SEGMENTS
+    SAI_SWITCH_ATTR_SEGMENTROUTE_MAX_SIDLISTS
+    SAI_SWITCH_ATTR_SEGMENTROUTE_TLV_TYPE
+
```
shihjeff: Done

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+## Behavioral Model
+
+In order to add IPv6 Segment Routing, it requires three mechanisms:
+1. Way to specify which flows will be marked for SR source or transit
+2. Way to add SR header (source / transit) or modify / remove SR header (endpoint) before normal L2/L3 processing
+
+For the first mechanism, ACL is used to match on any flow characteristics to determine the policy "color".  The color and packet DIP is used in the ACL match lookup to determine the Policy's Binding Segment ID (BSID) which is enabled in the next-hop group table. Alternatively, one can use the route table to just use DIP alone to determine the BSID.  The BSID is used in the next-hop group table to group Segment ID Lists (SID Lists) and ECMP hash between the possible paths
+
+For the second mechanism, the route / next-hop group / next-hop tables are used to identify the SR DIP and manipulate via endpoint function or source/transit SID lists as objects in the next-hop table.  Afterwards the flow will head to the egress pipeline for futher lookup on the modified packet.  Thus, the only structural changes needed is in the next-hop table to handle the transit, source, and endpoint actions.
+
```
shihjeff: Elaborated a bit more

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+This specification proposes the following points:
+1. Introduce the concept of Segment Routing using IPv6.
+2. Introduce behavioral model modifications to support SR source, transit, and endpoint.
+3. Introduce SAI APIs to define SR properties including segment and TLV definitions.
+
+## Behavioral Model
+
+In order to add IPv6 Segment Routing, it requires three mechanisms:
+1. Way to specify which flows will be marked for SR source or transit
```
shihjeff: Done

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,207 @@
...
+3. https://tools.ietf.org/html/draft-filsfils-spring-segment-routing-policy-00
+
+This specification proposes the following points:
+1. Introduce the concept of Segment Routing using IPv6.
+2. Introduce behavioral model modifications to support SR source, transit, and endpoint.
+3. Introduce SAI APIs to define SR properties including segment and TLV definitions.
+
+## Behavioral Model
+
+In order to add IPv6 Segment Routing, it requires three mechanisms:
```
shihjeff: Done

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
rodnymolina: ah ok

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
rodnymolina: got it, you're referring to SR-TE/policy draft. A note to clarify should be enough.

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
rodnymolina: If that's the naming convention we are following, then I'm fine with it. I would have picked a different one, but the important thing here is that we keep consistency across modules, so i'm good.

inc/saiswitch.h
rodnymolina: makes sense to remove it for now

kcudnik:
You are introducin union, please name it to pass check, look for other union examples in saitypes.h

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,210 @@
...
+
+This specification proposes the following points:
+1. Introduce the concept of Segment Routing using IPv6.
+2. Introduce behavioral model modifications to support SR source, transit, and endpoint.
+3. Introduce SAI APIs to define SR properties including segment and TLV definitions.
+
+## Behavioral Model
+
+To clarify the behaviors of different nodes:
+1. Source Node.  This node first encapsulates the SR header on a native IPv6 packet and routes on the outer-most address in SR list.
```
rodnymolina: As per SRv6-srh draft, the source-node can either be 1) A host originating an IPv6 packet, or 2) An SR domain ingress router encapsulating a received IPv6 packet into an outer IPv6 header followed by an SRH.

I guess you are referring to the 2), right? If so, please adjust this sentence accordingly.

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,210 @@
...
+2. Transit Node.  Typically is unaware of SR header and only does native routing.  But there is an option to append the packet with another SR header to create a stack of SR headers via mechanism similar to Source Node.
```
rodnymolina: I find the following "transit-node" description quite precise. If you agree, perhaps we can adjust yours to avoid any misunderstandings.

As per SRv6-srh draft:

"...the only node who is allowed to inspect the Routing Extension Header (and therefore the SRH), is the node
   corresponding to the DA of the packet.  Any other transit node MUST NOT inspect the underneath routing header and MUST forward the packet towards the DA and according to the IPv6 routing table."

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
```diff
@@ -0,0 +1,210 @@
...
+3. Endpoint Node.  Last node of the SR header stack.  User can choose an endpoint function to process the packet further such as removing the SR header and natively route.
```
rodnymolina: Based on this description you provided, would it be possible to rename this one to "egress node"? See that in SRv6 world, "endpoint node" semantics are different than what you are describing, and in fact "endpoint node" isn't usually the "egress node". An alternative to renaming this node type, and perhaps a better idea, is to change your "endpoint node" description to adjust to SRv6 semantics.

As per SRv6-srh draft: "The SR segment endpoint node is the node whose MyLocalSID table contains an entry for the DA of the packet."

inc/sainexthop.h
```diff
@@ -45,11 +45,64 @@ typedef enum _sai_next_hop_type_t
...
+    /** IPv6 Segment Route SID List */
+    SAI_NEXT_HOP_TYPE_SEGMENTROUTE_SIDLIST,
+
+    /** IPv6 Segment Route Endpoint Function */
+    SAI_NEXT_HOP_TYPE_SEGMENTROUTE_ENDPOINT
 
 } sai_next_hop_type_t;
 
 /**
+ * @brief Enum defining Endpoint Action types
```
rodnymolina: If not already done (don't think i've seen it), could you please write a comment about which endpoint-types are you supporting with this initial proposal?

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
shihjeff: Done

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
shihjeff: Done

doc/SAI-Proposal-IPv6_Segment_Routing-1.md
shihjeff: Done

inc/sainexthop.h
shihjeff: I'm confused on this.  It should be apparent from the enum list, which corresponds to the list in https://tools.ietf.org/html/draft-filsfils-spring-srv6-network-programming-00#section-4

rodnymolina: Ah ok, i thought you were planning to support only a subset of the endpoint functions, so no issues here.

Btw, you don't mention anything about the transit functions though. Are you planning to add them at a later time?

shihjeff: We do handle transit functions as described in https://tools.ietf.org/html/draft-filsfils-spring-srv6-network-programming-00#section-5.
T = normal Route / Next Hop behavior
T.insert = SAI_SEGMENTROUTE_SIDLIST_TYPE_INSERT
T.encap = SAI_SEGMENTROUTE_SIDLIST_TYPE_ENCAPS

I didn't add T.encap.l2 in version 1

shihjeff:
@lguohan / @kcudnik / @xinliu-seattle - Can we merge (since it didn't happen last Friday)?  I've fixed merge conflicts and all comments have been addressed.

kcudnik:
test this please

kcudnik:
this commit removes check_attr_default_attrvalue from sanity check

PR #507, opened 2017-04-24
saibridge: Add admin state forthe bridge port

vkochan-mlnx:
Add SAI_BRIDGE_PORT_ATTR_ADMIN_STATE attribute, it allows to control
admin state of the particular vport.


lguohan:
what is the main use case for having this attribute?

vkochan-mlnx:
Before removing a bridge port, need to disable it by setting bridge port admin mode to false, then flush the fdb entries for it, and then remove it.

vkochan-mlnx:
I did update commit with a "brief" explanation for this attribute.

PR #516, opened 2017-04-28
Introduce bulk route api

kcudnik:
Purpose of this api is to speed up operations on routes, since we can have a lot of routes in system, and setting next hop for most of them can be done in one operation

### DIFF
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -577,11 +577,16 @@ typedef struct _sai_attribute_t {
 
 typedef enum _sai_bulk_op_type_t {
 
-    /* bulk operation stops on the first failed creation. Rest of objects will use SAI_STATUS_NON_EXECUTED
-     * return status value. */
+    /*
+     * @brief Bulk operation stops on the first failed creation
+     *
+     * Rest of objects will use SAI_STATUS_NON_EXECUTED return status value.
+     */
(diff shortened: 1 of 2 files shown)


kcudnik:
Also to make it more generic we could use sai_object_key_t instead of sai_route_entry_t and we could use those declarations in other non object id objects

kcudnik:
test this please

kcudnik:
test this please

inc/saitypes.h
```diff
@@ -594,11 +594,16 @@ typedef struct _sai_attribute_t
 
 typedef enum _sai_bulk_op_type_t
```
tushar-ty: This should be called 'mode' as 'type' is a bit misleading.

tushar-ty: To be more specific it should be called 'error_mode'.

kcudnik: this name is not subject of this PR, but sure error mode sounds better

tushar-ty: Yes, missed correcting this name earlier. But since we are at it, better to fix it before it gets much widely utilized.

kcudnik: sure, please fix your pr #560

tushar-ty: Done. Thanks. Pls approve the other one so that this can be fixed as well.

PR #532, opened 2017-05-03
Add support for IP FRR

michaelsiuda:
Hi,

Please review my proposal for adding support for IP FRR into SAI.  I will be presenting this proposal at the next SAI call (5th May).

Best regards,
Michael Siuda
Metaswitch Networks


inc/sainexthopgroup.h
```diff
@@ -41,11 +41,45 @@ typedef enum _sai_next_hop_group_type_t
...
+    SAI_NEXT_HOP_GROUP_TYPE_PROTECTION,
+
     /* Other types of next hop group to be defined in the future, e.g., WCMP */
 
 } sai_next_hop_group_type_t;
 
 /**
+ * @brief No switchover triggers supported
+ */
+#define SAI_NEXT_HOP_SWITCHOVER_TRIGGER_NULL                         0x00000000
```
kcudnik: Please declare those as ENUM and not as define, inside sdk vendor can retlanslate those to uint8 if needed, but sai interface will be clear

inc/sainexthopgroup.h
```diff
@@ -81,6 +115,37 @@ typedef enum _sai_next_hop_group_attr_t
...
 
     /**
+     * @brief Events that can initiate switchover from primary to backup next hop
+     *
+     * Can be a combination of the following:
+     * SAI_NEXT_HOP_SWITCHOVER_TRIGGER_NULL,
+     * SAI_NEXT_HOP_SWITCHOVER_TRIGGER_BFD,
+     * SAI_NEXT_HOP_SWITCHOVER_TRIGGER_PORT_DOWN
+     *
+     * @type sai_uint8_t
```
kcudnik: declare those as enum and use enum type here

inc/sainexthopgroup.h
```diff
@@ -81,6 +115,37 @@ typedef enum _sai_next_hop_group_attr_t
...
+     * @validonly SAI_NEXT_HOP_GROUP_ATTR_TYPE == SAI_NEXT_HOP_GROUP_TYPE_PROTECTION
+     */
+    SAI_NEXT_HOP_GROUP_ATTR_SWITCHOVER_TYPE,
+
+    /**
+     * @brief Trigger a switchover from primary to backup next hop
+     *
+     * @type bool
+     * @default false
+     * @validonly SAI_NEXT_HOP_GROUP_ATTR_TYPE == SAI_NEXT_HOP_GROUP_TYPE_PROTECTION and SAI_NEXT_HOP_GROUP_ATTR_CLEAR_SWITCHOVER == false
```
kcudnik: "and" condition operand is not supported yet

kcudnik:
test this please

inc/sainexthopgroup.h
kcudnik: if those are flags then this parameter must be declared as sai_s32list_t enum, since sai don't uses flags

inc/sainexthopgroup.h
kcudnik: i added support of "and" condition

inc/sainexthopgroup.h
```diff
@@ -128,6 +193,27 @@ typedef enum _sai_next_hop_group_member_attr_t
...
     /**
+     * @brief Preferred role in the protection group
+     *
+     * Should only be used if the type of owning group is SAI_NEXT_HOP_GROUP_TYPE_PROTECTION
+     *
+     * @type sai_next_hop_group_member_protection_role_t
+     * @flags CREATE_ONLY
+     * @default SAI_NEXT_HOP_GROUP_MEMBER_PRIMARY
+     */
+    SAI_NEXT_HOP_GROUP_MEMBER_ATTR_PREFERRED_ROLE,
```
kcudnik: change to SAI_NEXT_HOP_GROUP_MEMBER_ATTR_PREFERRED_PROTECTION_ROLE,

inc/sainexthopgroup.h
```diff
@@ -128,6 +193,27 @@ typedef enum _sai_next_hop_group_member_attr_t
...
+
+    /**
+     * @brief The actual role in protection group
+     *
+     * Should only be used if the type of owning group is SAI_NEXT_HOP_GROUP_TYPE_PROTECTION
+     *
+     * @type sai_next_hop_group_member_protection_role_t
+     * @flags READ_ONLY
+     */
+    SAI_NEXT_HOP_GROUP_MEMBER_ATTR_OBSERVED_ROLE,
```
kcudnik: changet to SAI_NEXT_HOP_GROUP_MEMBER_ATTR_OBSERVED_PROTECTION_ROLE

inc/sainexthopgroup.h
michaelsiuda: As discussed in the call last week, the control plane will no longer provide these flags but identifiers for the port and BFD session.

inc/sainexthopgroup.h
michaelsiuda: Great, thank you.

inc/sainexthopgroup.h
michaelsiuda: As discussed in the call last week, the control plane will no longer provide these flags but identifiers for the port and BFD session.

inc/sainexthopgroup.h
kcudnik: Can you explain more? I not attended to that meeting so i don't have context here

kcudnik:
please fix errors

inc/sainexthopgroup.h
michaelsiuda: Hi Kamil, apologies for a late reply.  

During the call it was suggested that if control plane only indicates for example that the switching entity should trigger a switchover if the corresponding BFD session fails, then it would be non trivial for the switch to figure out which BFD session to monitor in this case.  
Therefore, we decided that it will be better if the control plane provides BFD session id rather than just a generic flag, when it wants the switch to trigger a switchover.  
The same logic applies to the port down event - control plane will provide an identifier for the associated port.

Let me know if you'd like more clarification.

inc/sainexthopgroup.h
```diff
@@ -128,6 +181,27 @@ typedef enum _sai_next_hop_group_member_attr_t
     SAI_NEXT_HOP_GROUP_MEMBER_ATTR_WEIGHT,
 
     /**
+     * @brief Preferred role in the protection group
+     *
+     * Should only be used if the type of owning group is SAI_NEXT_HOP_GROUP_TYPE_PROTECTION
```
kcudnik: add mandatory_on_create and condition

inc/sainexthopgroup.h
```diff
@@ -81,6 +100,40 @@ typedef enum _sai_next_hop_group_attr_t
...
+     */
+    SAI_NEXT_HOP_GROUP_ATTR_BFD_SESSION_ID,
+
+    /**
+     * @brief Identifier of the port associated with the primary next hop
+     *
+     * If the specified port fails, the switching entity triggers a switchover
+     * from the primary next hop to backup.
+     *
+     * @type sai_object_id_t
```
kcudnik: missing flags, default can be SAI_NULL_OBJECT_ID

inc/sainexthopgroup.h
```diff
@@ -128,6 +160,60 @@ typedef enum _sai_next_hop_group_member_attr_t
...
+     * @brief Identifier of the port associated with the next hop
+     *
+     * If the specified port fails, the switching entity marks this next
+     * hop as SAI_NEXT_HOP_GROUP_MEMBER_PROTECTION_ROLE_FAILED and does not use
+     * it to forward traffic. If there is a backup next hop available in this
+     * group then the backup's observed role is set to
+     * SAI_NEXT_HOP_GROUP_MEMBER_PROTECTION_ROLE_FORWARDING and it is used to
+     * forward traffic.
+     *
+     * @type sai_uint32_t
```
lguohan: can this be sai_object_id_t?

inc/sainexthopgroup.h
```diff
@@ -128,6 +160,60 @@ typedef enum _sai_next_hop_group_member_attr_t
...
+     * @flags CREATE_AND_SET
+     * @default 0
```
lguohan: should this be the sai_object_id_t and the object id type is port?

inc/sainexthopgroup.h
lguohan: @kcudnik, do you have specific suggestion?

should this be @validonly SAI_NEXT_HOP_GROUP_ATTR_TYPE = SAI_NEXT_HOP_GROUP_TYPE_PROTECTION

kcudnik: yea an be like this

michaelsiuda: The problem with what you suggested is that you're trying to enforce a condition on an attribute on Next Hop Group Member based on the value of attribute of Next Hop Group.  Is there any way to express this with the @validonly syntax, or should I just add a comment saying something in the lines of "The field is only considered if the parent next hop group is PROTECTION.  Otherwise it is ignored"?

michaelsiuda:
Hi @kcudnik, @lguohan, this is the (intended) final version of the API.  Could you please review and let me know if you find any problems?  

*The pipeline picture is still TBD, but it shouldn't affect the specs

Thanks,
Michael

inc/sainexthopgroup.h
lguohan: @kcudnik, can we add condition on the other object's attribute? It seems it is not possible to do it right now?

---
In reply to: [124805486](https://github.com/opencomputeproject/SAI/pull/532#discussion_r124805486) [](ancestors = 124805486)

lguohan:
@kcudnik , I have resolve the conflict, can you check? Also, there is a question from Michael on meta data for attribute SAI_NEXT_HOP_GROUP_MEMBER_ATTR_CONFIGURED_ROLE. can you check?

inc/sainexthopgroup.h
```diff
@@ -128,6 +167,45 @@ typedef enum _sai_next_hop_group_member_attr_t
...
+     * If the specified objects fails, the switching entity marks this
+     * next hop as SAI_NEXT_HOP_GROUP_MEMBER_PROTECTION_ROLE_FAILED and does
+     * not use it to forward traffic. If there is a backup next hop available
+     * in this group then the backup's observed role is set to
+     * SAI_NEXT_HOP_GROUP_MEMBER_PROTECTION_ROLE_FORWARDING and it is used to
+     * forward traffic.
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_LAG, SAI_OBJECT_TYPE_ROUTER_INTERFACE, SAI_OBJECT_TYPE_VLAN_MEMBER, SAI_OBJECT_TYPE_TUNNEL
```
kcudnik: should this also contain bridge_port object type ?

inc/saiswitch.h
```diff
@@ -1460,6 +1460,15 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_PFC_TC_DLR_INTERVAL,
 
     /**
+     * @brief Get the list of supported protected object types.
+     *        See comment for SAI_NEXT_HOP_GROUP_MEMBER_ATTR_MONITORED_OBJECT for more details.
+     *
+     * @type sai_s32_list_t
```
kcudnik: @type sai_s32_list_t sai_object_type_t since this will be enum

inc/sainexthopgroup.h
michaelsiuda: Added.

inc/saiswitch.h
michaelsiuda: I've added this as you suggested, but then had to change the name of the attribute such that it ends with OBJECT_TYPE.  Is it not misleading if we're talking about  a list of object types?

lguohan:
@kcudnik, can you review and approve if it is ok for you?

inc/saiswitch.h
kcudnik: you can add "_LIST" at the end if you want, it should also be accepted

PR #541, opened 2017-05-05
add following vlan related attribute for LAG port in order to keep

lguohan:
add following vlan related attribute for LAG port in order to keep consistency among all the ports within the LAG.

When a port joins a LAG, it should use these LAG attributes instead
of its corresponding port attributes. When it leaves the LAG, corresponding
attributes should be restored to the SAI default value.

1. SAI_LAG_ATTR_PORT_VLAN_ID
2. SAI_LAG_ATTR_DEFAULT_VLAN_PRIORITY
3. SAI_LAG_ATTR_DROP_UNTAGGED
4. SAI_LAG_ATTR_DROP_TAGGED

### DIFF
--- a/inc/sailag.h
+++ b/inc/sailag.h
@@ -130,15 +130,6 @@ typedef enum _sai_lag_attr_t
     SAI_LAG_ATTR_DEFAULT_VLAN_PRIORITY,
 
     /**
-     * @brief Ingress Filtering (Drop Frames with Unknown VLANs)
-     *
-     * @type bool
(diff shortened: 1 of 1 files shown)


lguohan:
@itaibaz , @AshokDaparthi , please review.

itaibaz:
since there is another commit to remove port bind mode, you might want to remove lag bind mode from this commit

itaibaz:
otherwise looks ok

PR #561, opened 2017-05-11
Add member and entry lists to several objects

kcudnik:
Purpose is to have ability to query all object members/entries associated with current object, like in VLAN object we have read only attribute vlan member list, so we can obtain all vlan members for given vlan.
This is useful for objects which are created automatically when creating switch. For user created objects (like in this PR) user can track all objects when creating them, but it would be nice to have those attributes also present in SAI headers.


kcudnik:
test this please

PR #566, opened 2017-05-16
PFC RX/TX pause duration counters

itaibaz:


inc/saiport.h
```diff
@@ -1416,6 +1416,54 @@ typedef enum _sai_port_stat_t
     /** SAI port stat PFC 7 tx pkts */
     SAI_PORT_STAT_PFC_7_TX_PKTS,
 
+    /** PFC pause duration for RX and TX per PFC priority [uint64_t] */
+    SAI_PORT_STAT_PFC_0_RX_PAUSE_DURATION,
```
lguohan: can you explain the difference between TX versue RX pause duration?

itaibaz: RX pause duration for certain priority is a the duration quanta in ingress pause frame for that priority (a pause frame received by the switch)
While TX pause duration for certain priority is the duration quanta in egress pause frame for that priority (a pause frame sent by the switch)

lguohan: the description is clear to me. can you add the description in the comment?

itaibaz: done

PR #573, opened 2017-05-18
User defined traps action

itaibaz:
1) clarify neighbor trap usage
2) clarify the packet action
3) add user trap ID attribute to route entry, neighbor entry and fdb
entry


inc/saihostif.h
```diff
@@ -452,16 +455,20 @@ typedef enum _sai_hostif_user_defined_trap_type_t
...
+    /** Router traps */
     SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_ROUTER = SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_START,
 
-    /** Neighbor table traps (default packet action is drop) */
+    /**
+     * @brief Neighbor table traps
+     *
+     * Generated by neighbor table entry hit with action trap/log, or by neighbor table miss
+     */
     SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_NEIGH,
```
kcudnik: Can we also change this to SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_NEIGHBOR instad of having short name?

inc/saihostif.h
```diff
@@ -452,16 +455,20 @@ typedef enum _sai_hostif_user_defined_trap_type_t
...
     SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_ROUTER = SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_START,
 
-    /** Neighbor table traps (default packet action is drop) */
-    SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_NEIGH,
+    /**
+     * @brief Neighbor table traps
+     *
+     * Generated by neighbor table entry hit with action trap/log, or by neighbor table miss
+     */
+    SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_NEIGHBOR,
```
lguohan: i prefer to keep the name as it is to maintain backward compatibility

kcudnik: Is anyone using that yet ?
I was also thinking about backward compatibility and i have proposal of adding redefinitions of enums in sailegacy.h like that:
#define SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_NEIGH ((sai_hostif_user_defined_trap_type_t)SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_NEIGHBOR)

PR #574, opened 2017-05-18
Critical resource proposal

lguohan:


inc/saiacl.h
```diff
@@ -396,12 +396,26 @@ typedef enum _sai_acl_table_attr_t
...
+     * @flags READ_ONLY
+     */
+    SAI_ACL_TABLE_ATTR_AVAILABLE_ACL_ENTRY = SAI_ACL_TABLE_ATTR_START,
+
+    /**
+     * @brief Available ACL counters for this table
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_ACL_TABLE_ATTR_AVAILABLE_ACL_COUNTER,
```
lguohan: one acl table can be bind to different binding ports, there may be multiple instances of such table and each instance may be its own limit. Therefore, we should add this attribute in the acl group member.

inc/saiswitch.h
```diff
@@ -635,6 +635,86 @@ typedef enum _sai_switch_attr_t
...
+     */
+    SAI_SWITCH_ATTR_AVAILABLE_IPV6_NEIGHBOR_ENTRY,
+
+    /**
+     * @brief Available Next hop group member entries
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_AVAILABLE_NEXT_HOP_GROUP_MEMBER_ENTRY,
```
tushar-ty: Does the application want to know the total remaining members and not the members remaining per group ?

PR #579, opened 2017-05-23
PFC deadlock recovery config and event.

tushar-ty:


tushar-ty:
What does the following error mean:

perl parse.pl
doxygen comment can't be upper sticked: saitypes.h:
count;
    /** Map list */
    sai_map_t *list;

kcudnik:
Put a empty line between count and /** Map list */, as error states, doxygen comment cant be upper sticked (to other items)

meta/parse.pl
```diff
@@ -284,6 +285,7 @@ sub ExtractDescription
 sai_vlan_list_t         VLAN_LIST
 sai_qos_map_list_t      QOS_MAP_LIST
 sai_tunnel_map_list_t   TUNNEL_MAP_LIST
+sai_map_list_t          MAP_LIST
```
kcudnik: missing attr_value_type enum value in meta/saimetadatatypes.h

inc/saiqueue.h
```diff
@@ -258,6 +271,32 @@ typedef enum _sai_queue_stat_t
 } sai_queue_stat_t;
 
 /**
+ * @brief Enum defining Queue deadlock event state.
+ */
+typedef enum _sai_queue_pfc_deadlock_event_type_t
+{
+    /** PFC deadlock detected */
+    SAI_QUEUE_PFC_DL_START = 0x00000001,
```
kcudnik: all enums must start with 0

meta/parse.pl
```diff
@@ -251,8 +251,9 @@ sub ExtractDescription
 sai_vlan_list_t         vlanlist
 sai_acl_field_data_t    aclfield
 sai_acl_action_data_t   aclaction
-sai_qos_map_list_t      qosmap
-sai_tunnel_map_list_t   tunnelmap
```
kcudnik: dont change those, those must match sai_attribute_value_t values names!

inc/saiqueue.h
```diff
@@ -258,6 +271,32 @@ typedef enum _sai_queue_stat_t
...
+
+    /** PFC deadlock recovery ended */
+    SAI_QUEUE_PFC_DL_END
+
+} sai_queue_pfc_deadlock_event_type_t;
+
+/**
+ * @brief Notification data format received from SAI queue deadlock event callback
+ */
+typedef struct _sai_queue_deadlock_notification_t
```
kcudnik: can you add _data_t at the end of struct name ? like in fdb notification ?

inc/saiqueue.h
kcudnik: please came up with other names cince suffix _START and _END is destined to mark start and end of ranges, and its actually removed by parser, you can use _STARTED, _ENDED

meta/parse.pl
kcudnik: i already added auto extraction here, you can remove that

meta/parse.pl
kcudnik: i already added auto extraction here, you can remove that

inc/saiqueue.h
tushar-ty: The reason why I wanted to start with a non-zero value was to be able to distinguish if something was not initialized properly and just passed in as zero vs an actual configured value. 
I think this is a general issue in SAI all over and should be addressed. For now I'll set this to zero.

meta/parse.pl
tushar-ty: ok

meta/parse.pl
tushar-ty: ok

inc/saiqueue.h
tushar-ty: ok

inc/saiqueue.h
tushar-ty: ok

meta/parse.pl
tushar-ty: Still seeing one issue:

saimetadata.c:29581:38: error: 'SAI_ATTR_VALUE_TYPE_MAP_LIST' undeclared here (not in a function)
     .attrvaluetype                 = SAI_ATTR_VALUE_TYPE_MAP_LIST,
                                      ^
Makefile:69: recipe for target 'saimetadata.o' failed

inc/saiqueue.h
kcudnik: For this problem there are always 2 sides And each one claims they are right. Lets Bering this topic to sai meeting And see. As for is fomething id initialized or not, you write unittests once And you are set to go. You should write them im any of those cases

kcudnik:
You are missing attr value type declaration on saimetadatatypea.h to run tests locally just enter meta directory And type make

meta/parse.pl
kcudnik: Yes you need to explicitly declare IT, read my last comments

meta/saimetadatatypes.h
```diff
@@ -190,11 +190,6 @@ typedef enum _sai_attr_value_type_t
     SAI_ATTR_VALUE_TYPE_INT32_RANGE,
 
     /**
-     * @brief Attribute value is vlan list.
-     */
-    SAI_ATTR_VALUE_TYPE_VLAN_LIST,
-
-    /**
```
kcudnik: please do not shuffle sort or reorder already existing enums since you are changin their numeric values and lib is not backward compatible in that way

tushar-ty: I think we should fix the lib as the enums should not be defined haphazardly. Keeping similar types organized together will help in the long run.

tushar-ty:
Hi Kamil, whats the issue here: ?

 ASSERT FAIL(on line 580) SAI_SWITCH_ATTR_PFC_COS_DLD_INTERVAL_MAP_LIST: attr value type is not supported, FIXME
Makefile:52: recipe for target 'all' failed

-thanks

kcudnik:
Yes, assert shows you line 580, since all lists etc must be explicitly added to sanity check please go to that line in saisanitycheck.c 
And take look around, there are usually some comments whats is going on, and it should be straight forward how to add new enum.
I will also review that on github when you update PR

inc/saiswitch.h
```diff
@@ -1269,6 +1280,55 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * If the monitored queue is in XOFF state for more than this duration then
+     * its considered to be in a PFC deadlock state and recovery process is kicked off.
+     * Note: Use COS value as key and timer interval as value.
+     *
+     * @type sai_map_list_t
+     * @flags CREATE_AND_SET
+     * @default empty
+     */
+    SAI_SWITCH_ATTR_PFC_COS_DLD_INTERVAL_MAP_LIST,
```
kcudnik: i saw error message that it force you to add MAP_LIST at the end, but thats oly when type is enum, but this map_list is struct, you can try to remove it and see if it will work should not complain about that, if it will complain i will take a deep look whats hapenning

tushar-ty: Removed the trailing MAP_LIST, but still seeing this error:

should this be CREATE_AND_SET?
 ASSERT FAIL(on line 828) SAI_SWITCH_ATTR_PFC_COS_DLD_INTERVAL: default value is required but this attr value type is not supported yet
Makefile:52: recipe for target 'all' failed
make: *** [all] Error 1

I think the default=empty is not working for this yet.

inc/saiswitch.h
```diff
@@ -1229,6 +1229,17 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_TAM_EVENT_NOTIFY,
 
     /**
+     * @brief Set Switch PFC deadlock event notification callback function passed to the adapter.
```
tushar-ty: Need to move this to line 1281 .

meta/saisanitycheck.c
```diff
@@ -821,6 +821,7 @@ void check_attr_default_required(
             break;
 
         case SAI_ATTR_VALUE_TYPE_POINTER:
+        case SAI_ATTR_VALUE_TYPE_MAP_LIST:
```
kcudnik: please move this to line 788

kcudnik:
test this please

PR #595, opened 2017-05-28
Fix acronyms

kcudnik:


meta/acronyms.txt
```diff
@@ -55,7 +55,7 @@ RSPAN - Remote Span
 SAI - Switch Abstraction Interface
 SDK - Software Development Kit
 SG
-SNMP - Services and Protocols for Advanced Networks
```
stcheng: what is SG above?

kcudnik: in saiswitch.h "Both *G/SG lookup supported" not familiar with technology no clue what SG mean :)

PR #598, opened 2017-05-29
Modified attr name to get max ports supported on switch

AshokDaparthi:
Discussed as part of PR #487 and decided to add new attribute in master.  

Issue: 
Previous attribute SAI_SWITCH_ATTR_PORT_NUMBER is not clear.
is this for active ports or max supported ports by NPU ?
Max supported ports are more relevant, since apps can query SAI_SWITCH_ATTR_PORT_LIST to get active list or adapter will create/delete ports as part of switch_init.
They will have active port count in DB.


inc/saiswitch.h
```diff
@@ -237,12 +237,20 @@ typedef enum _sai_switch_attr_t
...
 
     /**
-     * @brief The number of ports on the switch
+     * @brief Number of active ports on the switch
      *
      * @type sai_uint32_t
      * @flags READ_ONLY
      */
-    SAI_SWITCH_ATTR_PORT_NUMBER = SAI_SWITCH_ATTR_START,
+    SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS = SAI_SWITCH_ATTR_START,
```
lguohan: can you define "active port"? what is the definition of "active port"?

lguohan: for backward compatibility, please follow commit https://github.com/opencomputeproject/SAI/commit/3cc57f24b0d5df1b0e41abc5450cbbe2592352bf

AshokDaparthi: Done

lguohan:
use the @ignore for backward compatibility

lguohan:
should be ACTIVE NUMBER of ports for backward compatiblity

inc/saiswitch.h
```diff
@@ -1379,6 +1387,13 @@ typedef enum _sai_switch_attr_t
 #define SAI_KEY_HW_PORT_PROFILE_ID_CONFIG_FILE    "SAI_HW_PORT_PROFILE_ID_CONFIG_FILE"
 
 /**
+ * @def SAI_SWITCH_ATTR_PORT_NUMBER
+ *
+ * For backward compatibility.
+ */
+#define SAI_SWITCH_ATTR_PORT_NUMBER     SAI_SWITCH_ATTR_MAX_NUMBER_OF_SUPPORTED_PORTS
```
kcudnik: declare this as enum and use @ignore  tag

PR #599, opened 2017-05-29
Organizationally Unique Identifier for 25G/50G auto negotiation

AshokDaparthi:
Original PR#427.


AshokDaparthi:
@kcudnik -  "I added support for number conditions in #547. Please update your PR". But still i am seeing the error for condition . could you please check it.

PR #600, opened 2017-05-29
WRED support to buffer pool and port.

AshokDaparthi:
At present WRED supports in queue level. As part of this proposal we are extending WRED/ECN support to buffer pools and port.

Buffer Pool WRED: 
WRED thresholds are allowed to configure per each buffer pool. When WRED thresholds crosses the pool level WRD threshold, system will start drop/mark packets. 

Port Level WRED:
Port is part of multiple buffer pools (SAI_PORT_ATTR_QOS_EGRESS_BUFFER_PROFILE_LIST) with different buffer limits, extending the WRED also to per port per buffer pool i.e Key for port WRED are port oid and buffer pool oid.

WRED configuration are supported at multiple level now but DROP/MARK will be controlled at more granular level i.e Queue.  i.e Even though wred/ecn is enabled at buffer pool/port level, drop or marking will happen only on WRED/ECN  enabled queues referring to port/buffer pool.

### DIFF
--- a/inc/saiwred.h
+++ b/inc/saiwred.h
@@ -227,6 +227,20 @@ typedef enum _sai_wred_attr_t
     SAI_WRED_ATTR_ECN_MARK_MODE = 0x0000000d,
 
     /**
+     * @brief Pointer to buffer pool object id
+     *
+     * Pool id != #SAI_NULL_OBJECT_ID when profile is associated with port.
+     * with specific pool.
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_BUFFER_POOL
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_WRED_ATTR_POOL_ID = 0x0000000d,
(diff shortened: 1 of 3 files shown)


inc/saiwred.h
```diff
@@ -227,6 +227,20 @@ typedef enum _sai_wred_attr_t
...
+     * Pool id != #SAI_NULL_OBJECT_ID when profile is associated with port.
+     * with specific pool.
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_BUFFER_POOL
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_WRED_ATTR_POOL_ID = 0x0000000d,
```
kcudnik: repealted value as SAI_WRED_ATTR_ECN_MARK_MODE, thats why check fails

kcudnik:
You created a loop, from your attributes you can set WRED on BUFFER_POOL and you can set BUFFER_POOL on WRED object. Only one of those should be possible.

PR #602, opened 2017-05-30
Remove port Bind mode - Port can be part of .1D bridge and vlans at same time, Hence bind mode is not necessary.

AshokDaparthi:
Port can be part of .1D bridge and vlans at same time,  Hence bind mode is not necessary.
As discussed in OCP weekly meeting removing it.


lguohan:
can you put this comment into the commit message?

PR #603, opened 2017-05-30
Removing sai_tunnel_map_list_t

AshokDaparthi:
The tunnel map attribute SAI_TUNNEL_MAP_ATTR_MAP_TO_VALUE_LIST seems to be redundant with the tunnel_map_entry object as the latter can be used to achieve the same functionality as tunnel map list and also supports incremental addition or removal of mappings. The former seems to be an older representation of mapping key to value,

As discussed in OCP weekly call removing the tunnel map list.


lguohan:
@AshokDaparthi , conflict, can you rebase?

lguohan:
@itaibaz , @zubinshah , @kram, @tushar-ty . Are you aware of any usage of this API? can we remove it or do we need to give graceful period for people to move to the new API?

itaibaz:
our code is already ported to use tunnel map entry object
so we no longer use tunnel map list

kcudnik:
I didn't noticed at first that sai_map_t don't belong to tunnel map :) not should be fine

lguohan:
retest this please

PR #604, opened 2017-05-30
Flood disable on vlan  and bridge

AshokDaparthi:


inc/saibridge.h
```diff
@@ -326,6 +326,16 @@ typedef enum _sai_bridge_attr_t
...
 
     /**
+     * @brief To disable flooding traffic (Broadcast, unknown unicast,
+     * unknown multicast) on a bridge
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_BRIDGE_ATTR_FLOOD_DISABLE,
```
stcheng: what's the behavior when setting this attribute along with the `sai_bridge_port_fdb_learning_mode_t SAI_BRIDGE_PORT_FDB_LEARNING_MODE_DROP`? Will there be any conflicts? e.g. an unknown unicast packet comes and will it be flooded or dropped?

lguohan: they are unrelated, learning mode is about the src mac, flood disable is about dst mac.

PR #608, opened 2017-05-31
Add SG/XG acronyms

stcheng:


kcudnik:
not much helpful XD

stcheng:
this info is enough for googling 🔍

kcudnik:
i didn't found exact explanation, but seems like S i source (ip), and can be * as any, and G i multicast group ip

lguohan:
SG is the source specific multicast
*G is the source nonspecific multicast

kcudnik:
Can you add Guohan description to acronyms ?

stcheng:
@kcudnik updated.

PR #609, opened 2017-05-31
Wred for buffer pool and port

AshokDaparthi:
At present WRED supports in queue level. As part of this proposal we are extending WRED/ECN support to buffer pools and port.

Buffer Pool WRED:
WRED thresholds are allowed to configure per each buffer pool. When WRED thresholds crosses the pool level WRD threshold, system will start drop/mark packets.

Port Level WRED:
Port is part of multiple buffer pools (SAI_PORT_ATTR_QOS_EGRESS_BUFFER_PROFILE_LIST) with different buffer limits, extending the WRED also to per port per buffer pool i.e Key for port WRED are port oid and buffer pool oid.

WRED configuration are supported at multiple level now but DROP/MARK will be controlled at more granular level i.e Queue. i.e Even though wred/ecn is enabled at buffer pool/port level, drop or marking will happen only on WRED/ECN enabled queues referring to port/buffer pool.


kcudnik:
missing declaration SAI_OBJECT_TYPE_PORT_MEMBER as new object type

AshokDaparthi:
@itaibaz , @lguohan @atitjain - Please review and let me know first level of comments.

lguohan:
meta checking failure.

 ASSERT FAILED (on line 3809): SAI_OBJECT_TYPE_PORT_POOL not present on any object list (eg. VLAN_MEMBER is present on SAI_VLAN_ATTR_MEMBER_LIST)

@kcudnik , is this real or false alarm?

itaibaz:
I think we are missing an iterator for Port Pool objects
Should add a RO list of Port pool objects under Port

lguohan:
@itaibaz , I see, @AshokDaparthi , can you add?

AshokDaparthi:
is this really required? if you create these objects as part of init(like VLAN_MEMBER_LIST) then it is needed otherwise host adapter will have list.

AshokDaparthi:
@itaibaz @lguohan @kcudnik - Could you please reply. based on that i will update PR

lguohan:
@kcudnik, can you clarify this meta data test. is it neccessary?

kcudnik:
since port obejct id is mandatory on port_pool, please add new READ_ONLT attribute to SAI_PORT_ATTR_PORT_POOL_LIST as object list with SAI_OBJECT_TYPO_PORT_POOL, so user can obtain all pool list from given port, by default this list probably should be empty if there are no port_pools allocated at switch init

or add excetption in saisanitycheck.c:3766 (not reccomended)

lguohan:
@AshokDaparthi , please follow the 1st approach suggested by kamil.

aspirine92:
Hi @AshokDaparthi / @MELLANOX-zhichenr / @karthik-krishnamurthy ,

I need some clarifications according to these changes.

Q1: As per SAI_PORT_POOL_ATTR_QOS_WRED_PROFILE_ID and  SAI_BUFFER_POOL_ATTR_WRED_PROFILE_ID descriptions, we have some Port threshold and Buffer Pool threshold. Please point where these values could be took from.

Q2: What should happen if several PortPool entities with different WREDs are applied on the same Buffer Pool but different Ports? Which Buffer Pool threshold should be configured?

Q3: The similar to Q2 - one Port and several Buffer Pools (if it possible).

Regards,
Oleh

AshokDaparthi:
Hi,

We can apply buffer profiles on buffer pool and portpool. Buffering capacity is controlled by buffer profiles.  Next is WRED on pool, WRED thresholds will take from WRED profile. Behavior is if overall buffering on pool exceeds WRED min and MAX applied with SAI_BUFFER_POOL_ATTR_WRED_PROFILE_ID, WRED will kick in and  same with portpool.

Regards
Ashok D

PR #634, opened 2017-06-19
Sai mirror ptf

MELLANOX-zhichenr:
1. Update mirror ptf tests to match the new headers.
2. Modify data type on server side to receive the correct data.
    Please see switch.py line 55 for more details about datatype. (u16 on client is implemented using i16, which has the range of s16)


test/saithrift/tests/saimirror.py
```diff
@@ -273,20 +283,50 @@ def runTest(self):
                                     ip_ihl=5,
                                     ip_src='17.18.19.0',
                                     ip_dst='33.19.20.0',
+                                    version=2,
+                                    mirror_id=(erspanid & 0x3FFFFFFF),
                                     inner_frame=pkt3
                                     )
         m1=Mask(exp_pkt1)
-        m2=Mask(exp_pkt2)
         m1.set_do_not_care_scapy(ptf.packet.IP,'tos')
```
MELLANOX-zhichenr: What is the reason to ignore 'tos' here?

test/saithrift/src/switch_sai_rpc_server.cpp
```diff
@@ -2386,25 +2386,28 @@ class switch_sai_rpcHandler : virtual public switch_sai_rpcIf {
                   attr_list[i].value.u8 = attribute.value.u8;
                   break;
               case SAI_MIRROR_SESSION_ATTR_VLAN_TPID:
-                  attr_list[i].value.u16 = attribute.value.u16;
```
MELLANOX-zhichenr: u16 on client is implemented using i16, which is actually s16. Please see the comment in switch.py line 55 for reference.

test/saithrift/src/switch_sai_rpc_server.cpp
```diff
@@ -2386,25 +2386,28 @@ class switch_sai_rpcHandler : virtual public switch_sai_rpcIf {
...
+                  attr_list[i].value.u16 = attribute.value.u32;
```
lguohan: the u16 does not u32?

MELLANOX-zhichenr: On client, TPID has to be stored using u32 because:
1. u16 is implemented using i16 on client, which has the range of s16
2. One common value for TPID is 0x8100, which is out of the range of s16

So TPID is now stored using u32 on client instead.

PR #635, opened 2017-06-20
Add pipeline discard traps to host IF + Add mirror capability for traps

itaibaz:


inc/saihostif.h
```diff
@@ -378,6 +400,16 @@ typedef enum _sai_hostif_trap_attr_t
     SAI_HOSTIF_TRAP_ATTR_TRAP_GROUP,
 
     /**
+     * @brief Mirror session for the trap
+     *
+     * @type sai_object_list_t
```
kcudnik: do you need a list of mirror sessions here ?

itaibaz: Probably one mirror session is enough
However, I prefer to keep compatibility with ACL mirror action, and port ingress/egress mirror, where a list of sessions is defined, therefor i also allow here a list of mirror sessions

PR #637, opened 2017-06-22
SAI Proposal for Data Plane Telemetry

xiaozhou:
Data plane telemetry (DTEL) -- experimental track


doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,421 @@
...
+    SAI_TWL_ENTRY_ATTR_FIELD_INNER_L4_PORT_SRC,
+    /** L4 destination port UDP/TCP */
+    SAI_TWL_ENTRY_ATTR_FIELD_INNER_L4_PORT_DST,
+
+    /** Watch or not */
+    SAI_TWL_ENTRY_ATTR_ACTION_WATCH,
+    /** Report every matched packet (no triggering event detection) */
+    SAI_TWL_ENTRY_ATTR_ACTION_REPORT_ALL,
+    /** INT config session ID */
+    SAI_TWL_ENTRY_ATTR_ACTION_INT_CONFIG_SESSION_ID,
```
kcudnik: enum entries should be prefixed as enum name SI_TELEMETRY_WATCHLIST_ENTRY_ATTR_**

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,421 @@
...
+    sai_telemetry_report_trigger_attr_get_fn  telemetry_report_trigger_attr_get;
+    sai_telemetry_report_trigger_attr_set_fn  telemetry_report_trigger_attr_set;
+
+    sai_telemetry_int_session_create_fn     telemetry_int_session_create;
+    sai_telemetry_int_session_remove_fn     telemetry_int_session_remove;
+    sai_telemetry_int_session_attr_get_fn   telemetry_int_session_attr_get;
+    sai_telemetry_int_session_attr_set_fn   telemetry_int_session_attr_set;
+
+    sai_telemetry_int_sink_port_create_fn   telemetry_int_sink_port_create;
+    sai_telemetry_int_sink_port_remove_fn   telemetry_int_sink_port_remove;
```
kcudnik: missing set/get functions for telemetry_int_sink_port (we need quad api even if those functions will be dummy)

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,421 @@
...
+~~~cpp
+/** Queue alert thresholds */
+typedef struct _sai_telemetry_queue_alert_t {
+    /** egress port */
+    uint16_t port;
+    /** queue id */
+    uint16_t queue;
+    uint32_t depth_threshold;
+    uint32_t latency_threshold;
+} sai_telemetry_queue_alert_t;
```
kcudnik: sai_telemetry_queue_alert_t seem to extra entry in sai_attribute_value_t
and this contains port and queue, should this be object id's instead of uint's ? if yes please convert this sai_telemetry_queue_alert_t to as new SAI_OBJECT_TYPE_TELEMETRY_QUEUE_ALERT, and all those fileds mark as attirbutes of that object, and mark SAI_TELEMETRY_TRIGGER_ATTR_QUEUE_ALERT as sai_object_id_t
since we don't want to have object_id fields insice attr_value structure other than oid and objlist since it will be very hard for metadata to track

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,421 @@
...
+
+Figure 3 shows an example workflow of Packet Postcards, an alternative telemetry approach. Each switch makes its own decision and reports packets info (aka “postcards”) to the monitor. Unlike INT, a Postcard switch never modifies the original data packets. 
+
+## Mirror on Drop ##
+
+![MoD](figures/mod.png "Figure 4: Mirror on Drop")
+
+__Figure 4: Mirror on Drop__
+
+Figure 4 depicts the mirror on drop capability. Switches mirror packets dropped by the ingress pipe, egress pipe or queueing buffer to the monitor for network diagnosis. The report messages include packet header, switch metadata, and drop reason.
```
lguohan: mirror on drop is very similiar to this proposal here. https://github.com/opencomputeproject/SAI/commit/8b4e0ff32e017a30b8082cdd234ccd887ea6df07

this other proposal has another feature is to trap those dropped packet to CPU.

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
xiaozhou: I updated the proposal to use ACL for telemetry watchlist

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
xiaozhou: Added the missing functions.

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
xiaozhou: Updated to use sai_telemetry_queue_alert_attr_t

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
xiaozhou: Mirror on drop (MoD) is similar to hostif trap in terms of capturing dropped packets. The main difference is that MoD in this proposal is integrated with the telemetry framework. It unifies all drop conditions, sends dropped packets with drop reasons and associated switch metadata to the (**remote**) monitor through ERSPAN sessions. I am not sure how to nicely integrate it into the existing hostif framework. Please let me know if you have any suggestions. Thanks!

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,740 @@
...
+An __INT transit__ is a device along the path of a packet from source to sink. INT transit device adds its own INT information to the packet as requested by an INT source.
+
+
+## Packet Postcards
+
+![Postcard](figures/postcard.png "Figure 3: Postcard")
+
+__Figure 3: Packet Postcards__
+
+Figure 3 shows an example workflow of packet Postcards. Each switch makes its own decision and reports packets info (aka “postcards”) to the monitor individually. Unlike INT, a Postcard switch never modifies the original data packets. INT and Postcard achieve the same flow monitoring capabilities with different data plane approaches. They are __mutually exclusive__ in a network deployment.
```
lguohan: post card and int should not be mutually exclusive.

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,740 @@
...
+    SAI_ACL_TABLE_ATTR_FIELD_INNER_L4_SRC_PORT,
+
+    /**
+     * @brief Inner L4 Dst Port
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_INNER_L4_DST_PORT,
```
lguohan: do we also need INNER_SRC_IP and INNER_DST_IP as well?

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,740 @@
...
+
+
+## Telemetry data plane workflow
+
+![workflow](figures/workflow.png "Figure 1: Telemetry data plane workflow")
+
+__Figure 1: Telemetry data plane workflow__
+
+Figure 1 describes the telemetry data plane functionality at a high level. The telemetry module inspects every data packet without interfering with normal data forwarding pipeline. Different components may be located at different switches in the network. A telemetry watchlist table specifies the flows to monitor. It performs ternary match on the packet headers and switch ports, and provides telemetry action parameters. Packets belonging to the specified flow spaces will be processed by the event detection logic. If a triggering event is detected, the switch will generate a report message to the monitor. The report message includes packet header and switch metadata associated with the packet (e.g., timestamp, ingress/egress ports, queue depth/latency).
+
```
lguohan: can you define watchlist? it seems to be acl?

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,740 @@
...
+    SAI_SWITCH_ATTR_TELEMETRY_REPORT_UDP_DST_PORT,
+
+    /**
+     * @brief Telemetry flow state clear cycle
+     *
+     * @type sai_uint16_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_SWITCH_ATTR_TELEMETRY_FLOW_STATE_CLEAR_CYCLE,
```
lguohan: what is the time granularity here?

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,740 @@
...
+} sai_switch_attr_t;
+~~~
+
+## Changes to `saiacl.h` for INT, Postcard and MoD watchlist
+~~~cpp
+typedef enum _sai_acl_action_type_t
+{
+    ......
+    /** Enable mirror on drop */
+    SAI_ACL_ACTION_TYPE_TELEMETRY_MOD_ENABLE,
```
lguohan: do you have an example to enable MIRROR on drop? since it is an acl action, you also need a match entry, what is the match entry?

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,740 @@
...
+    SAI_SWITCH_ATTR_TELEMETRY_FLOW_STATE_CLEAR_CYCLE,
+
+    /**
+     * @brief Latency sensitivity for flow state change detection
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 15
+     */
+    SAI_SWITCH_ATTR_TELEMETRY_LATENCY_SENSITIVITY,
```
lguohan: what is latency sensitivity for flow state change? what does this 15 mean?

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
```diff
@@ -0,0 +1,740 @@
...
+ */
+typedef enum _sai_telemetry_int_session_attr_t 
+{
+    /**
+     * @brief INT max hop count
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     */
+    SAI_TELEMETRY_INT_SESSION_ATTR_MAX_HOP_COUNT,
```
lguohan: what is the max_hop_count meta data in the session attribute?

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
xiaozhou: Removed the sentence

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
xiaozhou: These are already in the saiacl.h

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
xiaozhou: Added an example

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
xiaozhou: Currently second in the example. May need to change to millisecond to increase flexibility. Should we specify the granularity in SAI?

doc/inband-telemetry/SAI-Proposal-Advanced-Telemetry-draft.md
xiaozhou: The maximum number of hops that are allowed to add their metadata to the packet. Added comments.

doc/inband-telemetry/SAI-Proposal-Data-Plane-Telemetry.md
```diff
@@ -0,0 +1,879 @@
...
+     * @flags CREATE_AND_SET
+     * @default SAI_ACL_DTEL_FLOW_OP_NOP
+     */
+    SAI_ACL_ENTRY_ATTR_ACTION_DTEL_FLOW_OP,
+
+    /**
+     * @brief INT session ID
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
```
kcudnik: missing object type here

doc/inband-telemetry/SAI-Proposal-Data-Plane-Telemetry.md
```diff
@@ -0,0 +1,879 @@
...
+/**
+ * @brief Queue report trigger attributes
+ */
+typedef enum _sai_dtel_queue_report_attr_t
+{
+    /**
+     * @brief Queue object ID
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
```
kcudnik: missing object type

doc/inband-telemetry/SAI-Proposal-Data-Plane-Telemetry.md
```diff
@@ -0,0 +1,879 @@
...
+     * @type sai_ip4_list_t
+     * @flags CREATE_AND_SET
+     */
+    SAI_DTEL_REPORT_SESSION_ATTR_DST_IP_LIST,
+
+    /**
+     * @brief Telemetry report virtual router ID
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
```
kcudnik: missing object type here @objects

doc/inband-telemetry/SAI-Proposal-Data-Plane-Telemetry.md
```diff
@@ -0,0 +1,879 @@
...
+     * @type sai_dtel_event_type_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_DTEL_EVENT_ATTR_TYPE,
+
+    /**
+     * @brief DTel report session
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE
```
kcudnik: missing object type here @objects

doc/inband-telemetry/SAI-Proposal-Data-Plane-Telemetry.md
```diff
@@ -476,15 +478,15 @@ typedef enum _sai_dtel_report_session_attr_t
...
+     * @type sai_ip_address_t
      * @flags CREATE_AND_SET
      */
     SAI_DTEL_REPORT_SESSION_ATTR_SRC_IP,
 
     /**
      * @brief Telemetry report destination IP addresses
      *
-     * @type sai_ip4_list_t
+     * @type sai_ip_address_list_t
```
kcudnik: there is no sai_ip_address_list_t type in saitypes.h

inc/sai.h
```diff
@@ -120,6 +120,9 @@ typedef enum _sai_api_t
     SAI_API_MPLS             = 36, /**< sai_mpls_api_t */
     SAI_API_UBURST           = 37, /**< sai_uburst_api_t */
     SAI_API_MAX              = 38, /**< total number of APIs */
+    /* Start of experimental APIs */
+    SAI_API_DTEL             = SAI_API_MAX, /**< sai_dtel_api_t */
+    SAI_API_EXPERIMENTAL_MAX = SAI_API_MAX + 1,
```
kcudnik: Such construct after api max is not supported please add this as normal api

inc/saitypes.h
```diff
@@ -239,6 +239,12 @@ typedef enum _sai_object_type_t
     SAI_OBJECT_TYPE_TAM_HISTOGRAM            = 68,
     SAI_OBJECT_TYPE_TAM_MICROBURST           = 69,
     SAI_OBJECT_TYPE_MAX                      = 70,
+    /* Start of experimental object types */
+    SAI_OBJECT_TYPE_DTEL_QUEUE_REPORT        = SAI_OBJECT_TYPE_MAX,
+    SAI_OBJECT_TYPE_DTEL_INT_SESSION         = SAI_OBJECT_TYPE_MAX + 1,
+    SAI_OBJECT_TYPE_DTEL_REPORT_SESSION      = SAI_OBJECT_TYPE_MAX + 2,
+    SAI_OBJECT_TYPE_DTEL_EVENT               = SAI_OBJECT_TYPE_MAX + 3,
+    SAI_OBJECT_TYPE_EXPERIMENTAL_MAX         = SAI_OBJECT_TYPE_MAX + 4,
```
kcudnik: Such construct are not supported also you need to have quad API for each object type to create remove set And get

inc/saitypes.h
```diff
@@ -677,6 +688,19 @@ typedef struct _sai_segment_list_t
...
+        sai_uint16_t u16;
+        sai_uint32_t u32;
+    } value;
+    union {
+        sai_uint8_t u8;
+        sai_uint16_t u16;
+        sai_uint32_t u32;
+    } mask;
+} sai_ternary_field_t;
+
```
kcudnik: There is no type defined here so serialize method will not know which value should be serialized. why not reuse acl_field struct ?

kcudnik: Also this could be just 2 uint32 for simplification

JipanYanga:
While reading through this PR,  one thing I'm not able to get a clear idea, what are the exact packet format for flow report, drop report and queue report?    I could see they will be udp packet,  are the format of switch info/drop reason/que info fields supposed to vary with different vendor?

Probably I missed something, could some example report packets be shared?

xiaozhou:
@JipanYanga Thanks for the question. My understanding is that reporting format is out of the scope of SAI definition. SAI only defines the control API for network devices, not the data plane packet format. 

The actual reporting format could be different for different vendors and can evolve while SAI remain the same. Here is an example [draft proposal](https://github.com/p4lang/p4-spec/blob/master/applications/telemetry/DRAFT_Telemetry_Report_Format.pdf) of data plane telemetry report format.

JipanYanga:
@xiaozhou Thanks for the response.  For drop report and queue report,  external server might be able to handle the different report formats from different vendors. But for flow report, if the packet format is not unified, or at least compatible among vendors,  the usage of data path telemetry would be limited.

xiaozhou:
@JipanYanga Same with drop report and queue report, a flow report is sent from a network device and then consumed by the monitor server. Switches along the routing path of the report packet do not need to parse the telemetry report format, so there is no vendor compatibility issue.

I think you are actually talking about the packet format for in-band telemetry (e.g., INT, iOAM) rather than the report packet format. In-band telemetry does require network devices agree on the same packet format because telemetry information is inserted into the data packet and need to be parsed and modified by all participating network devices. Here are the specification of [INT](https://p4.org/p4-spec/docs/In-band%20Network%20Telemetry%20(INT).pdf) and [iOAM](https://tools.ietf.org/html/draft-ietf-ippm-ioam-data-01). Again, they should not be specified by SAI.

inc/experimental/saidtel.h
```diff
@@ -0,0 +1,562 @@
...
+     */
+    SAI_DTEL_ATTR_SINK_PORT_LIST,
+
+    /**
+     * @brief Reserved DSCP value for INT over L4
+     *
+     * @type sai_acl_field_data_t
+     * @flags CREATE_AND_SET
+     */
+    SAI_DTEL_ATTR_INT_L4_DSCP,
```
JipanYanga: For INT over L4, is this attribute mandatory?   
There is scenario where DSCP field is used for choosing packet queue.  If the original DSCP got changed,  packet may go through different queue which causes the report having wrong egress timestamp and queue info.

xiaozhou: It is not mandatory. One current design choice is to use DSCP field to identify INT information after L4 header. Only one DSCP bit is needed to be reserved for INT, while other DSCP bits can still be used for choosing packet queue. People can choose to use other fields before L4 header to differentiate INT packets, which requires a new attribute in SAI_DTEL.

JipanYanga: Since DSCP is a 6-bit field, taking one DSCP bit means 32 dscp values are gone.  I think we'll need new attribute in SAI_DTEL to support other INT design choices like the probe marker solution mentioned in https://tools.ietf.org/html/draft-lapukhov-dataplane-probe-01

JipanYanga: So there is no standard on the INT packet transportation format yet, but sink and transit nodes need mechanism to identify them.  Could some attribute with opaque value be defined besides SAI_DTEL_ATTR_INT_L4_DSCP?   The opaque value is up to ASIC SDK to interpret then act accordingly?  SAI just provides the configuration channel.

xiaozhou: That's a nice idea. I suggest us to give more thoughts and discussions on whether it is better to add an opaque value now, or add another specific field later when someone plans to use INT with some other specific format. One great thing about SAI is that it is not RFC. It is more flexible for the users/community to add more fields as the network applications evolve.

meta/saisanitycheck.c
```diff
@@ -1661,6 +1666,11 @@ void check_attr_acl_fields(
                 break;
             }
 
+            if (md->objecttype == SAI_OBJECT_TYPE_DTEL)
+            {
+                break;
+            }
+
```
lguohan: why is this needed?

xiaozhou: Because I used sai_acl_field_data_t for [SAI_DTEL_ATTR_INT_L4_DSCP](https://github.com/xiaozhou/SAI/blob/master/inc/saidtel.h#L143-L151) as Kamil suggested. It is same as the special case UDF MATCH in lines above. Should I make it more detailed with specific fields allowed which is similar to the UFD MATCH?

kcudnik: please remove this check, there is something wrong here it captures too much data scope,
whats the error when you remove this ?

kcudnik: if you want to make it simlar to udf match, please use only explicit attributes and specific valu types like for udf match in lines 1637..1662

xiaozhou: The error messages would be: *ASSERT FAILED (on line 1669): SAI_DTEL_ATTR_INT_L4_DSCP: acl field may only be set on acl field and udf match*.

I have updated the check to only match on specified attr and value types

PR #639, opened 2017-06-22
Mark acl table range type as create only (#539)

yakivh:


kcudnik:
Ah it's already changed on the master

PR #641, opened 2017-06-22
Add FEC AN request support to saiport.h 0.9.6

shihjeff:
Signed-off-by: shihjeff <jechen@cavium.com>

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -349,6 +349,14 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_REMOTE_ADVERTISED_FEC_MODE,
 
     /**
+     * @brief Advertised remote port FEC AN request bit
+     *
+     * @type bool
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_REMOTE_ADVERTISED_FEC_MODE_REQUEST,
+
+    /**
      * @brief Query list of Remote Port's Advertised HALF-Duplex speed in Mbps
      *
      * @type sai_u32_list_t
@@ -469,13 +477,22 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_ADVERTISED_SPEED,
 
     /**
-     * @brief Query/Configure list of Advertised port FEC Mode
+     * @brief Query/Configure list of Advertised port FEC Mode. AN ability bit implicitly set to true.
      *
      * @type sai_s32_list_t sai_port_fec_mode_t
      * @flags CREATE_AND_SET
      * @default empty
      */
     SAI_PORT_ATTR_ADVERTISED_FEC_MODE,
+
+    /**
+     * @brief Advertised port FEC AN request bit.  True means FEC is requested for 10/40/100G
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_PORT_ATTR_ADVERTISED_FEC_MODE_REQUEST,
 
     /**
      * @brief Query/Configure list of Advertised HALF-Duplex speed in Mbps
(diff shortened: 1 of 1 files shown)


inc/saiport.h
```diff
@@ -478,6 +486,15 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_ADVERTISED_FEC_MODE,
 
     /**
+     * @brief Advertised port FEC AN request bit.  True means FEC is requested for 10/40/100G
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_PORT_ATTR_ADVERTISED_FEC_MODE_REQUEST,
```
lguohan: what is the relationship between this one and SAI_PORT_ATTR_ADVERTISED_FEC_MODE? Is it true and the SAI_PORT_ATTR_ADVERTISED_FEC_MODE is valid only when SAI_PORT_ATTR_ADVERTISED_FEC_MODE_REQUEST is true?

should we add @validonly SAI_PORT_ATTR_ADVERTISED_FEC_MODE_REQUEST=true for attribute SAI_PORT_ATTR_ADVERTISED_FEC_MODE?

shihjeff: So in IEEE 802.3by-2016 Clause 73.6.5, there are 4 bits (F0 - F3)

1. F0 = 10 Gb/s - FEC ability
2. F1 = 10 Gb/s - FEC requested
3. F2 = 25G RS-FEC requested
4. F3 = 25G FC-FEC requested

These are the bits SAI_PORT_ATTR_ADVERTISED_FEC_MODE_* modify.  The interpretation is:

1. SAI_PORT_ATTR_ADVERTISED_FEC_MODE modifies F0, F2, F3
2. SAI_PORT_ATTR_ADVERTISED_FEC_MODE_REQUEST modifies F1

Your question is why do we need REQUEST?  Why not just have F0 & F1 always both == 0 or both == 1 (thus @validonly would make sense)?

Answer is some customers want ability == 1 and request == 0 (saying I'm capable of FEC but I don't wan't to use it).  So there is no need to use @validonly.

25G consortium has multiple bits for requesting RS vs FC vs none.  But since I don't see SAI making a distinction between consortium vs IEEE, folks we work with default to IEEE standard and switch to consortium via sai.ini file

If we want to support 25G consortium request via this new attribute, we could change it to sai_port_fec_mode_t

shihjeff: @lguohan - what is SAI's position between 25G IEEE & Consortium?  I noticed that there is a port attribute that mentions consortium but I don't see an explicit definition.

CC @mikelazar

lguohan: correct if me if i am wrong

SAI_PORT_FEC_MODE_RS -> F2 = 1
SAI_PORT_FEC_MODE_FC -> F3 = 1 and F0 = 1

Can we just add a new FEC_MODE instead of adding a NEW attribute? 
SAI_PORT_FED_MODE_10G_FEC_REQUEST -> F1 = 1

I think F1 is only for 10G specifically, so we probably want to make it to 10G specficially.

shihjeff: In the current IEEE standard, that is true.  But 25G Consortium is different and more flexible.  Wanted to make attribute generic, not specific to the IEEE standard.  So if above is set for 25G for IEEE standard, it does nothing in the adapter.

shihjeff:
Not needed for customers... closing PR

shihjeff:
Not needed by customers.. Closing for now

amranjan44:
@mikelazar @AshokDaparthi , Can you please help get this pull request merged to support both Ethernet 25G Consortium and IEEE FEC modes.

shihjeff:
Do not merge.  Changes in progress

PR #650, opened 2017-07-03
Port autoneg meta check

AshokDaparthi:
Auto neg and Media type can be set even after port create.


shihjeff:
Please make another PR for 0.9.6

inc/saiport.h
```diff
@@ -438,7 +438,7 @@ typedef enum _sai_port_attr_t
      * @brief Media Type
      *
      * @type sai_port_media_type_t
-     * @flags CREATE_ONLY
+     * @flags CREATE_AND_SET
```
kcudnik: i think this attribute should be create_only and reflects physical hardware of port which cant be change ad hoc (probably?)

shihjeff: I think the argument is that SFPs are modular.  You could swap out fiber & copper based SFPs dynamically.

lguohan: yes, in case you change the transceiver, you will need to change this attribute. @kcudnik

kcudnik: Should that operation require to recreate port again ? since if they are hot swap then other (not all?) 
 attributes on port should be create_and_set also

lguohan: no it does not require to recreate the port again. what other attributes? HW_LANE_LIST is related to break out so it is create only. even you replace the port module, it port will not automatically break out into 4, you still need to call remove_port() and create_port()

PR #651, opened 2017-07-03
Control flooding on .1D bridges

karthik-krishnamurthy:
As discussed in the OCP weekly call, adding support for controlling the unknown unicast, unknown multicast and broadcast traffic on .1D bridges.


karthik-krishnamurthy:
@lguohan  @mattyk @itaibaz @atitjain  Please review the changes.

inc/sail2mcgroup.h
```diff
@@ -99,6 +99,16 @@ typedef enum _sai_l2mc_group_member_attr_t
     SAI_L2MC_GROUP_MEMBER_ATTR_L2MC_OUTPUT_ID,
 
     /**
+     * @brief L2MC Tunnel Endpoint IP.Valid and mandatory if bridge port
+     * is of type SAI_BRIDGE_PORT_TYPE_TUNNEL.
+     *
+     * @type sai_ip_address_t
+     * @flags CREATE_ONLY
+     * @default 0.0.0.0
```
lguohan: here in the comment the bridge port refers to the SAI_L2MC_GROUP_MEMBER_ATTR_L2MC_OUTPUT_ID attribute. is it possible to use meta data to specify the condition? @kcudnik

inc/saibridge.h
```diff
@@ -335,14 +335,52 @@ typedef enum _sai_bridge_attr_t
...
     /**
-     * @brief To disable flooding traffic (Broadcast, unknown unicast,
-     * unknown multicast) on a bridge
+     * @brief Unknown unicast flood group.
      *
-     * @type bool
+     * Provides control on the set of bridge ports on which unknown
+     * unicast packets need to be flooded.If null object id is passed,
+     * then flooding will be disabled. Default is to disable flooding.
+     * Valid for SAI_BRIDGE_TYPE_1D.
```
lguohan: validonly SAI_BRIDGE_ATTR_TYPE  == SAI_BRIDGE_TYPE_1D

inc/saibridge.h
```diff
@@ -335,14 +335,55 @@ typedef enum _sai_bridge_attr_t
...
+     * @validonly SAI_BRIDGE_ATTR_TYPE == SAI_BRIDGE_TYPE_1D
+     */
+    SAI_BRIDGE_ATTR_UNKNOWN_MULTICAST_FLOOD_GROUP,
+
+    /**
+     * @brief Broadcast flood group.
+     *
+     * Provides control on the set of bridge ports on which broadcast
+     * packets need to be flooded.If null object id is passed,
+     * then flooding will be disabled. Default is to disable flooding.
```
lguohan: default should be the flooding to all local ports (matty)

inc/sail2mcgroup.h
kcudnik: What would condition look like ?

karthik-krishnamurthy: The condition is that the new attribute is mandatory and valid if bridge port object id (given in SAI_L2MC_GROUP_MEMBER_ATTR_L2MC_OUTPUT_ID) was created with SAI_BRIDGE_PORT_ATTR_TYPE == SAI_BRIDGE_PORT_TYPE_TUNNEL.
This seems like an indirect condition so not sure on how this can be conveyed using meta data

kcudnik: It's not possible to make that condition, since this object depends on other object attribute. In general we should not have such dependencies, we will need special logic to handle such cases.

inc/saibridge.h
karthik-krishnamurthy: Edited the comments to make the default behavior as flood to sub-ports.

karthik-krishnamurthy: @lguohan @mattyk  please review the updated comments

inc/saibridge.h
```diff
@@ -428,9 +428,9 @@ typedef enum _sai_bridge_attr_t
      *
      * Provides control on the set of bridge ports on which broadcast
      * packets need to be flooded.If null object id is passed,
-     * then flooding will be disabled. Default is to flood to subports
+     * then flooding will be disabled. Default is to flood to sub-ports
      * of the bridge.During bridge create, SAI would internally create
-     * an L2MC group and whenever subports are added to the bridge, SAI
+     * an L2MC group and whenever sub-ports are added to the bridge, SAI
      * would automatically add them to that L2MC group.
```
karthik-krishnamurthy: @lguohan @mattyk  Please review the updated comments

PR #652, opened 2017-07-03
Rename buffer pool size -> reserved size, in backward compatible way

itaibaz:


inc/saibuffer.h
```diff
@@ -266,7 +266,10 @@ typedef enum _sai_buffer_pool_attr_t
      * @type sai_uint32_t
      * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
      */
-    SAI_BUFFER_POOL_ATTR_SIZE,
+    SAI_BUFFER_POOL_ATTR_RESERVED_SIZE,
```
AshokDaparthi: Buffer pool size is not reserved size as per proposal, this is total pool size.  SAI_BUFFER_POOL_ATTR_SHARED_SIZE = SAI_BUFFER_POOL_ATTR_SIZE - Reserved_size(from buffer profiles )+ SAI_BUFFER_POOL_XOFF_SIZE.

@lguohan - Let me know if i am missing anything?

itaibaz: Thanks for noticing
I actually planned to change buffer profile and not buffer pool, but made a merge error
closing this and opening a new one, thanks

PR #658, opened 2017-07-06
Changes for flush based on bridge or vlan object

dgsudharsan:
Replaced VLAN ID with BV_ID which is vlan object or bridge object id


lguohan:
can fix the meta data checking?

dgsudharsan:
Fixed meta condition

PR #662, opened 2017-07-11
Fix metadata for buffer profile dynamic/static thresholds

itaibaz:
They are conditionally mandaotry based on profile threshold mode or
inheritance from pool
Note inheritance from pool can't be properly expressed with current meta
data infrastructure

As conditionally mandatory can be create only, changed threshold mode
and threshold valus to be create only

### DIFF
--- a/inc/saibuffer.h
+++ b/inc/saibuffer.h
@@ -475,7 +475,7 @@ typedef enum _sai_buffer_profile_attr_t
      * If set, this overrides #SAI_BUFFER_POOL_ATTR_THRESHOLD_MODE.
      *
      * @type sai_buffer_profile_threshold_mode_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default SAI_BUFFER_PROFILE_THRESHOLD_MODE_INHERIT_BUFFER_POOL_MODE
      */
     SAI_BUFFER_PROFILE_ATTR_THRESHOLD_MODE,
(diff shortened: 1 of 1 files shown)


inc/saibuffer.h
```diff
@@ -484,21 +484,29 @@ typedef enum _sai_buffer_profile_attr_t
      * @brief Dynamic threshold for the shared usage
      *
      * The threshold is set to the 2^n of available buffer of the pool.
-     * Mandatory when SAI_BUFFER_POOL_THRESHOLD_MODE = SAI_BUFFER_THRESHOLD_MODE_DYNAMIC
```
kcudnik: previously this was conditional depending on pool id passed SAI_BUFFER_PROFILE_ATTR_POOL_ID, so it was indirect dependency, did that changed ?

itaibaz: Buffer profile threshold mode has 3 values
Inherit buffer pool mode, which is the default value, means look at buffer pool which the profile is attached to (through SAI_BUFFER_PROFILE_ATTR_POOL_ID), and inherit from that pool threshold mode (SAI_BUFFER_POOL_ATTR_THRESHOLD_MODE)
There are 2 explicit modes - SAI_BUFFER_PROFILE_THRESHOLD_MODE_STATIC, SAI_BUFFER_PROFILE_THRESHOLD_MODE_DYNAMIC, that basically set explicit buffer profile mode

Once the mode is decided according to this logic, if the mode is static only static attributes are relevant (SAI_BUFFER_PROFILE_ATTR_SHARED_STATIC_TH). If it's dynamic, only dynamic attributes are relevant (SAI_BUFFER_PROFILE_ATTR_SHARED_DYNAMIC_TH)

This logic hasn't changed

In the original buffer proposal we had just the 2 explicit buffer profile threshold modes
Several months ago Dell added the concept of inherit from pool
This pull request doesn't change anything in the logic IMO, just fixes the metadata

kcudnik: @lguohan can you comment on that ?

lguohan: ideally, i am looking for 

(SAI_BUFFER_PROFILE_ATTR_THRESHOLD_MODE == SAI_BUFFER_PROFILE_THRESHOLD_MODE_INHERIT_BUFFER_POOL_MODE &&
SAI_BUFFER_PROFILE_ATTR_POOL_ID.SAI_BUFFER_POOL_ATTR_THRESHOLD_MODE == SAI_BUFFER_POOL_THRESHOLD_MODE_STATIC) ||
(SAI_BUFFER_PROFILE_ATTR_THRESHOLD_MODE  == SAI_BUFFER_PROFILE_THRESHOLD_MODE_STATIC)

It my understanding that the first part is not supported, so itai make them in the comment and only added the second part in the metadata.

itaibaz: Kamil,
As I think we don't have the ability to check another object
SAI_BUFFER_PROFILE_ATTR_POOL_ID.SAI_BUFFER_POOL_ATTR_THRESHOLD_MODE
I think the current pull request is the best that can be done now
Can you approve ?

inc/saibuffer.h
```diff
@@ -475,7 +475,7 @@ typedef enum _sai_buffer_profile_attr_t
      * If set, this overrides #SAI_BUFFER_POOL_ATTR_THRESHOLD_MODE.
      *
      * @type sai_buffer_profile_threshold_mode_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
```
kcudnik: just not that you are limiting change capability changing this to create only

PR #666, opened 2017-07-12
Fix hash objects meta data - allow setting hash objects

itaibaz:
Switch has 2 default objects - SAI_SWITCH_ATTR_ECMP_HASH,
SAI_SWITCH_ATTR_LAG_HASH. They are read only objects created on system
initialization, with the attributres modifiable
The rest of the hash objects matching specific traffic types for
ECMP/LAG, should be editable and null by default, otherwise
create/remove hash API has no usage as the hash object can't be assigned

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -971,9 +971,10 @@ typedef enum _sai_switch_attr_t
      * @brief The hash object for IPv4 packets going through ECMP
      *
      * @type sai_object_id_t
-     * @flags READ_ONLY
+     * @flags CREATE_AND_SET
      * @objects SAI_OBJECT_TYPE_HASH
-     * @default internal
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
      */
     SAI_SWITCH_ATTR_ECMP_HASH_IPV4,
 
@@ -981,9 +982,10 @@ typedef enum _sai_switch_attr_t
      * @brief The hash object for IPv4 in IPv4 packets going through ECMP
      *
      * @type sai_object_id_t
-     * @flags READ_ONLY
+     * @flags CREATE_AND_SET
      * @objects SAI_OBJECT_TYPE_HASH
-     * @default internal
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
      */
     SAI_SWITCH_ATTR_ECMP_HASH_IPV4_IN_IPV4,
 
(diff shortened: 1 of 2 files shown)


inc/saiswitch.h
```diff
@@ -971,29 +971,32 @@ typedef enum _sai_switch_attr_t
      * @brief The hash object for IPv4 packets going through ECMP
      *
      * @type sai_object_id_t
-     * @flags READ_ONLY
+     * @flags CREATE_AND_SET
      * @objects SAI_OBJECT_TYPE_HASH
-     * @default internal
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
```
kcudnik: not sure if there will be null instead of some hash set after switch create, maybe we could relax this add support "internal" as default value on non read_only object types

itaibaz: It might make sense to keep the default as internal for these objects, if the platform has specific out of the box settings for these traffic (although it's not defined what they should be in the spec)

However, the sanity check fails, with :
ASSERT FAILED (on line 998): SAI_SWITCH_ATTR_ECMP_HASH_IPV4: default internal currently can be set only on read only objects

What do you suggest ?

kcudnik: yes, i as proposed to relax this condition to allow "internal" on non read only objects

inc/saiswitch.h
```diff
@@ -1052,8 +1057,9 @@ typedef enum _sai_switch_attr_t
      * @brief The hash object for IPv6 packets going through LAG
      *
      * @type sai_object_id_t
-     * @flags READ_ONLY
+     * @flags CREATE_AND_SET
      * @objects SAI_OBJECT_TYPE_HASH
+     * @allownull true
```
kcudnik: just a small comment here why we allow null? what will null mean when hash is not assigned ?

itaibaz: As I understand
When assigning null to ECMP/LAG specific traffic type (ipv4, ipv6, ipv4inipv4)
The policy applied for that traffic will be decided by the default ECMP/LAG hash - SAI_SWITCH_ATTR_ECMP_HASH, SAI_SWITCH_ATTR_LAG_HASH

kcudnik: Yea so that default hash is the internal prezent value do we should nie allow null but set that internal value to bring back default

kcudnik: Actually eather approach is fine but id we allow null then there should be null after switch create

itaibaz: In this case, I return to the original commit
Allow null and default null
I think it's more straight forwards to the user

PR #674, opened 2017-07-13
Fix wrong condition policer mode

itaibaz:


kcudnik:
I will try to add check for that

PR #677, opened 2017-07-14
MPLS proposal API

itaibaz:


inc/saitypes.h
```diff
@@ -54,6 +54,7 @@ typedef UINT8   sai_mac_t[6];
 typedef UINT32  sai_ip4_t;
 typedef UINT8   sai_ip6_t[16];
 typedef UINT32  sai_switch_hash_seed_t;
+typedef UINT32  sai_label_id_t;
```
kcudnik: Wht not use uint32 directly? Variable name will stilo be label

lguohan: i guess it is like the vlan_id

PR #678, opened 2017-07-17
Media and AutoNeg metadate change

skanagas:
Changes for pushing the pull request #650 to v0.9.6.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -436,7 +436,7 @@ typedef enum _sai_port_attr_t
      * @brief Auto Negotiation configuration
      *
      * @type bool
-     * @flags CREATE_ONLY
+     * @flags CREATE_AND_SET
      * @default false
      */
     SAI_PORT_ATTR_AUTO_NEG_MODE,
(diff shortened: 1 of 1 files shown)


inc/saiport.h
```diff
@@ -454,7 +454,7 @@ typedef enum _sai_port_attr_t
      * @brief Media Type
      *
      * @type sai_port_media_type_t
-     * @flags CREATE_ONLY
+     * @flags CREATE_AND_SET
```
kcudnik: so it can be copper and suddenly optic ?

pkarashchenko: I see it can be possible. The copper cable can be replaced by fibre and vice versa.
I see two strategies of upper layer application based on autoneg and media type attributes:
1. Port is created only when media type is detected. This mean that if switch start-up and no cables are plugged we will see 0 interfaces or ports with default properties for autoneg and media type are created and port is re-created (remove/add) when media type is detected. In this case I see that many related properties will be reset to default values (like CoS configuration, PVID, L2/L3 configuration, etc.)
2. Ports are created with default properties for autoneg and media type and autoneg and media type attributes are changed when media type is detected. With this approach all other port related configuration is kept untouched.

In current interface we have "CREATE_ONLY", this means that strategy 1) is used.
Proposed changes will comply with both 1) and 2)

shihjeff: This change was already pushed into the master branch in PR#650.  There is need to make same change in 0.9.6 for integration purposes.

CC @kcudnik / @lguohan

PR #679, opened 2017-07-17
IGMP snooping enable/disable control for VLAN as custom attribute

ragu-raman:


inc/saivlan.h
```diff
@@ -281,6 +281,18 @@ typedef enum _sai_vlan_attr_t
...
+     * @brief IGMP Snooping enable or disable control for VLAN
+     *
+     * IGMP Snooping enable control for VLAN. Default is
+     * disabled
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE = SAI_VLAN_ATTR_CUSTOM_RANGE_START,
```
kcudnik: custom range attributes are not supported here, just add this as regular attribute

ragu-raman: @kcudnik, in last week sai community meeting @lguohan suggested to add it as custom attribute as the discussion was not complete and dell need this attribute in 1.2. My inital pull request for the same which was merged by you was revereted by @lguohan (#671). Please let  me know how to proceed

kcudnik: Add this as regural attribute, some vendors will just. Not implement it

inc/saivlan.h
```diff
@@ -281,6 +281,18 @@ typedef enum _sai_vlan_attr_t
...
+     * @brief IGMP Snooping enable or disable control for VLAN
+     *
+     * IGMP Snooping enable control for VLAN. Default is
+     * disabled
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE,
```
kcudnik: move this before SAI_VLAN_ATTR_CUSTOM_RANGE_START attribute

lguohan: since this is a temp solution, I think it is then better to move this attribute into CUSTOM range.  Once we work out a final solution, we'll then remove this attribute. what is your concern?

ragu-raman: @kcudnik Build is failing when I add it in the custom range with "sai_vlan_attr_t: values are not increasing by 1: last: 14 current: 268435457". Can you please help to fix this?

kcudnik: yes: move this attribute before SAI_VLAN_ATTR_CUSTOM_RANGE_START attribute

lguohan: if you add @flags like this one, it should pass the metadata check.

https://github.com/opencomputeproject/SAI/blob/master/inc/saihostif.h#L155

ragu-raman: Hi @lguohan , I tried the changes you suggested. Now I am getting the following error
 ASSERT FAILED (on line 178): expected true 'value < 0x10000': enum value is too big, range?
Makefile:52: recipe for target 'all' failed
Should the SAI_VLAN_ATTR_CUSTOM_RANGE_START value be changed < 0x10000 ?

kcudnik: ahh, there are too many checks, let me fix that quickly

kcudnik:
retest this please

inc/saivlan.h
kcudnik: fixed

ragu-raman:
@kcudnik @lguohan , can you please merge this pull request?

PR #697, opened 2017-08-01
Acl cpu action

AshokDaparthi:


kcudnik:
Ahh, one more, since you added SAI_ACL_ACTION_TYPE_SET_CPU_QUEUE t hey are matched with acl_entry to be consistent, so acl_entry can't be "ignored" must exist, or mark with "@ignore" SAI_ACL_ACTION_TYPE_SET_CPU_QUEUE  but this will also not work, since id's must be continuous of all enums, so remove "ignore" flag?

vemulabalaji:
Seems these changes are approved in Aug 2017, still these changes are not yet available in latest Code base. 
Am i correct ?

PR #698, opened 2017-08-01
Fix buffer profile attribute to allow inherit mode

stcheng:


kcudnik:
Probably not needed since SAI_BUFFER_PROFILE_ATTR_THRESHOLD_MODE is "create_only" and can't change during object life time, and when buffers are inherited then you should not be able to set them anyway since change will not have effect

itaibaz:
Now that I look at it, I agree probably this change is not needed
Static threshold value should be mandatory if (mode is static) OR (mode is inherit form pool and pool is static)
The previous version was more relaxed, saying it is mandatory only if mode is static
if we change to mandatory also when mode is inherit, this is actually wrong. because when mode will be inherit, you will need to provide both static and dynamic values

So I think this change shouldn't go in
And with the existing code, you could still pass either static or dynamic value when using mod inherit

kcudnik:
please refer to recent change: https://github.com/opencomputeproject/SAI/pull/672/files

stcheng:
if the value  is also inherited from the pool, then which attribute shall I use to set the pool's threshold value?

itaibaz:
You can still use SAI_BUFFER_PROFILE_ATTR_SHARED_DYNAMIC_TH and
SAI_BUFFER_PROFILE_ATTR_SHARED_STATIC_TH

So sample Sonic flow could be 
             else if (fvField(*i) == buffer_dynamic_th_field_name)		              else if (fvField(*i) == buffer_dynamic_th_field_name)
              {		              {
                  attr.id = SAI_BUFFER_PROFILE_ATTR_SHARED_DYNAMIC_TH;		                  attr.id = SAI_BUFFER_PROFILE_ATTR_SHARED_DYNAMIC_TH;
                  attr.value.u32 = (uint32_t)stoul(fvValue(*i));		                  attr.value.u32 = (uint32_t)stoul(fvValue(*i));
                  attribs.push_back(attr);		                  attribs.push_back(attr);
              }
That flow doesn't set the buffer profile mode (since by default it is inherit)

Or you could write this explicitley
             else if (fvField(*i) == buffer_dynamic_th_field_name)		              else if (fvField(*i) == buffer_dynamic_th_field_name)
              {		              {
                 attr.id = SAI_BUFFER_PROFILE_ATTR_THRESHOLD_MODE;		
                 attr.value.s32 = SAI_BUFFER_PROFILE_THRESHOLD_MODE_INHERIT_BUFFER_POOL_MODE;		
                 attribs.push_back(attr);		
 		
                  attr.id = SAI_BUFFER_PROFILE_ATTR_SHARED_DYNAMIC_TH;		                  attr.id = SAI_BUFFER_PROFILE_ATTR_SHARED_DYNAMIC_TH;
                  attr.value.u32 = (uint32_t)stoul(fvValue(*i));		                  attr.value.u32 = (uint32_t)stoul(fvValue(*i));
                  attribs.push_back(attr);		                  attribs.push_back(attr);
              }

stcheng:
Hi Itai, could you format your response a little bit? the code format is hard to read.

lguohan:
I revert this commit in v1.0 branch. I think itai is probably right.

lguohan:
this commit is not merged into v1.0 branch.

PR #706, opened 2017-08-08
Removing condition in bridge port bridge id attribute

karthik-krishnamurthy:
Removing condition in bridge port bridge id attribute as it is required for all types of bridge ports


stcheng:
I think the purpose of this condition is that by default, there is only one default 1Q bridge. If this attribute is not provided, the bridge port is going to be part of the default 1Q bridge.

lguohan:
this is to create a bridge port that bridges a 1D bridge and router, so bridge ID is needed. not sure why we need to remove the condition.

dgsudharsan:
I think the bridge id should be a mandatory attribute for all bridge port types. That's the reason why the condition is removed.

karthik-krishnamurthy:
@stcheng @lguohan 
The understanding of bridge port is that it associates/connects either
a)  Port to a .1Q bridge               (SAI_BRIDGE_PORT_TYPE_PORT)
b) (Port, VLAN) to a .1D bridge   (SAI_BRIDGE_PORT_TYPE_SUB_PORT) 
c)  RIF to a .1Q bridge                 (SAI_BRIDGE_PORT_TYPE_1Q_ROUTER)
d) RIF to a .1D bridge                  (SAI_BRIDGE_PORT_TYPE_1D_ROUTER)
e) Tunnel to a .1D/.1Q bridge     (SAI_BRIDGE_PORT_TYPE_TUNNEL)

Since in all the types of bridge ports we need bridge id, we thought of removing the condition.

itaibaz:
IMO, in case it's a .1Q bridge, we don't need the bridge ID
The reason is, there can only be up to a single instance of .1Q bridge at any given time, so giving the bridge ID is redundant
For .1D bridge, we do need the bridge ID

Indeed looks like there is a problem with the condition, so perhaps it should be fixed to 
SAI_BRIDGE_PORT_ATTR_TYPE == SAI_BRIDGE_PORT_TYPE_SUB_PORT OR
SAI_BRIDGE_PORT_ATTR_TYPE == SAI_BRIDGE_PORT_TYPE_1D_ROUTER OR
SAI_BRIDGE_PORT_ATTR_TYPE == SAI_BRIDGE_PORT_TYPE_TUNNEL

karthik-krishnamurthy:
@itaibaz That seems fine as well

lguohan:
@mattyk

karthik-krishnamurthy:
@stcheng @lguohan @itaibaz @mattyk  Updated the pull request based on the OCP discussion

inc/saibridge.h
```diff
@@ -178,7 +178,9 @@ typedef enum _sai_bridge_port_attr_t
      * @type sai_object_id_t
      * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
      * @objects SAI_OBJECT_TYPE_BRIDGE
-     * @condition SAI_BRIDGE_PORT_ATTR_TYPE == SAI_BRIDGE_PORT_TYPE_1D_ROUTER
+     * @condition SAI_BRIDGE_PORT_ATTR_TYPE == SAI_BRIDGE_PORT_TYPE_SUB_PORT
+     * or SAI_BRIDGE_PORT_ATTR_TYPE == SAI_BRIDGE_PORT_TYPE_1D_ROUTER or
+     * SAI_BRIDGE_PORT_ATTR_TYPE == SAI_BRIDGE_PORT_TYPE_TUNNEL
```
kcudnik: can you move "or" to the same line as above so they are eather both on left or right side, or inline entire expression

karthik-krishnamurthy: @kcudnik Sure. Done

PR #721, opened 2017-09-27
add SAI_SWITCH_ATTR_AVAILABLE_ACL_{TABLE,TABLE_GROUP}

lguohan:
SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE counts the available entires
for ACL TABLE objects

SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE_GROUP counts the available
entries for ACL TABLE GROUP objects

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -726,6 +726,22 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_AVAILABLE_IPMC_ENTRY,
 
     /**
+     * @brief Available ACL Tables
+     *
+     * @type sai_uint32_t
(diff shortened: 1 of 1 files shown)


zubinshah:
@lguohan  -- sorry if i missed any of the discussions that happened around this attribute. 

What is the expectation here : SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE -- will count the available new entries that can be added for a specific ACL table. 

How do we compute SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE_GROUP -- it can be tricky. 
Can we clarify that this will only compute the aggregate ACL Table Entries available that are bound to a group?

PR #731, opened 2017-10-30
ACL CRM changes

prsunny:
The following are the changes as part of this commit for ACL Critical Resource Monitoring:

1. Available ACL_ENTRY and ACL_COUNTER are moved from Table Group Member attribute to be a Table attribute.
2. For the Switch attributes, AVAILABLE_ACL_TABLE and AVAILABLE_ACL_TABLE_GROUP, modified the API to return a _list_ of resources available instead of a single _uint32_ value. ACL_TABLE and ACL_GROUP can be associated to different _stage_ and _bind_points_. Hence a single global _uint32_ value could be ambiguous in this scenario. The callee shall return a list of entries with the 'available_numbers' at each 'stage' and 'bind_point'.
3. ACL _enums_ for 'stage' and 'bind_point' are moved to `saitypes.h`.

Notes: 

1. If the acl table is in different acl table group members, then the available acl entry should be the minimum free entries of all acl table group members.
2. For AVAILABLE_ACL_TABLE and AVAILABLE_ACL_TABLE_GROUP, if certain 'stage' and 'bind_point' are _not supported_, then the implementation is not required to include the combinations in the _list_. If the implementation include the combinations in the list, then the avail_num must be 0.

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -34,44 +34,6 @@
  */
 
 /**
- * @brief Attribute data for SAI_ACL_TABLE_ATTR_STAGE
- */
-typedef enum _sai_acl_stage_t
-{
-    /** Ingress Stage */
-    SAI_ACL_STAGE_INGRESS,
-
-    /** Egress Stage */
-    SAI_ACL_STAGE_EGRESS,
-
-} sai_acl_stage_t;
-
-/**
- * @brief Attribute data for SAI_ACL_TABLE_ATTR_BIND_POINT
- */
-typedef enum _sai_acl_bind_point_type_t
-{
-    /** Bind Point Type Port */
-    SAI_ACL_BIND_POINT_TYPE_PORT,
-
-    /** Bind Point Type LAG */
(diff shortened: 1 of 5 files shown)


inc/saiacl.h
```diff
@@ -34,44 +34,6 @@
...
-    SAI_ACL_BIND_POINT_TYPE_LAG,
-
-    /** Bind Point Type VLAN */
-    SAI_ACL_BIND_POINT_TYPE_VLAN,
-
-    /** Bind Point Type RIF */
-    SAI_ACL_BIND_POINT_TYPE_ROUTER_INTFERFACE,
-
-    /** @ignore - for backward compatibility */
-    SAI_ACL_BIND_POINT_TYPE_ROUTER_INTF = SAI_ACL_BIND_POINT_TYPE_ROUTER_INTFERFACE,
```
kcudnik: missing in moved item

prsunny: Yes, fixing it

PR #736, opened 2017-11-16
PTF: alignment with 1.2 headers & fixes

yakivh:
* Update for FDB entry
* Bunch of renames
* SAI_LAG_ATTR_PORT_VLAN_ID usage
* Fixes for PVID reset


IGordynskyi:
@lguohan / @yakivh 
I guess if PTF aligned with 1.2 headers it should be pushed to [v1.2 branch](https://github.com/opencomputeproject/SAI/tree/v1.2) as well.
How do you think?

PR #741, opened 2017-11-30
Bfd proposal V0.1

ragu-raman:
Added changes for hw based lookup.
Added switch level bfd capability


kcudnik:
Instead of validonly use condition on attributes where they are mandatory on create

kcudnik:
instead sai_bfd_capability_t use 3 attributes readonly on switch, then you dont need to introduce new type on serialize in metadata

PR #742, opened 2017-12-04
Fixed SAI tunnel attributes definitions

IGordynskyi:
Fixed definitions for SAI tunnel attributes:
SAI_TUNNEL_ATTR_UNDERLAY_INTERFACE,
SAI_TUNNEL_ATTR_OVERLAY_INTERFACE,
SAI_TUNNEL_ATTR_DECAP_TTL_MODE,
SAI_TUNNEL_ATTR_DECAP_DSCP_MODE.

### DIFF
--- a/inc/saitunnel.h
+++ b/inc/saitunnel.h
@@ -451,7 +451,6 @@ typedef enum _sai_tunnel_attr_t
      * @type sai_object_id_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      * @objects SAI_OBJECT_TYPE_ROUTER_INTERFACE
-     * @condition SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_GRE
      */
     SAI_TUNNEL_ATTR_UNDERLAY_INTERFACE,
 
@@ -463,7 +462,6 @@ typedef enum _sai_tunnel_attr_t
      * @type sai_object_id_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      * @objects SAI_OBJECT_TYPE_ROUTER_INTERFACE
-     * @condition SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_GRE
      */
     SAI_TUNNEL_ATTR_OVERLAY_INTERFACE,
 
@@ -480,8 +478,6 @@ typedef enum _sai_tunnel_attr_t
 
     /**
      * @brief Tunnel TTL mode (pipe or uniform model)
-     *
-     * Default would be
      *
      * @type sai_tunnel_ttl_mode_t
      * @flags CREATE_ONLY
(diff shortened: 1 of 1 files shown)


IGordynskyi:
@lguohan / @kcudnik / @itaibaz / @karthik-krishnamurthy Could you please review this PR?

inc/saitunnel.h
```diff
@@ -463,7 +462,6 @@ typedef enum _sai_tunnel_attr_t
      * @type sai_object_id_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      * @objects SAI_OBJECT_TYPE_ROUTER_INTERFACE
-     * @condition SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_GRE
```
MELLANOX-zhichenr: For VXLAN, do we need overlay interface?

IGordynskyi: @MELLANOX-zhichenr I also thought that we do not need that. But configuration of SAI_TUNNEL_OVERLAY_INTERFACE attribute for VXLAN tunnel is given in [SAI-Proposal-tunnel_v6](https://github.com/opencomputeproject/SAI/blob/master/doc/tunnel/SAI-Proposal-tunnel_v6.docx) (page 23).

MELLANOX-zhichenr: There is an updated proposal for VXLAN tunnel with .1D bridge. In this case, I guess we need to use tunnel map <vni, bridge if> instead of overlay interface.

https://github.com/opencomputeproject/SAI/blob/master/doc/bridge/SAI_Proposal_Bridge_port_v0.9.docx

I guess the original tunnel proposal needs to be updated.

Thus I think we need to keep the comment for overlay interface. The other changes look fine to me.

IGordynskyi: @MELLANOX-zhichenr, ok. I will revert condition section for SAI_TUNNEL_OVERLAY_INTERFACE attribute.

MELLANOX-zhichenr: Thanks! The new change looks good to me.

karthik-krishnamurthy: @MELLANOX-zhichenr  @IGordynskyi Wouldn't overlay interface needed for VXLAN for L3 gateway or RIOT feature ? Say after  VXLAN termination we want to route the inner packet then the overlay interface would be needed i guess ?

IGordynskyi: @karthik-krishnamurthy / @MELLANOX-zhichenr , yes it looks like we need SAI_TUNNEL_OVERLAY_INTERFACE attribute for L3 gate way functionality. Please see VXLAN decap pipeline object model: 
vxlan_decap

IGordynskyi: @MELLANOX-zhichenr Do you have any concerns on this?

MELLANOX-zhichenr: I am OK with the change since there is a use case for it.

IGordynskyi:
@lguohan / @kcudnik, Could you please merge this PR?
Thanks,
Iurii

IGordynskyi:
Hi @kcudnik / @lguohan, 
If you have any concerns on this PR or if this PR need to be declined, please let me know.
Thanks

jcmathews:
Is HEAD END REPLICATION supported for VXLAN, Any SAI attributes are defined for them?

PR #744, opened 2017-12-06
Modify return type for pfc deadlock event.

tushar-ty:
This is required to provide some handshake mechanism between the application and SAI/SDK in the event flow.


kcudnik:
Please remove, all notifications should return void, id you need extra flag please add it as parameter

kcudnik:
If you want to change some option about some notifications on object then there should be attribute on that object to set it

PR #745, opened 2017-12-07
Add return flag in notification data structure.

tushar-ty:
Add return flag in notification data structure to indicate app behavior on every deadlock situation.

### DIFF
--- a/inc/saiqueue.h
+++ b/inc/saiqueue.h
@@ -318,7 +318,8 @@ typedef struct _sai_queue_deadlock_notification_data_t
     /** Deadlock event */
     sai_queue_pfc_deadlock_event_type_t event;
 
-    /** @brief Application based recovery management indicator.
+    /**
+     * @brief Application based recovery management indicator.
      *
      * This is a return value from host adapter.
      * If set to TRUE then host application will manage deadlock recovery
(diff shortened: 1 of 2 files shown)


tushar-ty:
Kamil, could you pls explain whats missing in the change thats causing the latest error.

kcudnik:
unittest fails since you added extra field to queue struct:

```./saiserializetest >/dev/null
ASSERT STR_EQ FAILED(test_serialize_notifications:1052): is:
{"count":1,"data":[{"queue_id":"oid:0x0","event":"SAI_QUEUE_PFC_DEADLOCK_EVENT_TYPE_DETECTED","app_managed_recovery":false}]}
expected:
{"count":1,"data":[{"queue_id":"oid:0x0","event":"SAI_QUEUE_PFC_DEADLOCK_EVENT_TYPE_DETECTED"}]}
```

take a look at meta/saiserializetest.c at line 1052

lguohan:
@sihuihan88 to review

inc/saiqueue.h
```diff
@@ -318,6 +318,16 @@ typedef struct _sai_queue_deadlock_notification_data_t
...
     sai_queue_pfc_deadlock_event_type_t event;
 
+    /**
+     * @brief Application based recovery management indicator.
+     *
+     * This is a return value from host adapter.
+     * If set to TRUE then host application will manage deadlock recovery
+     * else SAI adapter or SDK will manage deadlock recovery
+     * and also generate recovery ended notification.
+     */
```
lguohan: should we also add a comment that this attribute is only APPLICABLE when event == SAI_QUEUE_PFC_DEADLOCK_EVENT_TYPE_DETECTED?

PR #746, opened 2017-12-12
Bfd proposal

ragu-raman:
Addressed the comments from @kcudnik 
Added examples for sai lookup case and non lookup case
Added C-BIT attribute
Added Num of BFD sessions attribute in saiswitch.c


kcudnik:
file header is not matching other headers,
also: sai_bfd_encapsulation_type_t == SAI_BFD_SESSION_ATTR_ENCAPSULATION_TYPE not ending on enum name BFD_ENCAPSULATION_TYPE

lguohan:
@ragu-raman, can you resolve the conflict and also pass the meta data checking?

ragu-raman:
@kcudnik @lguohan Can you please help me to fix the below error?
Creating serialzie for sai_bfd_session_state_change_notification_fn
not handled sai_bfd_session_state_notification_t data in sai_bfd_session_state_change_notification_fn, FIXME

kcudnik:
Please take a look at my fix https://github.com/kcudnik/SAI/commit/9bf68246adc93eb911d63a25519f341040ae95f8 i fixed all your issues, but i also had to change some validonly/conditions so you should revisit that, since we dont allow neasted validonly conditions

You should be able to cherrypick those changes to your branch and build should be succeess + resolve conflicts

kcudnik:
As error says:
last doxygen line should not be empty saibfd.h:  *

inc/saibfd.h
```diff
@@ -0,0 +1,563 @@
...
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_BFD_SESSION_ATTR_MULTIHOP,
+
+    /**
+     * @brief Control Plane Independent
+     *
+     * @type bool
+     * @flags CREATE_ONLY
```
lguohan: why CREATE_ONLY, can it be create and set?

inc/saibfd.h
```diff
@@ -0,0 +1,563 @@
...
+    SAI_BFD_SESSION_ATTR_ECHO_ENABLE,
+
+    /**
+     * @brief Multi hop BFD session
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_BFD_SESSION_ATTR_MULTIHOP,
```
lguohan: what is multihop attribute here?

inc/saibfd.h
```diff
@@ -0,0 +1,563 @@
...
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_BFD_SESSION_ATTR_TYPE = SAI_BFD_SESSION_ATTR_START,
+
+    /**
+     * @brief Hardware lookup valid
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
```
lguohan: can you make the default to be true since that is more desired mode?

inc/saibfd.h
```diff
@@ -0,0 +1,563 @@
...
+
+    /**
+     * @brief Virtual Router
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_VIRTUAL_ROUTER
+     * @condition SAI_BFD_SESSION_ATTR_HW_LOOKUP_VALID == true
+     */
+    SAI_BFD_SESSION_ATTR_VIRTUAL_ROUTER,
```
lguohan: can you group all HW_LOOKUP_VALID related attribute together?

kcudnik:
as i mentioned before, please revisit attribute flags:
ASSERT FAILED (on line 1390): SAI_BFD_SESSION_ATTR_VLAN_TPID: validonly attribute is also validonly attribute, not allowed

we currently don't allow that one attribute depends on another attribute
on validonly case we can probably remove that check but i need to revisit that why this was added on the first place

inc/saibfd.h
ragu-raman: This is a type of BFD session where the packet is expected to go over more than one ip hop

inc/saibfd.h
ragu-raman: Changed

inc/saibfd.h
ragu-raman: The attributes are grouped based on the headers of the bfd packet. Changing that order might be more confusing

PR #756, opened 2018-01-16
Added SAI_OBJECT_TYPE_BRIDGE_PORT support in ACL

manish-volansys:


inc/saiacl.h
```diff
@@ -1195,7 +1195,7 @@ typedef enum _sai_acl_entry_attr_t
      *
      * @type sai_acl_field_data_t sai_object_list_t
      * @flags CREATE_AND_SET
-     * @objects SAI_OBJECT_TYPE_PORT
+     * @objects SAI_OBJECT_TYPE_BRIDGE_PORT
```
lguohan: we still need to match the physical port in some cases, we can allow to match extra bridge_ports

manish-volansys: Thanks @lguohan I made the changes accordingly.

manish-volansys: @lguohan Looks like after adding both the objects together (i.e. "@objects SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_BRIDGE_PORT") sanity build fails with following error: 

**ASSERT FAILED (on line 3191): SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS: allowed object types on object id list is more then 1, not supported yet**

When I looked at **check_mixed_object_list_types() in meta/saisanitycheck.c**, it looks like this type of syntax is currently not supported. 

Do we have any specific syntax for such case or should we add that?

kcudnik: this was added to not allow different object types on on object list for example in this case should it be possible that this list will contain "port and bridge" object ids? so far no SAI was allowing that and all lists were supporting single object types, but i had this in mind and in such case i was preparind "allowmixedobjecttypes" attribute on list, i will add this tag shortly so you can use that

kcudnik:
relaxed on #759

inc/saiacl.h
kcudnik: relaxed on #759

kcudnik:
test this please

manish-volansys:
@lguohan @kcudnik Since this pending too long, could you please merge this?

kcudnik:
test this please

PR #758, opened 2018-01-18
Add SAI_OBJECT_TYPE_BRIDGE_PORT in saiacl.h

manish-volansys:


kcudnik:
ASSERT FAILED (on line 3191): SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS: allowed object types on object id list is more then 1, not supported yet

i think this restriction could be relaxed to support different object types @lguohan can you confirm?

kcudnik:
closing, this change is for v1.2 branch, if needed just reopen

PR #760, opened 2018-01-19
RIF loopback packet action

SathishKumarTR-Dell:
Updated the RIF loopback attribute to type sai_packet_action_t as it defines the packet action for the packets that ingress and gets routed on the same RIF. Possible packet actions would be SAI_PACKET_ACTION_DROP or SAI_PACKET_ACTION_COPY or SAI_PACKET_ACTION_FORWARD.


SathishKumarTR-Dell:
@lguohan @mattyk  Can you approve this?

PR #766, opened 2018-01-31
Adding ECN threshold and marking probability attributes for ECT traffic

karthik-krishnamurthy:
The pull request is for having ECN min, max thresholds and marking probability for ECT traffic. By default, without using the new attributes both non-ECT/ ECT taffic would use the WRED min/max thresholds and drop probability. The new attributes provide options to have separate ECN thresholds for ECT traffic


karthik-krishnamurthy:
@lguohan @kcudnik  Tried to the additional condition SAI_SWITCH_ATTR_ECN_ECT_THRESHOLD_ENABLE, however the meta checks fail. Should i revert the change and document it in the comment ?

The condition which we want to add contains mixed conditions and/or and also references a attribute which is not part of this object. Doubt if this is supported currently.

SAI_SWITCH_ATTR_ECN_ECT_THRESHOLD_ENABLE == true and ( SAI_WRED_ATTR_ECN_MARK_MODE == SAI_ECN_MARK_MODE_GREEN or SAI_WRED_ATTR_ECN_MARK_MODE == SAI_ECN_MARK_MODE_GREEN_YELLOW or SAI_WRED_ATTR_ECN_MARK_MODE == SAI_ECN_MARK_MODE_GREEN_RED or SAI_WRED_ATTR_ECN_MARK_MODE == SAI_ECN_MARK_MODE_ALL )

lguohan:
@kcudnik, does out meta support such condition?

kcudnik:
No, you can eather use "or" or "and" condition, there are not combinations like this, this would require to support "parser" which would allow that, this condition is too comples for now

karthik-krishnamurthy:
@kcudnik @lguohan ok I will revert the change and update the comments

karthik-krishnamurthy:
@kcudnik I am not able to add the condition in comments as well as the description text seems to be limited to 200 characters.

karthik-krishnamurthy:
@lguohan Please approve and merge this commit

kcudnik:
Description is limited only to 200 characters - but only first line of "brief" if you add empty line after attribute description like this:

@brief this attribute doest ...

But condition only appliest "here put long condition"

kcudnik:
FIY, i also realized that SAI_SWITCH_ATTR_ECN_ECT_THRESHOLD_ENABLE is attribute from another object type, and right now all conditions belongs to the same object type, in which we already know on which object we perform condition, in case of adding different object type in condition, user would have to specify also object id on which this condition should be tested, and that would involve sole other logic to thest if that other attribute from other object type is also conditional and if is that condition met, which would be challenging to implement such logic.

in case of switchobject id we could assume that we use switch_id on which we perform operation, but still challenging logic problem would stand

karthik-krishnamurthy:
@kcudnik Ok so looks like it is difficult to implement a logic for such a condition.
@lguohan I have added the condition in the description. Please let me know if this looks fine

PR #767, opened 2018-02-04
Add mirror sample rate in mirror session attributes

zhenggen-xu:
This PR is to add the mirror sample rate attribute for mirror session, so we could do 1: N mirror/sampling instead of 1:1 in some cases.  This does not change the default behavior.

### DIFF
--- a/inc/saimirror.h
+++ b/inc/saimirror.h
@@ -101,6 +101,18 @@ typedef enum _sai_mirror_session_attr_t
     SAI_MIRROR_SESSION_ATTR_TRUNCATE_SIZE,
 
     /**
+     * @brief Mirror sample rate. Every 1/sample_rate the packets will be mirrored.
+     *
+     * Value 0 to no sampling
+     * Value 1 to every packet sampling (normal mirror)
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 1
+     */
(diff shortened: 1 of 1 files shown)


inc/saimirror.h
```diff
@@ -100,6 +100,18 @@ typedef enum _sai_mirror_session_attr_t
      */
     SAI_MIRROR_SESSION_ATTR_TRUNCATE_SIZE,
 
+    /**
+     * @brief Mirror sample rate. Every 1/sampe_rate the packets will be mirrored.
+     *
+     * Value 0 to no sampling
+     * Value 1 to every packet sampling (normal mirror)
+     *
+     * @type sai_uint32_t
```
kcudnik: use bool ?

lguohan:
the target branch should be master.

inc/saimirror.h
```diff
@@ -100,6 +100,18 @@ typedef enum _sai_mirror_session_attr_t
      */
     SAI_MIRROR_SESSION_ATTR_TRUNCATE_SIZE,
 
+    /**
+     * @brief Mirror sample rate. Every 1/sampe_rate the packets will be mirrored.
```
lguohan: sample_rate

lguohan:
broadcom will further look at this and comment back.

kcudnik:
closing, this change is for v1.2 branch, if needed just reopen

PR #769, opened 2018-02-04
Add trap type for subnet route

zhenggen-xu:
This trap is for packets hit the subnet routes on HW.


lguohan:
per community discussion, use SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_NEIGHBOR

PR #770, opened 2018-02-05
Add comment for consolidated fdb flush notification

kcudnik:


inc/saifdb.h
```diff
@@ -264,6 +264,37 @@ typedef enum _sai_fdb_flush_attr_t
 /**
  * @brief Notification data format received from SAI FDB callback
  *
+ * When FDB flush API is called (for example with no parameters) and switch
+ * learned a lot of MAC addresses, then calling this API can cause to generate
+ * a lot of notifications.
+ *
+ * Vendor can decide whether in that case send notifications 1 by 1 and
```
prsunny: If it is one-by-one, what is the expectation for event type?, Is it still SAI_FDB_EVENT_FLUSHED?

lguohan: I think it should be SAI_FDB_EVENT_FLUSHED

kcudnik: yes, it should still be SAI_FDB_EVENT_FLUSHED

PR #771, opened 2018-02-09
Add mirror sample rate in mirror session attributes

zhenggen-xu:
Per request, this PR moved to master.

This PR is to add the mirror sample rate attribute for mirror session, so we could do 1: N mirror/sampling instead of 1:1 in some cases. This does not change the default behavior.


zhenggen-xu:
Guohan,

I also cherry-picked the change to v1.2 branch for PR:  https://github.com/opencomputeproject/SAI/pull/767
I assume SAI 3.2 from BRCM will use v1.2?  If so, we need merge PR 767 as well.

Regards
/Xu

PR #777, opened 2018-02-15
Add SAI_PORT_ATTR_EYE_VALUES

nikos-li:


kcudnik:
A new value type SAI_ATTR_VALUE_TYPE_PORT_EYE_VALUES_LIST must be added after this: https://github.com/opencomputeproject/SAI/blob/master/meta/saimetadatatypes.h#L360

inc/saitypes.h
```diff
@@ -691,6 +691,40 @@ typedef struct _sai_segment_list_t
...
+ * must set the count member to the actual number of objects filled in the
+ * list member and return #SAI_STATUS_BUFFER_OVERFLOW. Once the caller
+ * gets such a return code, it may use the returned count member to
+ * re-allocate the list and retry.
+ */
+typedef struct _sai_port_eye_values_list_t
+{
+    uint32_t count;
+    sai_port_lane_eye_values_t *list;
+} sai_port_eye_values_list_t;
```
kcudnik: proposal to change name to sai_port_lane_eye_values_list_t or remove "lane" from name sai_port_lane_eye_values_t

inc/saitypes.h
```diff
@@ -691,6 +691,40 @@ typedef struct _sai_segment_list_t
     sai_ip6_t *list;
 } sai_segment_list_t;
 
+/**
+ * @brief Defines a lane with its eye values
+ */
+typedef struct _sai_port_lane_eye_values_t
+{
+    uint32_t lane_number;
```
kcudnik: would "lane" be enough ? you can add comment to that filed that thie is lane number, but single "lane" in this context is probably self explaining

nikos-li: I'm fine with just lane.

inc/saitypes.h
nikos-li: sai_port_lane_eye_values_list_t can imply the values for one lane which is not the case (despite the list in the name). I can do sai_port_lanes_eye_values_list_t if you like - I don't want any confusion on such APIs.

nikos-li:
>A new value type SAI_ATTR_VALUE_TYPE_PORT_EYE_VALUES_LIST must be added after this: >https://github.com/opencomputeproject/SAI/blob/master/meta/saimetadatatypes.h#L360

Correct. I did a git add . and was in the wrong directory and missed that it didn't pick that change. Will correct tomorrow.

inc/saitypes.h
lguohan: what is the unit for these for left/right/up/down values?

nikos-li:
@lguohan Please assign someone to fix the SAI dictionary unless I'm misreading the build error.

lguohan:
@nikos-li , can you take a look at this PR and add word to dictionary? https://github.com/opencomputeproject/SAI/pull/776/files

kcudnik:
please add word "millivolts" to meta/aspell.en.pws

nikos-li:
@lguohan @kcudnik meta/acronyms.txt is not allowing mV acronym. If acronyms.txt is only allowing all capital acronyms then that's wrong and needs to be corrected.

inc/saitypes.h
```diff
@@ -691,6 +691,41 @@ typedef struct _sai_segment_list_t
...
 
+/**
+ * @brief Defines a lane with its eye values with the values being in
+ * millivolts.
+ */
+typedef struct _sai_port_lane_eye_values_t
+{
+    uint32_t lane;
+    int32_t left;
+    int32_t right;
```
SAI-shared-user: left, right seems to be related to time instead of voltage.

lguohan:
one thing we can do is that use millivolts instead of mV here.

nikos-li:
@kcudnik @lguohan How do I add mV and mUI in acronyms.txt? It seems to accept only all upper case and that is not right for those units. What else needs to change for acronyms to accept mixed lower/upper case?

kcudnik:
add mV to meta/aspell.en.pws

inc/saitypes.h
```diff
@@ -691,6 +691,41 @@ typedef struct _sai_segment_list_t
     sai_ip6_t *list;
 } sai_segment_list_t;
 
+/**
+ * @brief Defines a lane with its eye values with the up and down values
+ * being in mV and left and right being in mUI.
```
lguohan: I am not sure about mUI. https://en.wikipedia.org/wiki/Unit_interval_(data_transmission)

I think it should be UI.

nikos-li: UI with int32_t will cause rounding up/down. mUI is correct.

lguohan: what does 'm' stands for here? milli UI, like UI/1000?

nikos-li: Yes. m in measurements stands for milli.

PR #781, opened 2018-02-22
Add a comment about port PVID matching LAG PVID before adding it to LAG

itaibaz:
Application should set the port PVID to the same of the lag PVID before
adding the port as a LAG member. This simplifies SAI implementation,
since SAI does not need to bookkeep the port original PVID, and to reset
the port PVID to the original value after removing the port from the LAG


lguohan:
mellanox, dell, microsoft, metaswitch agree this is right approach to go, while linkedin/broadcom in favor of  other approach. A vote is done in favor of this approach.

PR #786, opened 2018-03-08
Add mirror session congestion mode

itaibaz:
Controls whether mirroring traffic can cause back pressure and packet
drop of the original traffic

### DIFF
--- a/inc/saimirror.h
+++ b/inc/saimirror.h
@@ -62,6 +62,19 @@ typedef enum _sai_erspan_encapsulation_type_t
 } sai_erspan_encapsulation_type_t;
 
 /**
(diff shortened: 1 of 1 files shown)


atitjain:
How is this ensured? The mirror traffic then needs to take different queue on the monitor port?

itaibaz:
In correlated mode, every packet that matches mirroring, is mirrored, no matter what the buffer status is. This eventually can cause back pressure, such as buffers filling up, or pause frames being generated for loseless traffic.

In independent mode (best effort), when a packet matches mirroring, there is a check on the relevant queue of the analyzer port - if there is space in the buffer, the packet is mirrored. if the buffer is full, the packet isn't mirrored

PR #787, opened 2018-03-14
Proposed API to query generic capability of attribute and enum attrib…

itaibaz:
…ute values

Generic API to query capability (is implemented per operation) for sai
attribute ID
Generic API to query enum values capability (list of implemented enum
values) for enumerated attributes (enum or enum list)

### DIFF
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -62,6 +62,27 @@ typedef struct _sai_object_key_t
 } sai_object_key_t;
 
 /**
+ * @brief Structure for attribute capabilities per operation
+ */
+typedef struct _sai_attr_capability_t
+{
+    /**
+     * @brief Create is implemented
+     */
+    bool create_implemented;
+
+    /**
+     * @brief Set is implemented
+     */
+    bool set_implemented;
+
+    /**
+     * @brief Get is implemented
+     */
+    bool get_implemented;
+} sai_attr_capability_t;
+
+/**
  * @brief Get maximum number of attributes for an object type
  *
  * @param[in] switch_id SAI Switch object id
@@ -144,6 +165,34 @@ sai_status_t sai_bulk_get_attribute(
         _Inout_ uint32_t *attr_count,
         _Inout_ sai_attribute_t **attr_list,
         _Inout_ sai_status_t *object_statuses);
+
+/**
+ * @brief Query attribute capability
+ *
+ * @param[in] object_type SAI object type
+ * @param[in] attr_id SAI attribute ID
+ * @param[out] attr_capability Capability per operation
+ *
+ * @return #SAI_STATUS_SUCCESS on success, failure status code on error
+ */
+sai_status_t sai_query_attribute_capability(
+        _In_ sai_object_type_t object_type,
+        _In_ sai_attr_id_t attr_id,
+        _Out_ sai_attr_capability_t *attr_capability);
+
+/**
+ * @brief Query an enum attribute (enum or enum list) list of implemented enum values
+ *
+ * @param[in] object_type SAI object type
+ * @param[in] attr_id SAI attribute ID
+ * @param[inout] enum_values_capability List of implemented enum values
+ *
+ * @return #SAI_STATUS_SUCCESS on success, #SAI_STATUS_BUFFER_OVERFLOW if list size insufficient, failure status code on error
+ */
+sai_status_t sai_query_attribute_enum_values_capability(
+        _In_ sai_object_type_t object_type,
+        _In_ sai_attr_id_t attr_id,
+        _Inout_ sai_s32_list_t *enum_values_capability);
 
 /**
  * @}
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -169,6 +169,7 @@ sai_status_t sai_bulk_get_attribute(
 /**
  * @brief Query attribute capability
  *
+ * @param[in] switch_id SAI Switch object id
  * @param[in] object_type SAI object type
  * @param[in] attr_id SAI attribute ID
  * @param[out] attr_capability Capability per operation
@@ -176,6 +177,7 @@ sai_status_t sai_bulk_get_attribute(
  * @return #SAI_STATUS_SUCCESS on success, failure status code on error
  */
 sai_status_t sai_query_attribute_capability(
+        _In_ sai_object_id_t switch_id,
         _In_ sai_object_type_t object_type,
         _In_ sai_attr_id_t attr_id,
         _Out_ sai_attr_capability_t *attr_capability);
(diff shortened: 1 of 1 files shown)


inc/saiobject.h
```diff
@@ -61,6 +61,27 @@ typedef struct _sai_object_key_t
...
+
+    /**
+     * @brief Set is implemented
+     */
+    bool set_implemented;
+
+    /**
+     * @brief Get is implemented
+     */
+    bool get_implemented;
```
kcudnik: why remove is missing ? Is create implemented implies that remove is implemented as well?

inc/saiobject.h
```diff
@@ -145,6 +166,34 @@ sai_status_t sai_bulk_get_attribute(
...
+ * @param[in] object_type SAI object type
+ * @param[in] attr_id SAI attribute ID
+ * @param[inout] enum_values_capability List of implemented enum values
+ *
+ * @return #SAI_STATUS_SUCCESS on success, #SAI_STATUS_BUFFER_OVERFLOW if list size insufficient, failure status code on error
+ */
+sai_status_t sai_query_attribute_enum_values_capability(
+        _In_ sai_object_type_t object_type,
+        _In_ sai_attr_id_t attr_id,
+        _Inout_ sai_s32_list_t *enum_values_capability);
```
kcudnik: Im not following description here, can you give example of such api ?

From current description im assumming that if attribute is enum or enum list, this api will return only supported enum values for that specific attribute id ?

inc/saiobject.h
```diff
@@ -145,6 +166,34 @@ sai_status_t sai_bulk_get_attribute(
...
+ * @param[in] object_type SAI object type
+ * @param[in] attr_id SAI attribute ID
+ * @param[out] attr_capability Capability per operation
+ *
+ * @return #SAI_STATUS_SUCCESS on success, failure status code on error
+ */
+sai_status_t sai_query_attribute_capability(
+        _In_ sai_object_type_t object_type,
+        _In_ sai_attr_id_t attr_id,
+        _Out_ sai_attr_capability_t *attr_capability);
```
kcudnik: Also if libsai will support multiple asic at the same time (multiple switches created) there can be possibility that same attribute is not implemented/not supported on one of those asics, and this api has no way to tell the difference

inc/saiobject.h
itaibaz: As for remove operation, remove operation doesn't accept an attribute list, remove operation is without any attribute, so checking whether an attribute is implemented for remove is not relevant

As for the reason having this API - we won't to avoid having to call create/get/set and get an error, and then understand that the attribute isn't implemented. AFAIK, for example, Sonic now crashes on such error. And such error also gets the log dirty. So we want a clean way for the application to query capabilities

inc/saiobject.h
itaibaz: Let's look as an example on

/**
 * @brief Priority flow control mode
 */
typedef enum _sai_port_priority_flow_control_mode_t
{
    /** Same value for RX/TX */
    SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_COMBINED,

    /** Separate values for RX/TX */
    SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_SEPARATE,

} sai_port_priority_flow_control_mode_t;

    /**
     * @brief Combined or separate Bit vectors for port PFC RX/TX
     *
     * @type sai_port_priority_flow_control_mode_t
     * @flags CREATE_AND_SET
     * @default SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_COMBINED
     */
    SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL_MODE,

Now let's say your SAI implementation supports only combined mode, and not separated mode

Application will :
sai_s32_list_t attribs
attribs.count = 2
attribs.list = malloc 2 items
call the query enum values api

SAI implementation in this example, will fill
count = 1
list[0] = combined mode

inc/saiobject.h
itaibaz: Thanks for the input, I will add switch ID as parameter

inc/saiobject.h
lguohan: @kcudnik , one challenge of using create api is that it will taint the asic sometimes with failed operation. For example, to implement a feature, there are two approaches. We can try to use the first approach, it may fail at certain stage, then we have to roll back all previous actions and try the 2nd approach. The roll back operations might be challenage. 

Instead if we have the query API, we can query multiple capabilities at once and decide which way to go.

kcudnik: of course, but as we spoke some time ago that when create fails rollback should be easy, rollback could be challenge when we would set multiple attributes at once thatswhy we have only one attribute for set api. And query first is way to go good enough

inc/saiobject.h
kcudnik: great thanks for explaining

inc/saiobject.h
lguohan: yes, but an approach may require creating multiple objects, and you failed for the last object creation. then you need to rollout back all previous created objects. Sometime, the object cannot be created because the resource constraint not the asic capability, in this case you need to check the error code and why it fails, and then decide how to react.

inc/saiobject.h
judyjoseph: Hi @itaibaz

  I came across this explanation and would to understand how can we achieve a capability check where there are multiple values for an attribute, and the SAI capability check will return back "True" only if it supports all of those attributes together.

  The scenario is like this --> for a particular ASIC family, the SAI does not support both v4 (SRC_IP, DST_IP) and V6(SRC_IPV6, DST_IPV6) attributes in one field group for user ACL tables. The key size is exceeding the HW max-key.
 
So can we have a capability query which checks if the combination of "SAI_ACL_ACTION_TYPE_SET_SRC_IP, SAI_ACL_ACTION_TYPE_SET_DST_IP, SAI_ACL_ACTION_TYPE_SET_SRC_IPV6, SAI_ACL_ACTION_TYPE_SET_DST _IPV6" for the attribute SAI_ACL_TABLE_ATTR_ACL_ACTION_TYPE_LIST is supported OR not. 

Currently I understand that we have the capability check individually for a capability of an attribute, or as in your eg: above the capability query can return back what all capabilities SAI supports for an attribute. 

But how do we do this capability check for a combination of capabilities that could be used together. Pls share your thoughts - I can share more details too.

 -thanks

itaibaz: Hi,
The current method is not intended to query relationship between values and more complicated limitations
In the example you mentioned, the query should return all values, since they are individually supported. And the additional limitations can't be expressed
In addition, the query capability is intended to query specific vendor implementation, that might be a subset of the full SAI API. But the case you describe is generic to all vendors, you can't set both v4 and v6 fields, as a packet is either v4 or v6. So this could theoretically be described in the SAI API itself IMO

judyjoseph: Thanks @itaibaz

PR #790, opened 2018-03-14
Update sai.h

trignesh:
Updated doxygen comment in sai.h

### DIFF
--- a/inc/sai.h
+++ b/inc/sai.h
@@ -149,8 +149,8 @@ typedef enum _sai_log_level_t
 
 /**
 * @brief Defines sai profile get value function pointer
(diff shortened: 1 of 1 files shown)


inc/sai.h
```diff
@@ -147,10 +147,16 @@ typedef enum _sai_log_level_t
 
 } sai_log_level_t;
 
+/**
+* @brief Defines sai profile get value function pointer
+*/
```
kcudnik: if you are doxygen adding comments, also params[] should be added with description, see other methods for examples and error logs when building

kcudnik:
Please fix compilation issues

PR #791, opened 2018-03-14
Add extended statistics get functions

itaibaz:
Extended funtions have extra mode paramter. For now the options are
regular read and read with clear. Read and clear allows atomic operation
without loosing counters in the gap between the read and clear functions
in existing API.
Mode can be furthermore extended upon future need.
Old get statistics functions are deprecated but left for backward
compatability

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1651,7 +1651,7 @@ typedef sai_status_t (*sai_get_port_attribute_fn)(
         _Inout_ sai_attribute_t *attr_list);
 
 /**
- * @brief Get port statistics counters.
+ * @brief Get port statistics counters. Deprecated for backward compatibility.
  *
  * @param[in] port_id Port id
  * @param[in] number_of_counters Number of counters in the array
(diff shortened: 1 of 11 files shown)


inc/saitypes.h
```diff
@@ -802,6 +802,19 @@ typedef sai_status_t (*sai_bulk_object_remove_fn)(
...
+{
+    /**
+     * @brief Read statistics
+     */
+    SAI_STATS_MODE_READ,
+
+    /**
+     * @brief Read and clear after reading
+     */
+    SAI_STATS_MODE_READ_AND_CLEAR,
```
kcudnik: for this to be atomic operation, actual asic needs to support such capability and maybe not all asic can support that

for example if asic dont support that, there is no way to guarantee atomic on that, in port stat when port is running like 100gbps, then after just reading stats all the byte counters will gain a lot until clear will be called

inc/saiport.h
```diff
@@ -1928,13 +1964,15 @@ typedef struct _sai_port_api_t
     sai_set_port_attribute_fn         set_port_attribute;
     sai_get_port_attribute_fn         get_port_attribute;
     sai_get_port_stats_fn             get_port_stats;
+    sai_get_port_stats_ext_fn         get_port_stats_ext;
```
atitjain: if we are retaining instead of passing mode as a parameter,

itaibaz:
Add query capability

PR #792, opened 2018-03-15
VRRP support in RIF create call

zubinshah:
VRRP support during RIF create calls. Currently it seems we don't have a way to manage multiple Virtual MAC entries to be added for Native RIF entries. This addition in the router_interface attributes will allow NOS to pass in a Port or VLAN rif -- create_router_interface call, and only use the SRC_MAC to program the ingress MAC entry for routing and avoid using this Virtual MAC for any other purpose in the SAI adapter.

### DIFF
--- a/inc/sairouterinterface.h
+++ b/inc/sairouterinterface.h
@@ -240,6 +240,19 @@ typedef enum _sai_router_interface_attr_t
     SAI_ROUTER_INTERFACE_ATTR_LOOPBACK_PACKET_ACTION,
 
     /**
+     * @brief RIF creation is a virtual RIF.
+     *
+     * Create a Virtual RIF object, which only programs the ingress router mac.
+     * This simplifies the management of VRRP backup role/master role as defined
+     * by RFC 5798 (or similar proprietary protocols).
+     *
+     * @type bool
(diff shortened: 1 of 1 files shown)


kcudnik:
Word 'mac' should use capital letters

AshokDaparthi:
@lguohan , @mattyk @itaibaz - Please review

inc/sairouterinterface.h
```diff
@@ -239,6 +239,19 @@ typedef enum _sai_router_interface_attr_t
      */
     SAI_ROUTER_INTERFACE_ATTR_LOOPBACK_PACKET_ACTION,
 
+    /**
```
zubinshah: ++ specific comments to not allow nbr entries , and this is simply about resource optimization. 
++ comment for using the native rif can also be used for vrrp , simply not using the egress resouces or not learning nbr on this vrrp rif.

zubinshah: ++ on virtual rif created by using this flag, attribute are invalid - ADMIN state for v4 and v6, mtu size, v4 and v6 mcast enable, neighbor packet actions. 

@lguohan and @AshokDaparthi  please check.

PR #794, opened 2018-03-28
Add capability metadata for attributes

kcudnik:


itaibaz:
I cloned meta dir from your fork, metacap branch
When trying to run make, I get errors with serialize.pm

Creating serialzie for sai_acl_resource_t
Creating serialzie for sai_attr_capability_metadata_t
ERROR: not handled sai_attr_capability_t operationcapability in sai_attr_capability_metadata_t, FIXME
Odd number of elements in hash assignment at serialize.pm line 363 (#1)
    (W misc) You specified an odd number of elements to initialize a hash,
    which is odd, because hashes come in key/value pairs.

Use of uninitialized value in list assignment at serialize.pm line 363 (#2)
    (W uninitialized) An undefined value was used as if it were already
    defined.  It was interpreted as a "" or a 0, but maybe it was a mistake.
    To suppress this warning assign a defined value to your variables.

    To help you figure out what was undefined, perl will try to tell you
    the name of the variable (if any) that was undefined.  In some cases
    it cannot do this, so it also tells you what operation you used the
    undefined value in.  Note, however, that perl optimizes your program
    and the operation displayed in the warning may not necessarily appear
    literally in your program.  For example, "that $foo" is usually
    optimized into "that " . $foo, and the warning will refer to the
    concatenation (.) operator, even though there is no . in
    your program.

Any idea what is wrong ?

itaibaz:
P.S. : otherwise I see there is already good validation for using incorrect enum value
So my comment re that issue is not relevant

kcudnik:
As you can see build succeeded on this branch, so probably you are mergin this commit with your code which is not synced with master?

PR #807, opened 2018-04-19
Add support  for experimental attributes

kcudnik:


experimental/saitypesextensions.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef __SAITYPESEXTENSIONS_H_
+#define __SAITYPESEXTENSIONS_H_
+
+#include <saitypes.h>
+
+/**
+ * @brief SAI object type extensions
+ *
+ * @flags
```
mikelazar: Do we need the flags tag ? Don't we have a contiguous range ?

experimental/saitypesextensions.h
```diff
@@ -0,0 +1,47 @@
...
+ */
+typedef enum _sai_object_type_extensions_t
+{
+    SAI_OBJECT_TYPE_EXTENSIONS_RANGE_START = SAI_OBJECT_TYPE_MAX,
+
+    SAI_OBJECT_TYPE_FPGA = SAI_OBJECT_TYPE_EXTENSIONS_RANGE_START,
+
+    /** Add new experimental object types above this line */
```
mikelazar: You may not want a doxygen comment here ... just a regular comment starting with a /*

experimental/saitypesextensions.h
kcudnik: we have, but @flags is enum bound, and this enum is not starting from "0" but from last enum of corresponding sai object type enum

experimental/saitypesextensions.h
kcudnik: ah yes

kcudnik:
as for empty enum, current constraints on parser don't allow to specify empty enums since they make no sense to the code (excluding start and end range)

experimental/saitypesextensions.h
mikelazar: OK, thanks

kcudnik:
test this please

PR #812, opened 2018-04-27
Add a new DTel attribute to provide fine-grained control while enabli…

shruthi9:
Add a new DTel attribute to provide fine-grained control while enabling drop reports. The new attribute selectively enables/disables queue tail drop reporting when drop reporting is enabled. If the new attribute is disabled, all other drops are still reported.

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -222,6 +222,9 @@ typedef enum _sai_acl_action_type_t
     /** Enable DTEL drop report (experimental) */
     SAI_ACL_ACTION_TYPE_DTEL_DROP_REPORT_ENABLE,
 
+    /** Enable DTEL tail drop reporting (experimental) */
+    SAI_ACL_ACTION_TYPE_DTEL_TAIL_DROP_REPORT_ENABLE,
+
     /** Set DTEL flow sampling (experimental) */
     SAI_ACL_ACTION_TYPE_DTEL_FLOW_SAMPLE_PERCENT,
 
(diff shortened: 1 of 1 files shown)


arkadiyshapiro:
@kcudnik - if its approved, how can we merge? Also do we need separate PR to merge to SAI master?

kcudnik:
yes, a separate PR is required for master

shruthi9:
@kcudnik: Thanks for the review! What is the procedure to merge this change into v1.3?

inc/saiacl.h
```diff
@@ -222,6 +222,9 @@ typedef enum _sai_acl_action_type_t
     /** Enable DTEL drop report (experimental) */
     SAI_ACL_ACTION_TYPE_DTEL_DROP_REPORT_ENABLE,
 
+    /** Enable DTEL tail drop reporting (experimental) */
+    SAI_ACL_ACTION_TYPE_DTEL_TAIL_DROP_REPORT_ENABLE,
```
lguohan: do you need to add TAIL_DROP in _sai_dtel_event_type_t

as well as introudce DTEL_TAIL_DROP_REPORT in saidtel.h?

shruthi9: Regarding the question that came up in the community review: we have this attribute as an ACL action because we want to selectively turn on/off tail drop reporting capability on specific flows (matched by the drop watchlist).

@lguohan: We already have SAI_DTEL_EVENT_TYPE_QUEUE_REPORT_TAIL_DROP which represents tail drop event irrespective of which mode (drop reporting or queue reporting) captured it.

lguohan: what is the difference between the TAIL_DROP_REPORT and QUEUE_REPORT_TAIL_DROP?

shruthi9: There are two complementary modes that can be independently enabled/disabled: Drop reporting and queue reporting. Drop reporting reports pipeline drops *and* queue tail drops. Queue reporting reports only queue tail drops, in addition to congestion events. They both have an overlap in functionality w.r.t queue tail drop reporting. 

TAIL_DROP_REPORT applies to queue tail drops as part of drop reporting. QUEUE_REPORT_TAIL_DROP applies to queue tail drops as part of queue reporting. These knobs independently control turning on/off these features in the two modes.

arkadiyshapiro:
@xinliu-seattle @lguohan - what else is needed to merge?

PR #818, opened 2018-05-09
Port isolation group enhancement

karthik-krishnamurthy:
The pull request adds support of port isolation group. The isolation group attribute can be set on port or bridge port object. Packets which ingress on port/bridge port should not forward packets to the members of port isolation group object.


lguohan:
comments from the community meeting:
- explain the relation between port isolation and bridge port isolation.
- add attribute for acl action
- add example how to use the port isolation.

karthik-krishnamurthy:
@lguohan  Added a document which explains the relation between port isolation and bridge port isolation.
Added example configurations. Also added an attribute ACL action.
Please review.

PR #821, opened 2018-05-28
l2mc/ipmc support multi-NPU

zhichuanhe:
I have discussed these revision with @Muthu through email,the detail information refer to http://lists.opencompute.org/pipermail/opencompute-sai/2018-May/subject.html. This submission involves two parts：
1. IPMC/L2MC/rpf Group ID allocataion in multiple NPU scenario,
 In multiple NPU scenario,the ingress NPU need to allocate a H/W resource called the IPMC index which will be indexed into table to give the ports list on which a multicast packet should go out, the other egress NPUs will use the IPMC index to read H/W table directly without going through the  lookup process based on packet's (S,G/x,G).From the point of view of the upper users(SONIC),the IPMC member are sets of router interface,and the same IPMC member can share a group,but from the point of view of NPU,finally,the group member need to convert to port list,so the relationship between IPMC Group and IPMC entry must be 1:1
2. Fix compilation  error in CheckInOutParam() in style.pm ,because the object "sai_object_id_t " can't use  __InOut__ type,

### DIFF
--- a/inc/saimirror.h
+++ b/inc/saimirror.h
@@ -62,6 +62,19 @@ typedef enum _sai_erspan_encapsulation_type_t
 } sai_erspan_encapsulation_type_t;
 
 /**
+ * @brief Mirror session congestion mode
+ */
+typedef enum _sai_mirror_session_congestion_mode_t
+{
+    /** Mirroring traffic is independent from original traffic, original traffic not affected by congestion of mirroring traffic */
+    SAI_MIRROR_SESSION_CONGESTION_MODE_INDEPENDENT,
+
+    /** Mirroring traffic is correlated with original traffic, and can cause back pressure and discards of original traffic if there is congestion */
+    SAI_MIRROR_SESSION_CONGESTION_MODE_CORRELATED,
+
+} sai_mirror_session_congestion_mode_t;
+
+/**
  * @brief SAI attributes for mirror session
  */
 typedef enum _sai_mirror_session_attr_t
@@ -111,6 +124,15 @@ typedef enum _sai_mirror_session_attr_t
      * @default 1
      */
     SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE,
+
+    /**
+     * @brief Controls whether mirroring traffic can cause back pressure and packet drop of the original traffic
+     *
+     * @type sai_mirror_session_congestion_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_MIRROR_SESSION_CONGESTION_MODE_INDEPENDENT
+     */
+    SAI_MIRROR_SESSION_ATTR_CONGESTION_MODE,
 
     /**
      * @brief Class-of-Service (Traffic Class)


meta/style.pm
```diff
@@ -607,10 +607,6 @@ sub CheckInOutParams
     my $ptr = $4;
     my $param = $5;
 
-    if ($type eq "sai_object_id_t" and not $line =~ /_In_ sai_object_id_t \w+|_In_ const sai_object_id_t \*\w+|_Out_ sai_object_id_t \*\w+/)
```
kcudnik: Please revert this back

kcudnik:
Use bulk functions to create multiple IPMC objects

meta/style.pm
kcudnik: this change will affect all objects, and for multiple ojects use bulk functions

zhichuanhe: ok

MaruthamuthuP:
Hi Kcudnik,
   The point is not about bulk creation of IPMC objects but in doing a Create IPMC Objects likes this creates separate IPMCGroupId/rpfGroupId/L2MCGroupId per every entry created, but by way making the create functions for these entities with a _InOut_ we expect the we could reuse IPMCGroupId/rpfGroupId/L2MCGroupId across multiple entries more so specifically on a Multi NPU scenario. 
i.e in Ingress NPU when the create is called since the argument is now changed to _InOut_ it could be filled with a NULL_OBJECT_ID(value of 0 lets say), which SAI Implentation could interpret the need for creating a new Id and for the egress NPU(in the context of S,G/*,G) it shall use the given Object Id and convert it into local H/W significant Index and create them, hence its better if we have this argument changed to _InOut_.

Thanks
Maruthamuthu.P

zhichuanhe:
In multi NPU scenario, next_hop_id and next_hop_group_id may also have the same problem. I think their parameters also need changed to InOut, right ?

kcudnik:
if the created object id can be reused in other object, then it should be an attribute, and if not attribute then it should be implemented like member and member_group etc

MaruthamuthuP:
I think my description in my earlier post was misleading, sorry about that.

My context of explanation is not to do with re-using this object ID in another Object , it is re-using that same ObjectID in the create request for other NPUs.
i.e.,

assuming say a *,G or S,G created in a ingress NPU resulted in creating the following (for NPU 1)
ipmcGroupId = 1 --> giving the list of L3 egress Interfaces for IPMC entry
rpfGroupId = 1 --> Giving the L3 Ingress Interface for IPMC entry
(or)
l2mcGroupId =1
Now for this NPU , the SAI layer could derive a H/W resource ID based on this ipmcGroupId/rpfGroupId to program some H/W tables.

Now if only this same Resource ID for other NPU are programmed appropriately, the packet when it ingresses this NPU will do a lookup and carry this H/W resource ID in the pipleline headers to inform the other NPU to use this resource ID to look up their DB instead of looking up the packet headers again in each of the egress NPUs.

Now inoder to achieve this we expect the SAI for other NPU also gets a create for ipmcGroupId=1 and rpfGroupId=1, by way of specifying the GroupID in the create itself explicitly. Because as such the sai_object_id is also a data type right,  which could have a default say SAI_NULL_OBJECT_ID(say 0), which shall be passed by the caller while wanting to create in the ingress NPU and expect a objectID created in return , and use this created SAI object ID for the other NPU's create request, by specifying explicitly in their creates so that the above mentioned resource sharing across NPU is achieved.

kcudnik:
from design point of view create function always should create new unique object id for each create call, and it should be bound to the npu that was created on, also this allows to count references on each object id usage and track each object. im not sure if i follow correctly this description but i don't see an issue with creating new object id on every call

zhichuanhe:
@MaruthamuthuP @kcudnik  can we add new property SAI_IPMC_GROUP_ATTR_IPMC_GROUP_ID in sai_ipmc_group_attr_t  to solve it? When we create a ipmc group with sai_create_ipmc_group_fn, we give an attribute list which include the attribute above. The SAI see the attribute and check if the value  is not SAI_NULL_OBJECT_ID(0), then use this value as return ipmc group ID, and give back the oid(_out_ sai_object_id_t *ipmc_group_id in sai_create_ipmc_group_fn). L2MC/RFP group ID also use the same solution.
    /**
     * @brief IPMC group id 
     *     
     *  This attribute only takes effect when the group id is not eqal to SAI_NULL_OBJECT_ID
     *
     * @type sai_object_id_t
     * @flags CREATE_ONLY
     * @objects SAI_OBJECT_TYPE_IPMC_GROUP
     */
    SAI_IPMC_GROUP_ATTR_IPMC_GROUP_ID 

typedef sai_status_t (*sai_create_ipmc_group_fn)(
        _out_ sai_object_id_t *ipmc_group_id,
        _In_ uint32_t attr_count,
        _In_ const sai_attribute_t *attr_list);

kcudnik:
yes, this solutions works for me, this will create loop in objects because you will be able to pass same object  type into attribute of the same object type, there is check for that in sai sanity check, so an exception for this attribute will be needed

PR #827, opened 2018-06-16
SAI Test cases

prsunny:
Extended list of test cases to cover certain use-cases that verifies the SAI pipeline model.    

Note:
As discussed in community meeting (6/14/18), reference to pipeline model are added to requested test cases

### DIFF
--- a/test/saithrift/tests/SAITest.md
+++ b/test/saithrift/tests/SAITest.md
@@ -1,0 +1,122 @@
+Overview   
+========
+
+|  Testcase ID 	|         Test Catagory	 		| Number of TCs |   Owned by	|   Comments	|
+|---		|---					|---		|---		|---		|
+|   	1-2	|  L3 Directed Broadcast		|   2		|   		|   		|
+|   	3	|  L3 Loopback subnet			|   1		|   		|   		|
+|   	4-5	|  L2/L3 MTU				|   2		|   		|   		|
+|   	6-8	|  IPv4 /32 route and same Neighbor IP 	|   3		|   		|   		|
+|   	9	|  Dual/Triple tagging 			|   1		|   		|   		|
+|   	10-12	|  MAC move 				|   3		|   		|   		|
+|   	13-16	|  Neighbor MAC change and age-out 	|   4		|   		|   		|
+|   	17-19	|  Traffic/FDB learning on Bridgeport 	|   3		|   		|   		|
+|   	20-21	|  LAG/ECMP Hash seed 			|   2		|   		|   		|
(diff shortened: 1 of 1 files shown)


test/saithrift/tests/SAITest.md
```diff
@@ -0,0 +1,122 @@
...
+
+4.	L3MtuTest  
+Send L3 destined packet with max interface MTU and verify forwarding. With L3 destined packet, verifying the forwarding behavior also ensures IP MTU value configured for the router interface. Increase the packet size beyond the configured MTU value and verify the behavior. Default action is to drop packets exceeding configured MTU
+
+5.	L2MtuTest  
+Send untagged packet with max MTU size and verify forwarding.  Add VLAN tag so that the total packet size exceeds the MTU value and verify behavior. Default action is to drop packets exceeding configured MTU value. 
+
+6.	L3IPv4/32Test - I  
+(Ref: https://github.com/opencomputeproject/SAI/blob/master/doc/behavioral%20model/pipeline_v6.pdf, router flow)
+Add a route entry with /32 prefix (e.g. 10.1.1.1/32 via NH1) and a neighbor entry for the same IP address (e.g. 10.1.1.1 MAC1 Port1). Send traffic and verify that the packet takes the path specified by the /32 route entry and _not_ the neighbor path.
```
SathishKumarTR-Dell: For the directly reachable routes, egress_rif will be filled in the SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID attribute.
And, the "egress_rif, pkt_dst_ip" would be matched in the "Neigh table" instead of the "egress_rif,NH_DstIP" as captured in the "Ingress router flow" diagram.
In that case, there will be two matches for the packet in this scenario (/32 route && neighbor entry coexist) - both in the Router table and in the Neigh table. Then, there should be a stage that determines which one of the matches will take precedence. It looks like that is not defined explicitly in the pipeline model currently. 
@lguohan @atitjain Comments?

atitjain:
Hi Sathish,

Thanks for the discussion. I agree, the referred document pipeline_v6.pdf<https://github.com/opencomputeproject/SAI/blob/master/doc/behavioral%20model/pipeline_v6.pdf> doesn’t seem to provide the resolution logic if both, route and neighbor, tables are hit. I am also trying to find a scenario when this can happen (from application’s perspective).

Regards,
Atit

test/saithrift/tests/SAITest.md
lguohan: @SathishKumarTR-Dell , the packet should first hit the route table, then it will hit neigh table, there is no precedence between route table and neighbor table.

PR #840, opened 2018-06-27
Add port attibute to pause and unpause the egress of a port

wendani:


inc/saiport.h
```diff
@@ -1086,6 +1086,17 @@ typedef enum _sai_port_attr_t
...
+    /**
+     * @brief Pause port egress
+     *
+     * Enable/Disable egress pause of a port.
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_PORT_ATTR_EGRESS_PAUSE_ENABLE,
```
itaibaz: I think we can achieve the same functionality by
SAI_PORT_ATTR_QOS_SCHEDULER_PROFILE_ID
And using a profile with SAI_SCHEDULER_ATTR_MAX_BANDWIDTH_RATE of 0
Is this different ?

wendani: This STOP_PORT_MAX_RATE approach does not work on brcm platform.

itaibaz: Perhaps this should be a vendor specific attribute then ?
The scheduler profile approach is something all vendors agreed upon long time ago, having 2 ways to control the same functionality isn't the best approach IMO

lguohan: scheduler profile approach is a mean rather than the purpose, usually SAI attribute defines the purpose not the means.

tushar-ty: Request name change to TX_ENABLE or STOP_TX.
Comments should specify that this should not drop frames, just stop transmit and buffer.

wendani:
retest this please

wendani:
retest this please

wendani:
retest this please

wendani:
retest this please.

lguohan:
please resolve build failure

wendani:
--- is not allowed in the comment

PR #841, opened 2018-06-27
Add PTP support

itaibaz:
PTP traps + RX/TX meatadata


kcudnik:
timespec_t struct must be prefixed sai_

PR #846, opened 2018-07-16
Supporting bridge type as a match field in ACL

karthik-krishnamurthy:
This pull request adds support to match packets based on the bridge type. This enables the application to apply/skip ACL entry lookup based on the bridge type. Certain ACLs bound at switch level might need to be applied only for the regular .1Q bridge and not for .1D bridges, this match criteria would help to solve such usecases.

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -1706,6 +1706,7 @@ typedef enum _sai_acl_entry_attr_t
      *
      * @type sai_acl_field_data_t sai_object_id_t
      * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_BRIDGE
      * @default disabled
      */
     SAI_ACL_ENTRY_ATTR_FIELD_BRIDGE_ID,


karthik-krishnamurthy:
@kcudnik The build for the pull request seems to be hanged state. Is there some issue with the build ?

mikelazar:
A suggestion made in the OCP SAI meeting is to use a new bind point (saitypes::SAI_ACL_TABLE_ATTR_BIND_POINT) for the bridge type.

This will need an update to the pipeline model.

inc/saiacl.h
```diff
@@ -998,10 +998,19 @@ typedef enum _sai_acl_table_attr_t
...
     SAI_ACL_TABLE_ATTR_FIELD_IPV6_NEXT_HEADER,
 
+    /**
+     * @brief Bridge Type
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_BRIDGE_TYPE,
```
atitjain: Please highlight the use case. Is the idea to match all the packets that are classified to either 1Q Bridge or 1D bridge? Or we want to match bridge OID. Also we might want to add qualifiers like bridge OID.

karthik-krishnamurthy: @atitjain The requirement for this pull request is not to match on the bridge OID, but just to match packets classified on a .1Q bridge or on any .1D bridge. We want to classify packets at the switch bind point level, hence this match field is needed.

Yes it would be useful to add bridge OID and also a new bindpoint for .1D bridge similar to vlan. But the requirement is to also differentiate at the switch bind point level.

karthik-krishnamurthy: Yes that's the idea.

inc/saiacl.h
```diff
@@ -1660,10 +1678,43 @@ typedef enum _sai_acl_entry_attr_t
...
+     * This would be useful when the ACL table/group is
+     * bound to a switch bind point and the ACL needs to
+     * be applied only on bridges of type 802.ID or only
+     * on the 802.1Q bridge.The value for this match will
+     * the enum values of sai_bridge_type_t.
+     * i.e. SAI_BRIDGE_TYPE_1Q or SAI_BRIDGE_TYPE_1D.
+     *
+     * @type sai_acl_field_data_t sai_int32_t
+     * @flags CREATE_AND_SET
+     * @default disabled
```
atitjain: _@default_ value doesn't match the type (_sai_bridge_type_t_). disable if maps to 0 then from the _sai_bridge_type_t_, it matches _SAI_BRIDGE_TYPE_1Q_. Not sure if we should add some INVALID/NONE type in the enum. Please think about it.

inc/saiacl.h
```diff
@@ -1660,10 +1678,43 @@ typedef enum _sai_acl_entry_attr_t
...
+     * This is to match the packet's Bridge ID.
+     * This will be useful when the ACL table/group
+     * is bound to a switch bind point and the rules need
+     * to identify the Bridge on which the packet is
+     * present.
+     *
+     * @type sai_acl_field_data_t sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_BRIDGE
+     * @default disabled
```
atitjain: I guess _@default_ here should be _SAI_NULL_OBJECT_ID_.

inc/saiacl.h
karthik-krishnamurthy: Hi @atitjain , 

I followed the existing convention for defining fields the ACL entry. For example the field SAI_ACL_ENTRY_ATTR_FIELD_ACL_IP_TYPE  uses the enum sai_acl_ip_type_t. 
I guess the disabled value represents the boolean in sai_acl_field_data_t which is present in all ACL fields.

inc/saiacl.h
karthik-krishnamurthy: Same reason as the above. The similar example is SAI_ACL_ENTRY_ATTR_FIELD_IN_PORT where the default state  for sai_acl_field_data_t  is disabled. I guess this is convention followed in the existing saiacl.h header.
@kcudnik Is the understanding correct ?

inc/saiacl.h
karthik-krishnamurthy: @lguohan @atitjain Can this pull request be merged ? We had discussed this OCP call and i don't see any additional comments from the OCP members.

PR #847, opened 2018-07-16
TAM 2.0 updates

tushar-ty:


rchandramouli:
With reference the changes to saitam.h via commit: d44e0bc, would it be possible to update the proposal document with illustration on how partial snapshots can be obtained for, say, a set of ingress port/pg objects? In particular, it would be helpful if the examples provided in SAI-Proposal-TAM-v0.10.docx can be reduced to objects/attributes with TAM v2.1.

PR #850, opened 2018-07-30
SAI Neighbor entry VRF leaking

SathishKumarTR-Dell:
Added virtual router id into sai_neighbor_entry_t structure to fit the VRF leaking use-case that needs to leak the neighbor entry learnt on a egress RIF from one VRF into another VRF context. Details discussed on the email forum.

### DIFF
--- a/inc/saineighbor.h
+++ b/inc/saineighbor.h
@@ -128,6 +128,13 @@ typedef struct _sai_neighbor_entry_t
     sai_object_id_t switch_id;
 
     /**
+     * @brief Virtual Router ID
+     *
+     * @objects SAI_OBJECT_TYPE_VIRTUAL_ROUTER
+     */
+    sai_object_id_t vr_id;
+
+    /**
      * @brief Router interface ID
      *
      * @objects SAI_OBJECT_TYPE_ROUTER_INTERFACE


SathishKumarTR-Dell:
@atitjain I am not sure i understand your comment fully. 
sai_neighbor_entry_t has to be passed with valid vr_id and rif_id always even with the pull request. The rif_id will belong to the passed vrf in regular cases whereas the rif_id will be part of a different vrf in the leaking case.

atitjain:
Hi Sathish, 

I just wanted to understand if the adapter needs to create 2 different entries implicitly
- For the valid vr_id (derived from the rif)
- For the leaked vr_id 
Or the user will have to configure 2 entries explicitly 
- For the valid vr_id (derived from the rif) and with vr_id = NULL 
- For the leaked vr_id with rif = NULL
Maybe we will discuss more on a call. 

Regards,
Atit

SathishKumarTR-Dell:
@atitjain  By adapter you mean the SAI adapter or Host adapter? There is no implicit creation with the pull request as the API struct will contain both the vrf_id and rif_id fields. 
Host adapter (NOS) will create two SAI Neighbor entries in the VRF leaking use-case when it needs to leak the entry into another leaked vrf, like this:
(source-vr-id, rif-id, ip-address) and 
(leaked-vr-id, rif-id, ip-address).

PR #854, opened 2018-08-07
Add warm boot recover attribute

itaibaz:


atitjain:
Could you please add some more explanation on what is the idea behind recovery. what are the scenarios when you will be triggering it?

tushar-ty:
I could not understand the purpose and when/how would this attribute will be required/used ?
An example will help.
-thx.

itaibaz:
As I briefly explained in today's meeting
Currently there is existing attribute SAI_SWITCH_ATTR_RESTART_WARM, which serves as a hint a warm restart procedure is restarting
After such boot, on some ASICs, SAI is reinitialized from scratch, and SAI_KEY_BOOT_TYPE has the value of the boot type done
On other ASIC, SAI is not reinitialized from scratch, rather the process is resumed. A hint that the process is resumed is needed, and that's where the new attribute comes into play

PR #855, opened 2018-08-09
Added attribute to get ACL table usage related information

PugalendranDell:
Added following ACL table attributes,
1) SAI_ACL_TABLE_ATTR_PRIORITY - Priority is passed to SAI during ACL table creation
2) SAI_ACL_TABLE_ATTR_FIELD_DROP_MARKED - To drop a packet through ACL table
3) SAI_ACL_TABLE_ATTR_GROUP_ID
4) SAI_ACL_TABLE_ATTR_USED_ACL_ENTRY_LIST - Number of used entries in the ACL table. Its defined as list to account number of used entries in all pipes corresponding to the same ACL table.
5) SAI_ACL_TABLE_ATTR_AVAILABLE_ACL_ENTRY_LIST - Number of free entries in the ACL table. Its defined as list to account number of free entries in all pipes corresponding to the same ACL table.

Added following ACL SLICE related attributes,
1) SAI_ACL_SLICE_ATTR_SLICE_ID - Slice number of the ACL hardware slice.
2) SAI_ACL_SLICE_ATTR_SLICE_PIPE_ID - Pipe number in which this ACL hardware slice is present.
3) SAI_ACL_SLICE_ATTR_SLICE_STAGE - Stage (INGRESS/EGRESS).
4) SAI_ACL_SLICE_ATTR_ACL_TABLE_LIST - Object ID of the ACL table present in this hardware slice.
5) SAI_ACL_SLICE_ATTR_USED_ACL_ENTRY - Number of entries used in the current hardware slice.
6)SAI_ACL_SLICE_ATTR_AVAILABLE_ACL_ENTRY - Number of free entry space available in the current hardware slice.

Added following switch attribute,
1) SAI_SWITCH_ATTR_ACL_SLICE_LIST - Object ID of all hardware ACL slices present.

1) SAI_OBJECT_TYPE_ACL_SLICE - Added new object for ACL hardware slices.

### DIFF
--- a/experimental/saiaclextensions.h
+++ b/experimental/saiaclextensions.h
@@ -1,0 +1,162 @@
+/**
+ * Copyright (c) 2014 Microsoft Open Technologies, Inc.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *    not use this file except in compliance with the License. You may obtain
+ *    a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    THIS CODE IS PROVIDED ON AN *AS IS* BASIS, WITHOUT WARRANTIES OR
+ *    CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT
+ *    LIMITATION ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS
+ *    FOR A PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
+ *
+ *    See the Apache Version 2.0 License for specific language governing
+ *    permissions and limitations under the License.
+ *
+ *    Microsoft would like to thank the following companies for their review and
+ *    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
+ *    Dell Products, L.P., Facebook, Inc., Marvell International Ltd.
+ *
+ * @file    saiaclextensions.h
+ *
+ * @brief   This module defines SAI ACL extentions interface
+ */
+
+#if !defined (__SAIACLEXTENSIONS_H_)
+#define __SAIACLEXTENSIONS_H_
+
+#include <saiacl.h>
+
+/**
+ * @defgroup SAIACL EXTENTIONS SAI - ACL specific API definitions
+ *
+ * @{
+ */
+
+/**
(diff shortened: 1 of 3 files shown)


kcudnik:
Missing sai_acl_slice_api_t declaration

PugalendranDell:
@itaibaz - Group of ports which shares the same ACL hw resources is named as pipe. It is indexed through pipe_id. 

For example, 
Assume there is 10 ports present. When you add ACL entry for first 5 ports it will go into pipe_0 TCAM, if you add ACL entry for second 5 ports it will go into pipe_1 TCAM.

PugalendranDell:
@itaibaz 
Assume there is 2 pipe present in the system and "APP_X" has added 10 entries for pipe_0 ports and 100 entries for pipe_1 ports and "APP_X" is mapped to 512 sized hw TCAM.

In case of SAI_ACL_TABLE_ATTR_SIZE=0, 
SAI_ACL_TABLE_ATTR_USED_ACL_ENTRY_LIST - This will return list of attributes, list_0 will have 10 and list_1 will have 100.

SAI_ACL_TABLE_ATTR_AVAILABLE_ACL_ENTRY_LIST - This will return list of attributes, list_0 will have 502 and list_1 will have 412.

=================

Assume there is 2 pipe present in the system and "APP_X" has added 100 entries which is common for all ports in the system and "APP_X" is mapped to 512 sized hw TCAM.

In case of SAI_ACL_TABLE_ATTR_SIZE=0, 
SAI_ACL_TABLE_ATTR_USED_ACL_ENTRY_LIST - This will return list of attributes, list_0 will have 100 and list_1 will have 100.

SAI_ACL_TABLE_ATTR_AVAILABLE_ACL_ENTRY_LIST - This will return list of attributes, list_0 will have 412 and list_1 will have 412.

If SAI_ACL_TABLE_ATTR_SIZE = 200, then AVAILABLE will be calculated as (200-USED)

itaibaz:
Thanks for the explanation
Yet, I'm not clear about pipe meaning, and think it will have to be explained in the SAI pipeline object model
You mention "Group of ports which shares the same ACL hw resources is named as pipe"
Why just a group of ports ?
We have several binding point types. Aren't the other binding point types (For example, a group of vlans) also a pipe ?
A table can be bounded to multiple binding points, so will the pipe be the multiplication of the groups ?

Until this concept is well defined in SAI object model, I think making any change to original sai ACL file isn't good

PR #857, opened 2018-08-12
[flexsai]: Make code compatible with python3

marian-pritsak:


hesingh:
@lguohan:  I filed a new issue yesterday.  Please see

https://github.com/opencomputeproject/SAI/issues/896

I think I may have issues with Python setup on my Ubuntu 18.04 I am using to build the flexsai backend.  I do have python 2.7 installed.  Then I have also installed python3.  However sai.cpp fails to compile for me.  the failure logs are included in the issue above.  Any idea what's wrong?

PR #866, opened 2018-08-28
Revert "Added SAI_OBJECT_TYPE_BRIDGE_PORT support in ACL (#756)"

stcheng:
This reverts commit e472cc22654c388e56d749464bde7462bc9d8bda.


stcheng:
Revert the commit that adds BRIDGE_PORT type on all match port attributes.

atitjain:
can you please explain this change. Maybe not in all cases but in some cases it makes sense to have bridgeport as part of match/action.

stcheng:
thanks

PR #867, opened 2018-08-29
Set the analyzer port as invalid when SPAN type is remote

jemifdo:
For both local and enhanced remote SPAN the mirrored packet is always unicasted (i.e., sent via a single port). Remote or enhanced remote SPAN in NPU perspective, is just adding the appropriate encapsulation (VLAN or GRE tunnel) to the data packet and sent out via the analyzer port.
But the RSPAN implementation by most of Network OS, normally assigns a special VLAN in which source learning is disabled and the packets to the analyzer is always flooded within the VLAN (i.e., Tx to all VLAN members). In order to achieve this behavior we now have to create multiple mirroring sessions, one for each member in the VLAN by adding the VLAN member as the analyzer port.
Instead we can consider the VLAN ID parameter alone and enable the packet to be mirrored to all its members and thereby making the analyzer port as invalid for remote SPAN.


itaibaz:
One mirror session can have multiple monitor port in it, just set per each port you want to monitor SAI_PORT_ATTR_INGRESS_MIRROR_SESSION / SAI_PORT_ATTR_EGRESS_MIRROR_SESSION to the session

In RSPAN, the selected traffic is transmitted on a monitoring VLAN ((pushing a VLAN, SAI_MIRROR_SESSION_ATTR_VLAN_ID, to the packets).
It is usually assumed that the network will flood all packets on the mirroring VLAN in a way that it will arrive to the analyzer (and the mirroring VLAN is only used for mirroring traffic).
Still an analyzer port (SAI_MIRROR_SESSION_ATTR_MONITOR_PORT, the name might be a bit confusing) is needed to define the egress port, and is relevant for RSPAN, so it can't be removed

Can you explain the motivation, Thanks

jemifdo:
One mirror session can have multiple monitor port only as per NPU capability but as per SAI definition the monitor port is a single object of type either Port or LAG.
Consider a case I need to have an RSPAN VLAN with two ports P1 and P2 and I want to mirror packets ingressing and egressing out port P3 to the RSPAN VLAN.
Now I need to create two RSPAN mirror sessions one with port P1 as monitor port and other with P2 as monitor port and then attach these two mirror sessions to the P3 port object as objlist and at both ingress and egress.
Two avoid this multiple mirror session create, we can make the monitor port as invalid only for RSPAN and use the VLAN ID thereby considering all its members as monitor ports for the mirror session or, we can make the monitor port attribute as type objlist. I proposed the former solution.

itaibaz:
While I understand your use case, I would say, perhaps, please leave the monitor port as valid for RSPAN
And consider the following usage. If monitor port is filled (not null), then it will be used. If monitor port is not filled, mirrored packet will be flooded on the vlan.
In such case, the needed change, is to allownull for monitor port

stcheng:
So the port ID is set to null only when we are configuring the RSPAN source switch? When it is the destination switch, do we still need the port ID?

jemifdo:
Monitor port object value can be NULL only when the mirroring type is RSPAN. But for RSPAN the monitor port can also be a valid one. I assume @itaibaz suggested the same in the above comment, so that the new change doesn't affect any application that uses the current SAI behavior and also at the same time we can flood the mirror packets to all the members of RSPAN VLAN by setting the monitor port object as NULL.
@stcheng sorry, I couldn't get what you mean by source switch and destination switch.

stcheng:
I think for RSPAN to be set up, there will be separate source switch and destination switch?

jemifdo:
@stcheng hope you mean the mirror source and mirror destination. If so, then here monitor port refers to mirror destination.

jemifdo:
@xinliu-seattle @lguohan I've addressed the comment raised by @rsivasankar during the OCP discussions. Is this pull request ready to be merged?

jemifdo:
@kcudnik the compilation fails because the META check doesn't support attribute value type "sai_object_list_t" when it is "MANDATORY_ON_CREATE".

kcudnik:
I think that was just in case condition, since so far we didn't had any mandatory_on_create attribute that will take object_list, let me add check for that

kcudnik:
test this please

kcudnik:
test this please

PR #869, opened 2018-09-06
SAI Get Stats function counter_id data-type

SathishKumarTR-Dell:
This commit is to use a common data type for all the SAI Stat Get APIs (sai_get_xxxx_stats_ext_fn) uniformly. Updated just the SAI Port Stat Get API to discuss this first in the community. 
Will update the same for other SAI objects after the discussion.

### DIFF
--- a/inc/saibfd.h
+++ b/inc/saibfd.h
@@ -510,7 +510,7 @@ typedef sai_status_t (*sai_get_bfd_session_attribute_fn)(
 typedef sai_status_t (*sai_get_bfd_session_stats_fn)(
         _In_ sai_object_id_t bfd_session_id,
         _In_ uint32_t number_of_counters,
-        _In_ const sai_bfd_session_stat_t *counter_ids,
+        _In_ const sai_stat_id_t *counter_ids,
         _Out_ uint64_t *counters);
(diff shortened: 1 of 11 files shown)


itaibaz:
How will sai_stat_id_t look ?
Will it be a big enum containing all the list of every counter for every functionality ?
Where will the definition sit ?

itaibaz:
Or did you mean to cast the specific functionality counter to uint ?

SathishKumarTR-Dell:
@itaibaz Yes, sai_stat_id_t is the data-type defined to uint32_t similar to the the sai_attr_id_t type used for attribute id. API implementation is just made uniform with a single data-type for all objects.

kcudnik:
i missed this commit, i i would have some objectsions, to this change, since having each separate enum would enforce user on compile time if he passed wrong stats id in the enum, and if we have unified this to stat id, we need to cast each enum to that tape

SathishKumarTR-Dell:
@kcudnik stat id was unified to a common type in the Stats API signature just like the attribute id type in the set API. It is to help generate any meta data and/or common code for handling all stats get APIs.

PR #870, opened 2018-09-06
added vxlan tunnel as possible condition for underlay interface attri…

YonatanPitz:
Added vxlan tunnel to SAI_TUNNEL_ATTR_UNDERLAY_INTERFACE condition.
When creating VxLAN tunnel, one needs to specify to RIF in which routing happens after encap.


inc/saitunnel.h
```diff
@@ -451,7 +451,7 @@ typedef enum _sai_tunnel_attr_t
      * @type sai_object_id_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      * @objects SAI_OBJECT_TYPE_ROUTER_INTERFACE
-     * @condition SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_GRE
+     * @condition SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_GRE or SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_VXLAN
```
lguohan: I have question here, what if we are using triditional l2 vxlan? then there is no underlay interface, but we can assume in that case we still need to supply underlay interface, but it will not bu used?

atitjain: How to account for VRF in the termination/decap?

lguohan:
after decap, it is the overlay router, so it is not this underlay
interface. for vrf after decap, we are using the tunnel mapper to map vxlan
into vrf.

Atit Jain <notifications@github.com> 于2018年9月7日周五 上午1:39写道：

karthik-krishnamurthy:
I guess if the tunnel mapper to map vnid to vrf is not present, we could use the tunnel mapper for vnid to bridge to obtain the bridge and use bridge RIF associated with the bridge the to get the VRF context

inc/saitunnel.h
YonatanPitz: I think that even in traditional l2 vxlan, after encapsulation, the packet is than routed in L3 router, and thus still needs an underlay interface to determine underlay vrf.
No?

lguohan:
@YonatanPitz, you are right.

YonatanPitz <notifications@github.com> 于2018年9月7日周五 上午9:39写道：

inc/saitunnel.h
jcmathews: Is HEAD END REPLICATION supported for VXLAN, Any SAI attributes are defined for them?

PR #878, opened 2018-10-16
Add RDMA related matching fileds to ACL

hzheng5:
- OpCode field in Base Transport Header (BTH)
- Syndrome field in The ACK Extended Transport Header (AETH)

ref: https://folk.uio.no/erikbra/hovedfag/companion/javadoc/no/uio/ifi/gruns/infiniband/IBAPacket.html


lguohan:
meta data check failed.

kcudnik:
you could also add those acronyms to acronyms.txt file and that should also work

hzheng5:
> you could also add those acronyms to acronyms.txt file and that should also work

Thanks kcudnik for the tip. Will try it next time. :)

PR #879, opened 2018-10-18
Change SAI_MIRROR_SESSION_ATTR_VLAN_HEADER_VALID to CREATE_AND_SET

stcheng:
This attribute can be modified because the destination could be
updated to either within a VLAN or not.

Signed-off-by: Shu0T1an ChenG <shuche@microsoft.com>

### DIFF
--- a/inc/saimirror.h
+++ b/inc/saimirror.h
@@ -162,9 +162,10 @@ typedef enum _sai_mirror_session_attr_t
      * Valid for RSPAN or ERSPAN with valid Vlan header.
      *
      * @type sai_uint16_t
-     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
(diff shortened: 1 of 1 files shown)


stcheng:
Commit message has been updated:
```
    Change SAI_MIRROR_SESSION_ATTR_VLAN_HEADER_VALID to CREATE_AND_SET

    This attribute can be modified because the destination could be
    updated to either within a VLAN or not.

    Because SAI_MIRROR_SESSION_ATTR_VLAN_ID is no longer set as
    mandatory on create, user needs to set this VLAN ID specifically
    when creating an RSPAN session. Otherwise, the default value 0
    will be used.

    Signed-off-by: Shu0T1an ChenG <shuche@microsoft.com>
```

PR #880, opened 2018-10-20
Add SAI_SWITCH_ATTR_CURRENT_TEMP to get average temperature from sensors

padmanarayana:
Retrieve the average of sensors value readings from the ASIC.

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -399,6 +399,24 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_OPER_STATUS,
 
     /**
+     * @brief Maximum number of temperature sensors available.
+     *
+     * @type sai_uint8_t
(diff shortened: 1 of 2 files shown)


inc/saiswitch.h
```diff
@@ -409,6 +409,17 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_MAX_TEMP,
 
+    /**
+     * @brief The average value of the temperature
+     * retrieved from the switch sensors
```
kcudnik: does this mean from all available sensors? not sure if description is good enough, if you have 1 sensor with 10 degree, and 2nd with 100 then average is 55 which maybe ok but system is defenetly not ok at 100

kcudnik: updating description should be fine

PR #884, opened 2018-10-29
Force stat functions to be consistent across all SAI

kcudnik:


kcudnik:
@VitalyVovnoboy  please review this change, since i didn't know whether we should have "tam_histogram_stat" changed to tam_microburst" (as in this change is, since stats name suggests that) or should we just change _sai_tam_microburst_stat_t to _sai_tam_histogram_stat_t

lguohan:
it looks like the previous name is not correct, it should be microburst stat

inc/saiuburst.h
```diff
@@ -380,23 +380,57 @@ typedef sai_status_t (*sai_get_tam_histogram_attribute_fn)(
...
+ */
+typedef sai_status_t (*sai_get_tam_microburst_stats_ext_fn)(
+        _In_ sai_object_id_t tam_microburst_id,
         _In_ uint32_t number_of_counters,
         _In_ const sai_stat_id_t *counter_ids,
+        _In_ sai_stats_mode_t mode,
         _Out_ uint64_t *counters);
 
+/**
+ * @brief Clear port pool statistics counters.
```
karthik-krishnamurthy: Typo in brief. I guess it should clear TAM microburst stats

kcudnik: fixed

PR #890, opened 2018-11-16
[warm boot] Define SAI_SWITCH_ATTR_PRE_SHUTDOWN for warm shutdown

yxieca:
Signed-off-by: Ying Xie <ying.xie@microsoft.com>

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -1706,6 +1706,25 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_UNINIT_DATA_PLANE_ON_REMOVAL,
 
     /**
+     * @brief Instruct SAI to execute switch pre-shutdown
+     *
+     * Indicates controlled switch pre-shutdown as first step of warm shutdown.
+     * This hint is optional, SAI application could skip this step and
+     * go directly to warm shutdown.
+     * This hint should be ignored, if at the time SAI receives this hint,
+     * SAI_SWITCH_ATTR_RESTART_WARM is NOT already set to TRUE.
+     * The scope of pre-shutdown is to backup SAI/SDK data, but leave CPU port
+     * active for some final control plane traffic to go out.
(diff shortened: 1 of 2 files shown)


kcudnik:
actually you could add "pre" and "shutdown" as 2 words instead of introducing new one and it would match attribute name "PRE_SHUTDOWN"

yxieca:
@kcudnik I thought about that. I am a bit worried about 'pre' becomes a more widely allowed word than we wanted. But I am open to your suggestion. Let's hear some more opinions before deciding.

kcudnik:
great

inc/saiswitch.h
```diff
@@ -1705,6 +1705,25 @@ typedef enum _sai_switch_attr_t
...
 
+    /**
+     * @brief Instruct SAI to execute switch preshutdown
+     *
+     * Indicates controlled switch preshutdown as first step of warm shutdown.
+     * This hint is optional, SAI application could skip this step and
+     * go directly to warm shutdown.
+     * This hint should be ignored, if at the time SAI receives this hint,
+     * SAI_SWITCH_ATTR_RESTART_WARM is NOT already set to TRUE.
+     * The scope of preshutdown is to backup SAI/SDK data, but leave CPU port
```
SathishKumarTR-Dell: CPU port should get disabled as part of the detach sequence in remove_switch() API. So, Isnt the CPU packets will be processed until the remove_switch() is called in the current model already? 
Is this attribute meant to introduce an intermediate stage in the "planned restart" operation so that the backup process can be done? Please clarify.

yxieca: What you said is true. However, remove_switch() without pre-shutdown could take very long time. And we don't known what is the moment when cpu port is shutdown. So we couldn't time ourselves to send out these 'last' messages. With 2 stage shutdown, we can precisely control when these packet will be sent out, and knowing that the remove_switch() will be a faster operation.

PR #894, opened 2018-11-23
Add check for params prefixed sai_

kcudnik:


inc/sai.h
```diff
@@ -201,14 +201,14 @@ sai_status_t sai_api_initialize(
...
- * @param[in] sai_api_id SAI API ID
+ * @param[in] api SAI API ID
  * @param[out] api_method_table Caller allocated method table. The table must
  * remain valid until the sai_api_uninitialize() is called.
  *
  * @return #SAI_STATUS_SUCCESS on success, failure status code on error
  */
 sai_status_t sai_api_query(
-        _In_ sai_api_t sai_api_id,
+        _In_ sai_api_t api,
```
stcheng: not api_id?

kcudnik: it's still api id, but in entire headers style is to name params after type so if type is sai_zzz_t param is named zzz,

PR #900, opened 2018-12-06
SAI Packet Action Group Proposal 

karthik-krishnamurthy:
The following pull request has the draft proposal for Packet Action Groups. The .md file contains the requirement and use-cases for this proposal.


doc/SAI-Packet-Action-Group.md
```diff
@@ -0,0 +1,707 @@
...
+  /**
+   * @brief Action Value based on the action type
+   */
+   union _action_value {
+     bool booldata;
+     sai_uint8_t u8;
+     sai_uint16_t u16;
+     sai_uint32_t u32;
+     sai_mac_t mac;
+     sai_object_id_t oid;
```
kcudnik: having oid here will be hard to implement since all list we assume are no oid members, and we have automated logic to track all oid dependencies but not in nested structs/unions of attr value
i would propose to came out with different approach for oid here

karthik-krishnamurthy: @kcudnik Sure once this is discussed in the call, we will think of a different way to solve this.

kcudnik: we have a lot of types already in sai_attribute_t and if possible lets keep to them using object attributes, and lets try no to introduce more complex structures

karthik-krishnamurthy: We have removed the oid from the structure and made it as an attribute. We want to have a list of packet modification actions in a single attribute and hence we added the structures sai_packet_action_group_action_t and sai_packet_action_group_action_list_t

kcudnik: looks good

doc/SAI-Packet-Action-Group.md
```diff
@@ -0,0 +1,729 @@
...
+     * SAI_PACKET_ACTION_GROUP_MEMBER_STATUS_INACTIVE and does
+     * not use it to forward traffic. If a next member exists (if any) whose monitored
+     * object in up , then the switching entity marks the failover status
+     * of the member as SAI_PACKET_ACTION_GROUP_MEMBER_STATUS_ACTIVE
+     * and it is used to forward traffic. Selection of the next active member is
+     * evaluated based on the order of creation of members.
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_LAG
```
atitjain: Can you use bridge port instead of LAG and PORT, that way it will be valid for Port, LAG, Sub port and tunnels.

doc/SAI-Packet-Action-Group.md
```diff
@@ -0,0 +1,729 @@
...
+     *
+     * This is the egress port/LAG interface out of which the packet needs to be
+     * sent out of the member. If this object is set as SAI_NULL_OBJECT_ID or
+     * is not specified. Then the packet needs to be dropped.
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_PACKET_ACTION_GROUP_MEMBER_ATTR_REDIRECT_INTERFACE,
```
atitjain: How does this impact the other attribute we have which support fast fail over natively

doc/SAI-Packet-Action-Group.md
```diff
@@ -0,0 +1,729 @@
...
+
+**In saitypes.h**
+1. Add new structure for packet action group actions
+2. Include the new structure in the sai_attribute_value_t structure
+
+```
+/**
+ * @brief Packet Action Group Action Type
+ */
+typedef enum _sai_packet_action_group_action_type_t
```
atitjain: Another idea could be to define a packet modification object. The attributes of the object could be the list fields to be enabled and their values. The OID can be used as action MODIFY. This gives a way to use the same set of modification for multiple groups and we can use it as direct ACL action too.

PR #920, opened 2019-02-27
Update SAI Port FEC modes 

SathishKumarTR-Dell:
Added new SAI Port RS-FEC modes that specifies the IEEE FEC Clauses for different port speeds and modulation. With this, application/user can configure new IEEE clause RS-FEC modes, CL119 and CL134,   relevant for the 50G PAM4 hardware lane speeds. Also, added specific CL108 and CL91 RS-FEC modes for the 25G NRZ hardware lane speeds.


inc/saiport.h
```diff
@@ -170,6 +170,18 @@ typedef enum _sai_port_fec_mode_t
 
     /** Enable FC-FEC - 10G, 25G, 40G, 50G ports */
     SAI_PORT_FEC_MODE_FC,
+
+    /** Enable RS-FEC - CL91/RS528 - 100G NRZ ports */
+    SAI_PORT_FEC_MODE_CL91_RS,
```
lguohan: what about the the original MODE_RS? what is the definition for that one?

SathishKumarTR-Dell: I have not modified the original MODE_RS for backward compatibility. Application can migrate to use the specific CLxx_RS FEC modes with the new changes.

lguohan: can you clarify the MODE_RS?

SathishKumarTR-Dell: CL91 was the baseline standard using RS(528,514) block size for 100G PHY types with four lanes of 25GbE NRZ links. 
CL108 was later added as the new clause also using RS(528,514) block size for single lane 25G PHY types specifically. Currently, the MODE_RS does not differentiate the CL91/CL108 clauses. 
We could do the below options:
1. Define the MODE_RS for CL91 RS-FEC and modify the comment to read /** Enable RS-FEC – CL91/RS528 -100G NRZ ports */.    (OR)
2. Define new MODE_CL91_RS and MODE_CL108_RS specific modes (as in the PR) and deprecate the MODE_RS in the current version. We can update the comment for MODE_RS to say it is "deprecated" and MODE_CLxxx_RS modes should be used.

Hope this clarifies. Please let me know which one of the two options you prefer.

lguohan: they are all RS, but for different lane/port speed, to simplify the user use, can we use MODE_RS and let the SAI to figure out the current speed and setup the RS mode accordingly?

for example, the port is 1x or 2x 25G NRZ, then use CL108, if the port is 4x25G NRZ, use CL91. if the port is 1x50G PAM4, use CL134 and if the port is 2x, 4x, 8x 50G NRZ, use CL119.

SathishKumarTR-Dell: Yes, the default configuration can be derived as in your example.
However, there are also some use-cases and HW capability that supports a different combination of FEC modes. 100GAUI-4 C2C, 100GAUI-4 C2M, 50GAUI-2 C2M, 50GAUI-2 C2C standard support 2 lanes of 26.5625GBaud and NRZ modulation, but with the CL134_RS544 FEC mode. Likewise, 51.5625GBaud and PAM4 modulation lanes that can support CL91_RS528 FEC mode.  
So, it might to useful to open up the FEC modes supported in HW to user in order to override the default configuration based on the link-partner capability, deployment etc?

meta/acronyms.txt
```diff
@@ -89,3 +89,5 @@ WWW - World Wide Web
 XG - (*,G) Any-source Multicast
 XOFF - Transmitter Off
 XON - Transmitter On
+NRZ - Non Return to Zero
+PAM4 - Pulse Amplitude Modulation
```
atitjain: I think we can just keep PAM here, PAM4 is more specific case.

inc/saiport.h
Avineus: Should SAI be aware of FEC clauses? 
Based on FEC mode, let the vendor specific code figure out the best Clauses based on the interface type.

inc/saiport.h
```diff
@@ -170,6 +170,18 @@ typedef enum _sai_port_fec_mode_t
...
+
+    /** Enable RS-FEC - CL108/RS528 - 25G, 50G NRZ ports */
+    SAI_PORT_FEC_MODE_CL108_RS,
+
+    /** Enable RS-FEC - CL134/RS544 - 50G PAM4 ports */
+    SAI_PORT_FEC_MODE_CL134_RS,
+
+    /** Enable RS-FEC - CL119/RS544 - 100G, 200G, 400G PAM4 ports */
+    SAI_PORT_FEC_MODE_CL119_RS,
```
Avineus: Can we just use RS or FC in SAI header file.

In vendor code, fec types/clauses can be addressed based on the interface type.
Other wise there would be lot of error conditions check need to be added on port speed and its fec combination mismatches

meta/acronyms.txt
SathishKumarTR-Dell: Updated the acronym. Added PAM4 as well to make the build meta check successful.

inc/saiport.h
SathishKumarTR-Dell: We can have the default FEC mode derived by SAI or Vendor SDK based on port speed/modulation for the typical cases. This pull request is for exposing the specific RS-FEC mode clauses to user and host adapter application to support any specific media/backward-compatibility use-cases and for debuggability. For instance, 200G PAM4 can operate with Clause 119 RS544_2xN FEC per IEEE 802.3bs as default or with Clause 91 RS544 FEC per IEEE 802.3cd. 
Likewise there seem to be use-cases for 100GAUI-4, 50GAUI-2 PHY types with NRZ lanes, operating with RS544 FEC mode. 
With that, I am proposing that we can call out the existing FEC_MODE_RS mode as the default mode derived by a vendor SAI implementation based on the port speed/modulation etc. The specific FEC modes proposed in the PR can be used to override the default mode for any specific use-case or for debug purpose after a link failure. I have updated the PR accordingly. 
@lguohan @Avineus  Please let me know your comments.

lguohan:
1. should be backward compatible.
2. get/set of a same SAI attribute should return the same value. if user set fec to rs and want to get a clause of rs, then it should be a different attribute like oper_fec.

SathishKumarTR-Dell:
Added a new attribute to return the port's operational FEC status. 
Existing implementation that set/get the SAI_PORT_FEC_MODE_RS value for SAI_PORT_ATTR_FEC_MODE should continue to work the same way if do not require the specific clause FEC modes. SAI implementation can return the specific clause FEC modes for the new operational FEC mode attribute.

mikeberesford:
I was proposing something similar to this in https://github.com/opencomputeproject/SAI/pull/1224 - can/should this PR be revived?

inc/saiport.h
rlhui: @lguohan  - what do you think for the above response?  There's a recent PR #1224  having similar idea to define more specific lanes/speeds.

rlhui:
> I was proposing something similar to this in #1224 - can/should this PR be revived?

@SathishKumarTR-Dell , @AshokDaparthi  - checking if you're still interested in discussing this? Thanks.

dgsudharsan:
Can we add SAI_PORT_FEC_MODE_INTERNAL to cover the case where we let SAI to chose the FEC mode based on autoneg or based on standards? The chosen FEC mode can be read from SAI_PORT_ATTR_OPER_FEC_MODE

SathishKumarTR-Dell:
Community revived the discussion on this topic and updated with a new API proposal under the PR: https://github.com/opencomputeproject/SAI/pull/1224
Hence, closing this one.

PR #921, opened 2019-02-28
Adding SAI PTF tests for Directed Broadcast

chitra-raghavan:
Adding L3 Directed Broadcast tests  into SAI PTF tests.

L3DirectedBroadcast - I
L3DirectedBroadcast - II


test/saithrift/tests/sail3.py
```diff
@@ -2025,3 +2025,145 @@ def runTest(self):
             attr_value = sai_thrift_attribute_value_t(u16=1)
             attr = sai_thrift_attribute_t(id=SAI_PORT_ATTR_PORT_VLAN_ID, value=attr_value)
             self.client.sai_thrift_set_port_attribute(port1, attr)
+
+@group('l3')
+class L3DirectedBroadcast (sai_base_test.ThriftInterfaceDataPlane):
```
atitjain: I don't think we have any related attributes in SAI. The Directed broadcast is typically enabled by configuration and its not there by default:

https://www.cisco.com/c/m/en_us/techdoc/dc/reference/cli/nxos/commands/l3/ip-directed-broadcast.html
https://www.juniper.net/documentation/en_US/junos/topics/topic-map/switches-interface-targeted.html#id-example-configuring-ip-directed-broadcast-on-a-switch

itaibaz: I agree. For directed broadcast to work, it should be configured by the application. 
For example, Application should add a route for the directed broadcast with certain nexthop ip. And add a neighbor for that ip with dst mac ff:ff:ff:ff:ff:ff

chitra-raghavan: The script has written in the same way as mentioned by @itaibaz .  Route has been created for broadcast ip with a neighbor entry having ff:ff:ff:ff:ff:ff as destination mac.

atitjain: I think we need to figure out 2 cases:
1. Transit boxes, where we just need to add the routes for subnet broad cast address with unicast nexthop. 
2. Penultimate box which connects to the target subnet, this is the one where we need to use one nh with broadcase MAC. 
@itaibaz can you please confirm.

itaibaz: Atit comment makes sense
On a transit switch, we can have the entire subnet configured for the route, with a unicast next hop
On the penultimate switch, we can configure directly reachable route with the entire subnet and the attached RIF as next hop. We then need to add a neighbor with this RIF, the direct broadcast IP, and mac address ff:ff:ff:ff:ff:ff.
Looks like current code does add this neighbor.
But doesn't add the directly reachable route.

chitra-raghavan: As per review comments ,  added steps to verify  traffic for  broadcast route in Transit switches via unicast nexthop. Testcase  splits into 2 parts 1. Transit  devices 2. Penumtimate devices

PR #924, opened 2019-03-04
Adding SAI PTF tests for IPv4/v6 ECMP Group member

chitra-raghavan:
Adding SAI PTF tests for IPv4/v6 ECMP Group member Tests

L3IPv4EcmpGroupMemberTest – I
L3IPv4EcmpGroupMemberTest – II
L3IPv4EcmpGroupMemberTest – III (Zero member)
L3IPv6EcmpGroupMemberTest – I
L3IPv6EcmpGroupMemberTest – II


test/saithrift/tests/sail3.py
```diff
@@ -2025,3 +2025,630 @@ def runTest(self):
...
+
+@group('l3')
+@group('ecmp')
+class L3IPv4EcmpGroupMemberTest(sai_base_test.ThriftInterfaceDataPlane):
+    def runTest(self):
+        '''
+        Description:
+        i) L3IPv4EcmpGroupMemberTest I:
+        Create a VRF with IPv4 and IPv6 enabled. Create three router interfaces in
+        the same VRF. Create a route (24 mask) with nhop and neighbor entry in two
```
atitjain: Here I think we need to rephrase. The RIFs are added to next-hops. Then we need to create a next-hop group and add next-hops to those groups, which

chitra-raghavan: rephrased as per comments

test/saithrift/tests/sail3.py
```diff
@@ -2025,3 +2025,632 @@ def runTest(self):
...
+            # Remove the nhop bound with router interface 1, and verify the traffic flows towards the ECMP route.
+            self.client.sai_thrift_remove_next_hop_group_member(nhop_gmember1)
+
+            count = [0, 0]
+            max_itrs = 101
+            itr_count = 0
+
+            print "sending %s packets to verify distribution on 2 ports after removing one ecmp group member" %(max_itrs - 1 )
+
+            for i in range(1, max_itrs):
```
prsunny: The for loop is repeated and looks like it is common code. Can you check if it can be put into one function to avoid duplication?

test/saithrift/tests/sail3.py
```diff
@@ -2025,3 +2025,632 @@ def runTest(self):
...
+                pkt = simple_tcp_packet(
+                                eth_dst=router_mac,
+                                eth_src=src_mac,
+                                ip_dst=ip_dst,
+                                ip_src=ip_src,
+                                tcp_sport=sport,
+                                tcp_dport=dport,
+                                ip_id=106,
+                                ip_ttl=64)
+                exp_pkt1 = simple_tcp_packet(
```
prsunny: exp_pkt1/2/3 is not relevant in this case. Can you remove this?

test/saithrift/tests/sail3.py
```diff
@@ -2025,3 +2025,632 @@ def runTest(self):
...
+            nhop_gmember3 = sai_thrift_create_next_hop_group_member(self.client, nhop_group1, nhop3)
+            sai_thrift_create_route(self.client, vr1, addr_family, ip_addr4_subnet, ip_mask1, nhop_group1)
+
+            var_gmember3 = True
+            count = [0, 0, 0]
+            max_itrs = 101
+
+            print "sending %s packets to verify distribution on 3 ports after adding one more ecmp group member" %(max_itrs - 1)
+
+            for i in range(1, max_itrs):
```
prsunny: Same as above for V4

test/saithrift/tests/sail3.py
chitra-raghavan: Addressed comments

test/saithrift/tests/sail3.py
chitra-raghavan: Addressed comments

test/saithrift/tests/sail3.py
chitra-raghavan: Addressed comments

PR #925, opened 2019-03-05
 Adding SAI PTF tests for LAG ACL Table/Group and bindings

chitra-raghavan:
Adding SAI PTF tests for LAG Acl.

L3AclTableTest – I (LAG Binding)
L3AclTableGroupTest – I (LAG Binding)

### DIFF
--- a/test/saithrift/tests/saiacl.py
+++ b/test/saithrift/tests/saiacl.py
@@ -1134,3 +1134,503 @@ class BindAclTableInGroupTest(sai_base_test.ThriftInterfaceDataPlane):
             self.client.sai_thrift_remove_router_interface(rif_id3)
             self.client.sai_thrift_remove_router_interface(rif_id4)
             self.client.sai_thrift_remove_virtual_router(vr_id)
+
+@group('acl')
+class L3AclTableTestI(sai_base_test.ThriftInterfaceDataPlane):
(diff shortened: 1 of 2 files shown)


test/saithrift/tests/saiacl.py
```diff
@@ -1134,3 +1134,503 @@ def runTest(self):
...
+          8. Create next hop to reach the neighbor entry.
+          9. Create route entry with /24 mask (i.e 11.11.11.0/24) through NHop1
+          10. Send IPv4 packet from port 1 (LAG member) with destination IP 11.11.11.1
+
+        Part1:
+          11. Create ACL table group "acl_table_group_id"
+          12. Create two ACL table to block the traffic based on Dst IP address (11.11.11.1) and Src IP address (10.10.10.1).
+          13. Associate the ACL table with the group members.
+          14. Bind the ACL table and entry with LAG Id1 with attribute SAI_LAG_ATTR_INGRESS_ACL
+          15. Send the IP packet again through port 1 and ensure the packet is dropped.
```
prsunny: May be in next iteration, we can consider sending two different packets, one matching source IP acl entry and another one matching destination IP acl entry and ensure both are dropped. In this test, since the same packet match both the entries, the parallel action part is not really verified.

PR #929, opened 2019-03-19
New flood vector combined type

itaibaz:
Adding clarification to the meaning of all flood vector - includes the router for broadcast and multicast, doesn't include the router for unknown unicast.
Add a new type, combined, to easily add L2MC group on top of the existing all definition

### DIFF
--- a/inc/saibridge.h
+++ b/inc/saibridge.h
@@ -412,7 +412,8 @@ typedef enum _sai_bridge_type_t
 typedef enum _sai_bridge_flood_control_type_t
 {
     /**
-     * Flood on all sub-ports
+     * @brief Flood on all sub-ports
+     *
      * When setting sub-ports to broadcast or unknown multicast flood, it also
      * includes flooding to the router. When setting sub-ports to unknown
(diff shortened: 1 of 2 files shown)


inc/saibridge.h
```diff
@@ -420,6 +426,14 @@ typedef enum _sai_bridge_flood_control_type_t
     /** Flood on the L2MC group */
     SAI_BRIDGE_FLOOD_CONTROL_TYPE_L2MC_GROUP,
 
+    /**
+     * @brief Flood on all sub-ports and L2MC group
```
atitjain: What is the use case? Is the idea to create a bunch of snooped interfaces with few statically added for bcast/mcast?

itaibaz: Using this API extension, you can easily add flooding to router (by adding a bridge port type router) or to a tunnel, on top of the regular behavior of flooding to all local ports, without needing to manage the list of local ports, by choosing the combined flood type.
A sample use case for flooding to router, is VXLAN L3 peering use case, demoed in OCP, where unknown unicast traffic is sent to router.

PR #936, opened 2019-04-04
Hostif Generic Netlink interface

padmanarayana:
A netdev tunnel channel is proposed where an application can create a virtual tunnel interface (over a loopback) to receive trapped packets/buffers from the SAI driver.  This can be used as an alternative to callback/FD based mechanisms.


inc/saihostif.h
```diff
@@ -875,10 +875,90 @@ typedef enum _sai_hostif_table_entry_channel_type_t
...
+    SAI_HOSTIF_TABLE_ENTRY_CHANNEL_TYPE_NETDEV_TUNNEL
 
 } sai_hostif_table_entry_channel_type_t;
 
+/**
+ * @brief Attribute data for SAI_HOSTIF_TABLE_ENTRY_ATTR_CHANNEL_ENCAP_TYPE
+ */
+typedef enum _sai_hostif_channel_encap_type_t
+{
+    /** Receive packets via GRE/ERSPAN Type 2*/
```
atitjain: Can we just call it ```SAI_HOSTIF_TABLE_CHANNEL_ENCAP_TYPE_GRE``` and add one for IPinIP, then it covers all tunnels:

    SAI_TUNNEL_TYPE_IPINIP,
    SAI_TUNNEL_TYPE_IPINIP_GRE,
    SAI_TUNNEL_TYPE_VXLAN,

inc/saihostif.h
```diff
@@ -875,10 +875,90 @@ typedef enum _sai_hostif_table_entry_channel_type_t
...
+     * @brief End of Attributes
+     */
+    SAI_HOSTIF_TUNNEL_ATTR_END,
+
+    /** Custom range base value */
+    SAI_HOSTIF_TUNNEL_ATTR_CUSTOM_RANGE_START = 0x10000000,
+
+    /** End of custom range base */
+    SAI_HOSTIF_TUNNEL_ATTR_CUSTOM_RANGE_END
+} sai_hostif_tunnel_attr_t;
```
atitjain: Where is this used?

atitjain:
> A netdev tunnel channel is proposed where an application can create a virtual tunnel interface (over a loopback) to receive trapped packets/buffers from the SAI driver. This can be used as an alternative to callback/FD based mechanisms.

In practice, the tunnel is created by using one of the loopback interfaces only and in that case, it's used to terminate the data traffic itself. So I couldn't quite get the use case you are trying.

kcudnik:
couple issues to fix:

WARNING: coment is ending without space saihostif.h 890:    /** Receive packets via GRE/ERSPAN Type 2*/

WARNING: @brief should start with capital letter: saihostif.h 899: * @brief hostif tunnel attributes

WARNING: @brief should start with capital letter: saihostif.h 942:     * @brief meta header valid

WARNING: Word 'hostif' is misspelled saihostif.h 899: * @brief hostif tunnel attributes

WARNING: Value SAI_HOSTIF_TABLE_CHANNEL_ENCAP_TYPE_ERSPAN_TYPE_II of sai_hostif_channel_encap_type_t is not prefixed as SAI_HOSTIF_CHANNEL_ENCAP_TYPE_

WARNING: Value SAI_HOSTIF_TABLE_CHANNEL_ENCAP_TYPE_UDP of sai_hostif_channel_encap_type_t is not prefixed as SAI_HOSTIF_CHANNEL_ENCAP_TYPE_

ERROR: enum sai_hostif_table_entry_channel_encap_type_t has no typedef enum sai_hostif_table_entry_channel_encap_type_t

WARNING: Value SAI_HOSTIF_TABLE_CHANNEL_ENCAP_TYPE_ERSPAN_TYPE_II of sai_hostif_channel_encap_type_t is not prefixed as SAI_HOSTIF_CHANNEL_ENCAP_TYPE_

WARNING: Value SAI_HOSTIF_TABLE_CHANNEL_ENCAP_TYPE_UDP of sai_hostif_channel_encap_type_t is not prefixed as SAI_HOSTIF_CHANNEL_ENCAP_TYPE_

lguohan:
I think we need example and scope in the documents.

kcudnik:
you added sai_hostif_tunnel_attr_t structure attributes but there is missing SAI_OBJECT_TYPE_HOSTIF_TUNNEL object type in sai_object_type_t

inc/saihostif.h
padmanarayana: This is a tunnel to lift packets from the SAI driver to applications. We do not plan to leverage the SAI_TUNNEL infra for this..

inc/saihostif.h
padmanarayana: The primary use case for this will be sFlow.

kcudnik:
you introduced new object type, but there are not api's to create/remove/set/get for this object type in sai_hostif_api_t, you need to add 4 new api's:
 
create_hostif_tunnel,
remove_hostif_tunnel
set_hostif_tunnel_attribute
get_hostif_tunnel_attribute

you can also try building metadata localy, just go to SAI/meta directory and type "make" to see if everything compiles

inc/saihostif.h
atitjain: Why not leverage the sFlow proposal for this. In this looks like you will do the sFlow from the CPU path? I think better is to just enable sFlow, like we do from Port, from tunnel id itself?

marian-pritsak:
Did you consider using the existing psample infrastructure in Linux as an alternative to protobuf? It allows you to carry metadata along with the packet and is already a part of the mainline kernel.

inc/saihostif.h
padmanarayana: I removed the use of tunnel since it's being misunderstood with the tunnel infra..

sflow-rt:
+1 for using psample netlink channel to deliver packet samples. 

psample is a generic (subsystem independent) method of delivering packet samples and associated metadata. It is also extensible so additional metadata can easily be added in future (queue depth, queuing delay, etc.) 

psample is supported by the open source [Host sFlow](https://sflow.net/) agent.

inc/saihostif.h
```diff
@@ -736,9 +739,11 @@ typedef enum _sai_hostif_attr_t
      * The maximum number of characters for the name is SAI_HOSTIF_NAME_SIZE - 1 since
      * it needs the terminating null byte ('\0') at the end.
      *
+     * If Hostif is a generic netlink, this indicates the generic netlink family name.
+     *
      * @type char
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
-     * @condition SAI_HOSTIF_ATTR_TYPE == SAI_HOSTIF_TYPE_NETDEV
+     * @condition SAI_HOSTIF_ATTR_TYPE == SAI_HOSTIF_TYPE_NETDEV or SAI_HOSTIF_ATTR_TYPE == SAI_HOSTIF_TYPE_GENETLINK
```
marian-pritsak: Is it always equal to "NETLINK_GENERIC"?

padmanarayana: This is set to the generic netlink family name whose multicast port id (SAI_HOSTIF_ATTR_GENETLINK_PORT_ID) is used to lift the samples. If the psample driver is used, it should be set to "psample" (refer https://github.com/torvalds/linux/blob/master/include/uapi/linux/psample.h#L33 ).

PR #937, opened 2019-04-08
SAI NAT Proposal

JaiOCP:
SAI NAT Proposal


lguohan:
as a general comments, we need a markdown file to describe the scope of the proposal and allow example code to demonstrate how to use the APIs

lguohan:
we also need to define the NAT module in the SAI behavior model/pipeline. Where the NAT module is in the pipeline.

JaiOCP:
Spec is not yet complete. Its an early commit for others to start looking at it.

JaiOCP:
Ack.

inc/saiacl.h
```diff
@@ -2148,6 +2151,16 @@ typedef enum _sai_acl_entry_attr_t
      */
     SAI_ACL_ENTRY_ATTR_ACTION_DTEL_REPORT_ALL_PACKETS,
 
+    /**
+     * @brief NAT Pool ID
+     *
+     * @type sai_acl_action_data_t sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_NAT
+     * @default disabled
```
atitjain: May be you want to call it NULL and not disabled as its an OID

inc/saiacl.h
```diff
@@ -231,6 +231,9 @@ typedef enum _sai_acl_action_type_t
     /** Enable DTEL report for all packets without filtering (experimental) */
     SAI_ACL_ACTION_TYPE_DTEL_REPORT_ALL_PACKETS,
 
+    /** Enable NAT */
+    SAI_ACL_ACTION_TYPE_NAT,
+
```
atitjain: Is the only way to achieve NAT as per this proposal is to apply through ACL action?

JaiOCP: Please review the updated diffs

inc/saiacl.h
JaiOCP: Please review updated diffs

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,297 @@
...
+sai_nat_entry_attr_list[12].value.u16 = 6000;
+
+sai_nat_entry_attr_list[13].id = SAI_NAT_ENTRY_ATTR_TO _PORT;
+sai_nat_entry_attr_list[13].value.u16 = 1026;
+
+sai_nat_entry_attr_list[14].id = SAI_NAT_ENTRY_ATTR_IP_PROTOCOL;
+sai_nat_entry_attr_list[14].value.u8 = 17; 
+
+nat_entry_attr_count = 15;
+sai_create_nat_entry_fn(
```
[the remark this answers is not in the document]
JaiOCP: Its upto the caller. If bulking is needed and all entries have same masks then there is no need to create 3 entry objects. If the entries do not have same attributes then 3 separate objects MUST be created.

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,297 @@
...
+Following per zone statistics is provided by the hardware for NAT feature.
+- DNAT_DISACARDS/SNAT_DISCARDS – If Packet is not TCP/UDP and/or is a fragmentated IP packet. 
+- DNAT_TRANSLATION_NEEDED/SNAT_TRANSLATION_NEEDED – If there is NAT table lookup miss for TCP/UDP packets, then this counter is incremented.
+- DNAT_TRANSLATIONS/SNAT_TRANSLATIONS – If NAT table lookup is a hit, then this counter is incremented.
+
+Counters can be attached to per NAT entry to gather per flow statistics. Per NAT entry stat is enabled in the entry attributes.
+
+> Per zone counters are created using sai_create_nat_counter_fn() API.
+
+NAT counters can be extracted using TAM streaming interface or by using TAM GET API.
```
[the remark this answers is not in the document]
JaiOCP: During the TAM review, there was a consensus that there should be a GET API to get the data and same data can be streamed as well. 
From SONiC implementation perspective this becomes very simple single API invocation.

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,297 @@
...
+HITBIT table data is retrieved using the TAM proposed generic GET API. 
+Advantage of using TAM GET API is that normalized data for the hitibit data can also be streamed to local NOS, or to an external collector or to multi chip system.
+Some hardware may not perform "clear on read" or NOS may want to control the operation "clear on read". For this reason a separate boolean attribute "CLEAR_ON_READ" is specified 
+
+TAM object is bound to NAT object for stats and hitbit table.
+
+Following is a simple GET API invocation for pulling NAT entry creates under source and destination nat entry objects. One or more objects can be specified in a single invocation. API is name 
+
+
+> sai_tam_telemetry_get_data(switch_id, objlist, clear_on_read, buffer_size, buffer);
```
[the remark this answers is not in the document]
JaiOCP: Adding a per entry processing using the flag is too expensive in terms of latency and convergence of NAT. There is already a significant delay in hw update because there is no bulking performed. There is a conscious effort to provide way to bulk updates in SAI API. 
Degenerate case though not suggested can still be achieved by create per entry objects

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,297 @@
...
+sai_nat_entry_attr_list[12].value.u16 = 1026;
+
+sai_nat_entry_attr_list[13].id = SAI_NAT_ENTRY_ATTR_TO _PORT;
+sai_nat_entry_attr_list[13].value.u16 = 6000;
+
+sai_nat_entry_attr_list[14].id = SAI_NAT_ENTRY_ATTR_IP_PROTOCOL;
+sai_nat_entry_attr_list[14].value.u8 = 17; 
+
+nat_entry_attr_count = 15;
+sai_create_nat_entry_fn(
```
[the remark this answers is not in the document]
JaiOCP: Same explanation as above

JaiOCP:
> Can you please also add NAT in the pipeline object model
> https://github.com/opencomputeproject/SAI/tree/master/doc/object-model
> 
> In addition, it seems there is no ACL based NAT with updated proposal ?

Yes, I will be updating the behavioral model soon. 
There is no ACL based NAT in this proposal. NAT entries can be mapped to ACL if there is such a need in the SAI driver.

doc/NAT/SAI-NAT-API.md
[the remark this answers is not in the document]
itaibaz: You are providing the same attribute ID several times in single create call function with different values. I am not familiar with any other API that allows that in SAI

There is bulk create definition in SAI
typedef sai_status_t (*sai_bulk_object_create_fn)(
        _In_ uint32_t object_count,
        _In_ const uint32_t *attr_count,
        _In_ const sai_attribute_t **attr_list,
        _In_ sai_bulk_op_error_mode_t mode,
        _Out_ sai_object_id_t *object_id,
        _Out_ sai_status_t *object_statuses);

If you want to allow bulk creation, I think we should have standard create and a bulk create
But using standard create as bulk create isn't something which is supported, IMO

doc/NAT/SAI-NAT-API.md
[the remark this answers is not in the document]
itaibaz: My comment was not referring to the TAM get API
I was referring to the fact we should be able to query the statistics as any other statistics in SAI, in a standard get_<objects>_stats_fn

doc/NAT/SAI-NAT-API.md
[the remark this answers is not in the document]
itaibaz: There is already a bulk API get
sai_status_t sai_bulk_get_attribute(
        _In_ sai_object_type_t object_type,
        _In_ uint32_t object_count,
        _In_ const sai_object_key_t *object_key,
        _Inout_ uint32_t *attr_count,
        _Inout_ sai_attribute_t **attr_list,
        _Inout_ sai_status_t *object_statuses);
which supports optimized get operation
So we only need to add read only activity attribute, and then we can use the bulk get to efficiently read the activity for multiple entries

doc/NAT/SAI-NAT-API.md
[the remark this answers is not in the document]
JaiOCP: This is not a usual case of attribute list. Its 1:1 mapping for NAPT.
Other way we can do is create each entry as an object then have sai_object_list_t. But this will create an explosion of objects.
Do you have a suggestion?

doc/NAT/SAI-NAT-API.md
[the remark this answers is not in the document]
JaiOCP: Yes, there was a consensus then instead of having a proliferation of get_stats API and then NOS parsing output of each API, can we have simpler API design hence the usage of GET API.
Think of it as an IOCTL for stats

doc/NAT/SAI-NAT-API.md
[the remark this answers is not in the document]
itaibaz: I think either to create each entry as an object. And I agree, that the downside will be many objects.
The other option is similar to qos maps sai_qos_map_list_t, to use a list of structs. The downside is you can't use strong parameter checking by the meta data.
I think the first option is the prefered way, and where most of the later specs have went (for example, tunnel maps which are composed of multiple entries each are an object)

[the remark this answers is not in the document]
JaiOCP: Lets discuss this in community.
Given the scale of NAT entries will >20K, 60k< , creating that many objects is prohibitively expensive.

inc/saitam.h
```diff
@@ -1560,17 +1560,17 @@ typedef void (*sai_tam_event_notification_fn)(
  * @brief TAM telemetry data get API
  *
  * @param[in] switch_id SAI Switch object id
- * @param[in] attr_count Number of attributes
- * @param[in] attr_list Array of attributes
+ * @param[in] obj_list SAI Switch object list
```
atitjain: This API is part of sai v1.4, there are compatibility implications with that version if the signature is changed.  Could you please check.

inc/sainat.h
```diff
@@ -0,0 +1,476 @@
...
+
+    /**
+     * @brief Start of Attributes
+     */
+    SAI_NAT_ENTRY_ATTR_START,
+
+    /**
+     * @brief NAT IP address translated from
+     *
+     * @type sai_ip_address_t
```
mickeyspiegel: During the call it was stated that the from and to IP addresses could be masked. Looking in saitypes.h, I don't see a mask on sai_ip_address_t. Did you mean to use sai_ip_prefix_t instead?

inc/saiswitch.h
```diff
@@ -1757,6 +1757,20 @@ typedef enum _sai_switch_attr_t
...
+     * Bind (or unbind) the NAT object.
+     * SAI_NULL_OBJECT_ID in the attribute value.
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_NAT
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_SWITCH_ATTR_NAT_OBJECT_ID,
```
mickeyspiegel: This needs to be a list to cover multiple pairs of NAT zones

hongrun007:
Hi, all,
Is it necessary for us to support NAT for IPv6

hesingh:
At first glance, since IPv6 has enough address space, including ULA, IPv6 has no need for NAT.  Also NAT is discouraged by the IETF due to many issues.  However, NAT has its uses, including NAT66.  Please read second paragraph in section 3 of https://tools.ietf.org/html/draft-mrw-behave-nat66-01

inc/sainat.h
```diff
@@ -0,0 +1,653 @@
...
+
+    /**
+     * @brief Mask for source IPv4 address in packet.
+     *
+     * @type sai_ip4_t
+     * @flags CREATE_AND_SET
+     * @default 0.0.0.0
+     * @validonly SAI_NAT_ENTRY_ATTR_NAT_TYPE == SAI_NAT_TYPE_SOURCE_NAT or SAI_NAT_ENTRY_ATTR_NAT_TYPE == SAI_NAT_TYPE_DOUBLE_NAT
+     */
+    SAI_NAT_ENTRY_ATTR_SRC_IP_MASK,
```
marian-pritsak: Why do we have a mask for setting IP port if the match in the entry is exact?

inc/sainat.h
```diff
@@ -0,0 +1,653 @@
...
+    SAI_NAT_ENTRY_ATTR_L4_DST_PORT,
+
+    /**
+     * @brief Enable/disable packet count
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_NAT_ENTRY_ATTR_ENABLE_PACKET_COUNT,
```
marian-pritsak: Use the generic counter for NAT entries

inc/sainat.h
```diff
@@ -0,0 +1,653 @@
...
+     * @brief IP layer 4 source port
+     */
+    sai_uint16_t l4_src_port;
+
+    /**
+     * @brief IP layer 4 destination port
+     */
+    sai_uint16_t l4_dst_port;
+
+} sai_nat_entry_key_t;
```
marian-pritsak: This entry is a 5-tuple. SNAT and DNAT use 3-tuples and there is no validation here.

inc/sainat.h
```diff
@@ -0,0 +1,653 @@
...
+     * @brief IP layer 4 source port mask
+     */
+    sai_uint16_t l4_src_port;
+
+    /**
+     * @brief IP layer 4 destination port mask
+     */
+    sai_uint16_t l4_dst_port;
+
+} sai_nat_entry_mask_t;
```
marian-pritsak: SONiC proposal states that the NAT entries are an exact match and not a ternary.

inc/sainat.h
```diff
@@ -0,0 +1,653 @@
...
+    SAI_NAT_ENTRY_ATTR_NAT_TYPE  = SAI_NAT_ENTRY_ATTR_START,
+
+    /**
+     * @brief NAT Inside Zone ID
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_NAT_ENTRY_ATTR_FROM_ZONE_ID,
```
marian-pritsak: What is the purpose of the zones in NAT entry?
According to the pipeline, we don't distinguish between the specific zones, but only when IN_ZONE!=OUT_ZONE, where IN_ZONE is an attribute of IRIF, and the OUT_ZONE - of ERIF.

inc/sainat.h
```diff
@@ -0,0 +1,653 @@
...
+        _In_ const sai_nat_entry_t *nat_entry,
+        _In_ const uint32_t *attr_count,
+        _Inout_ sai_attribute_t **attr_list,
+        _In_ sai_bulk_op_error_mode_t mode,
+        _Out_ sai_status_t *object_statuses);
+
+/**
+ * @brief NAT zone counters for each NAT type
+ */
+typedef enum _sai_nat_zone_counter_attr_t
```
marian-pritsak: Use generic counter

inc/sainat.h
```diff
@@ -0,0 +1,653 @@
...
+    sai_bulk_create_nat_entry_fn              create_nat_entries;
+    sai_bulk_remove_nat_entry_fn              remove_nat_entries;
+    sai_bulk_set_nat_entry_attribute_fn       set_nat_entries_attribute;
+    sai_bulk_get_nat_entry_attribute_fn       get_nat_entries_attribute;
+
+    sai_create_nat_zone_counter_fn            create_nat_zone_counter;
+    sai_remove_nat_zone_counter_fn            remove_nat_zone_counter;
+    sai_set_nat_zone_counter_attribute_fn     set_nat_zone_counter_attribute;
+    sai_get_nat_zone_counter_attribute_fn     get_nat_zone_counter_attribute;
+} sai_nat_api_t;
```
marian-pritsak: How to create the zone itself?

marian-pritsak:
How do we configure DNAT Pool?

marian-pritsak:
Need to add a new binding points - PRE_NAT and POST_NAT (before DNAT and after SNAT)

inc/saiswitch.h
```diff
@@ -788,6 +788,30 @@ typedef enum _sai_switch_attr_t
...
+     */
+    SAI_SWITCH_ATTR_AVAILABLE_DNAT_ENTRY,
+
+    /**
+     * @brief Available Double NAT entries
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_AVAILABLE_DOBULE_NAT_ENTRY,
```
marian-pritsak: SAI already has a general sai_object_type_get_availability function that allows you to achieve the same CRM functionality

inc/saitam.h
```diff
@@ -2038,17 +2038,17 @@ typedef void (*sai_tam_event_notification_fn)(
...
  * @param[out] buffer Data buffer
  *
  * @return #SAI_STATUS_SUCCESS on success, failure status code on error
  */
 sai_status_t sai_tam_telemetry_get_data(
         _In_ sai_object_id_t switch_id,
-        _In_ uint32_t attr_count,
-        _In_ const sai_attribute_t *attr_list,
+        _In_ sai_object_list_t obj_list,
+        _In_ bool clear_on_read,
```
marian-pritsak: How are the TAM changes related to this NAT proposal?

meta/acronyms.txt
```diff
@@ -94,6 +96,7 @@ UDP - User Datagram Protocol
 VNI - Virtual Network Interface
 VNID - Virtual Network Identifier
 VR - Virtual Router
+VRF - Virtual Route Forwarding
```
marian-pritsak: I guess it's "Virtual Routing and Forwarding"

inc/saitam.h
JaiOCP: Stale Changes

meta/acronyms.txt
JaiOCP: Merge Diff. Not related to NAT changes

inc/saitam.h
JaiOCP: Merge Diffs. Not related to NAT changes

inc/saiswitch.h
JaiOCP: Review it with Gouhan. My preference is to do it in separate clean up PR given the scope of changes and its impact on schedule

inc/sainat.h
JaiOCP: Zone is a namespace for operator and is configured as a string in schema. NAT application maps it to a uint_8 and provides this in SAI API.
Another good suggestion by Mike L. is to use say_nat_type_t and because of the impact of changes we are not considering this as well.

inc/sainat.h
JaiOCP: generic counter is just added and do not have infra support in SAI or SONiC. We will consider this in next feature set.

inc/sainat.h
JaiOCP: This we have discussed and closed in SAI pipeline

inc/sainat.h
JaiOCP: Yes, for NAT entries they are exact match. Mask is defined to mask out an individual key. Ranges etc can be created for the DO NOT NAT condition

inc/sainat.h
JaiOCP: condition check has been added. Please look at the latest updates.
  /**
     * @brief Mask for source IPv4 address in packet.
     *
     * @type sai_ip4_t
     * @flags CREATE_AND_SET
     * @default 0.0.0.0
     * @validonly SAI_NAT_ENTRY_ATTR_NAT_TYPE == SAI_NAT_TYPE_SOURCE_NAT or SAI_NAT_ENTRY_ATTR_NAT_TYPE == SAI_NAT_TYPE_DOUBLE_NAT
     */
    SAI_NAT_ENTRY_ATTR_SRC_IP_MASK,

inc/sainat.h
JaiOCP: There is SONiC or SAI implementation for generic counter infra. For this feature given the timelines plan is to with legacy approach.

inc/sainat.h
JaiOCP: This is for the entire field mask, though ranges can be configured for do not nat use case

inc/saiswitch.h
JaiOCP: Stale Changes

inc/sainat.h
JaiOCP: Stale Changes

inc/saiswitch.h
ritalhui: @lguohan , Hi Guohan, please comment here?  Thanks.

rck-innovium:
> > Is the pipeline file here the latest version reflecting all changes discussed so far?
> 
> Latest submitted file has the changes

Should ingress security ACLs that match on say srcIP or dstIP addresses match on the translated addresses or before NAT?

rck-innovium:
> > > Is the pipeline file here the latest version reflecting all changes discussed so far?
> > 
> > 
> > Latest submitted file has the changes
> 
> Should ingress security ACLs that match on say srcIP or dstIP addresses match on the translated addresses or before NAT?

In the pipeline, Policy-Based-Forwarding ACLs happen after DNAT/(SNAT+DNAT). However, SNAT happens after all ingress ACLs. There seems to be some inconsistency w.r.t ACLs. 
1) For packets that are either DNAT or double-NAT, ingress ACL lookup happens on translated DIP, SIP. 
2) For packets that are just SNAT, ingress ACL lookup happens before NAT translation.
So the behavior for ACLs mating on SIP is different for  double-NAT vs SNAT.

JaiOCP:
> > > Is the pipeline file here the latest version reflecting all changes discussed so far?
> > 
> > 
> > Latest submitted file has the changes
> 
> Should ingress security ACLs that match on say srcIP or dstIP addresses match on the translated addresses or before NAT?

Ingress ACL rules should be applied consistently for a given translation domain/zone. This gets tricky in NAT as the reverse traffic SIP/DIP will appear in the outer zone i.e. translated address. This gets even more tricky with dynamic xlations. Having said the pipeline is structured in way where ACL can be applied on original SIP/DIP and/or xlated SIP/DIP

JaiOCP:
> > > > Is the pipeline file here the latest version reflecting all changes discussed so far?
> > > 
> > > 
> > > Latest submitted file has the changes
> > 
> > 
> > Should ingress security ACLs that match on say srcIP or dstIP addresses match on the translated addresses or before NAT?
> 
> In the pipeline, Policy-Based-Forwarding ACLs happen after DNAT/(SNAT+DNAT). However, SNAT happens after all ingress ACLs. There seems to be some inconsistency w.r.t ACLs.
> 
> 1. For packets that are either DNAT or double-NAT, ingress ACL lookup happens on translated DIP, SIP.
> 2. For packets that are just SNAT, ingress ACL lookup happens before NAT translation.
>    So the behavior for ACLs mating on SIP is different for  double-NAT vs SNAT.

There are two ACL blocks, one pre NAT and other post NAT. Post NAT block also is responsible for any PBR like functions.
Post NAT block is not reflected for SNAT, I will add it in pipeline. Good catch.

mickeyspiegel:
Marian made a comment towards the end of the meeting yesterday morning about the representation of the NAT miss counters in the pipeline. Looking into this further, I assume that these are just the normal actions and trap counters. As described in the pipeline, the SAI_HOSTIF_TRAP_TYPE_NAT_HAIRPIN, SAI_HOSTIF_TRAP_TYPE_DNAT_MISS, and SAI_HOSTIF_TRAP_TYPE_SNAT_MISS traps may be triggered. Counters will be associated with those traps only if SAI_HOSTIF_TRAP_ATTR_COUNTER_ID is set (at creation time or later) in the corresponding hostif_trap object. The action associated with the trap (e.g. drop or trap) is specified by SAI_HOSTIF_TRAP_ATTR_PACKET_ACTION. Is this correct?

What Marian was objecting to was the representation of the counters in the pipeline diagram. Regardless of whether the action associated with the trap is drop or trap, it is the same counter. By calling it "DNAT_MISS counter" and "DNAT_MISS_DROP counter", it implies that there are two counters for DNAT_MISS traps, when in fact there is only one. We should either just remove all the stuff about "disable" and "enable" traps and trap counters since it is just normal trap logic, or adjust the terminology to something like "trap_packet_action == drop" and "trap_packet_action == trap" and move the counter up a block to the trap itself.

JaiOCP:
> Marian made a comment towards the end of the meeting yesterday morning about the representation of the NAT miss counters in the pipeline. Looking into this further, I assume that these are just the normal actions and trap counters. As described in the pipeline, the SAI_HOSTIF_TRAP_TYPE_NAT_HAIRPIN, SAI_HOSTIF_TRAP_TYPE_DNAT_MISS, and SAI_HOSTIF_TRAP_TYPE_SNAT_MISS traps may be triggered. Counters will be associated with those traps only if SAI_HOSTIF_TRAP_ATTR_COUNTER_ID is set (at creation time or later) in the corresponding hostif_trap object. The action associated with the trap (e.g. drop or trap) is specified by SAI_HOSTIF_TRAP_ATTR_PACKET_ACTION. Is this correct?
> 
> What Marian was objecting to was the representation of the counters in the pipeline diagram. Regardless of whether the action associated with the trap is drop or trap, it is the same counter. By calling it "DNAT_MISS counter" and "DNAT_MISS_DROP counter", it implies that there are two counters for DNAT_MISS traps, when in fact there is only one. We should either just remove all the stuff about "disable" and "enable" traps and trap counters since it is just normal trap logic, or adjust the terminology to something like "trap_packet_action == drop" and "trap_packet_action == trap" and move the counter up a block to the trap itself.

Thanks Mickey,
Yes this didn't got captured in the meeting minutes.
@marian-pritsak - We will do as Mickey is suggesting. Let us know if you still have question

JaiOCP:
Mickey,

 

Minor clarification.

Functionally traps are enabled/disabled and trap action corresponding to it will be punt/drop. Point about being single counter attached is correct. If the action is drop then counter maintains the drop counter else it is the punt (or trap) count. I will update the language as suggested.

 

-Jai

ritalhui:
> Is the pipeline file here the latest version reflecting all changes discussed so far?
> 
> Latest submitted file has the changes
> 
> Should ingress security ACLs that match on say srcIP or dstIP addresses match on the translated addresses or before NAT?
> 
> In the pipeline, Policy-Based-Forwarding ACLs happen after DNAT/(SNAT+DNAT). However, SNAT happens after all ingress ACLs. There seems to be some inconsistency w.r.t ACLs.
> 
> For packets that are either DNAT or double-NAT, ingress ACL lookup happens on translated DIP, SIP.
> For packets that are just SNAT, ingress ACL lookup happens before NAT translation.
> So the behavior for ACLs mating on SIP is different for  double-NAT vs SNAT.
> 
> 
> There are two ACL blocks, one pre NAT and other post NAT. Post NAT block also is responsible for any PBR like functions.
> Post NAT block is not reflected for SNAT, I will add it in pipeline. Good catch.

Hi Jai, we should also ensure ACL can be applied on original source address for double NAT case.  If we look at the "ingress router flow" at pipeline_v6.pdf page 4, there's an ACL block in between VRF table block and router table block.  The same ACL block can be reused as for no_nat table.  This ACL block can also be applied for regular ACL on original addresses (can include orig DIP too.)

meta/acronyms.txt
lguohan: @JaiOCP , better to do rebase instead of merge so that other commits won't show up.

meta/saisanitycheck.c
```diff
@@ -623,6 +623,7 @@ void check_attr_object_type_provided(
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_IPV4:
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_IPV6:
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_IP_ADDRESS:
+        case SAI_ATTR_VALUE_TYPE_IPV4:
```
kcudnik: move it to line 602, this set only contains ACL_ACTION entries

meta/parse.pl
```diff
@@ -1247,6 +1247,10 @@ sub ProcessDefaultValue
     {
         WriteSource "$val = { .chardata = { 0 } };";
     }
+    elsif ($default =~ /^0\.0\.0\.0$/ and $type =~ /^(sai_ip4_t)/)
```
kcudnik: why use explicit ipv4 instead of sai_ip_address_t? which is already supported on line 1236

meta/saisanitycheck.c
JaiOCP: Kamil, This is caused by accidental merge

meta/parse.pl
JaiOCP: Accidental merge. I will be doing rebase

JaiOCP:
> Thanks a lot Jai. I think we should still have a diagram to show how the new NAT block starts/integrates with existing pipeline. E.g. what is the previous block that'd point to "ingress zone table"? After NAT block is added, router flow on page 4 in pipeline_v6.pdf is no longer the latest. The page (and perhaps few other router flow pages) need to be updated to indicate at which point that the "global_nat_enable" flag is checked leading to a detour to an additional NAT processing.

Rita, Yes there is a preNAT block between VRF and router block and then a post NAT ACL block. This takes care of any pre NAT ACL (SNAT/DNAT/Double NAT or NoNAT) and as well post NAT after translations

inc/sainat.h
```diff
@@ -0,0 +1,653 @@
...
+     */
+    SAI_NAT_ENTRY_ATTR_START,
+
+    /**
+     * @brief NAT Type defined in sai_nat_type_t
+     * @type sai_nat_type_t
+     * @flags CREATE_AND_SET
+     * @default SAI_NAT_TYPE_NONE
+     */
+    SAI_NAT_ENTRY_ATTR_NAT_TYPE  = SAI_NAT_ENTRY_ATTR_START,
```
rck-innovium: If we are not able to accommodate Marian's proposal of different APIs, the next best thing would be to move the SAI_NAT_ENTRY_ATTR_NAT_TYPE to be part of sai_nat_entry_key_t.
This allows for easier documentation.

typedef struct _sai_nat_entry_key_t
	{
	sai_nat_type_t nat_type;
	    /**
	     * @brief IPv4 source address
>> valid only if nat_type is DoubleNAT or SNAT
	     */
	    sai_ip4_t src_ip;
	
	    /**
	     * @brief IPv4 destination address
>> valid only if nat_type is DNAT or DoubleNAT
	     */
	    sai_ip4_t dst_ip;
	:
        :

Further, SAI_NAT_ENTRY_ATTR_NAT_TYPE  is marked as CREATE_AND_SET -- modifying this at run-time actually means moving the entry from one logical block to another. It looks better to mark it as create_only.

JaiOCP: Ravi, 
As we discussed in the community meeting, moving NAT type to NAT entry key is correct from the pipeline perspective. NAT type is an attribute which is a result of the match.

Regarding CREATE_AND_SET, thats a good suggestion. I will change it CREATE_ONLY. This does not look for a good candidate for add/modify so the proposed change makes sense

JaiOCP: Ravi,

I just realized why I marked this as CREATE_AND_SET. metadata checker allows validly check only with flags CREATE_AND_SET and not CREATE_ONLY. Sorry I have revert back this change

kcudnik:
retest this please

inc/sainat.h
```diff
@@ -0,0 +1,660 @@
...
+    SAI_NAT_ENTRY_ATTR_HIT_BIT_COR,
+
+    /**
+     * @brief Per NAT entry hit bit state
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_NAT_ENTRY_ATTR_HIT_BIT,
```
mickeyspiegel: What is the difference between SAI_NAT_ENTRY_ATTR_HIT_BIT and SAI_NAT_ENTRY_ATTR_HIT_BIT_COR?

inc/sainat.h
```diff
@@ -0,0 +1,660 @@
...
+    SAI_NAT_ENTRY_ATTR_FROM_ZONE_ID,
+
+    /**
+     * @brief NAT Outside Zone ID
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_NAT_ENTRY_ATTR_TO_ZONE_ID,
```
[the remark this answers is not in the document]
JaiOCP: Fixed. Removed ZONE attributes and added VRF attribute

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+attr_count = 1;
+
+memset(&dnat_pool_entry, 0, sizeof(nat_etnry));
+
+dnat_pool_entry.data.key.dst_ip = 65.55.42.1;
+dnat_pool_entry.data.mask.dst_ip = 0xffffffff;
+dnat_pool_entry.data.key.l4_dst_port = 1024;
+dnat_pool_entry.data.mask.l4_dst_port = 0xffff;
+dnat_pool_entry.data.key.proto = 17;
+dnat_pool_entry.data.mask.proto = 0xff;
```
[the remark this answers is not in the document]
JaiOCP: yes dnat pool entry just needs vrf and dip as keys. Cleaned to the doc

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+nat_entry_attr[6].value.bool = true;
+
+attr_count = 7;
+
+memset(&subnet_nat_entry, 0, sizeof(nat_entry));
+
+subnet_nat_entry.data.key.dst_ip = 200.0.0.0;
+subnet_nat_entry.data.mask.dst_ip = 0xffffff00;
+
+create_nat_entry(&subnet_nat_entry, attr_count, nat_entry_attr);
```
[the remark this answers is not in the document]
JaiOCP: Done

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+
+memset(&dbl_nat_etnry, 0, sizeof(nat_etnry));
+
+dbl_nat_entry.data.key.src_ip = 200.200.200.1;
+dbl_nat_entry.data.mask.src_ip = 0xffffffff;
+dbl_nat_entry.data.key.dst_ip = 172.16.1.100;
+dbl_nat_entry.data.mask.dst_ip = 0xffffffff;
+
+create_nat_entry(&dbl_nat_entry, attr_count, nat_entry_attr);
+
```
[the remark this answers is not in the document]
JaiOCP: Done

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+A new boolean field is added to sai_switch_attr in saiswitch.h
+> “SAI_SWITCH_ATTR_NAT_ENABLE”
+> set_switch_attribute() API is used to enable/disable NAT feature.
+
+## 4.0 Enable Traps for SNAT and DNAT Miss Packets
+Following two traps are added to hostif. Hostif driver MUST enable these traps for SNAT and DNAT miss in hw for receiving the packets. 
+SNAT/DNAT miss packets are received on a regular netdev channel to the application.
+
+- “SAI_HOSTIF_TRAP_TYPE_SNAT_MISS”
+- “SAI_HOSTIF_TRAP_TYPE_DNAT_MISS”
```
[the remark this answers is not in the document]
JaiOCP: Done

inc/saiswitch.h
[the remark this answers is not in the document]
JaiOCP: Done. Thanks for catching it.

JaiOCP:
> Three major comments:
> 
> 1. Why are FROM_ZONE and TO_ZONE attributes included in sai_nat_entry_attr? I thought we agreed to remove them. The pipeline does not describe what an implementation is expected to do with these attributes.
> 2. The pipeline says that the DNAT table sets "vrf" in addition to "dip" and "dport", but there is no attribute specifying the virtual router to use after DNAT translation. How is a switch supposed to know which "vrf" to use for the subsequent router table lookup?
> 3. The pipeline says that the DNAT pool prefix lookup is only based on "vrf" and "dip", but SAI-NAT-API.md has an example that specifies l4_dst_port and proto. These should be removed.
> 
> The other comments are minor.

1. Removed FROM_ZONE and TO_ZONE from sai_nat_entry_attr.
2. Added SAI_NAT_ENTRY_ATTR_VR_ID attribute only for dnat or double nat entry type.
3. Cleaned up the spec

inc/sainat.h
JaiOCP: NOS may perform read on various counters and HIT_BIT value for each nat entry. Along with it may choose to perform on clear on read using COR attribute.

Explain in doc in following section.
## Get Hit Bit, NAT Zone and Per NAT Entry Stats
Besides per NAT entry and per Zone counters there is hit bit value as. Hit bit is used to age the dynamic entries installed in HW.
In this proposal aging is performed by the NOS. NOS retrieves the hitbit table periodically at a configured time aging interface. If the hitbit is not SET for a given entry, the entry is removed from the NAT table using the sai_remove_entry_fn() API.

Per NAT entry stats and hit bit is retrieved by “get_nat_entry_attributes()”.
##### Step 1: Read SNAT entry hit bit and packet/byte count:
```sh
sai_attribute_t nat_entry_attr[10];

nat_entry_attr[0].id = SAI_NAT_ENTRY_ATTR_HIT_BIT;

nat_entry_attr[1].id = SAI_NAT_ENTRY_ATTR_BYTE_COUNT;

nat_entry_attr[2].id = SAI_NAT_ENTRY_ATTR_PACKET_COUNT;

attr_count = 3;

File: doc/NAT/SAI-NAT-API.md [hunk 29]
```diff
@@ -50,10 +51,11 @@ A new boolean field is added to sai_switch_attr in saiswitch.h
 
 ## 4.0 Enable Traps for SNAT and DNAT Miss Packets
 Following two traps are added to hostif. Hostif driver MUST enable these traps for SNAT and DNAT miss in hw for receiving the packets. 
-SNAT/DNAT miss packets are received on a regular netdev channel to the application.
+SNAT/DNAT/HARIPIN miss packets are received on a regular netdev channel to the application.
```
mickeyspiegel: New typo: HARIPIN --> HAIRPIN

doc/NAT/SAI-NAT-API.md
```diff
@@ -50,10 +51,11 @@ A new boolean field is added to sai_switch_attr in saiswitch.h
 
 ## 4.0 Enable Traps for SNAT and DNAT Miss Packets
 Following two traps are added to hostif. Hostif driver MUST enable these traps for SNAT and DNAT miss in hw for receiving the packets.
```
mickeyspiegel: two --> three

inc/sainat.h
```diff
@@ -114,6 +96,20 @@ typedef enum _sai_nat_entry_attr_t
      */
     SAI_NAT_ENTRY_ATTR_SRC_IP_MASK,
 
+    /**
+     * @brief Replace destination IPv4 address in packet.
```
mickeyspiegel: Replace virtual router associated with the IPv4 addresses in the packet.

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+nat_entry_attr[0].value = SAI_NAT_TYPE_SOURCE_NAT;
+
+nat_entry_attr[1].id = SAI_NAT_ENTRY_ATTR_SRC_IP;
+nat_entry_attr[1].value.u32 = 65.55.42.1; //example string
+
+nat_entry_attr[2].id = SAI_NAT_ENTRY_ATTR_L4_SRC_PORT;
+nat_entry_attr[2].value.u16 = 1024;
+
+nat_entry_attr[3].id = SAI_NAT_ENTRY_ATTR_TO_ZONE;
+nat_entry_attr[3].value.u32 = 200;
```
mickeyspiegel: This attribute no longer exists, should be deleted

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+
+nat_entry_attr[4].id = SAI_NAT_ENTRY_ATTR_FROM_ZONE;
+nat_entry_attr[4].value.u32 = 100;
```
mickeyspiegel: This attribute no longer exists, should be deleted

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+nat_entry_attr[0].value = SAI_NAT_TYPE_DESTINATION_NAT;
+
+nat_entry_attr[1].id = SAI_NAT_ENTRY_ATTR_DST_IP;
+nat_entry_attr[1].value.u32 = 10.0.0.1; //example string
+
+nat_entry_attr[2].id = SAI_NAT_ENTRY_ATTR_L4_DST_PORT;
+nat_entry_attr[2].value.u16 = 6000;
+
+nat_entry_attr[3].id = SAI_NAT_ENTRY_ATTR_TO_ZONE;
+nat_entry_attr[3].value.u32 = 100;
```
mickeyspiegel: This attribute no longer exists, should be deleted

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+
+nat_entry_attr[4].id = SAI_NAT_ENTRY_ATTR_FROM_ZONE;
+nat_entry_attr[4].value.u32 =200;
```
mickeyspiegel: This attribute no longer exists, should be deleted

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+nat_entry_attr[0].value = SAI_NAT_TYPE_DOUBLE_NAT;
+
+nat_entry_attr[1].id = SAI_NAT_ENTRY_ATTR_SRC_IP;
+nat_entry_attr[1].value.u32 = 138.76.28.1; //example string
+
+nat_entry_attr[2].id = SAI_NAT_ENTRY_ATTR_DST_IP;
+nat_entry_attr[2].value.u32 = 200.200.200.100; //example string
+
+nat_entry_attr[3].id = SAI_NAT_ENTRY_ATTR_TO_ZONE;
+nat_entry_attr[3].value.u32 = 200;
```
mickeyspiegel: This attribute no longer exists, should be deleted

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+
+nat_entry_attr[4].id = SAI_NAT_ENTRY_ATTR_FROM_ZONE;
+nat_entry_attr[4].value.u32 =100;
```
mickeyspiegel: This attribute no longer exists, should be deleted

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+nat_entry_attr[1].id = SAI_NAT_ENTRY_ATTR_DST_IP;
+nat_entry_attr[1].value.u32 = 128.17.18.0; //example string
+
+nat_entry_attr[2].id = SAI_NAT_ENTRY_ATTR_DST_IP_MASK;
+nat_entry_attr[2].value.u32 = 0xffffff00;
+
+nat_entry_attr[3].id = SAI_NAT_ENTRY_ATTR_TO_ZONE;
+nat_entry_attr[3].value.u32 = 200;
+nat_entry_attr[4].id = SAI_NAT_ENTRY_ATTR_FROM_ZONE;
+nat_entry_attr[4].value.u32 =100;
```
mickeyspiegel: These attributes no longer exist, should be deleted

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+
+## 6.0 NAT Exceptions
+NAT exceptions can be defined using the NAT entry. In that case they will be placed in the SNAT/DNAT table. Operator MUST consider defining the exceptions so as not to collide with other NAT entries.
+An override NAT exception can be created using the ACL rule. A new ACL action is defined in saiacl.h
+
+> SAI_ACL_ENTRY_ATTR_ACTION_NO_NAT
+
+In SAI pipeline if there is a ACL match with action NO_NAT and there is also a match in SNAT/DNAT table then ACL match result always takes precedence.
+
+### 6.1 NAT exception using NAT Entry
```
mickeyspiegel: The details in this section seem incorrect. It is not a NAT entry that creates the exceptions, it is an ACL entry. There is no SAI_NAT_TYPE_NO_NAT.

doc/NAT/SAI-NAT-API.md
```diff
@@ -0,0 +1,443 @@
...
+nat_entry_attr[1].id = SAI_NAT_ENTRY_ATTR_DST_IP;
+nat_entry_attr[1].value.u32 = 128.17.18.0; //example string
+
+nat_entry_attr[2].id = SAI_NAT_ENTRY_ATTR_DST_IP_MASK;
+nat_entry_attr[2].value.u32 = 0xffffff00;
+
+nat_entry_attr[3].id = SAI_NAT_ENTRY_ATTR_TO_ZONE;
+nat_entry_attr[3].value.u32 = 200;
+nat_entry_attr[4].id = SAI_NAT_ENTRY_ATTR_FROM_ZONE;
+nat_entry_attr[4].value.u32 =100;
```
mickeyspiegel: These attributes no longer exist, should be deleted

doc/NAT/SAI-NAT-API.md
JaiOCP: Done

doc/NAT/SAI-NAT-API.md
JaiOCP: done

doc/NAT/SAI-NAT-API.md
JaiOCP: removed

doc/NAT/SAI-NAT-API.md
JaiOCP: removed

doc/NAT/SAI-NAT-API.md
JaiOCP: removed

doc/NAT/SAI-NAT-API.md
JaiOCP: removed

doc/NAT/SAI-NAT-API.md
JaiOCP: remvoed

doc/NAT/SAI-NAT-API.md
JaiOCP: removed

doc/NAT/SAI-NAT-API.md
JaiOCP: removed

doc/NAT/SAI-NAT-API.md
JaiOCP: removed

doc/NAT/SAI-NAT-API.md
JaiOCP: Left over from old API. Removed section 6.1

inc/sainat.h
JaiOCP: done

lguohan:
@JaiOCP , thanks for the efforts for developing this nat spec. can you also write a document to explain the pipeline? when I look at the pipeline graph along, still have many questions. It will be good if you can have a document to describe the pipeline.

JaiOCP:
> @JaiOCP , thanks for the efforts for developing this nat spec. can you also write a document to explain the pipeline? when I look at the pipeline graph along, still have many questions. It will be good if you can have a document to describe the pipeline.

@lguohan Committed a document explaining the pipeline.

PR #941, opened 2019-04-11
[bmtor]: Add META_TUNNEL table

marian-pritsak:


marian-pritsak:
L3_VXLAN table entry is iused instead of fdb in cases, where the overlay destination MAC is global for all VTEPs

atitjain:
> L3_VXLAN table entry is iused instead of fdb in cases, where the overlay destination MAC is global for all VTEPs

What is being achieved here, is it for routing the vxlan tunnel packet out of the gateway instead of switching it?

PR #942, opened 2019-04-15
SAI Proposal generic resource monitoring

marian-pritsak:


doc/SAI-Proposal-Generic-Resource-Monitoring.md
```diff
@@ -0,0 +1,51 @@
+Switch Abstraction Interface Change Proposal
```
atitjain: Why cant we just use following attributes from ```saiswitch.h``` 
```

    /**
     * @brief Available IPv4 routes
     *
     * @type sai_uint32_t
     * @flags READ_ONLY
     */
    SAI_SWITCH_ATTR_AVAILABLE_IPV4_ROUTE_ENTRY,

    /**
     * @brief Available IPv6 routes
     *
     * @type sai_uint32_t
     * @flags READ_ONLY
     */
    SAI_SWITCH_ATTR_AVAILABLE_IPV6_ROUTE_ENTRY,

    /**
     * @brief Available IPv4 Nexthop entries
     *
     * @type sai_uint32_t
     * @flags READ_ONLY
     */
    SAI_SWITCH_ATTR_AVAILABLE_IPV4_NEXTHOP_ENTRY,
```

marian-pritsak: These attributes are hard coded. The purpose of this proposal is to introduce a generic API that will allow the query of any type of resource without introducing more attributes to the switch object.

atitjain: So you intend to deprecate these attributes in switch once this proposal is approved.

marian-pritsak: Yes, that's correct.

meta/saisanitycheck.c
```diff
@@ -3100,6 +3100,11 @@ void check_non_object_id_object_attrs()
 
             META_ASSERT_NOT_NULL(m);
 
+            if (m->isresourcetype)
+            {
+                continue;
```
kcudnik: why this is added here? Do those resource types have different attributes ?

marian-pritsak: Currently, non-object entries do not allow attributes other than CREATE_AND_SET. I introduced a READ_ONLY attribute to a route entry that tells the addr_family of the entry to be used to distinguish between different pools of resources in the function `sai_object_type_get_availability`
So this is needed to allow that kind of a READ_ONLY attribute of an entry.

And it already behaves that way - if absent, it is assumed that isresourcetype is false.

kcudnik: yes, but your code allows all possible attributes, read_only, create_and_set etc, so please add extra limiting here theh when isresourcetype is true then attribute must be your expected type

marian-pritsak: Done

lguohan:
can you check the tunnel type? is that needed for counting the resources?

PR #945, opened 2019-04-26
Support for trapping BFD packets to CPU

bandaru-viswanath:


bandaru-viswanath:
Will submit independent pull requests for BFD and STP attribute

PR #946, opened 2019-04-27
Support for Bfd traps

bandaru-viswanath:
Support for BFD packets to be trapped to CPU

### DIFF
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -317,6 +317,18 @@ typedef enum _sai_hostif_trap_type_t
      */
     SAI_HOSTIF_TRAP_TYPE_BGPV6 = 0x00004004,
 
+    /**
+     * @brief BFD traffic (UDP dst port == 3784 or UDP dst port == 4784) to local
+     * router IP address (default packet action is drop)
+     */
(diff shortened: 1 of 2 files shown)


inc/saihostif.h
```diff
@@ -317,6 +317,18 @@ typedef enum _sai_hostif_trap_type_t
...
+     * router IP address (default packet action is drop)
+     */
+    SAI_HOSTIF_TRAP_TYPE_BFD = 0x00004005,
+
+    /**
+     * @brief BFDV6 traffic (UDP dst port == 3784 or UDP dst port == 4784) to
+     * local router IP address (default packet action is drop)
+     */
+    SAI_HOSTIF_TRAP_TYPE_BFDV6 = 0x00004006,
+
```
atitjain: Could you please check how did bfd (sai/inc/saibfd.h) work without this trap.

bandaru-viswanath: These traps are meant for applications, and not for sai itself. 

For those SAI implementations which supported BFD already, they already installed a TRAP_TO_CPU and consumed them inside/under SAI. 

This pull-request is for supporting applications which intend to implement BFD themselves (all in application).

PR #953, opened 2019-05-06
Comments updated for outer TTL/DSCP value if inner packet is non-IP

prsunny:


inc/saitunnel.h
```diff
@@ -335,7 +335,9 @@ typedef enum _sai_tunnel_ttl_mode_t
      *
      * Where the TTL field is preserved end-to-end by copying into the outer
      * header on encapsulation and copying from the outer header on
-     * decapsulation.
+     * decapsulation. This is applicable for inner IP packets. If the inner
+     * packet is a non-IP packet, then the value is undefined and implementation
+     * can chose a valid/meaningful outer TTL value, say in the case of VXLAN encap.
```
atitjain: Can't we define something, say on Switch/port level which is a default for all such cases? Or maybe follow the COS map rules as default behavior.

prsunny: @atit, this comment is solely to avoid the ambiguity. I don't think we would need to callout explicit default values in this particular case

lguohan: @atitjain, do you have some more concrete idea how to handle this case. we'd love to hear from you.

atitjain: Sure I will try.

atitjain: I think we need to update the behavioral model [Pipeline]((https://github.com/opencomputeproject/SAI/blob/master/doc/behavioral%20model/pipeline_v6.pdf)) with
1. CoS mapping at various stages (Port/ACL etc)
2. Pipeline Metadata part to reflect PCP/DSCP/TC/Color etc
Once this is done, this behavior will become deterministic.

atitjain: I think we need to update the [behavioral model](https://github.com/opencomputeproject/SAI/blob/master/doc/behavioral%20model/pipeline_v6.pdf) with
1. CoS mapping at various stages (Port/ACL etc)
2. Pipeline Metadata part to reflect PCP/DSCP/TC/Color etc
Once this is done, this behavior will become deterministic.

lguohan: @prsunny , can you update?

PR #956, opened 2019-05-13
Enhancement for L2VlanBcastUcastTest

PJHsieh:
1. During the testing, the mac "00:00:00:00:00:01"is learned on the
   ingress port. After testing, this mac is not removed from mac table.
   Modification:
   Remove this mac after testing.

2. Before testing, all port are removed from default vlan.
   After testing, just add the testing ports into default vlan.
   Modification:
   Remove only the testing ports from default vlan.


PJHsieh:
retest this please

PR #957, opened 2019-05-13
For BCM platform, if the aging time is n, the worst case shall be n*2.

PJHsieh:
1. Add input parameter chip_type
2. Modify the pause time from n+2 to n*2 if chip_type is 'bcm'
3. Fix the wrong argument for api sai_thrift_flush_fdb_by_vlan
   The last parameter shall be vlan_oid, not vlan_id
4. Add the description for chip_type in README.txt

### DIFF
--- a/test/saithrift/README.txt
+++ b/test/saithrift/README.txt
@@ -67,7 +67,7 @@ Otherwise, please setup an appropriate cross-compile environment to generate 'sa
 
     3. Copy tests directory to the test machine (client)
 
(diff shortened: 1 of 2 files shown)


PJHsieh:
retest this please

atitjain:
In most of the implementations, the worst case should be n*2 as the aging thread will run once every aging interval. Why is it kept as n + 2 for other implementations?

PJHsieh:
> In most of the implementations, the worst case should be n*2 as the aging thread will run once every aging interval. Why is it kept as n + 2 for other implementations?

Just test for BCM chips. Therefore, keep the original code.

PJHsieh:
@atitjain:
Is there any concern, or can you merge this PR ?
Thanks.

PR #959, opened 2019-05-13
TAM INT Support in SAI

JaiOCP:


lguohan:
retest this please

JaiOCP:
> retest this please

Fixup and retest done.

mickeyspiegel:
In sai_tam_collector_attr_t, I would like to see a TC or queue index attribute:

```
    /**
     * @brief Traffic class used for TAM reports to the collector
     *
     * @type sai_uint8_t
     * @flags CREATE_AND_SET
     * @default 0
     */
    SAI_TAM_COLLECTOR_ATTR_TC,
```

mickeyspiegel:
The word doc has markup of all the changes in the last round or two of updates. These changes should all be accepted before finalizing the document.

Also there are 3 comments in the word doc that need to be resolved:

1. Figure 1 is marked "Broadcom proprietary and confidential". Please remove the company specific text and logos.
2. The document has an example using the SAI_TAM_TEL_TYPE_ATTR_FLOW_ID attribute which does not exist in the current saitam.h.
3. The microburst detection example describes units of microseconds, but it is not clear where those units came from. The corresponding even threshold used units of nanoseconds.

PR #960, opened 2019-05-14
Bfd offload support

bandaru-viswanath:
There are hardware offloads available for establishing and sustaining BFD sessions. This pull request aims to add capabilities for the SAI/BFD API to allow applications to query the list of supported offloads and choose an offload while creating a BFD session.

*) Session Offload Type is a new attribute for the BFD session object.

The offload type 'SAI_BFD_SESSION_OFFLOAD_TYPE_FULL' offload indicates that underlying silicon is capable of establishing and sustaining BFD sessions with given parameters.
The offload type 'SAI_BFD_SESSION_OFFLOAD_TYPE_SUSTENANCE' offload indicates that underlying silicon is capable of sustaining an control-plane established BFD session
The offload type 'SAI_BFD_SESSION_OFFLOAD_TYPE_NONE' indicates that there is no offload available. 

BFD API already has mechanisms to report session establishment and tear-down via notifications.

*) The switch object has new RO attributes to obtain a list of supported BFD offload types.

### DIFF
--- a/inc/saibfd.h
+++ b/inc/saibfd.h
@@ -436,7 +436,7 @@ typedef enum _sai_bfd_session_attr_t
      * @flags CREATE_ONLY
      * @default SAI_BFD_SESSION_OFFLOAD_TYPE_NONE
      */
-    SAI_BFD_SESSION_ATTR_BFD_SESSION_OFFLOAD_TYPE,
+    SAI_BFD_SESSION_ATTR_OFFLOAD_TYPE,
 
     /**
      * @brief End of attributes
--- a/inc/saibfd.h
+++ b/inc/saibfd.h
@@ -53,6 +53,22 @@ typedef enum _sai_bfd_session_type_t
 } sai_bfd_session_type_t;
 
 /**
+ * @brief SAI offload type of BFD session
+ */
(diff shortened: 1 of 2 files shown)


inc/saibfd.h
```diff
@@ -412,6 +429,15 @@ typedef enum _sai_bfd_session_attr_t
...
     SAI_BFD_SESSION_ATTR_STATE,
 
+    /**
+     * @brief Offload type
+     *
+     * @type sai_bfd_session_offload_type_t
+     * @flags CREATE_ONLY
+     * @default SAI_BFD_SESSION_OFFLOAD_TYPE_NONE
+     */
+    SAI_BFD_SESSION_ATTR_BFD_SESSION_OFFLOAD_TYPE,
```
kcudnik: maybe it should be name SAI_BFD_SESSION_ATTR_OFFLOAD_TYPE ? no need to duplicate bfd_session prefix

bandaru-viswanath: Thanks. I renamed the Attribute as SAI_BFD_SESSION_ATTR_OFFLOAD_TYPE

PR #966, opened 2019-05-29
Added Redirect cancel packet action

PugalendranDell:
Added Redirect cancel packet action. User willing to cancel the Redirect packet action can use this new action.


lguohan:
discussed in the SAI meeting. it is better to be an ACL action instead of a generic packet action that can be used in various blocks such as fdb and fib.

PR #969, opened 2019-06-11
ARM32 bit fixes, in the usage of printf format specifiers

antony-rheneus:
Signed-off-by: Antony Rheneus <arheneus@marvell.com>


kcudnik:
try adding
#define __STDC_FORMAT_MACROS
#include <inttypes.h>

PR #939, opened 2019-04-10
Create SAI-Proposal-Generic-Counters.md

marian-pritsak:

PR #970, opened 2019-06-12
Add a RO attr to trap object for Packet Count

bandaru-viswanath:
Add a RO attribute to the trap object to get Trap Packet Count : SAI_HOSTIF_TRAP_ATTR_RX_PKTS

### DIFF
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -455,6 +455,14 @@ typedef enum _sai_hostif_trap_attr_t
     SAI_HOSTIF_TRAP_ATTR_MIRROR_SESSION,
 
     /**
+     * @brief Number of packets received for the trap
+     *
+     * @type sai_uint64_t
+     * @flags READ_ONLY
+     */
+    SAI_HOSTIF_TRAP_ATTR_RX_PKTS,
+
+    /**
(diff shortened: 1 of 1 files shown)


inc/saihostif.h
```diff
@@ -454,6 +454,14 @@ typedef enum _sai_hostif_trap_attr_t
      */
     SAI_HOSTIF_TRAP_ATTR_MIRROR_SESSION,
 
+    /**
+     * @brief Number of packets received for the trap
+     *
+     * @type sai_uint64_t
+     * @flags READ_ONLY
+     */
+    SAI_HOSTIF_TRAP_ATTR_RX_PKTS,
```
lguohan: since trap can be set to drop, forward, does this count all cases or it only counts packets with action trap?

bandaru-viswanath: The intent was to count only the packets trapped to CPU. If there is a need, I am open to generalise it to count all packets that hit the trap.

bandaru-viswanath:
This PR is being closed as discussed in the community meeting on 6/13, in favour of more generic "Generic Counters" proposal #939  from Marian. #939 is now approved and merged.

PR #971, opened 2019-06-12
Add attr to port-obj for timestamping for PTP Pkts

bandaru-viswanath:
Add an optional attribute to port-object to support timestamping for PTP Packets.


lguohan:
is there any reason we do not timestamping ptp packets?

bandaru-viswanath:
> is there any reason we do not timestamping ptp packets?

Yes. It is possible that the switch is set to just switch the packets. It is a valid use case.

bandaru-viswanath:
Based on the feedback in community meeting today, I have made the one-step and two-step time-stamping modes explicit.

PR #979, opened 2019-06-26
Add variable declaration for ACL PTF tests

MELLANOX-zhichenr:
Add object declaration so that the life time of the objects will be
extended to cleanup stage


MELLANOX-zhichenr:
Without the variable declaration, the following error message might appear in the finally block if the traffic test failed.

======================================================================
ERROR: saiacl.L3AclTableGroupTestI
----------------------------------------------------------------------
Traceback (most recent call last):
  File "SAI_thrift/tests/saiacl.py", line 1634, in runTest
    self.client.sai_thrift_remove_acl_table_group_member(acl_table_group_member_id1)
UnboundLocalError: local variable 'acl_table_group_member_id1' referenced before assignment


Thus, adding the declaration can help to get clear error message.

PR #983, opened 2019-07-03
Add additional hash inner fields

itaibaz:
add inner ip protocol, ethertype, l4 src/dst port, src/dst mac


lguohan:
@stcheng  to look at

PR #985, opened 2019-07-18
Debug counters proposal

itaibaz:


kcudnik:
/var/johnar/workspace/sai-meta-pr-build/inc/saidebugcounter.h:124: warning: Found unknown command `\isresourcetype'

missing from doxygen ? or wrong param name

lguohan:
@marian-pritsak , it is because the asic may not have so many different counters to count.

mickeyspiegel:
I am fine with the proposed approach of automatic bind method and Base object : SAI_OBJECT_TYPE_PORT.

doc/SAI-Proposal-Debug-Counters.md
```diff
@@ -0,0 +1,239 @@
...
+    SAI_DEBUG_COUNTER_ATTR_TYPE = SAI_DEBUG_COUNTER_ATTR_START,
+
+    /**
+     * @brief Object stat index
+     * Index is added to base start
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_DEBUG_COUNTER_ATTR_INDEX,
```
bandaru-viswanath: As discussed in the community meeting, can you please consider making this a READ_ONLY and let SAI manage the indexing ?

lguohan:
retest this please

doc/SAI-Proposal-Debug-Counters.md
daall: +1, I think letting the SAI manage the indexing would be less prone to user error

kcudnik:
WARNING: Word 'DMAC' is misspelled or is acronym, add to acronyms.txt? saidebugcounter.h 76:    /** Destination MAC is Reserved (DMAC=01-80-C2-00-00-0x) */

add DMAC to acronyms

doc/SAI-Proposal-Debug-Counters.md
itaibaz: Changed to RO as requested

kcudnik:
you can go ahead and remove this IF 👍 https://github.com/opencomputeproject/SAI/blob/master/meta/saisanitycheck.c#L1548

also you will need to connect this debug counter somewhere to the graph because right now it's disconnected node

doc/SAI-Proposal-Debug-Counters.md
```diff
@@ -0,0 +1,264 @@
...
+     * @type sai_debug_counter_bind_method_t
+     * @flags CREATE_ONLY
+     * @default SAI_DEBUG_COUNTER_BIND_METHOD_AUTOMATIC
+     */
+    SAI_DEBUG_COUNTER_ATTR_BIND_METHOD,	
+
+    /**
+     * @brief List of port in drop reasons that will be counted
+     *
+     * @type sai_s32_list_t sai_port_in_drop_reason_t
```
bandaru-viswanath: Just to be explicit. When a packet is deemed to be 'eligible' for a drop, it can be for more than one drop reason. In that sense, this debug-counter is counting the number of times a given drop-reason has come into play, as opposed to number of packets that have been dropped. So this counter, for a given port, would be greater-than-or-equal to the number of packets dropped on that port (assuming a drop at INGRESS as an example). 

It would be good to document this behaviour in the MD file.

inc/saidebugcounter.h
```diff
@@ -0,0 +1,277 @@
...
+    /* L3 reasons */
+
+    /** IPv4 Unicast Destination IP is link local (Destination IP=169.254.0.0/16) */
+    SAI_PORT_IN_DROP_REASON_DIP_LINK_LOCAL,
+
+    /** IPv4 Source IP is link local (Source IP=169.254.0.0/16) */
+    SAI_PORT_IN_DROP_REASON_SIP_LINK_LOCAL,
+
+    /** Packet size is larger than the MTU */
+    SAI_PORT_IN_DROP_REASON_EXCEEDS_MTU,
```
daall: It might be helpful to call this `SAI_PORT_IN_DROP_REASON_EXCEEDS_L3_MTU` to avoid confusing this with oversized L2 frames.

daall: It might be helpful to call this `SAI_PORT_IN_DROP_REASON_EXCEEDS_L3_MTU` to avoid confusing this with oversized L2 frames.

inc/saidebugcounter.h
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * @brief Attribute data for port out drop reasons
+ */
+typedef enum _sai_port_out_drop_reason_t
+{
+    /** Egress VLAN filter */
+    SAI_PORT_OUT_DROP_REASON_EGRESS_VLAN_FILTER,
+
+    /** Packet is destined for neighboring device but neighbor device link is down */
+    SAI_PORT_OUT_DROP_REASON_L3_EGRESS_LINK_DOWN,
```
lguohan: can you add SAI_PORT_IN_DROP_REASON_L3_EGRESS_LINK_DOWN?

inc/saidebugcounter.h
itaibaz: Added L2 MTU reason, so now we have L2 MTU and L3 MTU reasons

inc/saidebugcounter.h
itaibaz: It is already there - in egress reasons - 
    /** Packet is destined for neighboring device but neighbor device link is down */
    SAI_PORT_OUT_DROP_REASON_L3_EGRESS_LINK_DOWN,

kcudnik:
so actually current implementation will not solve disconnected object with readonly attribute since there should be attribute that this counter can be set to dbg_counter, or some attribute on dbg counter that could be set to other object type

i guess we could make exception for this counter object if no other solution will be implemented

inc/saidebugcounter.h
daall: I believe he's asking to have it included as both an egress reason and an ingress reason, as different platforms could implement this behavior as an ingress counter or an egress counter (or both).

inc/saidebugcounter.h
```diff
@@ -0,0 +1,280 @@
...
+
+    /** IPv4 Unicast Destination IP is link local (Destination IP=169.254.0.0/16) */
+    SAI_PORT_IN_DROP_REASON_DIP_LINK_LOCAL,
+
+    /** IPv4 Source IP is link local (Source IP=169.254.0.0/16) */
+    SAI_PORT_IN_DROP_REASON_SIP_LINK_LOCAL,
+
+    /** Packet size is larger than the L3 (Router Interface) MTU */
+    SAI_PORT_IN_DROP_REASON_EXCEEDS_L3_MTU,
+
```
rck-innovium: Pls add a reason for TTL errors like SAI_PORT_IN_DROP_REASON_TTL.

inc/saidebugcounter.h
```diff
@@ -0,0 +1,280 @@
...
+    SAI_PORT_IN_DROP_REASON_INGRESS_STP_FILTER,
+
+    /** Unicast FDB table action discard */
+    SAI_PORT_IN_DROP_REASON_FDB_UC_DISCARD,
+
+    /** Multicast FDB table empty tx list */
+    SAI_PORT_IN_DROP_REASON_FDB_MC_DISCARD,
+
+    /** Port loopback filter */
+    SAI_PORT_IN_DROP_REASON_LOOPBACK_FILTER,
```
rck-innovium: can you clarify on what this drop reason is?

inc/saidebugcounter.h
rck-innovium: L3 header errors.

doc/SAI-Proposal-Debug-Counters.md
rck-innovium: This is interesting. If the drop reasons are not meant to be mutually exclusive, then we can add few more drop reasons at a coarser granularity. For eg, along with the proposed fine grained drop reasons, we could have coarser drop-reasons like SAI_PORT_IN_DROP_REASON_L2_ANY, SAI_PORT_IN_DROP_REASON_L3_ANY etc. 
Use cases:
1) It allows NPU/ASIC vendors to accommodate ASIC specific drop counters which are not listed in the SAI spec.
2) As a user when I want to debug the drop reason for a given flow, I would first use the coarse grained counters like L2 or L3. Based on which of the counter got incremented, I can drill down to the other fine-grained counters.

doc/SAI-Proposal-Debug-Counters.md
```diff
@@ -0,0 +1,264 @@
...
+    SAI_DEBUG_COUNTER_TYPE_PORT_OUT_DROP_REASONS,
+
+} sai_debug_counter_type_t;
+```
+
+## Debug counter port in drop reason family
+```
+typedef enum _sai_port_in_drop_reason_t
+{
+    /* L2 reasons */
```
rck-innovium: The current enums would need changes to add drop-reasons as newer ASICs are added. This would keep constantly changing. An alternative is to do this generically and let vendors add new drop reasons without having to change SAI spec or SONIC.

Say, we have two switch read-only attributes  SAI_SWITCH_ATTR_INGRESS_PORT_DROP_REASON_LIST and 
SAI_SWITCH_ATTR_EGRESS_PORT_DROP_REASON_LIST. NOS queries these attributes and platform lists the drop port stat ids it supports along with descriptive names. NOS would just query the port stats with these ids and print the port, descriptive name of the drop reason, and the drop counter value.

typedef struct _sai_drop_reason_t
{
    /** drop reason Id */
    sai_uint32_t id;

    /** drop reason name */
    char reason_name[128];

} sai_drop_reason_t;

typedef struct _sai_drop_reason_list_t
{
    /** Number of entries in the list */
    uint32_t count;

    /**drop reason list */
    sai_drop_reason_t *list;

} sai_drop_reason_list_t;


//     * @type sai_switch_drop_reason_list_t
//      * @flags READ_ONLY
SAI_SWITCH_ATTR_INGRESS_PORT_DROP_REASON_LIST
SAI_SWITCH_ATTR_EGRESS_PORT_DROP_REASON_LIST

rck-innovium:
Say, a given platform uses the same counter to account for multiple drop reasons. For eg, let us consider the case that the underlying device has the same h/w counter for both of the  below reasons:
SAI_PORT_IN_DROP_REASON_FDB_UC_DISCARD.
SAI_PORT_IN_DROP_REASON_FDB_MC_DISCARD

 Now, how does this platform expose this capability using sai_query_attribute_enum_values_capability(). 
The end goal of debug counters should be to avoid silent drops, and hence the SONIC user should know that there exists a counter that would account for both FDB_UC_DISCARD and FDB_MC_DISCARD.

doc/SAI-Proposal-Debug-Counters.md
itaibaz: Added a section about counting a packet dropped because of multiple reasons in the MD
Packet will be counted once per debug counter instance

Added drop reason any L2, L3 as suggested

doc/SAI-Proposal-Debug-Counters.md
itaibaz: IMO, you can have the same comment on any SAI feature, such as all counters, or even all attributes. Sure, there are many differences between different ASICs. So we try to capture a common basis as the standard, and allow any vendor to extend the list with custom range (I will go ahead and add that to the proposal)
Having a completely custom list takes all the approach behind unified API, and makes SAI just a dispatcher for vendor specific API

kcudnik:
ASSERT FAILED (on line 4166): SAI_OBJECT_TYPE_DEBUG_COUNTER not present on any object list (eg. VLAN_MEMBER is present on SAI_VLAN_ATTR_MEMBER_LIST)

you will also need to add exception for iterator for debug counter
also can you put warning messages in those 2 places where you added exceptions ?

kcudnik:
can you have multiple counter objects with the same type sai_debug_counter_type_t??

itaibaz:
> can you have multiple counter objects with the same type sai_debug_counter_type_t??

Yes you can, please take a look on the example in the MD
Each instance, both of the same type, tracks different reasons

doc/SAI-Proposal-Debug-Counters.md
rck-innovium: IMO, debug_counters are different from other counters. Here, the goal is to not have silent packet drops. The SONIC code we are discussing here today really doesn't need to actually understand the drop_reason to do anything. The SONIC code just needs  to
1) let the user know about the debug counters supported by the platform
2) let the user configure different combination of drop counters on a given port identify the actual drop reason.
3) drop reasons vary widely across ASICs and there are no standards for it as against port stats etc.

inc/saidebugcounter.h
rck-innovium: Can you explain what SAI_PORT_IN_DROP_REASON_LOOPBACK_FILTER means?

doc/SAI-Proposal-Debug-Counters.md
lguohan: in the spec, we show example to query the drop reasons for a platform. what you can do is to add your own list of reason to the enum and once a nos query it will use those enum supported by your platform for debugging.

lguohan:
retest this please

lguohan:
retest this please

PR #988, opened 2019-07-22
Add support for INT type tail timestamping

bandaru-viswanath:
Add support for an INT type where the metadata (ingress and egress timestamps) are inserted at the end of the packet before the Ethernet CRC.

The developer work flow is identical as described in the TAM documentation, and only a new INT type is being added.

### DIFF
--- a/inc/saitam.h
+++ b/inc/saitam.h
@@ -469,7 +469,12 @@ typedef enum _sai_tam_int_type_t
     /**
      * @brief Direct Export (aka postcard)
      */
-    SAI_TAM_INT_TYPE_DIRECT_EXPORT
+    SAI_TAM_INT_TYPE_DIRECT_EXPORT,
+
+    /**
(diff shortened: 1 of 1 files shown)


inc/saitam.h
```diff
@@ -469,7 +469,12 @@ typedef enum _sai_tam_int_type_t
     /**
      * @brief Direct Export (aka postcard)
      */
-    SAI_TAM_INT_TYPE_DIRECT_EXPORT
+    SAI_TAM_INT_TYPE_DIRECT_EXPORT,
+
+    /**
+     * @brief Telemetry data at the end of the packet
+     */
+    SAI_TAM_INT_TYPE_TAIL_TIMESTAMP,
```
lguohan: make it more specific, IFA?

bandaru-viswanath: Sure, Also the same comment from Mickey in the community discussion, will change to be more specific.

PR #991, opened 2019-07-29
Add support for Aggregate reporting for TAM Events

bandaru-viswanath:
Add an optional attribute to tam/report object to add support for Aggregate reporting for TAM Events. Since it is an optional attribute, the current functionality of reporting every event is unaffected. User can select 'aggregate' mode for the report mode attribute and specify an interval. When chosen, aggregate-reports are sent at the chosen interval.

### DIFF
--- a/inc/saitam.h
+++ b/inc/saitam.h
@@ -1159,8 +1159,8 @@ typedef enum _sai_tam_report_mode_t
     /** Report all events */
     SAI_TAM_REPORT_MODE_ALL = 0,
 
-    /** Report in an aggregate mode */
-    SAI_TAM_REPORT_MODE_AGGREGATE,
+    /** Report in a bulk mode */
+    SAI_TAM_REPORT_MODE_BULK,
 
 } sai_tam_report_mode_t;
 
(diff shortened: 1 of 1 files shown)


bandaru-viswanath:
@marian-pritsak @mickeyspiegel Could you please approve this PR. I believe I have incorporated your feedback from the community discussion. Let me know if you have any further feedback.

inc/saitam.h
```diff
@@ -1207,6 +1221,25 @@ typedef enum _sai_tam_report_attr_t
...
+     * @type sai_tam_report_mode_t
+     * @flags CREATE_ONLY
+     * @default SAI_TAM_REPORT_MODE_ALL
+     */
+    SAI_TAM_REPORT_ATTR_REPORT_MODE,
+
+    /**
+     * @brief Report Interval in micro seconds
+     *
+     * Note: valid only if SAI_TAM_REPORT_ATTR_REPORT_MODE == SAI_TAM_REPORT_MODE_BULK
```
marian-pritsak: use tag "@validonly"

bandaru-viswanath: I have modified. Thanks @marian-pritsak.

PR #993, opened 2019-07-31
Warning: Fix for Unsigned long long usage in 32bit arch using ULL macro

antony-rheneus:
Signed-off-by: Antony Rheneus <arheneus@marvell.com>

### DIFF
--- a/meta/Makefile
+++ b/meta/Makefile
@@ -41,7 +41,6 @@ WARNINGS = \
 	-Winit-self \
 	-Winline \
 	-Winvalid-pch \
-	-Wlong-long \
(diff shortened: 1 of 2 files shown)


meta/Makefile
```diff
@@ -41,7 +41,6 @@ WARNINGS = \
 	-Winit-self \
 	-Winline \
 	-Winvalid-pch \
-	-Wlong-long \
```
kcudnik: why this is removed ?

antony-rheneus: Need  ULL macro suffix for unsigned long long incase of 32 bit arch, for a 64bit const

PR #996, opened 2019-08-07
Add support for deep/large buffers in buffer objects

bandaru-viswanath:
The currently defined 32bit type can't hold large sizes for the buffers. Changed the buffer size data type to 64bits.


bandaru-viswanath:
> I noticed SAI_SWITCH_ATTR_SUPPORTED_OBJECT_TYPE_LIST is also here
> Are these 2 PRs mixed together ?

This is my issue. It should not have been here. I will change. Thanks for pointing out.
> 
> Also, not that there is really anything against increasing to 64 bits, but is your buffer size really over 4 GB that uint32 covers ? AFAIK, ASIC buffers are in tens of MBs and not 4 GB

Yes. we have some chips that support buffers that 32bits can't hold, more than 4GB as you say.

bandaru-viswanath:
> > I noticed SAI_SWITCH_ATTR_SUPPORTED_OBJECT_TYPE_LIST is also here
> > Are these 2 PRs mixed together ?
> 
> This is my issue. It should not have been here. I will change. Thanks for pointing out.
> 

Corrected. Thanks Itai.

lguohan:
@itaibaz , can you approve?

PR #997, opened 2019-08-10
Modify GENETLINK multicast group attribute from id to name

padmanarayana:
Since it is convenient to lookup the multicast group based on name rather than id (both from SAI driver and sflow daemon perspective), changing the genetlink multicast sampling group attribute to char. For e.g, if the psample driver is used, SAI_HOSTIF_ATTR_NAME="psample" and SAI_HOSTIF_ATTR_GENETLINK_MCGRP_NAME="packets" (https://github.com/torvalds/linux/blob/master/include/uapi/linux/psample.h#L32 )

### DIFF
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -46,6 +46,11 @@
 #define SAI_HOSTIF_NAME_SIZE 16
 
 /**
+ * @brief Defines maximum length of generic netlink multicast group name
+ */
+#define SAI_HOSTIF_GENETLINK_MCGRP_NAME_SIZE 16
+
+/**
  * @brief Host interface trap group attributes
  */
 typedef enum _sai_hostif_trap_group_attr_t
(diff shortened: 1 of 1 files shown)


inc/saihostif.h
```diff
@@ -801,14 +806,17 @@ typedef enum _sai_hostif_attr_t
...
      * are received on this host interface
      *
-     * @type sai_uint32_t
+     * @type char
      * @flags CREATE_AND_SET
-     * @default 0
+     * @default ""
      */
-    SAI_HOSTIF_ATTR_GENETLINK_PORT_ID,
+    SAI_HOSTIF_ATTR_GENETLINK_MCGRP_NAME,
```
lguohan: should we make this mandatory?

padmanarayana: Done.

padmanarayana:
Including reviewers of #936 @kcudnik  @marian-pritsak @michaelli10 : please review

padmanarayana:
w.r.t the clarifications in today's SAI meeting:

1. The multicast group name length restriction comes from the kernel genetlink infra which limits the family name as well as the multicast group name to 16 chars:
linux/genetlink.h:
#define GENL_NAMSIZ	16	/* length of family name */

include/net/genetlink.h 

struct genl_multicast_group {
	char			name[GENL_NAMSIZ];
};

struct genl_family {
	unsigned int		id;
	unsigned int		hdrsize;
	char			name[GENL_NAMSIZ];
...

2. The genlmsg_multicast_netns() uses the portid (instead of name). However, from a persistence point of view (after warm/fast reboot), since the id might change (based on the driver implementation / order in which the mc groups are created), it is better to configure the multicast group name instead of id.

marian-pritsak:
Besides the MC group, how do we pass the PSAMPLE_ATTR_SAMPLE_GROUP group id?

padmanarayana:
Currently, bundling flows based on some criteria (say, ACL ) is not defined and hence PSAMPLE_ATTR_SAMPLE_GROUP can be set to a default 1 (since existing sflow agents (hsflowd)) use this default.

PR #998, opened 2019-08-12
Remove the need to create an object id for vlan in creating a sub port router interface

wendani:
Instead we only provide the vlan id value


wendani:
I assume SONiC is the only user of SAI. If this is the case, it should be fine to make this correction as we are still at the high-level design phase to have the sub port interface feature in SONiC.

https://github.com/opencomputeproject/SAI/pull/998#pullrequestreview-275278451

PR #1002, opened 2019-09-10
Defining new serdes attributes

dgsudharsan:
Defined new Serdes attributes. The new serdes range is reserved so that in future the attributes can be extended to include new settings in the future.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1223,6 +1223,7 @@ typedef enum _sai_port_attr_t
      * @type sai_object_id_t
      * @flags READ_ONLY
      * @objects SAI_OBJECT_TYPE_PORT_SERDES
+     * @default internal
      */
     SAI_PORT_ATTR_PORT_SERDES_ID,
 
@@ -2147,7 +2148,7 @@ typedef enum _sai_port_serdes_attr_t
      * to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_PREEMPHASIS,
@@ -2160,7 +2161,7 @@ typedef enum _sai_port_serdes_attr_t
      * to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_IDRIVER,
@@ -2173,7 +2174,7 @@ typedef enum _sai_port_serdes_attr_t
      * to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_IPREDRIVER,
@@ -2186,7 +2187,7 @@ typedef enum _sai_port_serdes_attr_t
      * a port and the list specifies list of values to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_TX_FIR_PRE1,
@@ -2199,7 +2200,7 @@ typedef enum _sai_port_serdes_attr_t
      * a port and the list specifies list of values to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_TX_FIR_PRE2,
@@ -2212,7 +2213,7 @@ typedef enum _sai_port_serdes_attr_t
      * a port and the list specifies list of values to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_TX_FIR_PRE3,
@@ -2225,7 +2226,7 @@ typedef enum _sai_port_serdes_attr_t
      * a port and the list specifies list of values to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_TX_FIR_MAIN,
@@ -2238,7 +2239,7 @@ typedef enum _sai_port_serdes_attr_t
      * a port and the list specifies list of values to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_TX_FIR_POST1,
@@ -2251,7 +2252,7 @@ typedef enum _sai_port_serdes_attr_t
      * a port and the list specifies list of values to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_TX_FIR_POST2,
@@ -2264,7 +2265,7 @@ typedef enum _sai_port_serdes_attr_t
      * a port and the list specifies list of values to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_TX_FIR_POST3,
@@ -2277,7 +2278,7 @@ typedef enum _sai_port_serdes_attr_t
      * a port and the list specifies list of values to be applied to each lane.
      *
      * @type sai_u32_list_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
      * @default internal
      */
     SAI_PORT_SERDES_ATTR_TX_FIR_ATTN,
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1160,6 +1160,7 @@ typedef enum _sai_port_attr_t
     /**
      * @brief Port serdes control pre-emphasis
      *
+     * Deprecated. Use SAI_OBJECT_TYPE_PORT_SERDES
      * List of port serdes pre-emphasis values. The values are of type sai_u32_list_t
      * where the count is number lanes in a port and the list specifies list of values
      * to be applied to each lane.
@@ -1173,6 +1174,7 @@ typedef enum _sai_port_attr_t
     /**
      * @brief Port serdes control idriver
      *
+     * Deprecated. Use SAI_OBJECT_TYPE_PORT_SERDES
      * List of port serdes idriver values. The values are of type sai_u32_list_t
      * where the count is number lanes in a port and the list specifies list of values
      * to be applied to each lane.
@@ -1186,6 +1188,7 @@ typedef enum _sai_port_attr_t
     /**
      * @brief Port serdes control ipredriver
      *
+     * Deprecated. Use SAI_OBJECT_TYPE_PORT_SERDES
      * List of port serdes ipredriver values. The values are of type sai_u32_list_t
      * where the count is number lanes in a port and the list specifies list of values
      * to be applied to each lane.
@@ -1213,6 +1216,15 @@ typedef enum _sai_port_attr_t
      * @default SAI_PORT_PTP_MODE_NONE
      */
     SAI_PORT_ATTR_PTP_MODE,
+
+    /**
+     * @brief Serdes object ID for the port
+     *
+     * @type sai_object_id_t
+     * @flags READ_ONLY
+     * @objects SAI_OBJECT_TYPE_PORT_SERDES
+     */
+    SAI_PORT_ATTR_PORT_SERDES_ID,
 
     /**
      * @brief End of attributes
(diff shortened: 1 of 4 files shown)


inc/saiport.h
```diff
@@ -201,9 +201,53 @@ typedef enum _sai_port_ptp_mode_t
 
 } sai_port_ptp_mode_t;
 
+/**
+ * @brief Port Serdes Attribute ID Range
+ */
+#define SAI_PORT_ATTR_SERDES_RANGE 32
```
kcudnik: this is not needed

inc/saiport.h
```diff
@@ -201,9 +201,53 @@ typedef enum _sai_port_ptp_mode_t
...
+    SAI_PORT_ATTR_OFFSET_SERDES_TX_FIR_POST1 = 6,
+
+    /** Tx fir Postcursor2 tap-filter offset */
+    SAI_PORT_ATTR_OFFSET_SERDES_TX_FIR_POST2 = 7,
+
+    /** Tx fir Postcursor3 tap-filter offset */
+    SAI_PORT_ATTR_OFFSET_SERDES_TX_FIR_POST3 = 8,
+
+    /** Max serdes attribute offset */
+    SAI_PORT_ATTR_OFFSET_SERDES_END = SAI_PORT_ATTR_SERDES_RANGE - 1
```
kcudnik: not need need to set END to the range end, since those attributes 9+ are not defined yet, leave end as next enum, take a look at other attr definitions

inc/saiport.h
```diff
@@ -201,9 +201,53 @@ typedef enum _sai_port_ptp_mode_t
...
+
+} sai_port_attr_offset_serdes_t;
+
 /**
  * @brief Attribute Id in sai_set_port_attribute() and
  * sai_get_port_attribute() calls
+ *
+ * @flags Contains flags
```
kcudnik: you changed this to be flags, and you added ranges for attributes, but none of those are defined, instead of adding min/max ranges, which at this point is incomplete, just add new attributes one by one, instead introducing incomplete ranges

inc/saiport.h
```diff
@@ -1214,6 +1261,33 @@ typedef enum _sai_port_attr_t
...
+     * Start of port serdes control values. The values are of type sai_u32_list_t
+     * where the count is number lanes in a port and the list specifies list of values
+     * to be applied to each lane.
+     *
+     * @type sai_u32_list_t
+     * @flags CREATE_AND_SET
+     * @default internal
+     * @range SAI_PORT_ATTR_SERDES_RANGE
+     */
+    SAI_PORT_ATTR_SERDES_MIN,
```
SathishKumarTR-Dell: 1. It might be good to have a "struct" type defined for the "list of different pre-emphasis taps" and add it into the sai_attribute_value_t union. 
2. Rename the attribute to denote TX_SERDES_PARAMS or TX_PRE_EMPHASIS

dgsudharsan:
Can someone from Broadcom confirm whether the new attributes would work for their 6 tap filter mechanism? 
@ben-gale

SathishKumarTR-Dell:
[1] With this model, it has to be mandated that the host adapter needs to set all the 6 parameters whenever the configuration has to be changed on the port. 
[2] Suggest to change the type of the attributes to "sai_s32_list_t"

dgsudharsan:
> [1] With this model, it has to be mandated that the host adapter needs to set all the 6 parameters whenever the configuration has to be changed on the port.

Hi Sathish,
The set attribute API can set only one attribute at a time. Hence setting all 6 together is impossible. Can someone from broadcom confirm if all 6 are required to be set in a single shot or are there APIs to set them individually?

tonytitus:
Will it be possible to add below SAI serdes params as well(needed for Innovium chip):
    SAI_PORT_ATTR_SERDES_TX_EQ_PRE1
    SAI_PORT_ATTR_SERDES_TX_EQ_PRE2
    SAI_PORT_ATTR_SERDES_TX_EQ_PRE3
    SAI_PORT_ATTR_SERDES_TX_EQ_ATTN
    SAI_PORT_ATTR_SERDES_TX_EQ_MAIN
    SAI_PORT_ATTR_SERDES_TX_EQ_POST
    SAI_PORT_ATTR_SERDES_TX_EQ_POST2

dgsudharsan:
> Will it be possible to add below SAI serdes params as well(needed for Innovium chip):
> SAI_PORT_ATTR_SERDES_TX_EQ_PRE1
> SAI_PORT_ATTR_SERDES_TX_EQ_PRE2
> SAI_PORT_ATTR_SERDES_TX_EQ_PRE3
> SAI_PORT_ATTR_SERDES_TX_EQ_ATTN
> SAI_PORT_ATTR_SERDES_TX_EQ_MAIN
> SAI_PORT_ATTR_SERDES_TX_EQ_POST
> SAI_PORT_ATTR_SERDES_TX_EQ_POST2

Hi,
Whether SAI_PORT_ATTR_SERDES_TX_FIR_MAIN and SAI_PORT_ATTR_SERDES_TX_EQ_MAIN refer the same? Can't we just have one attribute?

tonytitus:
> > Will it be possible to add below SAI serdes params as well(needed for Innovium chip):
> > SAI_PORT_ATTR_SERDES_TX_EQ_PRE1
> > SAI_PORT_ATTR_SERDES_TX_EQ_PRE2
> > SAI_PORT_ATTR_SERDES_TX_EQ_PRE3
> > SAI_PORT_ATTR_SERDES_TX_EQ_ATTN
> > SAI_PORT_ATTR_SERDES_TX_EQ_MAIN
> > SAI_PORT_ATTR_SERDES_TX_EQ_POST
> > SAI_PORT_ATTR_SERDES_TX_EQ_POST2
> 
> Hi,
> Whether SAI_PORT_ATTR_SERDES_TX_FIR_MAIN and SAI_PORT_ATTR_SERDES_TX_EQ_MAIN refer the same? Can't we just have one attribute?

What does FIR stands for?

dgsudharsan:
> > > Will it be possible to add below SAI serdes params as well(needed for Innovium chip):
> > > SAI_PORT_ATTR_SERDES_TX_EQ_PRE1
> > > SAI_PORT_ATTR_SERDES_TX_EQ_PRE2
> > > SAI_PORT_ATTR_SERDES_TX_EQ_PRE3
> > > SAI_PORT_ATTR_SERDES_TX_EQ_ATTN
> > > SAI_PORT_ATTR_SERDES_TX_EQ_MAIN
> > > SAI_PORT_ATTR_SERDES_TX_EQ_POST
> > > SAI_PORT_ATTR_SERDES_TX_EQ_POST2
> > 
> > 
> > Hi,
> > Whether SAI_PORT_ATTR_SERDES_TX_FIR_MAIN and SAI_PORT_ATTR_SERDES_TX_EQ_MAIN refer the same? Can't we just have one attribute?
> 
> What does FIR stands for?

Finite Impulse Response.
https://www.edn.com/Home/PrintView?contentItemId=4458265

tonytitus:
Yes, SAI_PORT_ATTR_SERDES_TX_FIR_MAIN works for us.

kcudnik:
please take a look at errors and warnings and correct:
https://sonic-jenkins.westus2.cloudapp.azure.com/job/sai-meta-pr-build/1779/console

inc/saiport.h
```diff
@@ -2108,6 +2120,233 @@ typedef sai_status_t (*sai_clear_port_pool_stats_fn)(
...
+    /**
+     * @brief Port serdes control TX FIR POST3 filter
+     *
+     * List of port serdes TX fir postcursor3 tap-filter values.
+     * The values are of type sai_u32_list_t where the count is number lanes in
+     * a port and the list specifies list of values to be applied to each lane.
+     *
+     * @type sai_u32_list_t
+     * @flags CREATE_AND_SET
+     * @default internal
```
kcudnik: i would propose to use "empty" value here, if this port_serdes object is created are those list populated internally ? or they empty and user must set values ?

inc/saiport.h
```diff
@@ -1214,6 +1217,15 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_PTP_MODE,
 
+    /**
+     * @brief Serdes object ID for the port
+     *
+     * @type sai_object_id_t
+     * @flags READ_ONLY
```
bandaru-viswanath: Shouldn't this be write-able ? If this is read-only, how would the caller supply the serdes settings ?

inc/saiport.h
```diff
@@ -2108,6 +2120,233 @@ typedef sai_status_t (*sai_clear_port_pool_stats_fn)(
...
+    SAI_PORT_SERDES_ATTR_START,
+
+    /**
+     * @brief Port ID
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_PORT
+     */
+    SAI_PORT_SERDES_ATTR_PORD_ID = SAI_PORT_SERDES_ATTR_START,
```
bandaru-viswanath: Typo : PORD ==> PORT

Can we not reuse a serdes object (basically serdes settings) across similar ports ? Why would we want to tie to a specific port ?

inc/saiport.h
```diff
@@ -2108,6 +2120,233 @@ typedef sai_status_t (*sai_clear_port_pool_stats_fn)(
...
+
+    /**
+     * @brief Port serdes control pre-emphasis
+     *
+     * List of port serdes pre-emphasis values. The values are of type sai_u32_list_t
+     * where the count is number lanes in a port and the list specifies list of values
+     * to be applied to each lane.
+     *
+     * @type sai_u32_list_t
+     * @flags CREATE_AND_SET
```
bandaru-viswanath: Just thinking - Is it preferred to allow CREATE_AND_SET here as opposed to CREATE_ONLY ? I believe that for majority use cases CREATE_ONLY is sufficient, and caller can any way associate a new serdes object for a port in the event that some changes are needed at run-time. On the other hand, supporting SET requires maintain state in SAI implementations. 
I don't have any strong preference either way, but just wanted to highlight an alternate path.

inc/saiport.h
dgsudharsan: Hi kamil,
These values can be internally populated from ASIC specific files (for e.g config.bcm). That's why these were introduced earlier for pre-emphasis(refer existing SAI_PORT_ATTR_SERDES_PREEMPHASIS)

inc/saiport.h
dgsudharsan: No. The SAI_PORT_SERDES object has attribute SAI_PORT_SERDES_ATTRIBUTE_PORT_ID. This attribute is for getting the reverse mapping. much like what we have for getting vlan_member_list in vlan object

inc/saiport.h
dgsudharsan: I am ok with create_only flag.  @kcudnik @tonytitus @lguohan let me know if you are fine with this one.

inc/saiport.h
bandaru-viswanath: The way the serdes object came into being, serdes object is an attribute of port and not the other way round. So if you were to have a reverse-mapping, it should be in the serdes object (READ_ONLY port obj) instead of here.

inc/saiport.h
dgsudharsan: Done

inc/saiport.h
```diff
@@ -1214,6 +1217,15 @@ typedef enum _sai_port_attr_t
...
     SAI_PORT_ATTR_PTP_MODE,
 
+    /**
+     * @brief Serdes object ID for the port
+     *
+     * @type sai_object_id_t
+     * @flags READ_ONLY
+     * @objects SAI_OBJECT_TYPE_PORT_SERDES
+     */
+    SAI_PORT_ATTR_PORT_SERDES_ID,
```
lguohan: usually a port has a number of serdeses, why this not a list of serdes objects and each serdes has its attributes to control.

inc/saiport.h
```diff
@@ -1214,6 +1217,15 @@ typedef enum _sai_port_attr_t
...
+
```
lguohan: serdes is a specific technology, lane would be better IMHO.

inc/saitypes.h
```diff
@@ -262,7 +262,8 @@ typedef enum _sai_object_type_t
     SAI_OBJECT_TYPE_TAM_INT                  = 83,
     SAI_OBJECT_TYPE_COUNTER                  = 84,
     SAI_OBJECT_TYPE_DEBUG_COUNTER            = 85,
-    SAI_OBJECT_TYPE_MAX                      = 86,
+    SAI_OBJECT_TYPE_PORT_SERDES              = 86,
```
lguohan: -> SAI_OBJECT_TYPE_PORT_LANE

inc/saiport.h
```diff
@@ -2128,6 +2367,10 @@ typedef struct _sai_port_api_t
     sai_get_port_pool_stats_fn        get_port_pool_stats;
     sai_get_port_pool_stats_ext_fn    get_port_pool_stats_ext;
     sai_clear_port_pool_stats_fn      clear_port_pool_stats;
+    sai_create_port_serdes_fn         create_port_serdes;
```
lguohan: -> sai_create_port_lane_fn

inc/saiport.h
```diff
@@ -1160,6 +1160,7 @@ typedef enum _sai_port_attr_t
     /**
      * @brief Port serdes control pre-emphasis
      *
+     * Deprecated. Use SAI_OBJECT_TYPE_PORT_SERDES
```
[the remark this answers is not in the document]
kcudnik: Not possible, since compile/linking meta library is independent of user SAI code

inc/saiport.h
lguohan: by default, does the SAI create the port serdes object?

dgsudharsan: Yes. SAI needs to populate serdes object

inc/saiport.h
BrandonCheng0121: hi,
I have questions about usage of  typedef sai_status_t (*sai_create_port_serdes_fn)( ).
What attributes should be provided to create serdes object?   Specify switch_id,  SAI_PORT_SERDES_ATTR_PORT_ID, and what else?  
We know a port has a number of serdeses, why the output parameter is only one serdes ID instead of serdes ID list?

BR.
@dgsudharsan 
@lguohan

BrandonCheng0121:
> [1] With this model, it has to be mandated that the host adapter needs to set all the 6 parameters whenever the configuration has to be changed on the port.
> 
> Hi Sathish,
> The set attribute API can set only one attribute at a time. Hence setting all 6 together is impossible. Can someone from broadcom confirm if all 6 are required to be set in a single shot or are there APIs to set them individually?

Hi dgsudharsan,
In my experience, broadcom has a API to set all 6 attributes in a single shot.

inc/saiport.h
dgsudharsan: > hi,
> I have questions about usage of typedef sai_status_t (*sai_create_port_serdes_fn)( ).
> What attributes should be provided to create serdes object? Specify switch_id, SAI_PORT_SERDES_ATTR_PORT_ID, and what else?
> We know a port has a number of serdeses, why the output parameter is only one serdes ID instead of serdes ID list?
> 
> BR.
> @dgsudharsan
> @lguohan

Hi Brandon,
If you check the attributes are per port and not per lane. The attributes are set as a list for a port(list size equal to number of lanes). Hence we would be having only one serdes attribute per port.

dgsudharsan:
> > [1] With this model, it has to be mandated that the host adapter needs to set all the 6 parameters whenever the configuration has to be changed on the port.
> > Hi Sathish,
> > The set attribute API can set only one attribute at a time. Hence setting all 6 together is impossible. Can someone from broadcom confirm if all 6 are required to be set in a single shot or are there APIs to set them individually?
> 
> Hi dgsudharsan,
> In my experience, broadcom has a API to set all 6 attributes in a single shot.

Hi Brandon,
Yes. That is the final approach taken.  During the discussion it is decided that if serdes settings need to be changed for port, the existing object is deleted and new object is created with new attributes.

inc/saiport.h
BrandonCheng0121: Hi dgsudharsan,
In your opinion, a lane and serdes aren't similar things?
Each port has only one serdes object ID representing/including a number of lanes, right?

inc/saiport.h
BrandonCheng0121: Hi dgsudharsan,
These values can be populated from ASIC specific files (for e.g config.bcm) indeed.
If applications want to SET the better values in runtime, the flags should be CREATE_AND_SET, am I right?

@dgsudharsan

inc/saiport.h
BrandonCheng0121: Hi dgsudharsan,
If execute sai_remove_port_serdes_fn)() to remove the serdes objecet for a port, what will happen?
Does it mean this port has no serdes to use and this port's status is down?

BrandonCheng0121:
> [1] With this model, it has to be mandated that the host adapter needs to set all the 6 parameters whenever the configuration has to be changed on the port.
> Hi Sathish,
> The set attribute API can set only one attribute at a time. Hence setting all 6 together is impossible. Can someone from broadcom confirm if all 6 are required to be set in a single shot or are there APIs to set them individually?
> 
> Hi dgsudharsan,
> In my experience, broadcom has a API to set all 6 attributes in a single shot.
> 
> Hi Brandon,
> Yes. That is the final approach taken.  During the discussion it is decided that if serdes settings need to be changed for port, the existing object is deleted and new object is created with new attributes.

Hi dgsudharsan,
In my experience, when using attribute SAI_PORT_ATTR_SERDES_PREEMPHASIS, the value used by a lane is only 32 bits length. This length may be not enough for 6 tap.  Now some BRCM chips only need 3 tap and each tap value is not very long, which could put 3 tap in the 32 bits values. Now your this pull request has more flexible for each tap to set its own 32 bits value.

inc/saiport.h
BrandonCheng0121: Hi dgsudharsan,
Defining the type of these port_serdes_attr (pre1, pre2...) as **sai_s32_list_t** may be better. Because they are possible to be negative values.

PR #1004, opened 2019-09-20
Gearbox proposal spec 

AshokDaparthi:
What I Did:
Proposal draft spec for External PHY abstraction interface.


inc/saiswitch.h
```diff
@@ -1827,6 +1880,195 @@ typedef enum _sai_switch_attr_t
...
+     * @brief Platform adaption device write callback function passed to the adaptor.
+     * This is mandatory function for driver when device access not supported by file system.
+     *
+     * Use sai_switch_register_write_fn as write function.
+     *
+     * @type sai_pointer_t sai_switch_register_write_fn
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @condition SAI_SWITCH_ATTR_HARDWARE_ACCESS_SUPPORT_BY_SYSFS == false
+     */
+    SAI_SWITCH_ATTR_REGISTER_READ,
```
tushar-ty: typo - shud be write.

inc/saiswitch.h
```diff
@@ -1995,6 +2237,62 @@ typedef void (*sai_switch_state_change_notification_fn)(
...
+ * @objects switch_id SAI_OBJECT_TYPE_SWITCH
+ *
+ * @param[in] platform_context Platform context information.
+ * @param[in] switch_id Switch Id
+ * @param[in] device_addr Device address(PHY/lane/port MDIO address)
+ * @param[in] start_reg_addr Starting register address to write
+ * @param[in] number_of_registers Number of consecutive registers to write
+ * @param[in] reg_val Register write values
+ */
+typedef sai_status_t (*sai_switch_register_write_fn)(
```
kcudnik: could those functions be somehow transformed to attributes ?, if not then they will need some extra handling in sai metadata since they dont follow any currently supported format

inc/saiswitch.h
```diff
@@ -1827,6 +1880,195 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * Platform context information provided by the host adaptor to driver.
+     * This information is Host adaptor specific, typically used for maintain
+     * synchronization and device information. Driver will give this context back
+     * to adaptor as part of call back sai_switch_register_read/write_fn API.
+     *
+     * Use void as pointer.
+     *
+     * @type sai_pointer_t void
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
```
kcudnik: why this is mandatory on created ? seems like platform context should be something that should be obtained from switch which was created after switch creation

inc/saiswitch.h
```diff
@@ -1995,6 +2237,62 @@ typedef void (*sai_switch_state_change_notification_fn)(
...
+ *
+ * @param[in] platform_context Platform context information.
+ * @param[in] switch_id Switch Id
+ * @param[in] device_addr Device address(PHY/lane/port MDIO address)
+ * @param[in] start_reg_addr Starting register address to read
+ * @param[in] number_of_registers Number of consecutive registers to read
+ * @param[out] reg_val Register read values
+ */
+typedef sai_status_t (*sai_switch_register_read_fn)(
+        _In_ const void *platform_context,
```
kcudnik: why platform context is extra parameter since you have switch id already for identifying switch itself

kcudnik: also please avoid void* abitrary data, since it will be impossible to guess whats there and how to pass this over RPC (which is currently done in sonic), please consider using switch attributes, without any pointers etc.

inc/saiswitch.h
AshokDaparthi: This can not be attribute, everything need to access PHY. Could you please help me to add special handling.

inc/saiswitch.h
AshokDaparthi: PHY driver will call back API to access registers to NOS. NOS will take care of reaching PHY, During this time we need protection(locking) for bus access. PHY register access will happen during the switch_create also. So this it opaque date to PHY driver, NOS will provide during switch create and NOS will get back from PHY driver. Data inside this specific to NOS implementation. Like other  notification call back API's, pointer can be exchanged between NOS and PHY driver.

kcudnik: not sure if i understand this fully, switch ID should be enough to identify registers for expected switch, internally SAI can associate any platform context with returned switch ID. there should be no callback for calling such APIs like this one. Those should be synchronous APIs and any necessary locking should happen inside internal vendor SAI implementation. Other vendors may not have concept of "context" etc. If you really want callback for reading registers (which i should advise to not go this path) then declare notification for switch like other notifications .
IMO this should be synchronous API

kcudnik: @lguohan Can can take a look at this? Since concept of registers would be per platform and it seems like it can't be stored as attributes, then some special cases and database would need to be added in SONIC if some of registers would be modified

inc/saiswitch.h
lguohan: agree, @AshokDaparthi , can you correct?

inc/saiswitch.h
kcudnik: easiest way is to add those 2 functions for ignore from check under style.pm, and look for function CheckFunctionNaming, you will see exceptions there (twice)

inc/saiswitch.h
```diff
@@ -1433,11 +1482,15 @@ typedef enum _sai_switch_attr_t
     /**
      * @brief Port state change notification callback function passed to the adapter.
      *
+     * In case driver does not support this attribute, The Host adaptor should pool
```
rameshsanth: did you mean poll ?

inc/saiswitch.h
```diff
@@ -1827,6 +1880,202 @@ typedef enum _sai_switch_attr_t
...
+    /**
+     * @brief Platform adaption device read callback function passed to the adaptor.
+     * This is mandatory function for driver when device access not supported by file system.
+     *
+     * Use sai_switch_register_read_fn as read function.
+     *
+     * @type sai_pointer_t sai_switch_register_read_fn
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @condition SAI_SWITCH_ATTR_HARDWARE_ACCESS_SUPPORT_BY_SYSFS == false
+     * @switchtype PHY
```
kcudnik: switchtype tag is not supported, what does it mean anyway ? please remove 
this seems like this tag applies to limited number of attributes, and cant be applied on any attribute

inc/saiswitch.h
lguohan: @AshokDaparthi , syncd with tian, can you check the type to sai_uint64_t?

inc/saiport.h
```diff
@@ -1214,6 +1252,33 @@ typedef enum _sai_port_attr_t
...
+     * @type sai_port_interface_type_t
+     * @flags CREATE_AND_SET
+     * @default internal
+     */
+    SAI_PORT_ATTR_INTERFACE_TYPE,
+
+    /**
+     * @brief Configure port reference clock in Megahertz
+     *
+     * @type sai_uint32_t
```
AshokDaparthi: Use unit64

inc/saiport.h
```diff
@@ -1214,6 +1252,33 @@ typedef enum _sai_port_attr_t
...
+     * @flags CREATE_ONLY
+     * @default internal
+     */
+    SAI_PORT_ATTR_REF_CLOCK,
```
lguohan: use uint64 and the unit is hz

inc/saiswitch.h
lguohan: also change this platform_context to u64

inc/saiswitch.h
```diff
@@ -1827,6 +1880,202 @@ typedef enum _sai_switch_attr_t
...
+     */
+    SAI_SWITCH_ATTR_PLATFROM_CONTEXT,
+
+    /**
+     * @brief Platform adaption device read callback function passed to the adaptor.
+     * This is mandatory function for driver when device access not supported by file system.
+     *
+     * Use sai_switch_register_read_fn as read function.
+     *
+     * @type sai_pointer_t sai_switch_register_read_fn
```
kcudnik: currently all pointers declared in switch are considered notifications and must have _notification_t suffix, and _NOTIFY suffix on attribute name, an notifications functions must be declared void at this point.
This will pop up after you fix existing errors

kcudnik: also im a little bit confused, why this function neds to be set here on attributes if we already have declared them in struct api:
 sai_switch_register_read_fn     switch_register_read;
 sai_switch_register_write_fn    switch_register_write;

AshokDaparthi: @kcudnik - Below 2 API's are given back by ASIC SAI implementation to provide PHY access by MDIO interface which is connected to NPU.
sai_switch_register_read_fn switch_register_read;
sai_switch_register_write_fn switch_register_write;

Platform adaption will have  has 2 parts.  PHY driver will call the call back API's to NOS, this is provided as part of NOTIFICATION attribute. After coming to NOS, in case PHY is connected through NPU then  NOS will call those API's to write data to PHY.

meta/parse.pl
```diff
@@ -120,6 +120,10 @@ sub ProcessTagType
 
     return $val if $val =~ /^sai_pointer_t sai_\w+_notification_fn$/;
 
+    return $val if $val =~ /^sai_pointer_t sai_\w+_register_read_fn$/;
```
kcudnik: not sure if this make sense here to add every function, as mentioned before all pointers in switch are a notification type and follow convention of notifications. I would suggest to change names to be named as notifications scheme. There is logic later on which generates automatic structures from that and enum values.

As asked before, why we have those exact functions here if we have same functions in API structure ?
Can you describe purpose of those functions on API structure and on attributes ?

inc/saiswitch.h
kcudnik: please then follow convention for notification scheme as other pointers

kcudnik:
(on line 217): sai_switch_attr_t: values are not increasing by 1: last: 164 current: 166, 
this error is probably happening because of attribute firmware_broadcast_end, since "end" suffix is special its removed automatically, and it should not be used, could you rename the attribute ?

inc/saiswitch.h
kcudnik: @AshokDaparthi Why those functions needs to be given back by SAI ?, you have declared those 2 functions in api structure, no need to have them as parameters, i would suggest to remove those or make those as notification
@lguohan Lets have meeting to explain what those are needed for

inc/saiswitch.h
```diff
@@ -1995,6 +2260,62 @@ typedef void (*sai_switch_state_change_notification_fn)(
...
+ * @param[in] platform_context Platform context information.
+ * @param[in] switch_id Switch Id
+ * @param[in] device_addr Device address(PHY/lane/port MDIO address)
+ * @param[in] start_reg_addr Starting register address to read
+ * @param[in] number_of_registers Number of consecutive registers to read
+ * @param[out] reg_val Register read values
+ */
+typedef sai_status_t (*sai_switch_register_read_fn)(
+        _In_ uint64_t platform_context,
+        _In_ sai_object_id_t switch_id,
```
lguohan: what is this switch_id?

kcudnik:
added some starting point fro callback functions #1008 it should be easier to pass some checks after fixes here, then i will add proper callback handling and auto generating structs for callbacks

inc/saiswitch.h
```diff
@@ -1827,6 +1893,205 @@ typedef enum _sai_switch_attr_t
...
+    SAI_SWITCH_ATTR_REGISTER_READ,
+
+    /**
+     * @brief Platform adaption device write callback function passed to the adaptor.
+     * This is mandatory function for driver when device access not supported by file system.
+     *
+     * Use sai_switch_register_write_fn as write function.
+     *
+     * @type sai_pointer_t sai_switch_register_write_fn
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
```
kcudnik: make create_and set?

inc/saiswitch.h
```diff
@@ -1827,6 +1893,205 @@ typedef enum _sai_switch_attr_t
...
+    SAI_SWITCH_ATTR_PLATFROM_CONTEXT,
+
+    /**
+     * @brief Platform adaption device read callback function passed to the adaptor.
+     * This is mandatory function for driver when device access not supported by file system.
+     *
+     * Use sai_switch_register_read_fn as read function.
+     *
+     * @type sai_pointer_t sai_switch_register_read_fn
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
```
kcudnik: make create_and set?

meta/parse.pl
kcudnik: i reformatted this in #1008

inc/saiswitch.h
AshokDaparthi: This API and mandatory on create and CREATE_ONLY on mentioned condition

inc/saiswitch.h
AshokDaparthi: @kamil , @lguohan -  Yes, we can remove switch_id, Because while switch_create. we do not have switch_id created in NOS.

inc/saiswitch.h
kcudnik: what if user want to change this field during runtime ? for example for notification functtion this is possible

inc/saiswitch.h
kcudnik: i was thinking more like this, if same callback function would be used in many gearboxes, then switch_id would uniquely identify which switch we want to talk to, but this role should also be the same for platform_context, so one of them seems to be redundant

inc/saiswitch.h
lguohan: the function is part of infrastructure, current assumption is that you cannot change it.

kcudnik:
Just to clarify, we can't have "validonly" or "condition" when condition includes read_only attribute, since value of that attribute is not know and compilation phase, only on runtime.
Also making conditions like that on non created switch make no sense since switch don't exists yet

meta/saisanitycheck.c
```diff
@@ -2289,6 +2290,9 @@ void check_attr_existing_objects(
 
             META_MD_ASSERT_FAIL(md, "object list is not supported on this object type");
 
+        case SAI_ATTR_VALUE_TYPE_POINTER:
+
```
kcudnik: add comment that this is for read/write api ?

meta/saisanitycheck.c
```diff
@@ -1664,7 +1664,8 @@ void check_attr_key(
             case SAI_ATTR_VALUE_TYPE_OBJECT_ID:
 
                 if ((md->objecttype == SAI_OBJECT_TYPE_QUEUE && md->attrid == SAI_QUEUE_ATTR_PORT) ||
-                        (md->objecttype == SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP && md->attrid == SAI_INGRESS_PRIORITY_GROUP_ATTR_PORT))
+                    (md->objecttype == SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP && md->attrid == SAI_INGRESS_PRIORITY_GROUP_ATTR_PORT) ||
+                    (md->objecttype == SAI_OBJECT_TYPE_PORT_CONNECTOR && md->attrid == SAI_PORT_CONNECTOR_ATTR_SYSTEM_SIDE_PORT_ID) || (md->objecttype == SAI_OBJECT_TYPE_PORT_CONNECTOR && md->attrid == SAI_PORT_CONNECTOR_ATTR_LINE_SIDE_PORT_ID))
```
kcudnik: minor to break the line on each condition

inc/saiport.h
```diff
@@ -1214,6 +1252,35 @@ typedef enum _sai_port_attr_t
...
+     * @default internal
+     */
+    SAI_PORT_ATTR_REFERENCE_CLOCK,
+
+    /**
+     * @brief Port PRBS Polynomial
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default internal
```
lguohan: can the default value be zero?

inc/saiswitch.h
kcudnik: just to clarify after meetings, switch id will be removed from this callback since that should be passed to other SAI as a callback, and it would make no sens for other npu understanding switch id from other npu

inc/saiswitch.h
```diff
@@ -1433,6 +1494,9 @@ typedef enum _sai_switch_attr_t
     /**
      * @brief Port state change notification callback function passed to the adapter.
      *
+     * In case driver does not support this attribute, The Host adaptor should poll
```
SidharajU: There are minor typos "adaptor" seen multiple places

inc/saiport.h
```diff
@@ -136,6 +136,9 @@ typedef enum _sai_port_media_type_t
 
     /** Media type copper. Remote advertise medium information as copper */
     SAI_PORT_MEDIA_TYPE_COPPER,
+
```
apogedik: This comment is applicable to all files.  The copyright statement is not up to date.  And I do not see Broadcom listed anywhere as a contributor.

Copyright (c) 2014 Microsoft Open Technologies, Inc.
 *
 *    Licensed under the Apache License, Version 2.0 (the "License"); you may
 *    not use this file except in compliance with the License. You may obtain
 *    a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
 *
 *    THIS CODE IS PROVIDED ON AN *AS IS* BASIS, WITHOUT WARRANTIES OR
 *    CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT
 *    LIMITATION ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS
 *    FOR A PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
 *
 *    See the Apache Version 2.0 License for specific language governing
 *    permissions and limitations under the License.
 *
 *    Microsoft would like to thank the following companies for their review and
 *    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
 *    Dell Products, L.P., Facebook, Inc., Marvell International Ltd.

lguohan: @apogedik, broadcom has been a big contributor to the SAI, and we can certainly add broadcom. Since it is not specific to gear box. I suggest to open another PR to add broadcom name to all header files.

apogedik: Thanks.  Who creates PRs?  I am not familiar with the process.

lguohan: can you sync with Tushar?

PR #1006, opened 2019-10-21
Improve debug counter enums

daall:
* Add START and END attributes for drop reason enums
* Add additional port stats for debug counters
* Add additional switch stats for debug counters

### DIFF
--- a/inc/saidebugcounter.h
+++ b/inc/saidebugcounter.h
@@ -87,7 +87,7 @@ typedef enum _sai_in_drop_reason_t
     /* L2 reasons */
 
     /** Any L2 pipeline drop */
-    SAI_IN_DROP_REASON_L2_ANY,
+    SAI_IN_DROP_REASON_L2_ANY = SAI_IN_DROP_REASON_START,
 
     /** Source MAC is multicast */
     SAI_IN_DROP_REASON_SMAC_MULTICAST,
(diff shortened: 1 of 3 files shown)


inc/saidebugcounter.h
```diff
@@ -81,6 +81,9 @@ typedef enum _sai_debug_counter_bind_method_t
  */
 typedef enum _sai_in_drop_reason_t
 {
+    /** Start of in drop reasons */
+    SAI_IN_DROP_REASON_START,
```
kcudnik: sai_in_drop_reason_t: first enum should start with zero
assign this value to the first item, check _sai_hostif_trap_type_t

inc/saidebugcounter.h
```diff
@@ -310,6 +319,9 @@ typedef enum _sai_in_drop_reason_t
  */
 typedef enum _sai_out_drop_reason_t
 {
+    /** Start of out drop reasons */
+    SAI_OUT_DROP_REASON_START,
```
kcudnik: same here

inc/saiport.h
```diff
@@ -1741,12 +1741,54 @@ typedef enum _sai_port_stat_t
...
+
+    /** Get in port packet drops configured by debug counter API at index 5 */
+    SAI_PORT_STAT_IN_CONFIGURED_DROP_REASONS_5_DROPPED_PKTS,
+
+    /** Get in port packet drops configured by debug counter API at index 6 */
+    SAI_PORT_STAT_IN_CONFIGURED_DROP_REASONS_6_DROPPED_PKTS,
+
+    /** Get in port packet drops configured by debug counter API at index 7 */
+    SAI_PORT_STAT_IN_CONFIGURED_DROP_REASONS_7_DROPPED_PKTS,
+
```
fengyingling, line 1764: What exactly do these attributes mean? It would be best if there are some examples -_-？

fengyingling, line 1764: What exactly do these attributes mean? It would be best if there are some examples -_-？

PR #1012, opened 2019-11-01
Add support for deprecated tag in attribute metadata

kcudnik:


kcudnik:
should o be, since we want to keep backward compatibility and enum order,  but this topic could be brought up on SAI weekly meeting

PR #1023, opened 2019-12-10
Added SAI_SWITCH_ATTR_CHIP_RESET_MODE

apogedik:


inc/saiswitch.h
```diff
@@ -303,6 +303,21 @@ typedef enum _sai_switch_type_t
 
 } sai_switch_type_t;
 
+/**
+ * @brief Run time chip reset mode options
```
atitjain: There are several attributes already present which are used for the reset type for the chip. Please check. 

` /**
     * @brief Set Type of reboot WARM/COLD
     *
     * Indicates controlled warm restart.
     * Since warm restart can be caused by crash
     * (therefore there are no guarantees for this call),
     * this hint is really a performance optimization.
     * This hint is set as part of the shutdown sequence, before boot.
     * TRUE - Warm Reboot
     * FALSE - Cold Reboot
     *
     * @type bool
     * @flags CREATE_AND_SET
     * @default false
     */
    SAI_SWITCH_ATTR_RESTART_WARM,

    /**
     * @brief Warm boot recovery
     *
     * Start warm boot recovery when set to true
     * This hint is set after boot.
     * In case of host adapter restart, host adapter can pass boot type in
     * #SAI_KEY_BOOT_TYPE. In case of host adapter recovery, host adapter can
     * pass a hint about the boot type and recovery, in this flag.
     *
     * @type bool
     * @flags CREATE_AND_SET
     * @default false
     */
    SAI_SWITCH_ATTR_WARM_RECOVER,

    /**
     * @brief Type of restart supported
     *
     * @type sai_switch_restart_type_t
     * @flags READ_ONLY
     */
    SAI_SWITCH_ATTR_RESTART_TYPE,`

apogedik: We presented and discussed the same points you mention above at the community meeting.  This proposal is not the same as Warm/cold boot.  This was specifically requested to enable user to soft/hard reset the chip at run time.

rlhui: @apogedik , wanted to circle back on this PR.  Is existing solution meeting the requirement?  Thanks.

PR #1024, opened 2019-12-15
Support for static FDB Entries to allow MAC Move

bandaru-viswanath:


inc/saifdb.h
```diff
@@ -178,6 +178,16 @@ typedef enum _sai_fdb_entry_attr_t
...
 
+    /**
+     * @brief Specifies whether a MAC move is allowed
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     * @validonly SAI_FDB_ENTRY_ATTR_TYPE == SAI_FDB_ENTRY_TYPE_STATIC
+     */
+    SAI_FDB_ENTRY_ATTR_ALLOW_MAC_MOVE,
```
rck-innovium: Can you elaborate on the use-case where we want to allow MAC_MOVE of a static MAC address?

dgsudharsan:
Have you checked this? https://github.com/opencomputeproject/SAI/pull/696
In this case when a static move is detected the application will be notified and it can re-program the static MAC on the port.

inc/saifdb.h
srj102: As discussed in the SAI community meeting on Dec 19 the use case is for EVPN wherein the MACs donot age out as they are created and deleted by the EVPN control plane (BGP) . In that respect they are static MACs. However these MACs must be subjected to a MAC Move. For e.g. a MAC installed by BGP moves from the remote to a local port. This should trigger a hardware learn event. The SAI attribute is to allow for the hardware to learn the MAC even though the MAC is programmed as static.

srj102:
> Have you checked this? #696
> In this case when a static move is detected the application will be notified and it can re-program the static MAC on the port.

696 is after a static move is detected. Here the setting is to allow for a static move as described in the previous reply.

srj102:
> Have you checked this? #696
> In this case when a static move is detected the application will be notified and it can re-program the static MAC on the port.

The #696 has a trap type added for static fdb mac move. 
   However I do not see the SAI_SWITCH_ATTR_STATIC_FDB_MOVE_PACKET_ACTION     merged to saiswitch.h though it is mentioned in the PR. 

   I am not sure if this was intentional. Without this the added trap  does not seem to make sense. 
 
   The above packet action even if added is a switch wide attribute which is applicable to locally configured static MACs as well. We may not want this and would like to 
   differentiate from an EVPN MAC.  In other words we would like a drop if there is a mac move experienced for a local static but at the same time would like to 
   allow EVPN MACs to be subjected to MAC Move. 

   The EVPN remote macs though programmed as static are not the same as locally configured static MACs. These are dynamically learnt MACs on a remote 
   switch.

inc/saifdb.h
lguohan: can you add some clarification on the TRAP attributed introduced in #696 ? What is the expectation of that notification when MAC_MOVE is enabled or disabled on this attribute.

lguohan:
@marian-pritsak  and @itaibaz  to check the behavior.

inc/saifdb.h
bandaru-viswanath: The trap introduced in #696 is a generic trap, where as this PR is about a specific MAC entry. So the specific entry would take precedence over a generic trap. I hope this clarifies.

bandaru-viswanath: (Additional clarification) When MAC_MOVE is explicitly disabled for a static mac-entry, the trap (introduced in #696) would also not be generated.

marian-pritsak: Is this also expected to work with MLAG?

The behavior of this entry is closer to being "dynamic but not ageable" than "static but movable" because static FDB is not supposed to be changed by anything besides the application that created it, whereas dynamic is expected to change, and we just say that it does not expire. Let me know if you considered this way of definition.

srj102: Yes this is expected to work with MCLAG as well. Please refer to the MC Lag enhancements HLD floated in sonic forum. Static vs dynamic refers to a MAC being ageable and not whether they are movable. Here we are adding an attribute to static mac specifying if this is moveable or not.

lguohan: @bandaru-viswanath, thanks for the clarification. can you make the clarification in the header?

rlhui:
@kcudnik , would you please review/approve also?  thanks

inc/saifdb.h
bandaru-viswanath: @lguohan Updated the header file.

PR #1026, opened 2019-12-15
Support for turning on/off ARP & NDP suppression per vlan

bandaru-viswanath:


inc/saivlan.h
```diff
@@ -386,6 +386,19 @@ typedef enum _sai_vlan_attr_t
...
+     *
+     * When enabled, the ARP and NDP messages are trapped to CPU instead of
+     * getting copied (and flooding the vlan). Default is set to false
+     * for backward compatibility reasons
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_VLAN_ATTR_NEIGH_SUPPRESSION_ENABLE,
```
kcudnik: why not use full "NEIGHBOR" instead NEIGH?

bandaru-viswanath: Thanks Kamil, I changed the macro name.

inc/saivlan.h
```diff
@@ -386,6 +386,19 @@ typedef enum _sai_vlan_attr_t
...
+     *
+     * When enabled, the ARP and NDP messages are trapped to CPU instead of
+     * getting copied (and flooding the vlan). Default is set to false
+     * for backward compatibility reasons
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_VLAN_ATTR_NEIGHBOR_SUPPRESSION_ENABLE,
```
lguohan: maybe an attribute SAi_VLAN_ATTR_TRAP_DISABLE_LIST, which is a trap object list.

atitjain: the behavior can be achieved by creating a trap of type 
`
    /** Default packet action is forward */
    SAI_HOSTIF_TRAP_TYPE_ARP_REQUEST = 0x00002000,

    /** Default packet action is forward */
    SAI_HOSTIF_TRAP_TYPE_ARP_RESPONSE = 0x00002001,
`
and then add the entry with type 
`
    /** Vlan based Host Interface entry Type */
    SAI_HOSTIF_TABLE_ENTRY_TYPE_VLAN,
`
and define the action on the entry.

bandaru-viswanath:
This PR is being closed down based on the community discussion earlier, and a new-and-generic PR #1055 is submitted for acceptance.

bluecmd:
@bandaru-viswanath Are there any plans on pushing ARP/NDP suppression further? It seems like it would be a pre-req. for having support for this in SONiC from what I can deduce from https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/EVPN/EVPN_VXLAN_HLD.md

PR #1029, opened 2020-01-12
Change debug counter index to mandatory on create

kcudnik:
Motivation here is, that instead of index assigned by vendor user can
specify index. attribute is also marked as KEY so each index must be
unique. This will also allow SONiC comparison logic to correctly match
debug counter.

### DIFF
--- a/inc/saidebugcounter.h
+++ b/inc/saidebugcounter.h
@@ -364,14 +364,14 @@ typedef enum _sai_debug_counter_attr_t
      */
     SAI_DEBUG_COUNTER_ATTR_START,
 
-    /* READ-ONLY */
+    /* CREATE-ONLY */
 
     /**
      * @brief Object stat index
      * Index is added to base start
      *
      * @type sai_uint32_t
-     * @flags READ_ONLY
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY | KEY
      */
     SAI_DEBUG_COUNTER_ATTR_INDEX = SAI_DEBUG_COUNTER_ATTR_START,
 


kcudnik:
Great, lets have community discussion again

inc/saidebugcounter.h
```diff
@@ -364,14 +364,14 @@ typedef enum _sai_debug_counter_attr_t
...
-    /* READ-ONLY */
+    /* CREATE-ONLY */
 
     /**
      * @brief Object stat index
      * Index is added to base start
      *
      * @type sai_uint32_t
-     * @flags READ_ONLY
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY | KEY
```
lguohan, line 374: to maintainer backward compatibility, I suggest to keep this attribute as it is.

Then, you can introduction another CREATE_ONLY | KEY attribute

SAI_DEBUG_COUNTER_ATTR_USER_INDEX

the default value is 0, which mean let hardware to choose the index. Then, you can query the SAI_DEBUG_COUNTER_ATTR_INDEX to get the actually index.

if the value is not 0, then if we read SAI_DEBUG_COUNTER_ATTR_INDEX it will be the USER_INDEX.

kcudnik: index 0 maybe also valid index, and since that will be key, then only 1 object with default value can be created. KEY attributtes must be mandatory on create and create only, there is no default values for them

lguohan: then, can we have a UNIQUE attribute which means unique only. the default value can be 0xffffffff

kcudnik: dont quite get that, can you explain more?

kcudnik:
discussed online, index is not used on any other object, only for user information

PR #1030, opened 2020-01-14
Add headers integrity check script

kcudnik:


meta/Makefile
```diff
@@ -76,6 +76,7 @@ OBJ = saimetadata.o saimetadatautils.o saiserialize.o
 SYMBOLS = $(OBJ:=.symbols)
 
 all: saisanitycheck saimetadatatest saiserializetest saidepgraph.svg $(SYMBOLS)
+	./checkheaders.pl ../inc ../inc
```
lguohan: same folder?

kcudnik:
Yes since Sai dont know any other folder and this check checks od script compiles

meta/Makefile
kcudnik: Yes since Sai dont know any other folder and this check checks od script compiles

PR #1032, opened 2020-01-21
sai macsec flow statistics added

prafull-brcm:


dipankar-ba:
I thought these statistics should be per interface.  That is why those are currently defined for sai_macsec_port.
Each sai_macsec_flow attaches to a set of ACL rules and (an SC or a set of ingress SC).  I do not think there is any statistics counter defined for such an object.

prafull-brcm:
New pull request will be opened for same change after more discussion.

PR #1035, opened 2020-01-22
Add more restart types for hard and soft reset to sai_switch_restart_…

apogedik:
Add more restart types for hard and soft reset


rlhui:
@apogedik , checking if existing header meet the needs? Thanks.

PR #1037, opened 2020-01-24
SAI MACsec port statistics are moved to flow statistics

prafull-brcm:


inc/saimacsec.h
```diff
@@ -1273,13 +1292,13 @@ typedef struct _sai_macsec_api_t
     sai_remove_macsec_port_fn           remove_macsec_port;
     sai_set_macsec_port_attribute_fn    set_macsec_port_attribute;
     sai_get_macsec_port_attribute_fn    get_macsec_port_attribute;
-    sai_get_macsec_port_stats_fn        get_macsec_port_stats;
```
atitjain: Any considerations for the backward compatibility. Cant these be retained along with finer granularity flow stats.

inc/saimacsec.h
```diff
@@ -345,117 +345,136 @@ typedef enum _sai_macsec_port_attr_t
...
     /**
-     * @brief IEEE 802.1ae defined ifOutOctets or ifInOctets for MACSEC uncontrolled port
+     * @brief List of MACsec ACL entries associated with this flow.
+     *
+     * @type sai_object_list_t
+     * @flags READ_ONLY
+     * @objects SAI_OBJECT_TYPE_ACL_ENTRY
      */
-    SAI_MACSEC_PORT_STAT_OCTETS_UNCONTROLLED,
+    SAI_MACSEC_FLOW_ATTR_ACL_ENTRY_LIST,
```
atitjain: Why not use the match list directly without hinting ACLs. ACLs might have scale impact for large number of flow and there might be an implementation that uses an exact match table or any other way.

inc/saimacsec.h
prafull-brcm: This is good point we can keep port stats function for backward compatibility.

inc/saimacsec.h
prafull-brcm: ACL entry is rule match table associated with flow. This list indicates number of ACL rule entry attached to this MACsec flow. ACL entry are rule match table it can be attached to any flow.

DeanKalla:
Hi Prafull,

Could you update the macsec document with the attached version 1.4 in your PR? Dipankar made changes to clarify macsec_flow object and its relation to macsec_sc.
[SAI_MACsec_API_Proposal-v1.4.docx](https://github.com/opencomputeproject/SAI/files/4161842/SAI_MACsec_API_Proposal-v1.4.docx)

PR #1040, opened 2020-01-28
Add API definitions to support a L4 session-aware load balancer.

michazag:
An L4 Service Load Balancer (L4LB) serves as a single destination for a service running on multiple functionally equivalent endpoints. Traffic that is received on the L4-LB service is distributed among multiple endpoints - servers, microservices etc.

### DIFF
--- a/inc/sainexthopgroup.h
+++ b/inc/sainexthopgroup.h
@@ -44,6 +44,12 @@ typedef enum _sai_next_hop_group_type_t
     /** Next hop protection group. Contains primary and backup next hops. */
     SAI_NEXT_HOP_GROUP_TYPE_PROTECTION,
 
+    /**
+     * @brief Next hop group is a L4 session aware load balancer. Guarantees the
+     * session consistency across the group updates.
+     */
+    SAI_NEXT_HOP_GROUP_TYPE_L4_SESSION_CONSISTENT,
+
     /* Other types of next hop group to be defined in the future, e.g., WCMP */
 
 } sai_next_hop_group_type_t;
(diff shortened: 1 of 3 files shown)


inc/sainexthopgroup.h
```diff
@@ -133,6 +202,17 @@ typedef enum _sai_next_hop_group_attr_t
      */
     SAI_NEXT_HOP_GROUP_ATTR_COUNTER_ID,
 
+    /**
+     * @brief Maximum number of entries that the group can handle.
+     * If each entry corresponds to a session, this is the maximal number of sessions.
```
kcudnik: too long brief message, add empty comment line before first "If"

inc/sainexthopgroup.h
```diff
@@ -219,6 +300,26 @@ typedef enum _sai_next_hop_group_member_attr_t
      */
     SAI_NEXT_HOP_GROUP_MEMBER_ATTR_MONITORED_OBJECT,
 
+    /**
+     * @brief Member admin state. Allows for disabling a member without removing it.
+     *
+     * Should only be used if the type of owning group is SAI_NEXT_HOP_GROUP_TYPE_L4_SESSION_CONSISTENT
+     *
+     * @type sai_next_hop_group_member_admin_state_t
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
```
tushar-ty: This is not backward compatible. 
Can this be optional with a default of SAI_NEXT_HOP_GROUP_MEMBER_OPER_STATE_ACTIVE ?

michazag: Thanks, good point. 
I think you meant ADMIN_STATE (OPER_STATE is READ_ONLY) - and yes, I'll change to optional with default as 'SAI_NEXT_HOP_GROUP_MEMBER_ATTR_ADMIN_STATE_ACTIVE'

lguohan:
@michazag , we will like need a spec for this proposal to discuss the motivation and behaviror model.

https://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Proposal-IPv6_Segment_Routing-1.md

michazag:
Uploaded the proposal document. Currently in word format, will convert to .md

marian-pritsak:
I noticed we are missing the weight attribute of the NG member. Should we add that?

michazag:
> I noticed we are missing the weight attribute of the NG member. Should we add that?

There is the attribute below which is part of "sai_next_hop_group_member_attr_t". Do you mean some other structure/API where it is missing?
  /**
     * @brief Member weights
     *
     * @type sai_uint32_t
     * @flags CREATE_AND_SET
     * @default 1
     */
    SAI_NEXT_HOP_GROUP_MEMBER_ATTR_WEIGHT,

marian-pritsak:
My mistake. Somehow I didn't find this attribute.

lguohan:
looks like we are missing examples.

inc/sainexthopgroup.h
```diff
@@ -74,6 +80,69 @@ typedef enum _sai_next_hop_group_member_observed_role_t
...
+
+/**
+ * @brief Attribute data for #SAI_NEXT_HOP_GROUP_MEMBER_ATTR_OPER_STATE
+ */
+typedef enum _sai_next_hop_group_member_oper_state_t
+{
+    /**
+     * @brief Inactive. Does not serve any connections and is safe to remove
+     */
+    SAI_NEXT_HOP_GROUP_MEMBER_OPER_STATE_INACTIVE,
```
lguohan: it is not clear the criteria to determine if a member is inactive or not. how to you decide if a session is active or not, where to define the timeout period? Do you do the tcp connection state tracking?

lguohan:
in order to track the session status, you need resources in the asic to track the sessions. there is no api to query those session status, user has no idea if it is reaching the resource limit or not. We need those APIs to know what are resource consumption in the hardware.

inc/sainexthopgroup.h
```diff
@@ -74,6 +80,69 @@ typedef enum _sai_next_hop_group_member_observed_role_t
...
+     * @brief Active. Included in the distribution of the new connections
+     */
+    SAI_NEXT_HOP_GROUP_MEMBER_OPER_STATE_ACTIVE,
+
+    /**
+     * @brief Pending becoming inactive. The new connections are no longer
+     * distributed to this member, but it still has some open connections,
+     * so it's not yet safe to remove.
+     */
+    SAI_NEXT_HOP_GROUP_MEMBER_OPER_STATE_INACTIVE_PENDING,
```
lguohan: in case of inactive pending, user would like to know how many connections are still on this member, so that user can decide whether to force inactive or not. without such status, if there is a single long tcp connection, then the group member can never be removed.

PR #1042, opened 2020-02-05
Update sainat.h

JaiOCP:
This change adds nat type in sai_nat_entry_t to distinguish between dnat pool and dnat entry. Its possible that for dnat pool and dnat entry all the key/attributes are same and in this case ASIC redis DB rejects the update as duplicate. NAT type will help distinguish two of them

### DIFF
--- a/inc/sainat.h
+++ b/inc/sainat.h
@@ -326,7 +326,7 @@ typedef struct _sai_nat_entry_t
      * @objects SAI_OBJECT_TYPE_VIRTUAL_ROUTER
      */
     sai_object_id_t vr_id;
-    
+
     /**
      * @brief NAT entry type
      */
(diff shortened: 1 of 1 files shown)


kcudnik:
WARNING: line ends in whitespace sainat.h 329:

inc/sainat.h
```diff
@@ -327,6 +327,11 @@ typedef struct _sai_nat_entry_t
      */
     sai_object_id_t vr_id;
 
+    /**
+     * @brief NAT entry type
+     */
+    sai_nat_type_t nat_type;
+
```
rlhui: @kcudnik , @lguohan please help review and approve, thanks.

stephenxs:
Hi @JaiOCP,  @lguohan, @rlhui, 
Can you cherry-pick this PR to v1.5 as well? 
I'm going to advance the SAI head in sonic-sairedis in SONiC 201911 branch. 
This PR is in tag 1.5.2 but not in 1.5. 
without being cherry-picked to 1.5, it will be lost.
Thanks.

PR #1043, opened 2020-02-06
Dpkg cache

Kalimuthu-Velappan:
Changing all the files and folders having space into '-'


lguohan:
can you specify the reason to remove the space?

Kalimuthu-Velappan:
DPKG caching framework detects the file modification on the SAI repo and rebuild the module only if there is any change.  In order to detect the file modification, all the files under SAI used as dependency files.  The framework  doesnt work if there is space in the filename.

lguohan:
there should be no requirement on the file name restriction for SAI repo.

PR #1044, opened 2020-02-13
Add local ip trap types: icmp and icmpv6

wendani:
This enables fine policing ICMPv4 and ICMPv6 packets destined to local IPs. Earlier, they are matched as general IP2ME trap type, and are subject to the policer under the IP2ME type.

Signed-off-by: Wenda Ni <wenda.ni@bytedance.com>

### DIFF
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -352,6 +352,18 @@ typedef enum _sai_hostif_trap_type_t
      */
     SAI_HOSTIF_TRAP_TYPE_BFDV6 = 0x00004006,
 
+    /**
+     * @brief ICMP traffic (IP protocol == 1) to local router IP address
+     * (default packet action is drop)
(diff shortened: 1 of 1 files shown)


wendani:
Updated the justification as per yesterday's SAI community meeting discussion @JaiOCP

lguohan:
it is better to define it as user defined acl trap as not many asic support this packet type as native trap.

JaiOCP:
> it is better to define it as user defined acl trap as not many asic support this packet type as native trap.

ICMP is very standard trap for host processing and needs direct mapping to CPU cosq. HostIF trap is a better place for it.

JaiOCP (approved):

JaiOCP:
@rlhui Please merge this PR (its a blocking issue for next SAI release)

PR #1054, opened 2020-03-25
SR-MPLS: extending API to support IP-in-MPLS tunneling

phshaikh:
Proposing API extensions to be able to use MPLS for Segment Routing L3 traffic.


kcudnik:
Please correct all warnings and errors: https://sonic-jenkins.westus2.cloudapp.azure.com/job/sai-meta-pr-build/1929/console
i also think this PR will require some changes to parser since new type label was used, i will take a look after errors will be corrected

phshaikh:
@kcudnik Thank you for response. Are there set of commands shall I run before I resubmit changes for review? 

I did following
# cd meta
# make

And it succeeded, I am trying to see if there're more commands I can run locally before sending updates

kcudnik:
that everything you can run locally, and if succeeded then you are good to go

inc/saimpls.h
```diff
@@ -81,6 +81,31 @@ typedef enum _sai_inseg_entry_attr_t
...
 
+    /**
+     * @brief Use TC for a label.
+     *
+     * Associate TC using SAI_SWITCH_ATTR_QOS_EXP_TO_TC_MAP or SAI_PORT_ATTR_QOS_EXP_TO_TC_MAP.
+     * Mainly used for E-LSP kind of MPLS tunnels, where EXP infers both TC and COLOR
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default true
```
lguohan: is this backward compatible? should true to be default behavior or false?

inc/sainexthop.h
```diff
@@ -229,6 +229,28 @@ typedef enum _sai_next_hop_attr_t
...
+    /**
+     * @brief Encap tunnel-id for MPLS PUSH
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_TUNNEL
+     * @allownull true
+     * @condition SAI_NEXT_HOP_ATTR_TYPE == SAI_NEXT_HOP_TYPE_MPLS
+     */
+    SAI_NEXT_HOP_ATTR_MPLS_ENCAP_TUNNEL_ID,
```
lguohan: is this behavior backward compatible? it looks now if the next hop is mpls, then user must include this new attribute? it seems like a non-backward compatible behavior.

inc/sainexthop.h
```diff
@@ -229,6 +229,28 @@ typedef enum _sai_next_hop_attr_t
...
+    /**
+     * @brief Decap tunnel-id for MPLS PHP
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_TUNNEL
+     * @allownull true
+     * @condition SAI_NEXT_HOP_ATTR_TYPE == SAI_NEXT_HOP_TYPE_IP
+     */
+    SAI_NEXT_HOP_ATTR_MPLS_DECAP_TUNNEL_ID,
```
lguohan: is this behavior backward compatible? it looks now if the next hop is IP, then user must include this new attribute? it seems like a non-backward compatible behavior.

phshaikh: For decapsulation (PHP) this new optional attribute is proposed. If provided, then tunnel modes can be used to identify TTL and QoS treatment for outgoing traffic. 

Next hop of type IP is used for PHP as per https://github.com/opencomputeproject/SAI/blob/master/doc/MPLS/SAI-Proposal-MPLS-ver4.docx

inc/sainexthop.h
phshaikh: this is not mandatory attribute, it is proposed for PUSH operation to indicate TTL and QoS mode

inc/saimpls.h
phshaikh: There was no QoS mapping in existing SAI spec for MPLS traffic. 
Default behavior proposal is to use "QoS map", i think comment is wrong here (Use TC for a label should be Use TC from a QoS Map).

phshaikh:
Some comments in the last meeting
- split this pull request in to smaller pull requests
- move decap tunnel id into inseg entry (by Haidong) so same next hop can be reused for tunnel termination at PHP nodes

inc/sainexthop.h
phshaikh: I plan to move this to InSegEntry as per comment by with Haidong last week. This is to be used in conjunction with IP next hop to provide decapsulation behavior for PHP

phshaikh:
Closing this, divided this pull request into many smaller pull requests - 
1) https://github.com/opencomputeproject/SAI/pull/1058
2) https://github.com/opencomputeproject/SAI/pull/1060
3) https://github.com/opencomputeproject/SAI/pull/1062
4) https://github.com/opencomputeproject/SAI/pull/1064
5) https://github.com/opencomputeproject/SAI/pull/1066

PR #1056, opened 2020-03-26
Support for per-vlan traps

bandaru-viswanath:
Add an ability for the user to specify a list of traps that are applicable (only) for that vlan. This list is in-addition to the traps that are otherwise generally defined.

### DIFF
--- a/inc/saivlan.h
+++ b/inc/saivlan.h
@@ -202,6 +202,19 @@ typedef enum _sai_vlan_attr_t
     SAI_VLAN_ATTR_UNKNOWN_NON_IP_MCAST_OUTPUT_GROUP_ID,
 
     /**
+     * @brief List of trap objects associated with this vlan
+     *
+     * When set, these traps are associated with this vlan. These are in addition
+     * to other traps that are common to all vlan.
+     *
+     * @type sai_object_list_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_HOSTIF_TRAP
+     * @default empty
+     */
+    SAI_VLAN_ATTR_TRAP_LIST,
+
+    /**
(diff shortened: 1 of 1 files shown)


rlhui:
What if same object is in both the host trap table and this vlan trap table?

rck-innovium:
As discussed during the meeting can you summarize the reasons for diverging from the behavior of SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST, i.e. why not have have SAI_HOSTIF_TRAP_ATTR_EXCLUDE_VLAN_LIST?

rck-innovium:
The key to create the SAI_HOSTIF_TRAP object is currently the SAI_HOSTIF_TRAP_ATTR_TRAP_TYPE. The use case where two VLANs that need different behavior for the same trap will not work as is today.

rlhui:
There's a feedback from Matty in sai meeting - packets going to cpu need to be via hostif_trap channel, it cannot be omitted, even though it is meant to be per-vlan trap.  @mattyk , please clarify if any.  @bandaru-viswanath should address this first.

JaiOCP:
> As discussed during the meeting can you summarize the reasons for diverging from the behavior of SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST, i.e. why not have have SAI_HOSTIF_TRAP_ATTR_EXCLUDE_VLAN_LIST?

Would it make sense to have then a common exclude LIST based on a type where TYPE can be PORT/VLAN/RIF ?

lguohan:
retest this please

JaiOCP:
+Vissu

 

This PR will to be closed by the Author.

 

-Jai

bandaru-viswanath:
Closing this PR as per recommendation.

PR #1058, opened 2020-03-30
Add new hash fields for MPLS labels

### LINKED ISSUE #1057: MPLS: label stack hash field for MPLS traffic
Current SAI spec does not speak about hashing MPLS traffic. A mechanism is needed to indicate underlying Adapters to hash MPLS traffic, and to use as many MPLS labels as possible for the hashing, along with other packet attributes.

Proposing a hash attribute which indicates using label stack for hashing for traffic with MPLS ether-type. 

The expectation here is that underlying Adapter (and hardware) will attempt to hash incoming MPLS traffic with MPLS labels along with other Hash Attributes such as Source/Destination IP, Source/Destination L4 ports, or any other hash field settings.

Entropy label handling is NOT covered under this attribute. The attribute is indication to use label stack for hashing MPLS traffic along with other attributes

phshaikh:
Summary: this commit fixes #1057  

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/inc/saihash.h
+++ b/inc/saihash.h
@@ -102,8 +102,21 @@ typedef enum _sai_native_hash_field_t
     /** Native hash field inner destination MAC */
     SAI_NATIVE_HASH_FIELD_INNER_DST_MAC = 17,
 
-    /** Native hash field MPLS labels */
-    SAI_NATIVE_HASH_FIELD_MPLS_LABEL_STACK = 18,
+    /** Native hash field the top MPLS label */
+    SAI_NATIVE_HASH_FIELD_MPLS_LABEL_0 = 18,
+
+    /** Native hash field second MPLS label from the top */
+    SAI_NATIVE_HASH_FIELD_MPLS_LABEL_1 = 19,
+
+    /** Native hash field third MPLS label from the top */
+    SAI_NATIVE_HASH_FIELD_MPLS_LABEL_2 = 20,
+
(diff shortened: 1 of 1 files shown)


inc/saihash.h
```diff
@@ -102,6 +102,8 @@ typedef enum _sai_native_hash_field_t
     /** Native hash field inner destination MAC */
     SAI_NATIVE_HASH_FIELD_INNER_DST_MAC = 17,
 
+    /** Native hash field MPLS labels */
+    SAI_NATIVE_HASH_FIELD_MPLS_LABEL_STACK = 18,
```
marian-pritsak: If we'll look at the rest of this enum's attributes, they all indicate a particular field, with the differentiation between inner and outer. I think we should follow the same pattern here - expand the stack to `SAI_NATIVE_HASH_FIELD_MPLS_LABEL_0`, `SAI_NATIVE_HASH_FIELD_MPLS_LABEL_1` and so on for a fine-grain control.

Then the application can query how many labels ASIC supports by calling sai_query_attribute_enum_values_capability and fill the hash attribute accordingly.

phshaikh: makes sense, i can provide up-to five labels to begin with, from _0 to _4.

phshaikh: Applied comments

PR #1060, opened 2020-03-30
MPLS QOS Maps

### LINKED ISSUE #1059: MPLS: QOS Map required for MPLS traffic
SAI has a QoS map for mapping DSCP to traffic class and to color. SAI also has a QoS map for mapping traffic class and color to DSCP. A similar QoS map is required for MPLS traffic. MPLS traffic carries a 3-bit QoS field in MPLS headers called "EXP". 

A map is needed to associate a traffic class to an MPLS packet based on EXP field in the top label.

Accordingly proposing following attributes for 
- SAI_QOS_MAP_TYPE_EXP_TO_TC
- SAI_QOS_MAP_TYPE_EXP_TO_COLOR
- SAI_QOS_MAP_TYPE_TC_AND_COLOR_TO_EXP

SAI has QoS map bind points at port and switch level, accordingly providing following attributes
- For Port - 
   -SAI_PORT_ATTR_QOS_EXP_TO_TC_MAP,
  - SAI_PORT_ATTR_QOS_EXP_TO_COLOR_MAP,
  - SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_EXP_MAP
- For Switch
  - SAI_SWITCH_ATTR_QOS_EXP_TO_TC_MAP
  - SAI_SWITCH_ATTR_QOS_EXP_TO_COLOR_MAP
  - SAI_SWITCH_ATTR_QOS_TC_AND_COLOR_TO_EXP_MAP

Expectation here is that port QoS Map settings override those in the switch. 

If SAI_PORT_ATTR_QOS_EXP_TO_TC_MAP or SAI_SWITCH_ATTR_QOS_EXP_TO_TC_MAP are not configured map every EXP (0-7) value to any default traffic class (left to Adapter implementation)

If SAI_PORT_ATTR_QOS_EXP_TO_COLOR_MAP or SAI_SWITCH_ATTR_QOS_EXP_TO_COLOR_MAP are not configured map every EXP (0-7) value to SAI_PACKET_COLOR_GREEN. 

If  SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_EXP_MAP or SAI_SWITCH_ATTR_QOS_TC_AND_COLOR_TO_EXP_MAP are not configured map any combination of TC and color to some default EXP value (left to Adapter implementation)

MPLS RFC 3270, Per-Hop-Behavior Scheduling Class (or Traffic Class) can be inferred either from
- EXP field in MPLS packet
- Label value 

Accordingly following attributes are proposed for 
- SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP : a boolean flag to indicate if traffic class should be inferred from EXP value  or label.
- SAI_INSEG_ENTRY_ATTR_QOS_TC: Valid only if above flag is  false, value of traffic class associated with label value.

phshaikh:
Summary: this commit fixes #1059

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/inc/saimpls.h
+++ b/inc/saimpls.h
@@ -82,6 +82,31 @@ typedef enum _sai_inseg_entry_attr_t
     SAI_INSEG_ENTRY_ATTR_NEXT_HOP_ID,
 
     /**
+     * @brief Use TC for a label.
+     *
+     * Associate TC using SAI_SWITCH_ATTR_QOS_EXP_TO_TC_MAP or SAI_PORT_ATTR_QOS_EXP_TO_TC_MAP.
+     * Mainly used for E-LSP kind of MPLS tunnels, where EXP infers both TC and COLOR
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default true
+     */
+    SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP,
+
+    /**
+     * @brief TC for a label.
+     *
+     * Associate TC by a label.
+     * Mainly used for L-LSP kind of MPLS tunnels, where label infers TC and EXP infers COLOR
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     * @validonly SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP == false
+     */
+    SAI_INSEG_ENTRY_ATTR_QOS_TC,
+
+    /**
      * @brief End of attributes
      */
     SAI_INSEG_ENTRY_ATTR_END,
--- a/inc/saimpls.h
+++ b/inc/saimpls.h
@@ -32,6 +32,18 @@
  *
  * @{
  */
+typedef enum _sai_inseg_entry_psc_type_t
+{
+    /**
+     * @brief EXP of MPLS label infers both TC and COLOR
+     */
+    SAI_INSEG_ENTRY_PSC_TYPE_ELSP,
+
+    /**
+     * @brief MPLS label infers TC and EXP of MPLS label infers COLOR
+     */
+    SAI_INSEG_ENTRY_PSC_TYPE_LLSP
+} sai_inseg_entry_psc_type_t;
 
 /**
  * @brief Attribute Id for SAI in segment
@@ -82,6 +94,17 @@ typedef enum _sai_inseg_entry_attr_t
     SAI_INSEG_ENTRY_ATTR_NEXT_HOP_ID,
 
     /**
+     * @brief Define PSC type for a label.
+     *
+     * Defines how to infer both TC and COLOR
+     *
+     * @type sai_inseg_entry_psc_type_t
+     * @flags CREATE_AND_SET
+     * @default SAI_INSEG_ENTRY_PSC_TYPE_ELSP
+     */
+    SAI_INSEG_ENTRY_ATTR_PSC_TYPE,
+
+    /**
      * @brief TC for a label.
      *
      * Associate TC by a label (override TC provided by QOS MAP)
@@ -90,8 +113,39 @@ typedef enum _sai_inseg_entry_attr_t
      * @type sai_uint8_t
      * @flags CREATE_AND_SET
      * @default 0
+     * @validonly SAI_INSEG_ENTRY_ATTR_PSC_TYPE == SAI_INSEG_ENTRY_PSC_TYPE_LLSP
      */
     SAI_INSEG_ENTRY_ATTR_QOS_TC,
+
+    /**
+     * @brief Enable EXP -> TC MAP on label.
+     *
+     * Associate TC by a QOS MAP
+     * Mainly used for E-LSP tunnels, where EXP infers both TC and COLOR
+     * Overrides SAI_SWITCH_ATTR_QOS_MPLS_EXP_TO_TC_MAP and SAI_PORT_ATTR_QOS_MPLS_EXP_TO_TC_MAP
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_QOS_MAP
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     * @validonly SAI_INSEG_ENTRY_ATTR_PSC_TYPE == SAI_INSEG_ENTRY_PSC_TYPE_ELSP
+     */
+    SAI_INSEG_ENTRY_ATTR_MPLS_EXP_TO_TC_MAP,
+
+    /**
+     * @brief Enable EXP -> COLOR MAP on label.
+     *
+     * Associate COLOR by a QOS MAP
+     * Overrides SAI_SWITCH_ATTR_QOS_MPLS_EXP_TO_COLOR_MAP and SAI_PORT_ATTR_QOS_MPLS_EXP_TO_COLOR_MAP
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_QOS_MAP
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_INSEG_ENTRY_ATTR_MPLS_EXP_TO_COLOR_MAP,
 
     /**
      * @brief End of attributes
--- a/inc/saimpls.h
+++ b/inc/saimpls.h
@@ -82,27 +82,14 @@ typedef enum _sai_inseg_entry_attr_t
     SAI_INSEG_ENTRY_ATTR_NEXT_HOP_ID,
 
     /**
-     * @brief Use TC for a label.
-     *
-     * Associate TC using SAI_SWITCH_ATTR_QOS_EXP_TO_TC_MAP or SAI_PORT_ATTR_QOS_EXP_TO_TC_MAP.
-     * Mainly used for E-LSP kind of MPLS tunnels, where EXP infers both TC and COLOR
-     *
-     * @type bool
-     * @flags CREATE_AND_SET
-     * @default true
-     */
-    SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP,
-
-    /**
      * @brief TC for a label.
      *
-     * Associate TC by a label.
-     * Mainly used for L-LSP kind of MPLS tunnels, where label infers TC and EXP infers COLOR
+     * Associate TC by a label (override TC provided by QOS MAP)
+     * Mainly used for L-LSP tunnels, where label infers TC and EXP infers COLOR
      *
      * @type sai_uint8_t
      * @flags CREATE_AND_SET
      * @default 0
-     * @validonly SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP == false
      */
     SAI_INSEG_ENTRY_ATTR_QOS_TC,
 
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1457,7 +1457,7 @@ typedef enum _sai_port_attr_t
      * Map id = #SAI_NULL_OBJECT_ID to disable map on port.
      * To enable/disable trust EXP, map ID should be added/removed on port.
      * Default no map.
-     * If present overrides SAI_SWITCH_ATTR_QOS_EXP_TO_TC_MAP.
+     * If present overrides SAI_SWITCH_ATTR_QOS_MPLS_EXP_TO_TC_MAP.
      *
      * @type sai_object_id_t
      * @flags CREATE_AND_SET
@@ -1465,7 +1465,7 @@ typedef enum _sai_port_attr_t
      * @allownull true
      * @default SAI_NULL_OBJECT_ID
      */
-    SAI_PORT_ATTR_QOS_EXP_TO_TC_MAP,
+    SAI_PORT_ATTR_QOS_MPLS_EXP_TO_TC_MAP,
 
     /**
      * @brief Enable EXP -> COLOR MAP on port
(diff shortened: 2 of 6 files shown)


meta/acronyms.txt
```diff
@@ -61,6 +61,7 @@ L2MC - Layer 2 Multi Cast
 L3 - Layer 3
 LAG - Link Aggregation Group
 LPM - Longest Prefix Match
+LSP - Label Switched Patch
```
smaheshm, line 64: s/Patch/Path

inc/saimpls.h
```diff
@@ -81,6 +81,31 @@ typedef enum _sai_inseg_entry_attr_t
...
+     * @brief Use TC for a label.
+     *
+     * Associate TC using SAI_SWITCH_ATTR_QOS_EXP_TO_TC_MAP or SAI_PORT_ATTR_QOS_EXP_TO_TC_MAP.
+     * Mainly used for E-LSP kind of MPLS tunnels, where EXP infers both TC and COLOR
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default true
+     */
+    SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP,
```
marian-pritsak, line 94: What is the use case for this flag? By default, if the application installs a QoS map, it is supposed to be used for every packet. So why would someone need to disable the QoS mapping later in the inseg entry and what is the expected behavior in this case?

phshaikh: In L-LSP tunnel, MPLS labels carry traffic class
In E-LSP tunnel, EXP carry traffic class

This flag facilitate programming either E-LSP or L-LSP kind of labels in ILM. For E-LSP, infer TC from EXP with this flag set to true. For L-LSP, infer TC from label with this flag set to false.

I can remove it if we intend to support only E-LSP.

marian-pritsak: Right, for E-LSP there is a QoS map attached to port or switch. For L-LSP, the `SAI_INSEG_ENTRY_ATTR_QOS_TC` is set, which **implies** **not to use** the QoS map but to set TC according to the label, which makes `SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP` redundant.

phshaikh: SAI_INSEG_ENTRY_ATTR_QOS_TC has default value zero. Perhaps it should be CREATE_ONLY and no default value? Giving default value mean it may always be set and always override QoS map.

marian-pritsak: It can stay `CREATE_AND_SET`, but `@default disabled`.

phshaikh: @default disabled doesn't work. Build fails saying (make in directory meta)

```
ERROR: invalid default value 'disabled' on SAI_INSEG_ENTRY_ATTR_QOS_TC (sai_uint8_t)
```
Without default it fails saying
```
 ASSERT FAILED (on line 445): SAI_INSEG_ENTRY_ATTR_QOS_TC: expected default value, but none provided
```
With CREATE_ONLY it fails
```
 ASSERT FAILED (on line 445): SAI_INSEG_ENTRY_ATTR_QOS_TC: expected default value, but none provided
```

I think we need boolean flag to make it "validonly" if L-LSP is desired. Otherwise some value is required. 

Traffic class is sai_cos_t or uint8_t, another way is to define invalid traffic class in saitypes.h

```
#define SAI_INVALID_TC 0xff
```

And use it as default value for it, but this I don't know is desirable.

marian-pritsak: How about `@default internal`? I see it is used for many optional port and switch attributes.

phshaikh: > How about `@default internal`? I see it is used for many optional port and switch attributes.

It produces error

```
 ASSERT FAILED (on line 1120): SAI_INSEG_ENTRY_ATTR_QOS_TC: default internal currently can be set only on read only objects
```

marian-pritsak: I see. Then we probably should follow some examples of other optional attributes in SAI. For instance, SAI next-hop has an optional attribute VNI for tunnel encapsulation, which overrides the VNI mapping provided with the tunnel:

```
    /**
     * @brief Next hop entry VNI (override tunnel mapper)
     *
     * @type sai_uint32_t
     * @flags CREATE_AND_SET
     * @default 0
     * @validonly SAI_NEXT_HOP_ATTR_TYPE == SAI_NEXT_HOP_TYPE_TUNNEL_ENCAP
     */
    SAI_NEXT_HOP_ATTR_TUNNEL_VNI,
```

It has a `@default 0` tag, but what it really means is that if the attribute is provided, then use it, otherwise just use the tunnel mapping, but not use value 0. It doesn't have any explicit attribute like `SAI_NEXT_HOP_ID_USE_TUNNEL_VNI_MAP`. SAI driver implementation makes the decision to override the mapping by the **presence** of this attribute in the `create_next_hop()` call.

So I think we should follow the same path having annotation `@default 0` and just mention in the comment of the `SAI_INSEG_ENTRY_ATTR_QOS_TC` attribute that it overrides EXP to TC mapping, and remove a flag.

phshaikh: makes sense, if attribute is present only then it overrides even with default value. If not present, its default value doesn't take an effect.

inc/saiport.h
```diff
@@ -1451,6 +1451,53 @@ typedef enum _sai_port_attr_t
...
+     * Default no map.
+     * If present overrides SAI_SWITCH_ATTR_QOS_EXP_TO_TC_MAP.
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_QOS_MAP
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_PORT_ATTR_QOS_EXP_TO_TC_MAP,
```
lguohan, line 1468: should we always use MPLS_EXP in the here to make it more clear?

phshaikh: > should we always use MPLS_EXP in the here to make it more clear?

sure, i used mpls_exp in defining struct because "exp" is reserved key word :) but for consistency these can all be _MPLS_EXP_

lguohan: thanks. LGTM once you rename it.

phshaikh: @lguohan please check if I have addressed your comments

inc/saimpls.h
```diff
@@ -81,28 +81,15 @@ typedef enum _sai_inseg_entry_attr_t
...
      *
-     * Associate TC by a label.
-     * Mainly used for L-LSP kind of MPLS tunnels, where label infers TC and EXP infers COLOR
+     * Associate TC by a label (override TC provided by QOS MAP)
+     * Mainly used for L-LSP tunnels, where label infers TC and EXP infers COLOR
      *
      * @type sai_uint8_t
      * @flags CREATE_AND_SET
      * @default 0
-     * @validonly SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP == false
```
Haidong988: @validonly SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP == false is removed, how to identify L-LSP now?

phshaikh: > @validonly SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP == false is removed, how to identify L-LSP now?

If an application passes SAI_INSEG_ENTRY_ATTR_QOS_TC then it is L-LSP. It is not mandatory argument, so if an application skips it, then it is E-LSP. @marian-pritsak recommended getting rid of validonly since its possible to achieve it without this.

phshaikh: @marian-pritsak should we add QOS map object id here along with this flag? this is to enable different QoS treatment for different labels but passing through same port? Basically two tunnels may terminate through same port but want to use different QOS maps. Thoughts? @Haidong988 provided this input

Haidong988: This is for MPLS VPN situation, so I suggest 

> > @validonly SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP == false is removed, how to identify L-LSP now?
> 
> If an application passes SAI_INSEG_ENTRY_ATTR_QOS_TC then it is L-LSP. It is not mandatory argument, so if an application skips it, then it is E-LSP. @marian-pritsak recommended getting rid of validonly since its possible to achieve it without this.

So if you update E-LSP into L-LSP, how could you update it back into E-LSP?

Haidong988: > @marian-pritsak should we add QOS map object id here along with this flag? this is to enable different QoS treatment for different labels but passing through same port? Basically two tunnels may terminate through same port but want to use different QOS maps. Thoughts? @Haidong988 provided this input

I think it's better to put it into tunnel attribute since MPLS insegment can get tunnel attribute in MPLS VPN case. It can also be used by tunnel terminate entry in VXLAN case.

phshaikh: > This is for MPLS VPN situation, so I suggest
> 
> > > @validonly SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP == false is removed, how to identify L-LSP now?
> > 
> > 
> > If an application passes SAI_INSEG_ENTRY_ATTR_QOS_TC then it is L-LSP. It is not mandatory argument, so if an application skips it, then it is E-LSP. @marian-pritsak recommended getting rid of validonly since its possible to achieve it without this.
> 
> So if you update E-LSP into L-LSP, how could you update it back into E-LSP?

I guess its a good question, on such a case, one has to either remove entry and add new one. @marian-pritsak  any thoughts here?

phshaikh: > > @marian-pritsak should we add QOS map object id here along with this flag? this is to enable different QoS treatment for different labels but passing through same port? Basically two tunnels may terminate through same port but want to use different QOS maps. Thoughts? @Haidong988 provided this input
> 
> I think it's better to put it into tunnel attribute since MPLS insegment can get tunnel attribute in MPLS VPN case. It can also be used by tunnel terminate entry in VXLAN case.

there is a thought of getting rid of tunnel object altogether and use next hop attributes, there is a thought that traffic class or qos map should be part of next hop instead of tunnel. tunnel PR is separate, we have to make such changes in that PR if we want to put QoS map in tunnel

Haidong988: In basic L3 case, QoS map is by port or global.
In traditional tunnel case, QoS map is by tunnel. It also works if you put it  in nexthop because nexthop means a logicport. 
In overlay tunnel with both L2 and L3 case,  QoS map is by tunnel. If you put it in nexthop, in inner l3 routing case, nexthop consist of inner layer3 edit and outer tunnel edit（DVP）. Then there could be nexthop update issue. 
The requirement extending QoS Map is caused by tunnel application actually. It's the nature model.

Haidong988: > > This is for MPLS VPN situation, so I suggest
> > > > @validonly SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP == false is removed, how to identify L-LSP now?
> > > 
> > > 
> > > If an application passes SAI_INSEG_ENTRY_ATTR_QOS_TC then it is L-LSP. It is not mandatory argument, so if an application skips it, then it is E-LSP. @marian-pritsak recommended getting rid of validonly since its possible to achieve it without this.
> > 
> > 
> > So if you update E-LSP into L-LSP, how could you update it back into E-LSP?
> 
> I guess its a good question, on such a case, one has to either remove entry and add new one. @marian-pritsak any thoughts here?

In my understanding, default value 0 is a valid value. If the application created without this attribute, I believe the value should be 0 otherwise no default value is needed.
Even SAI_INSEG_ENTRY_ATTR_USE_EXP_TO_TC_MAP is removed, it still exists under SAI layer （in adapt layer）. I think SAI_NEXT_HOP_ATTR_TUNNEL_VNI description need no default value, because it is needed for Inner layer3 case only. In inner L2 case, VNI is mapped from VSI. And no requirement updating a L3 nexthop into a L2 one exists I guess.

phshaikh: So I am not disagreeing here, MPLS tunnel attributes are part of separate PR and we can put QoS map object id as attribute for MPLS tunnel. This PR is only for QoS map which will be applicable to LSR or mid point MPLS switches.

So here is my initial idea and in the increasing order of preference 
- QoS map at switch level
- QoS map at InSeg entry level (for tunnel termination)
- QoS map at port level

About valid-only flag, I am open to introduce it back. 

However based on earlier conversation I understood an application can skip passing it to SAI and in that case it will be E-LSP (default value of 0 even if valid won't be used because application didn't pass it), if application passes it (with or without value of 0) it will be L-LSP. But I don't know if we can switch back and forth between E-LSP and L-LSP in this case without removing entry. Not a big deal for me, I am okay if application has to remove E-LSP entry and add L-LSP entry for same label.

Haidong988: > So I am not disagreeing here, MPLS tunnel attributes are part of separate PR and we can put QoS map object id as attribute for MPLS tunnel. This PR is only for QoS map which will be applicable to LSR or mid point MPLS switches.
> 
> So here is my initial idea and in the increasing order of priority for underlying switch
> 
> * QoS map at switch level
> * QoS map at port level - overrides above
> * QoS map at InSeg entry level (for tunnel termination) - overrides above
> 
> About valid-only flag, I am open to introduce it back.
> 
> However based on earlier conversation I understood an application can skip passing it to SAI and in that case it will be E-LSP (default value of 0 even if valid won't be used because application didn't pass it), if application passes it (with or without value of 0) it will be L-LSP. But I don't know if we can switch back and forth between E-LSP and L-LSP in this case without removing entry. Not a big deal for me, I am okay if application has to remove E-LSP entry and add L-LSP entry for same label.

I am also okay if application has to remove E-LSP entry and add L-LSP entry for same label.
I mean the action of Create/Get/Set API should be clear for the attribute. Under what conditions it chould be updated(Set API)? What return value should be for "Get API" for E-LSP? How cloud "Get API" know MPLS Insegment is ELSP or LLSP after MPLS insegment created?

phshaikh: Hi Haidong, I have applied your comments.

PR #1062, opened 2020-03-30
MPLS Host-If traps for packets with expiring TTL and Router Alert Label

### LINKED ISSUE #1061: MPLS: Host Traps for expiring TTL and Router Alert Label
Mechanism is needed to trap MPLS packets with expiring TTL. These are useful to implement diagnostic tools and error reporting.

Mechanism is needed to trap MPLS packets with Router Alert Label. This is as per RFC 3032. When a packet with Router Alert Label is received, trap it for processing in software.

Accordingly two traps are proposed
- SAI_HOSTIF_TRAP_TYPE_MPLS_TTL_ERROR
- SAI_HOSTIF_TRAP_TYPE_MPLS_ROUTER_ALERT_LABEL

phshaikh:
Summary: this commit fixes #1061

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -397,6 +397,18 @@ typedef enum _sai_hostif_trap_type_t
      */
     SAI_HOSTIF_TRAP_TYPE_PIPELINE_DISCARD_ROUTER = 0x00007002,
 
+    /**
+     * @brief MPLS packets with expiring TTL value of 1
+     * (default packet action is drop)
+     */
+    SAI_HOSTIF_TRAP_TYPE_MPLS_TTL_ERROR = 0x00007003,
+
+    /**
+     * @brief MPLS packet with router alert label
+     * (default packet action is forward)
+     */
+    SAI_HOSTIF_TRAP_TYPE_MPLS_ROUTER_ALERT_LABEL = 0x00007004,
+
     /** Exception traps custom range start */
     SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE = 0x00008000,
 
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -397,20 +397,20 @@ typedef enum _sai_hostif_trap_type_t
      */
     SAI_HOSTIF_TRAP_TYPE_PIPELINE_DISCARD_ROUTER = 0x00007002,
 
-    /** Exception traps custom range start */
-    SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE = 0x00008000,
-
     /**
      * @brief MPLS packets with expiring TTL value of 1
      * (default packet action is drop)
      */
-    SAI_HOSTIF_TRAP_TYPE_MPLS_TTL_ERROR = 0x00009000,
+    SAI_HOSTIF_TRAP_TYPE_MPLS_TTL_ERROR = 0x00008000,
 
     /**
      * @brief MPLS packet with router alert label
      * (default packet action is forward)
      */
-    SAI_HOSTIF_TRAP_TYPE_MPLS_ROUTER_ALERT_LABEL = 0x00009001,
+    SAI_HOSTIF_TRAP_TYPE_MPLS_ROUTER_ALERT_LABEL = 0x00008001,
+
+    /** Exception traps custom range start */
+    SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE = 0x00009000,
 
     /**
      * @brief End of trap types
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -401,20 +401,13 @@ typedef enum _sai_hostif_trap_type_t
      * @brief MPLS packets with expiring TTL value of 1
      * (default packet action is drop)
      */
-    SAI_HOSTIF_TRAP_TYPE_MPLS_TTL_ERROR = 0x00007003,
+    SAI_HOSTIF_TRAP_TYPE_MPLS_TTL_ERROR = 0x00007500,
 
     /**
      * @brief MPLS packet with router alert label
      * (default packet action is forward)
      */
-    SAI_HOSTIF_TRAP_TYPE_MPLS_ROUTER_ALERT_LABEL = 0x00007004,
-
-    /**
-     * @brief MPLS packet with label trapped to CPU
-     * This is the MPLS packet with incoming label map entry
-     * programmed to trap it to CPU
-     */
-    SAI_HOSTIF_TRAP_TYPE_MPLS_INSEG_ENTRY_TO_CPU = 0x00007005,
+    SAI_HOSTIF_TRAP_TYPE_MPLS_ROUTER_ALERT_LABEL = 0x00007501,
 
     /** Exception traps custom range start */
     SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE = 0x00008000,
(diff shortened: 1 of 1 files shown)


marian-pritsak:
SAI inseg entry attribute `SAI_INSEG_ENTRY_ATTR_PACKET_ACTION` can also be configured to trap packet to the CPU. Do you want to add that trap ID for those trapped packets too?

Otherwise, it looks OK

phshaikh:
> SAI inseg entry attribute `SAI_INSEG_ENTRY_ATTR_PACKET_ACTION` can also be configured to trap packet to the CPU. Do you want to add the trap ID for those trapped packets too?
> 
> Otherwise, it looks OK

I believe I can add a trap id for that as well, while we're at it. 

Do you mind double checking if the value of these MPLS traps are okay? 

SAI_HOSTIF_TRAP_TYPE_STATIC_FDB_MOVE has 0x00006002 value and then next one SAI_HOSTIF_TRAP_TYPE_PIPELINE_DISCARD_EGRESS_BUFFER  is jumped to 0x00007000.

Should I be putting it before 0x00007000 or the current values look okay?

phshaikh:
Applied comments

marian-pritsak:
> > SAI inseg entry attribute `SAI_INSEG_ENTRY_ATTR_PACKET_ACTION` can also be configured to trap packet to the CPU. Do you want to add the trap ID for those trapped packets too?
> > Otherwise, it looks OK
> 
> I believe I can add a trap id for that as well, while we're at it.
> 
> Do you mind double checking if the value of these MPLS traps are okay?
> 
> SAI_HOSTIF_TRAP_TYPE_STATIC_FDB_MOVE has 0x00006002 value and then next one SAI_HOSTIF_TRAP_TYPE_PIPELINE_DISCARD_EGRESS_BUFFER is jumped to 0x00007000.
> 
> Should I be putting it before 0x00007000 or the current values look okay?

Yeah, the 0x00007000 range is related to buffer and other non-lookup drops. I'd suggest starting a new range for MPLS - 0x00009000.

Another thing - `SAI_HOSTIF_TRAP_TYPE_MPLS_INSEG_ENTRY_TO_CPU` shouldn't be the part of `sai_hostif_trap_type_t` because it is caused by the packet_action of the inseg entry installed by a user. Such traps go to the `sai_hostif_user_defined_trap_type_t` and should be called something like `SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_MPLS`.

inc/saihostif.h
```diff
@@ -409,6 +409,13 @@ typedef enum _sai_hostif_trap_type_t
      */
     SAI_HOSTIF_TRAP_TYPE_MPLS_ROUTER_ALERT_LABEL = 0x00007004,
 
+    /**
+     * @brief MPLS packet with label trapped to CPU
```
smaheshm, line 413: What's the use case for this trap. Sorry if this has been covered earlier, still catching up with videos.

inc/saihostif.h
```diff
@@ -409,6 +409,13 @@ typedef enum _sai_hostif_trap_type_t
...
+     * This is the MPLS packet with incoming label map entry
+     * programmed to trap it to CPU
+     */
+    SAI_HOSTIF_TRAP_TYPE_MPLS_INSEG_ENTRY_TO_CPU = 0x00007005,
```
smaheshm, line 417: Just like above messages, can you make the macro more explicit. I guess here you meant the trap for 'TTL expiry'.

phshaikh: Didn't get you. This trap is for those packets which hit inseg entry with packet forward action is to CPU

phshaikh:
> SAI_HOSTIF_TRAP_TYPE_MPLS_INSEG_ENTRY_TO_CPU

@marian-pritsak  0x9000 is SAI_HOSTIF_TRAP_TYPE_END, I don't know if I should be touching _END and 0x8000 is SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE.

So I picked 0x7500 - 0x8000 for MPLS, let me know if its okay to pick 0x9000, I don't mind one way or other

marian-pritsak:
> > SAI_HOSTIF_TRAP_TYPE_MPLS_INSEG_ENTRY_TO_CPU
> 
> @marian-pritsak 0x9000 is SAI_HOSTIF_TRAP_TYPE_END, I don't know if I should be touching _END and 0x8000 is SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE.
> 
> So I picked 0x7500 - 0x8000 for MPLS, let me know if its okay to pick 0x9000, I don't mind one way or other

You can assign SAI_HOSTIF_TRAP_TYPE_END to 0xa000, and MPLS to 0x9000

phshaikh:
> > > SAI_HOSTIF_TRAP_TYPE_MPLS_INSEG_ENTRY_TO_CPU
> > 
> > 
> > @marian-pritsak 0x9000 is SAI_HOSTIF_TRAP_TYPE_END, I don't know if I should be touching _END and 0x8000 is SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE.
> > So I picked 0x7500 - 0x8000 for MPLS, let me know if its okay to pick 0x9000, I don't mind one way or other
> 
> You can assign SAI_HOSTIF_TRAP_TYPE_END to 0xa000, and MPLS to 0x9000

Will do :) Thanks

lguohan:
> You can assign SAI_HOSTIF_TRAP_TYPE_END to 0xa000, and MPLS to 0x9000

SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE should be move after MPLS.

phshaikh:
> > You can assign SAI_HOSTIF_TRAP_TYPE_END to 0xa000, and MPLS to 0x9000
> 
> SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE should be move after MPLS.

would that be backward compatible? since range is pushed ahead?

marian-pritsak:
> > > You can assign SAI_HOSTIF_TRAP_TYPE_END to 0xa000, and MPLS to 0x9000
> > 
> > 
> > SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE should be move after MPLS.
> 
> would that be backward compatible? since range is pushed ahead?

It will since the enum names are not changed.
MPLS -> 0x8000
SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE -> 0x9000

lguohan:
it is not binary backward compatible, but if the user recompile their application with the new SAI header, it will work.

phshaikh:
@lguohan @marian-pritsak  applied comments

inc/saihostif.h
```diff
@@ -397,20 +397,20 @@ typedef enum _sai_hostif_trap_type_t
      */
     SAI_HOSTIF_TRAP_TYPE_PIPELINE_DISCARD_ROUTER = 0x00007002,
 
-    /** Exception traps custom range start */
-    SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE = 0x00008000,
```
kcudnik, line 401: why remove this ? and not use it like this:
SAI_HOSTIF_TRAP_TYPE_MPLS_TTL_ERROR  = SAI_HOSTIF_TRAP_TYPE_CUSTOM_EXCEPTION_RANGE_BASE??

phshaikh: because it is not a custom trap but an MPLS error. MPLS traps are given range from 0x8000 to 0x8fff. Do you see any issue with the existing change?

marian-pritsak: It was not removed, only the value changed.

PR #1064, opened 2020-03-30
MPLS ACL filters

### LINKED ISSUE #1063: MPLS: ACL filters for MPLS packets
Packets can be filtered based on their contents. A mechanism is being proposed to filter MPLS packets based on their contents. Each MPLS packet contains a list of one or more MPLS label. Each label contains following fields 
- 20b Label value 
- 8b TTL value
- 3b EXP value
- 1b Bottom-Of-Stack bit

Proposing following ACL table attributes and ACL table entries to filter MPLS packets
1) SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_INDEX_LIST, SAI_ACL_ENTRY_ATTR_FIELD_MPLS_LABEL_INDEX_LIST

A list of indexes in to label stack, with 0 being the top-most label. This is to filter based on label at a particular index in label stack.

2) SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_LABEL_LIST, SAI_ACL_ENTRY_ATTR_FIELD_MPLS_LABEL_LABEL_LIST

A list of label values.  This is to filter based on label at a particular index in label stack matching label value.

3) SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_TTL_LIST, SAI_ACL_ENTRY_ATTR_FIELD_MPLS_LABEL_TTL_LIST

A list of TTL values.  This is to filter based on TTL value at a particular index in label stack matching label value.

4) SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_EXP_LIST, SAI_ACL_ENTRY_ATTR_FIELD_MPLS_LABEL_EXP_LIST

A list of EXP values.  This is to filter based on TTL value at a particular index in label stack matching label value.

Existing metadata supports passing list of 8b unsigned integers for ACL field. This is not sufficient for passing list of 32b unsigned integers (of which only 20b would matter) for ACL field of list of label values. 

This meta data needs to be extended

phshaikh:
Summary: this commit fixes #1063

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -951,6 +951,42 @@ typedef enum _sai_acl_table_attr_t
      */
     SAI_ACL_TABLE_ATTR_FIELD_MACSEC_SCI,
 
+    /**
+     * @brief A list of indexes into MPLS label stack (0 is top-of-the-stack)
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_INDEX_LIST,
+
+    /**
+     * @brief A list of label values of MPLS labels indexed by list of indexes into MPLS label stack
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_LABEL_LIST,
+
+    /**
+     * @brief A list of TTL values of MPLS labels indexed by list of indexes into MPLS label stack
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_TTL_LIST,
+
+    /**
+     * @brief A list of EXP values of MPLS labels indexed by list of indexes into MPLS label stack
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_EXP_LIST,
+
     /* User Based metadata */
 
     /* User Based metadata [bool] */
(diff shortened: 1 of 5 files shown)


inc/saiacl.h
```diff
@@ -951,6 +951,42 @@ typedef enum _sai_acl_table_attr_t
...
     SAI_ACL_TABLE_ATTR_FIELD_MACSEC_SCI,
 
+    /**
+     * @brief A list of indexes into MPLS label stack (0 is top-of-the-stack)
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_INDEX_LIST,
```
lguohan, line 961: when we create acl table, we add this match field, at that point of time, the SAI needs to know the width of the fields so that it can allocate the enough tcam resource for this table. the question here is it is unclear the width of field from this attribute.

ashutosh-agrawal: One suggestion is to add a separate attribute for each MPLS label instead of a list. Similar to what is being proposed for Hash attributes. That will allow SAI driver to unambiguously determine the  tcam resources and also if underlying hardware has the ability to match on that particular label.

phshaikh: thanks, will apply these comments

- SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL0
- SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL0

marian-pritsak: Yes, I didn't take into account that the current definition is not deterministic in terms of the key size. Expanding the array as suggested will solve the problem.

phshaikh: @ashutosh-agrawal @lguohan @marian-pritsak applied comments

inc/saitypes.h
```diff
@@ -429,6 +429,9 @@ typedef union _sai_acl_field_data_mask_t
 
     /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT8_LIST */
     sai_u8_list_t u8list;
+
+    /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT32_LIST */
+    sai_u32_list_t u32list;
```
lguohan, line 434: this can be removed.

inc/saitypes.h
```diff
@@ -483,6 +486,9 @@ typedef union _sai_acl_field_data_data_t
...
```
lguohan, line 491: this can be removed.

inc/saiacl.h
```diff
@@ -952,42 +952,184 @@ typedef enum _sai_acl_table_attr_t
...
      * @flags CREATE_ONLY
      * @default false
      */
-    SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_INDEX_LIST,
+    SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL0_LABEL,
 
     /**
-     * @brief A list of label values of MPLS labels indexed by list of indexes into MPLS label stack
+     * @brief TTL value for MPLS label on the top
      *
```
JaiOCP, line 965: Please add a generic label attribute for all labels
SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_ALL
and corresponding ttl/exp/bos attributes

phshaikh: what is data type of ACL entry of SAI_ACL_TABLE_ATTR_FIELD_MPLS_LABEL_ALL? If its a list then I think problem with this has been that application is not able to provide prefixed size of matchers, it depends at "run-time" and can be arbitrarily large. This was the reason earlier changes of list were removed. Copying @lguohan  who had raised this concern before

JaiOCP: Discard the comment, wrong PR

PR #1074, opened 2020-04-22
Policy based hash

marian-pritsak:


rlhui:

JaiOCP:
@marian-pritsak 
Why not use the UDF infrastructure. It provides same functionality, isn't it?
UDF match rules are based on sai_acl_field_data_t for L2 and L3 fields and can be arbitrarily combined to create a complex rule for hash.

marian-pritsak:
> @marian-pritsak
> Why not use the UDF infrastructure. It provides same functionality, isn't it?
> UDF match rules are based on sai_acl_field_data_t for L2 and L3 fields and can be arbitrarily combined to create a complex rule for hash.

The UDF match object is very limited - you can classify only by Ethertype, IP type, GRE type.
It does not cover anything related to tunnels and inner headers or other packet fields. Instead of defining the match all over again, the better way is to just use what ACL has.

doc/Hash/Policy-based-hash.md
```diff
@@ -0,0 +1,219 @@
...
+ * Create fine grained fields for custom hash
+ *****************************************************/
+ 
+ sai_object_id_t inner_v6_hash_oid;
+
+ attr.id = SAI_FINE_GRAINED_HASH_FIELD_ATTR_NATIVE_HASH_FIELD;
+ attr.value.s32 = SAI_NATIVE_HASH_FIELD_INNER_IP_PROTOCOL;
+ attrs.push_back(attr);
+ 
+ attr.id = SAI_FINE_GRAINED_HASH_FIELD_ATTR_U8_MASK;
```
lguohan, line 76: where can I find this attribute?

marian-pritsak: Thet's from the old proposal. Removed from md now.

doc/Hash/Policy-based-hash.md
```diff
@@ -0,0 +1,215 @@
...
+
+Optional ordering of the fields is available as well.
+
+```
+    /**
+     * @brief Optional field ordering.
+     *
+     * Specifies in which order the fields are hashed,
+     * and defines in which fields should be associative for CRC with the same sequence ID.
+     * If not provided, it's up to SAI driver to choose.
```
rlhui, line 54: Hi, can we please add in the doc why "sequence id" is needed for the hashing fields?  I recall this is needed so that bidirectional flows can get to the same hash.  What does it mean by "defines in which fields should be associative for CRC with the same sequence ID"?

marian-pritsak: Added a description. Also posting it here as a reply:
```
In the case where the traffic is represented by two different packet types for inbound and outbound direction, 
setting the hash order is neccessary for non-associative hash algorithms like CRC, so that the resulting hash 
value will be the same. For example, for the bare metal machine sending outbound IP packet to the ToR, the hash 
field array could look like this: {SRC_IP:0, DST_IP:1, SRC_PORT:2, DST_PORT:3, IP_TYPE:4}, and for the inbound, getting 
VxLAN packet from the T1 to the same ToR, the hash field array will look like this: {INNER_DST_IP:0, INNER_SRC_IP:1, 
INNER_DST_PORT:2, INNER_SRC_PORT:3, INNER_IP_TYPE:4}. Here the hash associativity is achieved by swapping 
source and destination sequence IDs. The other way to achieve this would be to assign the same shared sequence ID 
value for IP addresses, and another for L4 ports.
```

abdosi: @marian-pritsak Assigning same sequence Id translates to what order ? 
For ex : src and dest ip having sequence Id will the order be {src-ip, dest-ip } or {dest-ip, src-ip} How the order is guarantee from SDK perspective ?

PR #1075, opened 2020-04-22
Masked hash with optional ordering

marian-pritsak:


inc/saitypes.h
```diff
@@ -941,6 +941,161 @@ typedef struct _sai_port_eye_values_list_t
...
+    /** Flag indicating if field is native or UDF */
+    bool is_native;
+
+    /** Native field ID */
+    sai_native_hash_field_t native_field_id;
+
+    /** Mask */
+    sai_native_hash_field_mask_t mask;
+
+    /** UDF field ID */
```
kcudnik, line 1076: out metadata dont support object id's in complicated structures like this, are you able to remove this field ? and add workaround in other way?

kcudnik: also why you need some complicate structures ? can this entire thing be convereted to new object type ?

inc/saitypes.h
```diff
@@ -941,6 +941,161 @@ typedef struct _sai_port_eye_values_list_t
...
+
+    /** Mask for IPv6 address */
+    sai_ip6_t ip6;
+
+} sai_native_hash_field_mask_t;
+
+/**
+ * @brief Masked hash field
+ */
+typedef struct _sai_fine_grain_hash_field
```
kcudnik, line 1065: I would propose to make this object type, and then when required to use list of those objects, you could use sai_objlist_t which is already defined in sai_attribute_value_t

marian-pritsak: Thanks. I'll explore this approach.

inc/saihash.h
```diff
@@ -335,10 +387,14 @@ typedef sai_status_t (*sai_get_hash_attribute_fn)(
  */
 typedef struct _sai_hash_api_t
 {
-    sai_create_hash_fn          create_hash;
-    sai_remove_hash_fn          remove_hash;
-    sai_set_hash_attribute_fn   set_hash_attribute;
-    sai_get_hash_attribute_fn   get_hash_attribute;
+    sai_create_fine_grained_hash_field_fn          create_fine_grained_hash_field;
```
kcudnik, line 413: why this functions go first, and not after already 4 hash defined ?

inc/saihash.h
```diff
@@ -39,89 +39,216 @@
...
     /**
-     * @brief Native hash field source IP.
+     * @brief Native hash field source IPv4.
      *
-     * Also, refers to the outer source IP
+     * Also, refers to the outer source IPv4
      * in case for encapsulated packets
      */
-    SAI_NATIVE_HASH_FIELD_SRC_IP = 0,
+    SAI_NATIVE_HASH_FIELD_SRC_IPV4 = 0,
```
lguohan, line 47: it looks like there is no significance for adding the enum number. it is copied from original header. I think we can remove the enum number in this pr.

https://github.com/opencomputeproject/SAI/commit/6075277f5aca38478b919b7fa786db9b4c082eb0#diff-048b4a2106e6b4a072d5c388e078c273

marian-pritsak: Done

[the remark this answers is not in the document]
marian-pritsak: Returned the fields back

rlhui:

inc/saihash.h
marian-pritsak: Done

inc/saihash.h
```diff
@@ -42,86 +42,209 @@ typedef enum _sai_native_hash_field_t
...
+     *
+     * Specifies in which order the fields are hashed,
+     * and defines in which fields should be associative for CRC with the same sequence ID.
+     * If not provided, it's up to SAI driver to choose.
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_ONLY
+     * @default 0
+     */
+    SAI_FINE_GRAINED_HASH_FIELD_ATTR_SEQUENCE_ID,
```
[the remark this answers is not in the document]
marian-pritsak: So as the annotation states, it's a numerical value indicating the order in which the field is fed to the hash engine.

PR #1076, opened 2020-04-22
ACL GRE key match

marian-pritsak:

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -1094,6 +1094,15 @@ typedef enum _sai_acl_table_attr_t
     SAI_ACL_TABLE_ATTR_FIELD_IPV6_NEXT_HEADER,
 
     /**
+     * @brief GRE key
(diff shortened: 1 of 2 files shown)


marian-pritsak:
The key is dedicated for GRE packets.
The VNI key should not be used for matching on GRE packets or NVGRE packets.

abdosi:
> The key is dedicated for GRE packets.
> The VNI key should not be used for matching on GRE packets or NVGRE packets.

@marian-pritsak I think we should add in comment section atleast if there is no way to let user know about it.

rlhui:

abdosi:
@marian-pritsak Please add the comment as mentioned above.

rlhui:
@marian-pritsak , reminder to please address lgtm and build failures.

marian-pritsak:
@abdosi Added comment directly to header

PR #1066, opened 2020-03-30
MPLS tunnel encapsulation and decapsulation

phshaikh:
This fixes #1065 

Define missing tunnel attributes for MPLS tunnels.

Associate encapsulation tunnel with next hop.

Associate decapsulation tunnel with in-segment entry.

PR #1079, opened 2020-05-01
Provide TTL and QoS treatment during MPLS encap and decap

phshaikh:
Summary:

This is alternative for #1066 PR

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


doc/MPLS/SAI-Proposal-MPLS-Outsegment.md
```diff
@@ -0,0 +1,210 @@
...
+- outer traffic class = inner traffic class
+
+```
+nexthop = {
+    {SAI_NEXT_HOP_ATTR_TYPE, SAI_NEXT_HOP_TYPE_MPLS},
+    {SAI_NEXT_HOP_ATTR_IP, <next hop ip>},
+    {SAI_NEXT_HOP_ATTR_ROUTER_INTERFACE_ID, <interface object>},
+    {SAI_NEXT_HOP_ATTR_LABELSTACK, <outgoing label stack>},
+    _{SAI_NEXT_HOP_ATTR_OUTSEG_TYPE, SAI_OUTSEG_TYPE_PUSH}_,
+    _{SAI_NEXT_HOP_ATTR_OUTSEG_TTL_MODE, SAI_OUTSEG_TTL_MODE_UNIFORM}_,
```
Haidong988, line 45: Does SAI_NEXT_HOP_ATTR_OUTSEG_TTL_MODE work for all the labels?
PW label may have different policy from LSP labels.
I'm not talking about ttl mode itself, but the nexthop Arch.
In current Arch, label stack consist of pw label and lsp labels use the same nexthop.
I think it better to divide one nexthop into two levels, nexthop1 for LSP, nexthop2 for PW2, nexthop3 for PW3, nexthop2 and nexthop3 can use nexthop1 as its low level nexthop by adding atrribute SAI_NEXT_HOP_ATTR_NEXT_LEVEL_NEXT_HOP_ID.

phshaikh: OUTSEG is entire label stack in the current architecture, so this attribute does not distinguish between different labels of stack. This suits well for our SR-MPLS use case

Not sure I understand multi-level next hop approach, can you specify an example, how would it look like and how will application program it?

Simplistically I think next hop itself can have two more attributes
- SAI_NEXT_HOP_ATTR_PW_LABEL
- SAI_NEXT_HOP_ATTR_VC_LABEL

Now you may have two attributes
- SAI_NEXT_HOP_ATTR_PW_LABEL_TTL_MODE
- SAI_NEXT_HOP_ATTR_VC_LABEL_TTL_MODE

In this case label stack is empty, you can introduce either new next hop types
- SAI_NEXT_HOP_TYPE_PWLS
- SAI_NEXT_HOP_TYPE_PWWS
- SAI_NEXT_HOP_TYPE_EVPN

One more thing we can do here is, make following attributes list 
- SAI_NEXT_HOP_ATTR_OUTSEG_TTL_MODE
- SAI_NEXT_HOP_ATTR_OUTSEG_EXP_MODE

First TTL Mode apply to First label in Label stack
Second TTL Mode apply to Second label in Label stack and so forth

But this will complicate underlying implementation, how would you specify TTL values for PIPE modes e.g.
Say first and second label are uniform but third and fourth are pipe then how would you specify third and fourth label's TTL value, should that be list of four or two items. If four, value for first and second is to be ignored by underlying adapter implementation.

Haidong988: If there is an LSP label stack{100,200,300}, and 1000 PW labels{10001-11000} for 1000 L3VPN tunnels, you need cost 4000 label edit entries even if 1000 PW labels are carried by the same LSP label stack{100,200,300} ,right?

phshaikh: I am not sure I understand. As per my understanding for L2VPN there is typically optionally one VC label (at the bottom) and one PW label (at the top). An ingress router, will impose one VC label and one PW label. 

If you're referring about next hops with different label stacks, {1000, 2000, 3000, 10001}, {1000, 2000, 3000, 10002} ....{1000, 2000, 11000} then you'll need 1000 next hops. I am not sure I understand "label edit entries". 

Note - our focus here is L3-MPLS and not L2-MPLS. I don't think existing SAI spec has support for MPLS L2 VPN. Not sure if for L2 VPN, PHP next hop can be SAI_NEXT_HOP_TYPE_IP which is as per existing SAI architecture.

phshaikh: Consider A---B---C---D---E

A can push label La with uniform mode and B can push label Lb with pipe mode and C can swap, D can php label Lb, and D can php label La. In the current mechanism A can not push both label La and Lb with different modes such as La with uniform and Lb with pipe or La and Lb with pipe but different TTL or EXP values. 

If A is pushing labels, all labels will have same EXP and same TTL values. if different labels need different EXP and TTL values, then it has to be done in separate switches as above.

If it has to be done in same switch then TTL and EXP mode specification will also be lists
- label stack : <La, Lb>
- TTL mode stack <pipe, pipe>
- TTL value stack <TTLa, TTLb>
- EXP mode stack <pipe, pipe>
- EXP value stack <EXPa, EXPb>

For uniform, EXP value and TTL value can be ignored. This is what I was suggesting though I don't think this granularity of programming is really needed.

Haidong988: Forget the PW label mentioned above, Just name it "tunnel label" for PW Label of L2VPN and VC label of stardard L3VPN which means a tunnel identifies a VSI or VRF.
Name it "path label" for LSP labels or SR labels which means a path. 1000 nexthops 
{1000, 2000, 3000, 10001}, {1000, 2000, 3000, 10002} ....{1000, 2000, 3000,11000} are independent for you? In our solution, the path part{1000,2000,3000} of these nexthops share the same physical space and could be updated into another path. And the path part itself could be a nexthop group in SAI which is actually an APS group providing a working path as well as a protecting path .
https://wiki.mef.net/pages/viewpage.action?pageId=29229129
In carrier usage, it's hard to keep all policy match because there may be more than one provider.

phshaikh: For us VC/Route label is the bottom most label, the one that defines a label for prefix. Next comes one or more Tunnel labels. So 1000, followed by 2000 ,3000,10001, 2000,3000,10002 and so on are separate tunnel labels and to support such 1000 distinct need 1000 next hops. 

I don't know on what basis are you choosing 1000,2000, 3000 as route labels and 1001 as tunnel labels. As per my understanding of RFCs, every prefix has only one label which is VC/route label and intermediate routers impose one or more tunnel labels. 

Further given link is fully implementable with this solution as well. If not, which use case is not implementable? l

My understanding, is there're two problems you're trying to solve.

1. An ability to granularly program TTL and EXP for each label, is that right? 
2 Another problem you're suggesting is "scale" where 1000 next hops should actually use minimum "label edit entries" (which is your internal H/W resource it seems), so instead of using 4000 entries (4 entries per label stack) use 1003 entries (which is 3 entries for VC label stack and 1000 entries for tunnel label).  I think second one is implementation detail. 

In your model, label stack is split - VC/route label stack and PW/tunnel label stack. This means Push MPLS next hops are of two sub-types
- Route label stack providing next hop
- Tunnel label stack providing stack

Route label stack providing next hop references Tunnel label stack providing next hop. MPLS push operation is combination of stacks of two. This then has to be reflected in application as well.

I see this as total overhaul of existing SAI spec mainly targeted towards your implementation. While the purpose of this PR is to program how TTL and EXP should be set for all labels of the label stack.

Haidong988: An ability to granularly program TTL and EXP for each label, is that right?
-- Yes, per-level is enough for me, each label is ok.

Another problem you're suggesting is "scale" where 1000 next hops should actually use minimum "label edit entries" (which is your internal H/W resource it seems), so instead of using 4000 entries (4 entries per label stack) use 1003 entries (which is 3 entries for VC label stack and 1000 entries for tunnel label). I think second one is implementation detail.
-- We could also do it as the normal way, but you know, customers always want to do more things with existed resource...

This means Push MPLS next hops are of two sub-types
Route label stack providing next hop
Tunnel label stack providing next hop
-- We did not define it as a subType, If there is no  next_level_nhp, nexthop is organized as the normal way. We did it by just adding a new way to use nexthop. If the customers do not need it, they could just use the old way.It's a very old requirement used in metro networks since 2008 when there is no MPLS SR.

phshaikh: I think for PW label you may think of expanding on tunnel objects in future? For now can we consider the label stack in the next hop only as "VC labels" only. Do you think this distinction would help? Of course, these could also be used as "PW labels" by application if uniform TTL and QoS treatment is desired, across all labels of the stack?

Haidong988: For the label stack in next hop, I guess a list is needed for most attributes associated with the label stack.
Beside, I did not see QoS map from TC/Color to EXP in encap direction, is it still working on switch and port level only ?
I think you need to add it into nexthop since mpls insegment has “EXP to TC/Color” mapping in decap direction. And it should be a list as well.

phshaikh: I am propoosing same EXP and same TTL for all labels in the label stack and so there is only one attribute to indicate EXP and TTL for all.

There is SAI_NEXT_HOP_ATTR_QOS_TC_AND_COLOR_TO_MPLS_EXP_MAP, its not given in this document but present in file in this PR - look at sainexthop.h

Previously I mentioned, can we consider label stack of next hop as only VC labels? and you can then extend it for PW labels in your subsequent PRs?

Haidong988: It's ok for the current requirement. 
But I think, it's hard to make sure that there is only 1 policy in VC label stack. In carrier network, a operator may lease the line to another operator. They have different policy in operate levels for their own.

phshaikh: In the multi-level label stack setup earlier, I trust you are considering one policy for VC label stack and one for PW label stack, and considering label stack attribute of next hop as VC label stack fits in that? Did I misunderstand earlier? In multi-level set up can you specify more than one policies or policies per label for VC label stack?

Haidong988: -- Your earlier understanding is right.
There are 2 steps, copy policy from VC label to PW label,copy policy from PW label to inner packet. In traditional carrier, the number of VC label is usually 1. 
 
 In multi-level set up can you specify more than one policies or policies per label for VC label stack?

-- Yes, we can specify policies per label for VC label stack in SDK level.
 It's not in SAI yet, but I think the VC label stack path cross devices of different operators is possible especially in SR case.

phshaikh: LABELSTACK attribute of next hop is currently array of uint32_t (i.e. label values). It can be extended to have array of special struct say sai_mpls_label_t which contains 
1) union of QOS map, EXP value, 
2) TTL flag to uniformly decrement or TTL value
3) label (20-bit) value.

but this could be backward incompatible change :-(

Haidong988: How about keep the LABELSTACK unchanged.
Just change SAI_NEXT_HOP_ATTR_OUTSEG_EXP_MODE into list. If the EXP_MODE is pipe, use the EXP value configured in LABELSTACK. TTL mode can use the same way. SAI_NEXT_HOP_ATTR_OUTSEG_TTL_VALUE and SAI_NEXT_HOP_ATTR_OUTSEG_EXP_VALUE are not needed.
Change 
 SAI_NEXT_HOP_ATTR_QOS_TC_AND_COLOR_TO_MPLS_EXP_MAP into list as well.

phshaikh: I understood you were open to have only two level of TTL and QoS settings - one for VC labels and one for PW labels. But if we have consensus to specify different TTL and QoS setting for each label, I can extend it to list. it will be an extra work for application and SDK.

Haidong988: As I have discussed with you long time ago, we care about the tunnel level ttl/qos mode in traditional mpls l2vpn/l3vpn. The mpls label ttl/qos mode is defined as device default behavior in many products. But if you want to define it as attribute of mpls insegment and nexthop to make it flexible for user configuration, that will lead to extra work although ASIC can do it. The application does not care about it so much before MPLS SR, but I'm afraid that MPLS SR really need it especially when there is a controller need to control everything.

phshaikh: I am trying to understand the problem we're pursuing here.

In a label stack there're two types of label
- VC label - which is the bottom most label
- Tunnel label - which is at the top of VC label

One more tunnel labels can be present for hierarchical tunnel.

Now we have two approaches
- all labels same TTL and EXP (existing)
- VC label has some TTL and EXP value and Tunnel labels have different TTL and EXP values
- In this, all tunnel labels have same TTL and EXP values,
- Or different tunnel label have different TTL and EXP value

I am leaning towards - all tunnel labels have same TTL and EXP value - this is how it works in our existing implementation. If hierarchical tunnel is needed then additional LSR or switches must do "ingress push" operations.

Let me know if it helps / solves your concern.

jayantwu: if the existing SAI can support mpls l3vpn?   if vpn label and lsp label are  in the same labelstack?

FLYFLYFL: I'm puzzled that if the VPN label and LSP label are in the same labelstack ? In other words, if the current SAI attributes can support MPLS L3VPN well ? @phshaikh

PR #1080, opened 2020-05-05
Add Name attribute for LAG

chiourung:
- What I did
  Use the Identify attribute to fix the issue that view comparison for LAG [Azure/SONiC#354](https://github.com/Azure/sonic-sairedis/issues/354)

- How I did it
  Add SAI_LAG_ATTR_NAME to identify the LAG.
  It needs to set SAI_LAG_ATTR_NAME when create LAG.
  For view comparison, it uses this attribute to find the matched LAG.

Signed-off-by: chiourung_huang <chiourung_huang@edge-core.com>


inc/sailag.h
```diff
@@ -141,6 +141,14 @@ typedef enum _sai_lag_attr_t
      */
     SAI_LAG_ATTR_DROP_TAGGED,
 
+    /**
+     * @brief LAG Name
+     *
+     * @type char
+     * @flags IDENTIFY
```
kcudnik, line 148: what is identify ? please remove this flag, use any of the existing flags

chiourung: Identify flag: 
Attribute with this flag will indicate that this attribute is for the identification
purpose for the client of SAI. SAI implementation doesn't need to handle this attribute at all. 

Please reference the pull request [sonic-sairedis#609](https://github.com/Azure/sonic-sairedis/pull/609).

kcudnik: please use standard flags here, you can use create_and_set or create_only

inc/sailag.h
```diff
@@ -141,6 +141,14 @@ typedef enum _sai_lag_attr_t
      */
     SAI_LAG_ATTR_DROP_TAGGED,
 
+    /**
+     * @brief LAG Name
+     *
+     * @type char
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
```
kcudnik, line 148: just an info, that this breaks backward compatibility, since users will have to add this flag for every lag, maybe it would be better to mark this as "create_only" ? to preserver backward compatibility

bandaru-viswanath: @kcudnik Should we be adding a name for the 'port' object as well for consistency ?

kcudnik: ports can be uniquely identified by lanes

chiourung: @kcudnik  If the flag is "create_only", it would need to assign default value. This attribute should be unique and no suitable default value for this attribute.

chiourung:
@kcudnik  Would this pull request be merged for v1.6.1?

kcudnik:
you would need to generate PR for v1.6.1 branch

inc/sailag.h
kcudnik: this attribute is also mandatory_on_create so it should not have default value

inc/sailag.h
```diff
@@ -141,6 +141,15 @@ typedef enum _sai_lag_attr_t
      */
     SAI_LAG_ATTR_DROP_TAGGED,
 
+    /**
+     * @brief LAG Name
+     *
+     * @type char
+     * @flags CREATE_ONLY
+     * @default \"\"
```
kcudnik, line 149: you don't need to add '\'

chiourung: It would build fail if I don't add \"\".
build fail message:
ERROR: invalid default tag value '' on sai_lag_attr_t SAI_LAG_ATTR_NAME

rlhui:
Could we change the PR title - "Add Name attribute for LAG" is more clear.

inc/sailag.h
kcudnik: ok, leave it as is, after you merge that i will work on fixing parser to allow ""

rlhui (approved):

lguohan:
this pr is a little unique, i think. since techinical, it is not needed by the asic. I think we may need to reconsider this.

lguohan:
even if we plan to introduce such anchor attribute for view comparison, it is better to be unified, for example, instead of char, we should define uint32 and use some kind of index as the anchor attribute.

kcudnik:
seems to me like uint32/uint64 could be better instead of char, faster and easier handling internally, CRAETE_ONLY | KEY

lguohan:
correct. also it seems this is only needed by sairedis comparison logic. should we have a flag to indicate that? meaning normal sai should not need this attribute? or maybe even this is arguable?

kcudnik:
currently, yes, seems like only sairedis CL needs this, as a flag do you suggest to have special indicator in SAI header that this attribute is only for sairedis etc? internally each vendor could just implement this as a static field in function just to keep the value for possible GET operation, and no actual meaning for vendor internals
in sairedis we could possibly use custom_range attributes to add our own new attributes, but they would be incompatible with SAI headers, anyway we already have some of those for apply/init view, but those are global ones, not related to any object, ad as for custom attribute this would be currently hard to introduce in metadata, so having extra attribute seems fine here
also in CL there is a path to find right LAG using other members like LAG_MEMBERs and ports, this is called pre-match logic, and works well, and finds correct lags

lguohan:
Just to be clear, this PR has been reverted because of some issues, but in general we are considering this will be really helpful for warm reboot.

kcudnik:
yes, also we could take into consideration version with uint32_t field instead of string name for lag

PR #1081, opened 2020-05-07
Voq system

michaelli10:
The following pull request has the SAI draft proposal for VOQ support as discussed in the SAI VOQ subgroup. The doc file contains the overview and the spec for the changes proposed.


rlhui:
Thanks.  Could we please use .md format for the doc?

rlhui:
It's better to not to mandate/suggest that SONiC is the OS for this SAI proposal.  The SAI proposal should be OS agnostics.

rlhui:
Please add description on the model that's supported, e.g. the sai proposal supports distributed OS, and/or centralized OS, single SAI instance and/or multiple SAI instances.

rlhui:
@eswaranb, @abdosi, please help reserve this proposal. Thanks.

rlhui:
@skeesara-nokia, would you please help review this proposal? Thanks.

inc/saineighbor.h
```diff
@@ -115,6 +115,17 @@ typedef enum _sai_neighbor_entry_attr_t
...
+     *
+     * Defines the remote neighbor interface when the destination is on a remote line card
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_NEIGHBOR_ENTRY_ATTR_REMOTE_OUTLIF,
+
     /**
```
skeesara-nokia, line 129: During the review it was agreed that there would be an explicit attribute indicate whether or not the "REMOTE_OUTLIF" attribute is being imposed by the application or to be allocated by SAI. I dont see that attribute here.

Also the REMOTE_OUTLIF was going to be changed to "ENCAP_INDEX" or "ENCAP_ID".

inc/saiswitch.h
```diff
@@ -2141,6 +2147,69 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * Mandatory in VOQ Switch
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_SWITCH_ATTR_SWITCH_ID,
+
+    /**
+     * @brief Maximum number of cores in the VOQ Switch (chassis)
```
skeesara-nokia, line 2161: Suggest "VOQ System" instead of "VOQ Switch"

inc/saiswitch.h
```diff
@@ -2141,6 +2147,69 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
```
skeesara-nokia, line 2165: This attribute applies only if the switch type.  The condition should be specified here. 
"@condition SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_VOQ"

inc/saiswitch.h
```diff
@@ -2141,6 +2147,69 @@ typedef enum _sai_switch_attr_t
...
+    SAI_SWITCH_ATTR_MAX_SYSTEM_CORES,
+
+    /**
+     * @brief System port configuration list.
+     *
+     * @type sai_system_port_config_list_t
+     * @flags CREATE_ONLY
+     * @default empty
```
skeesara-nokia, line 2173: This attribute applies only if the switch type. The condition should be specified here.
"@condition SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_VOQ"

inc/saisystemport.h
```diff
@@ -0,0 +1,208 @@
...
+
+    /**
+     * @brief Admin Mode
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_SYSTEM_PORT_ATTR_ADMIN_STATE,
+
```
skeesara-nokia, line 119: The intent of ADMIN_STATE is not clear for a system port. For local ports - does it mean that we can enable/disable a port using this attribute. For remote ports what does ADMIN_STATE do any way. If there is no clear intended usage - the suggestion is to remove this attribute. It could be added later if a need for such an attribute is discovered.

inc/saiport.h
```diff
@@ -1498,6 +1501,71 @@ typedef enum _sai_port_attr_t
...
+    /**
+     * @brief Fabric port Attached
+     *
+     * Signifies if the fabric port is attached
+     *
+     * @type bool
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_FABRIC_ATTACHED,
+
```
skeesara-nokia, line 1521: What does this attribute FABRIC_ATTACHED mean ? Does it mean fabric port is up or down ? If so how is it different from oper status

skeesara-nokia:
> 
> 
> @skeesara-nokia, would you please help review this proposal? Thanks.

Done

inc/saineighbor.h
```diff
@@ -115,6 +115,17 @@ typedef enum _sai_neighbor_entry_attr_t
      */
     SAI_NEIGHBOR_ENTRY_ATTR_COUNTER_ID,
 
+    /**
```
eswaranb, line 118: For the SAI_NEIGHBOR_ENTRY_ATTR_DST_MAC_ADDRESS ATTR, I believe we discussed that we could pass in a DUMMY or a NULL mac address. Could we add a comment to that effect?

inc/saitypes.h
```diff
@@ -944,6 +944,88 @@ typedef struct _sai_port_eye_values_list_t
...
 } sai_port_eye_values_list_t;
 
+/**
+ * @brief System port configuration attributes
+ */
+typedef struct _sai_system_port_config_t
+{
+    uint32_t port_id;
+    uint32_t attached_switch_id;
+    uint32_t attached_core_index;
```
abdosi, line 954: please add some description for each field and also what are mandatory argument that basic VOQ systemport needs.  If I remember correctly speed was not mandatory.

inc/saiport.h
```diff
@@ -1498,6 +1501,71 @@ typedef enum _sai_port_attr_t
...
+    /**
+     * @brief Attached Switch type.
+     *
+     * Signifies the attached switch type
+     *
+     * @type sai_switch_type_t
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_FABRIC_ATTACHED_SWITCH_TYPE,
```
eswaranb, line 1530: Are the ATTACHED_SWITCH_TYPE, ATTACHED_SWITCH_ID etc attributes also available in FABRIC_REACHABILITY in the fabric_port_reachability_t structure?

inc/saitypes.h
```diff
@@ -944,6 +944,88 @@ typedef struct _sai_port_eye_values_list_t
...
+
+    /** Rate of data units with CRC errors passed its threshold */
+    SAI_PORT_ERR_STATUS_CRC_RATE,
+
+    /** Error remote fault indication */
+    SAI_PORT_ERR_STATUS_REMOTE_FAULT_STATUS,
+
+    /** Error status max */
+    SAI_PORT_ERR_STATUS_MAX,
+} sai_port_err_status_t;
```
abdosi, line 1015: https://github.com/opencomputeproject/SAI/blob/master/inc/saiport.h already have port errors define
i feel this need to be sai_system_port_err_status_t and also enum should have SAI_SYSTEM_*.

inc/saitypes.h
```diff
@@ -1091,6 +1173,18 @@ typedef union _sai_attribute_value_t
...
+    sai_system_port_config_t sysportconfig;
+
+    /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_SYSTEM_PORT_CONFIG_LIST */
+    sai_system_port_config_list_t sysportconfiglist;
+
+    /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_FABRIC_PORT_REACHABILITY */
+    sai_fabric_port_reachability_t reachability;
+
+    /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_PORT_ERR_STATUS_LIST */
+    sai_port_err_status_list_t porterror;
```
abdosi, line 1187: sysporterror

inc/saisystemport.h
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * @brief Speed in Mbps
+     *
+     * On get, returns the configured port speed.
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 10000
+     */
+    SAI_SYSTEM_PORT_ATTR_SPEED,
```
eswaranb, line 109: SPEED is also part of the system_port_config_t, which can lead to confusion if this attribute doesn't agree with the value in the system_port_config_t. Should we remove either this one or the other one until there is a need for it?

abdosi, line 109: is this needed ? already sai_system_port_config_t has speed ?

inc/saitypes.h
```diff
@@ -944,6 +944,88 @@ typedef struct _sai_port_eye_values_list_t
...
+    uint32_t switch_id;
+
+    /** Remote switch ID is reachable through the fabric port */
+    bool reachable;
+} sai_fabric_port_reachability_t;
+
+/**
+ * @brief Port error status
+ */
+typedef enum _sai_port_err_status_t
```
eswaranb, line 987: If these errors are specific to fabric ports, should we call that out in the name  - sai_fabric_port_err_status_t ?

inc/saiqueue.h
```diff
@@ -311,6 +317,12 @@ typedef enum _sai_queue_stat_t
     /** Get/set WRED red marked bytes count [uint64_t] */
     SAI_QUEUE_STAT_WRED_ECN_MARKED_BYTES = 0x00000023,
 
+    /** Get current queue occupancy percentage [uint64_t] */
+    SAI_QUEUE_STAT_CURR_OCCUPANCY_LEVEL = 0x00000024,
+
```
abdosi, line 322: we already have SAI_QUEUE_STAT_CURR_OCCUPANCY_BYTES  and SAI_QUEUE_STAT_WATERMARK_BYTES . Why we need this ? is it related to voq changes or general changes ?

inc/saiqueue.h
```diff
@@ -41,12 +41,18 @@ typedef enum _sai_queue_type_t
...
 
-    /** H/w Unicast Queue */
+    /** H/w Egress Unicast Queue */
     SAI_QUEUE_TYPE_UNICAST = 0x00000001,
 
-    /** H/w Multicast (Broadcast, Unknown unicast, Multicast) Queue */
+    /** H/w Multicast Egress (Broadcast, Unknown unicast, Multicast) Queue */
     SAI_QUEUE_TYPE_MULTICAST = 0x00000002,
 
+    /** H/w Virtual Output Queue (VOQ). This queue is ingress unicast queue */
```
abdosi, line 50: please add comment Multicast VOQ is not supported as of now

inc/saiacl.h
```diff
@@ -2206,7 +2206,7 @@ typedef enum _sai_acl_entry_attr_t
     SAI_ACL_ENTRY_ATTR_ACTION_START = 0x00002000,
 
     /**
-     * @brief Redirect Packet to a destination which can be a port,
+     * @brief Redirect Packet to a destination which can be a port, system port,
```
abdosi, line 2209: what about attribute SAI_PORT_ATTR_INGRESS_ACL/SAI_PORT_ATTR_EGRESS_ACL/... . Are they applicable for VOQ system ? In VOQ system where ACL will be binded SAI_PORT or SAI_SYSTEM_PORT

abdosi:
sai_port_pool_attr_t is it applicable to VOQ system ? In general how we will bind Buffer Profile/Wred Profile ? Will it be on System port level or port level ?

inc/saiport.h
```diff
@@ -2025,6 +2093,21 @@ typedef enum _sai_port_stat_t
...
     SAI_PORT_STAT_PRBS_ERROR_COUNT,
 
+    /** SAI port stat if in FEC correctable pkts */
+    SAI_PORT_STAT_IF_IN_FEC_CORRECTABLE_FRAMES,
+
+    /** SAI port stat if in FEC not correctable pkts */
+    SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES,
+
+    /** SAI port stat if in FEC symbol errors */
+    SAI_PORT_STAT_IF_IN_FEC_SYMBOL_ERRORS,
```
abdosi, line 2103: are these stats SAI_PORT_STAT_IF_IN_FEC_* applicable for fabric port or any reglar port ?

inc/saineighbor.h
```diff
@@ -115,6 +115,17 @@ typedef enum _sai_neighbor_entry_attr_t
...
+     * @brief Remote Out Logical Interface
+     *
+     * Defines the remote neighbor interface when the destination is on a remote line card
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 0
```
abdosi, line 125: we should add @validonly in general for all VOQ based attributed

michaelli10:
@dsatterw  Adding reviewer

inc/saiport.h
michaelli10: It is used to check if the fabric port is attached to a device .  You can skip checking for the fabric port status if it is not connected.

inc/saiport.h
michaelli10: *ATTACHED* attributes are used to check for the device/port directly connected to a local fabric port.  If checking on a local VoQ switch, the connected device could be another VoQ switch (back to back) or a fabric switch.   On a local fabric switch, the connected device is most likely a VoQ switch but certain multi-level fabric topologies could connect fabric switches to other fabric switches.

*REACHABILITY* attribute is used to check if there is fabric connectivity from a local VoQ switch to a remote VoQ switch (regardless of which fabric port the cells traverse through).

inc/saiport.h
michaelli10: The *FEC* stats are applicable to fabric ports in this VoQ context but can also apply to front panel ports as well.

inc/saiqueue.h
michaelli10: Multicast VoQ is out of scope in this initial VoQ system PR but shall we add the attribute now?

inc/saiqueue.h
michaelli10: DNX Fabric ports do not support queue stats in byte units.  Fabric port queue stats use general units for queue statistics.  This can also be used in general cases.  Percentage is a generic representation of queue occupancy and watermark that abstracts port speed and configured queue sizes.

inc/saiswitch.h
michaelli10: Will update the comment

inc/saiswitch.h
michaelli10: Will update the meta condition

inc/saiswitch.h
michaelli10: Will update the meta condition

inc/saitypes.h
michaelli10: Will update the comments.

inc/saineighbor.h
dsatterw: Changing to SAI_NEIGHBOR_ENTRY_ATTR_ENCAP_INDEX

inc/saiswitch.h
dsatterw: Agreed. Making the change.

inc/saiswitch.h
dsatterw: Agreed. Making the change.

inc/saiswitch.h
dsatterw: Agreed. Making the change. Also need to change to MANDATORY_ON_CREATE to satisfy metachecker.

inc/saisystemport.h
dsatterw: I would like to keep this attribute. I feel we may find a need for it in the future. We can leave it up to the implementation as to whether to support it or not.

inc/saitypes.h
dsatterw: Will add comments. Only optional member is num_voq. Speed is needed for rate class scheduling on VOQ.

inc/saisystemport.h
dsatterw: We can remove this.

dsatterw: We can remove this.

inc/saiqueue.h
dsatterw: Will add comment.

inc/saiacl.h
dsatterw: Thought was to add capability where it didn't exist today. PORT object can be used to INGRESS and EGRESS qualifiers. To be able to redirect or monitor port on a different line card, system port needed to be added.

dsatterw:
> sai_port_pool_attr_t is it applicable to VOQ system ? In general how we will bind Buffer Profile/Wred Profile ? Will it be on System port level or port level ?

Buffer profile and WRED profiles and linked to VOQ/System port.

abdosi:
> > sai_port_pool_attr_t is it applicable to VOQ system ? In general how we will bind Buffer Profile/Wred Profile ? Will it be on System port level or port level ?
> 
> Buffer profile and WRED profiles are linked to VOQ/System port.

then we need enhance _sai_port_pool_attr_t to have system port also ? What about this attributes 
SAI_PORT_ATTR_QOS_INGRESS_BUFFER_PROFILE_LIST and SAI_PORT_ATTR_QOS_EGRESS_BUFFER_PROFILE_LIST ?

SAI_PORT_ATTR_INGRESS_PRIORITY_GROUP_LIST (needed for PFC) 

Will all three above will be Port or sysport level ?

inc/saitypes.h
michaelli10: In the VoQ context, these apply to fabric ports but the errors are generic enough for front panel ports as well.   For example:  CRC, FEC, remote fault can apply to front panel ports as well.

inc/saitypes.h
michaelli10: In this VoQ PR, these errors apply to fabric ports but are generic enough to apply to front panel ports as well.  These port errors don't apply to system ports.

inc/saitypes.h
michaelli10: The port error type applies to fabric ports in the VoQ context but can apply to front panel ports.  This is not for system port.

dsatterw:
> SAI_PORT_ATTR_QOS_INGRESS_BUFFER_PROFILE_LIST

PFC will be on source port so it should be port object based but we may need to add the xxx_BUFFER_PROFILE_LIST attributes to the system port object. We will need to discuss.

dsatterw:
> > SAI_PORT_ATTR_QOS_INGRESS_BUFFER_PROFILE_LIST
> 
> PFC will be on source port so it should be port object based but we may need to add the xxx_BUFFER_PROFILE_LIST attributes to the system port object. We will need to discuss.

Response from the DNX Arch Team:
The  BUFFER_PROFILE  relate to various resources allocations and thresholds of a Port or Port-PG LOCALLY.
In DNX architecture, these are meaningless on a System-Port context and there is no hardware to supports this.

BUFFER_PROFILE  should refer to the Local port


The resource allocation and  congestion aspect of  a VOQ  are under WRED_PROFILE and should be associated with System_PORT

inc/saineighbor.h
michaelli10: Will add comment

inc/saineighbor.h
michaelli10: Will add

inc/saiqueue.h
michaelli10: Will add the attribute for multicast VOQ

inc/saineighbor.h
michaelli10: Added comment "validonly SAI_NEIGHBOR_ENTRY_ATTR_IS_LOCAL == true".  The "@validonly" meta field fails the metachecker since SAI_NEIGHBOR_ENTRY_ATTR_COUNTER_ID is mandatory on create.

inc/saiqueue.h
michaelli10: Added SAI_QUEUE_TYPE_MULTICAST_VOQ

inc/saitypes.h
michaelli10: Moved mandatory comment to top level data structure comment

inc/saisystemport.h
michaelli10: Removed SAI_SYSTEM_PORT_ATTR_SPEED

inc/saiport.h
```diff
@@ -1498,6 +1501,71 @@ typedef enum _sai_port_attr_t
...
+
+    /**
+     * @brief System port for the port
+     *
+     * @type sai_object_id_t
+     * @flags READ_ONLY
+     * @objects SAI_OBJECT_TYPE_SYSTEM_PORT
+     */
+    SAI_PORT_ATTR_SYSTEM_PORT,
+
```
skeesara-nokia, line 1568: I think it would be good if we can query - attached_core_index and attached_core_port_index using the port table for ports local to the asic. This will help an application call flow - where the application can query the port table for these two attributes and then use them when requesting the creation of the system_port.

Without this information being available from the port_table - the application would need additional knowledge upfront about which ports belong to which core. It would less error prone if instead it uses the information returned from SAI.

michaelli10: Are you thinking of a future dynamic port breakout case where a new local port is created without an associated system port?  We can consider this as an enhancement.

The initial use case was to provide the local port and system port mapping upfront to create_switch using SAI_SWITCH_ATTR_SYSTEM_PORT_CONFIG_LIST.   If this info is not saved by the app,  you can retrieve the info from the local port list using get "SAI_PORT_ATTR_SYSTEM_PORT" attribute followed by get "SAI_SYSTEM_PORT_ATTR_CONFIG_INFO".

skeesara-nokia: This is not for dynamic port breakout. Today the configuration specifies the lanes that for each port. The configuration flow that I am looking is something line this.

1. create_switch (switch_id,max_cores) /* for DNX the ".bcm" specifies the initial ports that are created. System ports are not created here  */
2. query SAI for port table /* this will return attached_core_index and attached_core_port_index. */
3. create_system_port (system_port_id,attached_core_index,attached_core_port_index, ...) /* system port id is application assigned. The other two attributes were previously queried from SAI*/

If we do this - there would not be a need to embed attached_core_index" and "attached_core_port_index" in application configuration files (only lane information would be necessary). They would be read from the SAI port tables and used during system port creation.

abdosi:
> sai_port_pool_attr_t is it applicable to VOQ system ? In general how we will bind Buffer Profile/Wred Profile ? Will it be on System port level or port level ?

@dsatterw Any update on this ?

abdosi:
> PFC will be on source port so it should be port object based but we may need to add the xxx_BUFFER_PROFILE_LIST attributes to the system port object. We will need to discuss.

@dsatterw I assume final conclusion this will be on Local Port only.

abdosi:
Based on last discussion of checking on list of SAI_PORT_ attribute currently being used:

SAI_PORT_ATTR_QOS_TC_TO_QUEUE_MAP
SAI_PORT_ATTR_QOS_NUMBER_OF_SCHEDULER_GROUPS
SAI_PORT_ATTR_QOS_SCHEDULER_GROUP_LIST
How ro co-related SAI_PORT_ATTR_SPEED and sysport speed ?

Will add complete list also.

abdosi:
We need to add SAI_HOSTIF_ATTR_OBJ_ID to have Systemport of HostIf Creation for representing Remote CPU Port

abdosi:
> Based on last discussion of checking on list of SAI_PORT_ attribute currently being used:
> 
> SAI_PORT_ATTR_QOS_TC_TO_QUEUE_MAP
> SAI_PORT_ATTR_QOS_NUMBER_OF_SCHEDULER_GROUPS
> SAI_PORT_ATTR_QOS_SCHEDULER_GROUP_LIST
> How are they co-related SAI_PORT_ATTR_SPEED and sysport speed ?
> 
> Will add complete list also.

As discussed SAI_PORT_ATTR_QOS_TC_TO_QUEUE_MAP will need to be done on System Port level for VOQ systems.

dsatterw:
> > sai_port_pool_attr_t is it applicable to VOQ system ? In general how we will bind Buffer Profile/Wred Profile ? Will it be on System port level or port level ?
> 
> @dsatterw Any update on this ?
This will remain at port level. But WRED profile will be applied to a VOQ not a port in a VOQ system.

dsatterw:
> > PFC will be on source port so it should be port object based but we may need to add the xxx_BUFFER_PROFILE_LIST attributes to the system port object. We will need to discuss.
> 
> @dsatterw I assume final conclusion this will be on Local Port only.

Correct. Local port only

inc/saineighbor.h
```diff
@@ -50,6 +50,7 @@ typedef enum _sai_neighbor_entry_attr_t
 
     /**
      * @brief Destination MAC address for the neighbor
+     * validonly SAI_NEIGHBOR_ENTRY_ATTR_IS_LOCAL == true
```
kcudnik, line 53: if this is mandatory_on_create and validonly, then you can change validonly to @condition and add it to the attributes if SAI_NEIGHBOR_ENTRY_ATTR_IS_LOCAL  is not conditional

michaelli10: Is the change below what you had in mind?  This is failing the metachecker:
ASSERT FAILED (on line 1309): SAI_NEIGHBOR_ENTRY_ATTR_DST_MAC_ADDRESS: conditional attribute must be create only

    /**                 
     * @brief Destination MAC address for the neighbor
     *
     * @type sai_mac_t        
     * @condition SAI_NEIGHBOR_ENTRY_ATTR_IS_LOCAL == true
     */
    SAI_NEIGHBOR_ENTRY_ATTR_DST_MAC_ADDRESS = SAI_NEIGHBOR_ENTRY_ATTR_START,

inc/sairouterinterface.h
```diff
@@ -285,6 +285,17 @@ typedef enum _sai_router_interface_attr_t
...
+     * @brief To flag that this interface is present locally
+     *
+     * In VOQ systems it is possible that the interface is remote
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default true
+     */
+    SAI_ROUTER_INTERFACE_ATTR_IS_LOCAL,
+
```
rlhui, line 298: Currently a router interface has a type attribute defined as port, vlan, sub_port etc., it seems we need to add a system_port for the type, as none of the existing router_interface_type is correct when the router interface is pointing to a system port.

inc/saineighbor.h
```diff
@@ -50,6 +50,7 @@ typedef enum _sai_neighbor_entry_attr_t
 
     /**
      * @brief Destination MAC address for the neighbor
+     * Valid only when SAI_NEIGHBOR_ENTRY_ATTR_IS_LOCAL == true
```
rlhui, line 53: Why change "validonly" to "Valid only when"? This might break metadata check.

michaelli10: "@validonly" (with the @) is used to enable the metadata check.  @kcudnik  brought this issue up as well but I mentioned to him that the metachecker fails since the validonly metacheck only works with "CREATE_ONLY" flags.

Kamil was OK keeping it as a comment.  I just made a minor change to cleanup the comment.

inc/sailag.h
```diff
@@ -141,6 +141,22 @@ typedef enum _sai_lag_attr_t
...
     SAI_LAG_ATTR_DROP_TAGGED,
 
+    /**
+     * @brief LAG system port ID
+     *
+     * The application must manage the allocation of the system port IDs
+     * associated with the LAG for consistency across all switches in a
+     * VOQ based system. The system port ID range is from 1 to
+     * SAI_SWITCH_ATTR_NUMBER_OF_LAGS. The default value of 0 means this
+     * field is not used and SAI will allocate system port ID internally.
```
rlhui, line 151: Did you mean SAI will allocate the SYSTEM_PORT_AGGREGATE_ID internally.  "system port ID" means different thing.
Can we make it more explicit to say "LAG system port aggregate ID", instead of just saying system port ID in this comment?

michaelli10: Sure.  Will make the comment explicitly say "system port aggregate ID".

abdosi:
In pipeline: doc/behavioral model/pipeline_v7.vsdx 

Can you please clarify How the egress l3 interface is passed from Ingress Pipeline/ASIC to Egress Pipeline/ASIC ?

PR #1082, opened 2020-05-08
mirroring sampled packets into sflow tunnel

srikrishnagopu:
In the current SAI model, packets can either be sampled to CPU using samplepacket or can be mirrored to a remote destination using ERSPAN encapsulation. This proposal is to sample packets and mirror to a remote destination using sflow encapsulation.

**SFLOW Encapsulation:**

This can be achieved in two ways:
1. Using only mirror
2. Using mirror and samplepacket

**1. Sampling using only mirror:**
This is very similar to l3 gre tunnel encapsulation for mirrored packets. Following is the sequence to create a mirror and sample it into the SFLOW tunnel.

* -	ERSPAN tunnel type is SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_SFLOW_UDP_TUNNEL.
* -	Set UDP source port using SAI_MIRROR_SESSION_ATTR_UDP_SRC_PORT
* -	Set UDP dst port using SAI_MIRROR_SESSION_ATTR_UDP_DST_PORT
* -	Set Sampling rate using SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE

Other mirror parameters like Src IP, Dst IP, TOS and TTL can be used to formulate the IP Header.

**2. Sampling using mirror and samplepacket:**
The sampling rate can be configured independently of the mirroring session which provides more flexibility. For instance, a mirroring session can be configured to create a SFLOW tunnel. Every port can have a different sample packet object with different sampling rate but can share the same mirroring session.  A new sample_packet_type_t enum has been added to the samplepacket.h to select a mirror as a sampling destination.  The mirror session object to use will be selected when the sample object is attached to a SAI port object via SAI_PORT_ATTR_INGRESS_MIRROR_SESSION or SAI_PORT_ATTR_EGRESS_MIRROR_SESSION port attributes.

Proposal Doc with example:
[SAI-Proposal-Mirroring Sampled packets.docx](https://github.com/opencomputeproject/SAI/files/4808755/SAI-Proposal-Mirroring.Sampled.packets.docx)


srikrishnagopu:
@lguohan  @JaiOCP Could you review it please ?

JaiOCP:
Changes looks good to me

inc/saimirror.h
```diff
@@ -321,6 +326,28 @@ typedef enum _sai_mirror_session_attr_t
...
+    /**
+     * @brief UDP source port. UDP source port to encapsulate
+     * SFLOW packets
+     *
+     * @type sai_uint16_t
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
+     * @isvlan false
+     * @condition SAI_MIRROR_SESSION_ATTR_TYPE == SAI_MIRROR_SESSION_TYPE_ENHANCED_REMOTE
+     */
+    SAI_MIRROR_SESSION_ATTR_UDP_SRC_PORT,
```
ashutosh-agrawal, line 338: In most IP-UDP tunnels, a hash calculated over the original packet fields can be used to populate the UDP source port for better load-balancing. Do we want a similar capability here?

srikrishnagopu: @ashutosh-agrawal. Sure. I can update this proposal or add a new one. What do you suggest ?

Shall I go ahead and add another attribute SAI_MIRROR_SESSION_ATTR_UDP_SRC_PORT_ENTROPY_ENABLE ? or setting SAI_MIRROR_SESSION_ATTR_UDP_SRC_PORT to 0 will enable entropy hashing ?

ashutosh-agrawal: A new attribute is not necessary. You can just document the expected behavior of the device when SRC_PORT is set to 0.

lguohan: this create backward compatibility issue since now existing application must to supply this attribute now, otherwise it will fail?

inc/saimirror.h
```diff
@@ -321,6 +326,28 @@ typedef enum _sai_mirror_session_attr_t
...
     SAI_MIRROR_SESSION_ATTR_POLICER,
 
+    /**
+     * @brief UDP source port. UDP source port to encapsulate
+     * SFLOW packets
+     *
+     * @type sai_uint16_t
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
+     * @isvlan false
+     * @condition SAI_MIRROR_SESSION_ATTR_TYPE == SAI_MIRROR_SESSION_TYPE_ENHANCED_REMOTE
```
lguohan, line 336: should we also add condition like

@condition SAI_MIRROR_SESSION_ATTR_TYPE == SAI_MIRROR_SESSION_TYPE_ENHANCED_REMOTE and
SAI_MIRROR_SESSION_ATTR_ERSPAN_ENCAPSULATION_TYPE == SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_SFLOW_TUNNEL

inc/saimirror.h
lguohan: also for existing attribute SAI_MIRROR_SESSION_ATTR_GRE_PROTOCOL_TYPE since it is not needed for SFLOW_TUNNEL type, we should add condition there as well.

inc/saimirror.h
srikrishnagopu: I did try this but the meta compilation was asserting. Any thoughts on how to fix this ?

 ASSERT FAILED (on line 1282): SAI_MIRROR_SESSION_ATTR_UDP_SRC_PORT: conditional attribute is also conditional, not allowed

inc/saimirror.h
srikrishnagopu: @lguohan Add the conditional check now. UDP src port/dst port is needed only for sflow tunnel and gre protocol is needed only for l3 gre tunnel.

meta/saisanitycheck.c
```diff
@@ -1279,7 +1279,20 @@ void check_attr_conditions(
 
         if (cmd->conditiontype != SAI_ATTR_CONDITION_TYPE_NONE)
         {
-            META_MD_ASSERT_FAIL(md, "conditional attribute is also conditional, not allowed");
+            if (md->objecttype == SAI_OBJECT_TYPE_MIRROR_SESSION &&
```
kcudnik, line 1282: please remove those exceptions from here, since conditional attributes cant be conditional event with exceptions

inc/saimirror.h
lguohan: @kcudnik , can you comment on this?

meta/saisanitycheck.c
srikrishnagopu: Mirror session UDP src/dst port has to be set only when the tunnel is SFLOW and  Mirror session gre type has to be set only when the tunnel is L3_GRE.

Can you clarify on how this can be handled ?

rck-innovium:
> **2. Sampling using mirror and samplepacket:**
>  The mirror session object to use will be selected when the sample object is attached to a SAI port object via SAI_PORT_ATTR_INGRESS_MIRROR_SESSION or SAI_PORT_ATTR_EGRESS_MIRROR_SESSION port attributes.

How do we support both mirroring and sflow on the same port? SAI_PORT_ATTR_INGRESS_MIRROR_SESSION points to a list of Mirror objects. A packet coming in on this port is mirrored to all the mirrors in the list.   In the proposed model, how do we find which Mirror objects should be used for sflow, and which mirror objects should be used for plain mirroring? It would be better to separate it out, say something like SAI_PORT_ATTR_INGRESS_SFLOW_MIRROR_SESSION

phshaikh:
> In the proposed model, how do we find which Mirror objects should be used for sflow, and which mirror objects should be used for plain mirroring? It would be better to separate it out, say something like

Could you clarify why do you want to treat sflow mirror different from other mirrors? There is no distinction between sflow mirror and other mirrors.

As far as I understand packet should be sent to all mirrors. Mirror only provides sampling(SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE) and encapsulation properties(SAI_MIRROR_SESSION_ATTR_TYPE) and that governs outgoing packet rate and format. Sflow will have UDP, ERSPAN will have GRE, RSPAN will be Vlan tagged, SPAN will be original packet. 

Sampling decision is either at ingress or egress side or both (depending on underlying ASIC capability). 

Sampling at ingress is done by virtue of sample packet object association with port(SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE or SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE) and sample object can have sampling rate greater than 1(SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE). 

Sampling at egress is done by virtue of sampling rate attribute of mirror object(SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE) and the sampling rate is greater 1.

If ingress or egress sample  happening at port (SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE and SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE/SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE) and sample object type is SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION, only if packet is sample candidate, it should mirrored (to all mirror objects)(SAI_PORT_ATTR_INGRESS_MIRROR_SESSION/SAI_PORT_ATTR_EGRESS_MIRROR_SESSION).

If sampling is happening at mirror session (SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE), then all mirrors should receive all packets and only those mirrors with sampling rate greater than 1, which will sample packet as per the rate. 

In short SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE and SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE, it will apply to all mirror sessions of port (and each mirror session can have their own sampling rate). Otherwise mirror sessions' sampling rate should decide sample decision. If that's not desirable, don't configure SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE/SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE and configure sample rate at mirror session.

meta/saisanitycheck.c
phshaikh: Use @validonly tag to specify UDP is only valid for SFLOW type of mirror

abdosi:
> > In the proposed model, how do we find which Mirror objects should be used for sflow, and which mirror objects should be used for plain mirroring? It would be better to separate it out, say something like
> 
> Could you clarify why do you want to treat sflow mirror different from other mirrors? There is no distinction between sflow mirror and other mirrors.
> 
> As far as I understand packet should be sent to all mirrors. Mirror only provides sampling(SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE) and encapsulation properties(SAI_MIRROR_SESSION_ATTR_TYPE) and that governs outgoing packet rate and format. Sflow will have UDP, ERSPAN will have GRE, RSPAN will be Vlan tagged, SPAN will be original packet.
> 
> Sampling decision is either at ingress or egress side or both (depending on underlying ASIC capability).
> 
> Sampling at ingress is done by virtue of sample packet object association with port(SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE or SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE) and sample object can have sampling rate greater than 1(SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE).
> 
> Sampling at egress is done by virtue of sampling rate attribute of mirror object(SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE) and the sampling rate is greater 1.
> 
> If ingress or egress sample happening at port (SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE and SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE/SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE) and sample object type is SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION, only if packet is sample candidate, it should mirrored (to all mirror objects)(SAI_PORT_ATTR_INGRESS_MIRROR_SESSION/SAI_PORT_ATTR_EGRESS_MIRROR_SESSION).
> 
> If sampling is happening at mirror session (SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE), then all mirrors should receive all packets and only those mirrors with sampling rate greater than 1, which will sample packet as per the rate.
> 
> In short SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE and SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE, it will apply to all mirror sessions of port (and each mirror session can have their own sampling rate). Otherwise mirror sessions' sampling rate should decide sample decision. If that's not desirable, don't configure SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE/SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE and configure sample rate at mirror session.

rck-innovium:
> > In the proposed model, how do we find which Mirror objects should be used for sflow, and which mirror objects should be used for plain mirroring? It would be better to separate it out, say something like
> 
> Could you clarify why do you want to treat sflow mirror different from other mirrors? There is no distinction between sflow mirror and other mirrors.
> 
> As far as I understand packet should be sent to all mirrors. Mirror only provides sampling(SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE) and encapsulation properties(SAI_MIRROR_SESSION_ATTR_TYPE) and that governs outgoing packet rate and format. Sflow will have UDP, ERSPAN will have GRE, RSPAN will be Vlan tagged, SPAN will be original packet.
> 
> Sampling decision is either at ingress or egress side or both (depending on underlying ASIC capability).
> 
> Sampling at ingress is done by virtue of sample packet object association with port(SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE or SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE) and sample object can have sampling rate greater than 1(SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE).
> 
> Sampling at egress is done by virtue of sampling rate attribute of mirror object(SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE) and the sampling rate is greater 1.
> 
> If ingress or egress sample happening at port (SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE and SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE/SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE) and sample object type is SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION, only if packet is sample candidate, it should mirrored (to all mirror objects)(SAI_PORT_ATTR_INGRESS_MIRROR_SESSION/SAI_PORT_ATTR_EGRESS_MIRROR_SESSION).
> 
> If sampling is happening at mirror session (SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE), then all mirrors should receive all packets and only those mirrors with sampling rate greater than 1, which will sample packet as per the rate.
> 
> In short SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE and SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE, it will apply to all mirror sessions of port (and each mirror session can have their own sampling rate). Otherwise mirror sessions' sampling rate should decide sample decision. If that's not desirable, don't configure SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE/SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE and configure sample rate at mirror session.

The original question is about the 2nd model, viz. __Sampling using mirror and samplepacket__

I think where we both differ is that I want to support both sFlow and mirroring on the same ingress port. For example, on port p1, I have multiple mirror sessions. On some of these mirrors, I dont want sampling, and in some mirror sessions I want the same sample rate.  And I want to use the 2nd model so that I can conserve my samplers to one per port since all these use the same rate.

With the proposed model, how do we realize the above example? Having a new attribute avoids this limitation with the proposed model.

abdosi:
@phshaikh:  

Regarding your below description:

If ingress or egress sample happening at port (SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE and SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE/SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE) and sample object type is SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION, only if packet is sample candidate, it should mirrored (to all mirror objects)(SAI_PORT_ATTR_INGRESS_MIRROR_SESSION/SAI_PORT_ATTR_EGRESS_MIRROR_SESSION).

>>> Ideally that means if sflow is enable on port then we cannot mirror all packets. In most of ASIC Mirroring and SFLOW are two different operation and each one can have is own session/destination. We still want to mirror all packets to one session/encapsualtion and sflow some packets at given rate to another session/encapsulation 
I think that is point @rck-innovium and we brought up in meeting.

rck-innovium:
> @phshaikh:
> 
> Regarding your below description:
> 
> If ingress or egress sample happening at port (SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE and SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE/SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE) and sample object type is SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION, only if packet is sample candidate, it should mirrored (to all mirror objects)(SAI_PORT_ATTR_INGRESS_MIRROR_SESSION/SAI_PORT_ATTR_EGRESS_MIRROR_SESSION).
> 
> > > > Ideally that means if sflow is enable on port then we cannot mirror all packets. In most of ASIC Mirroring and SFLOW are two different operation and each one can have is own session/destination. We still want to mirror all packets to one session/encapsualtion and sflow some packets at given rate to another session/encapsulation
> > > > I think that is point @rck-innovium and we brought up in meeting.

Exactly.

abdosi:
also on top what is expectation for below case:

Port base sampling is enable and SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION is configured.

a) If we have SAI_PORT_ATTR_INGRESS_MIRROR_SESSION point to two Session

Session 1 of type SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_SFLOW_TUNNEL and 
Session 2 of type SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_L3_GRE_TUNNEL

we will receive SFLOW copy on both but that is expected/desired ? since we have dedicated SFLOW Tunnel type do we still want to get Sflow copy on GRE tunnel also ? 

b) Also another question I had if mirror session  tunnel type is SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_SFLOW_TUNNEL do we want to support SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE for that session ? If we can configure then we have Sampling 2 times and am not sure that can be supported.

phshaikh:
@rck-innovium 

>The original question is about the 2nd model, viz. Sampling using mirror and samplepacket

>I think where we both differ is that I want to support both sFlow and mirroring on the same >ingress port. For example, on port p1, I have multiple mirror sessions. On some of these >mirrors, I dont want sampling, and in some mirror sessions I want the same sample rate. And I >want to use the 2nd model so that I can conserve my samplers to one per port since all these >use the same rate.

>With the proposed model, how do we realize the above example? Having a new attribute >avoids this limitation with the proposed model.

Sorry, if I missed but which attribute are you proposing?

@abdosi @rck-innovium 
>Ideally that means if sflow is enable on port then we cannot mirror all packets. In most of ASIC >Mirroring and SFLOW are two different operation and each one can have is own >session/destination. We still want to mirror all packets to one session/encapsualtion and sflow >some packets at given rate to another session/encapsulation
>I think that is point @rck-innovium and we brought up in meeting.

"sflow" here is an encapsulation format (mainly UDP + any other implementation defined metadata) and not any specific operation. Sampling at the source is an operation represented by sample packet object. Set up sample packet object with sample rate of 1 and use sample rate attribute of mirror session. Does that help? It allows you to set up multiple mirrors while some mirrors sending all traffic while one with mirror with sample rate sending traffic as per sample rate.


>also on top what is expectation for below case:
>
>Port base sampling is enable and SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION is configured.
>
>a) If we have SAI_PORT_ATTR_INGRESS_MIRROR_SESSION point to two Session
>
>Session 1 of type SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_SFLOW_TUNNEL and
>Session 2 of type SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_L3_GRE_TUNNEL

>we will receive SFLOW copy on both but that is expected/desired ? since we have dedicated >SFLOW Tunnel type do we still want to get Sflow copy on GRE tunnel also ?

Why will there be "SFLOW" copy? Both mirrors will get original packet and will encapsulate as per their own properties or attributes. SFLOW_TUNNEL mirror will emit packet with sflow/udp encapsulation and L3_GRE_TUNNEL mirror will emit packet with gre encapsulation. 

>b) Also another question I had if mirror session tunnel type is >SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_SFLOW_TUNNEL do we want to support >SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE for that session ? If we can configure then we >have Sampling 2 times and am not sure that can be supported.

There are ASIC which support setting sample rate at mirror. It is upto underlying adapter to verify configuration and throw invalid parameter error if it can not accept configuration which application wants. Whether to use sample rate from sample object or from mirror object is also entirely prerogative of underlying implementation. Spec has to be as generic as possible to allow different vendors to implement sampling with different encapsulation (sflow being one of them).

phshaikh:
@abdosi it would be better if you post new comments with markdown formatting, Now i am totally lost which one is your comment and which one is mine since i think you edited my comment :-)

phshaikh:
@abdosi 

Assuming this is from you :-)

>@phshaikh, Sorry may be I am missing some thing here. Basic Question using SAI_SAMPLEPACKET_ATTR_TYPE = SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION means what ? As I understand once we do that we start using mirror destinations define by SAI_PORT_ATTR_INGRESS_MIRROR_SESSION (ingress direction) as sflow copy destination (which is as of now default to CPU only) Is it correct ?

SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION means a sample (original packet) should be sent to mirror. There is no "encapsulation" indication here. Also I don't think spec specifies that sample is sent "encapsulated" in sflow to CPU for SAI_SAMPLEPACKET_TYPE_SLOW_PATH. In most cases, ASIC sends packet as is to CPU and application is expected to do sflow encapsulation. Here SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION means, instead of sending to CPU, send it to mirror and let mirror's encap property govern encapsulation.

>I think from your perspective SAI_PORT_ATTR_INGRESS_MIRROR_SESSION can use by two application sflow and mirroring simultaneously. when being used in mirroring context all packets will be mirrored and in using sflow context only sample packets will be sent. Is it true ?

SAI_PORT_ATTR_INGRESS_MIRROR_SESSION is mirror (or list of mirrors) associated with port which mirror "ingress" traffic.

Consider this as pseudo code of SAI objects

```
sai_mirror {SFLOW} // mirror object
sample_object {SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION} // sample object to mirror
sai_port {SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE : sample_object, } // sample to mirrors
sai_port{SAI_PORT_ATTR_INGRESS_MIRROR_SESSION: sai_mirror} // list of mirrors to sample to
```

This means application created mirror of SFLOW type, associated it with port. And if sai_port is also using sample object of type, SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION, sampled packets will go to sai_mirror object.

If there're more than one mirror, all those mirrors will get sampled packet

rck-innovium:
> @rck-innovium
> 
> > With the proposed model, how do we realize the above example? Having a new attribute >avoids this limitation with the proposed model.
> 
> Sorry, if I missed but which attribute are you proposing?
> 

 Pls refer to the last sentence of my first comment. SAI_PORT_ATTR_INGRESS_SFLOW_MIRROR_SESSION. 
And I was wrong to use the term sFlow instead of sampling. I am rephrasing my original comment:



How do we support both mirroring and sampling on the same port? SAI_PORT_ATTR_INGRESS_MIRROR_SESSION points to a list of Mirror objects. A packet coming in on this port is mirrored to all the mirrors in the list. In the proposed model, how do we find which Mirror objects should be used for sampling, and which mirror objects should be used for plain mirroring without sampling? It would be better to separate it out, say something like SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION

phshaikh:
> > @rck-innovium
> > > With the proposed model, how do we realize the above example? Having a new attribute >avoids this limitation with the proposed model.
> > 
> > 
> > Sorry, if I missed but which attribute are you proposing?
> 
> Pls refer to the last sentence of my first comment. SAI_PORT_ATTR_INGRESS_SFLOW_MIRROR_SESSION.
> And I was wrong to use the term sFlow instead of sampling. I am rephrasing my original comment:
> 
> How do we support both mirroring and sampling on the same port? SAI_PORT_ATTR_INGRESS_MIRROR_SESSION points to a list of Mirror objects. A packet coming in on this port is mirrored to all the mirrors in the list. In the proposed model, how do we find which Mirror objects should be used for sampling, and which mirror objects should be used for plain mirroring without sampling? It would be better to separate it out, say something like SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION

In your model, 
1) does the new attribute belong to sample packet object or port object? 
2) is the new attribute going to be list of mirrors or just one mirror to receive samples?
3) will the same mirror objects exist in both attributes - new attribute and port's list of mirror sessions?
4) what do we gain with this model? The existing model is simpler in which sampled traffic goes to all mirrors but in the new model, sampled traffic goes to subset of mirrors (identified) by new attribute while un-sampled traffic goes to other mirrors. It also calls for differentiated treatment both in application and in adapter code among mirrors based on which should get sampled traffic vs unsampled traffic. @srikrishnagopu / @JaiOCP  your thoughts on this?

srikrishnagopu:
I have attached a document with an example as per the discussion we had on thursday.

@rck-innovium @abdosi Can we get a closure on this please ?  As @phshaikh said, I dont think its useful to identify only a subset of mirrors to be sampled and let the remaining mirrors unsampled.

In order to sample only certain mirror sessions, it can be achieved by specifying the sampling rate (SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE) on the mirror object itself. The main motivation of this is to share the mirror sessions across different ports but use different sampling rates.

abdosi:
> I have attached a document with an example as per the discussion we had on thursday.
> 
> @rck-innovium @abdosi Can we get a closure on this please ? As @phshaikh said, I dont think its useful to identify only a subset of mirrors to be sampled and let the remaining mirrors unsampled.
> 
> In order to sample only certain mirror sessions, it can be achieved by specifying the sampling rate (SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE) on the mirror object itself. The main motivation of this is to share the mirror sessions across different ports but use different sampling rates.

@srikrishnagopu I also feel we need another attribute here. 
For eg: take your above pseudo code.
How can I define another regular mirror session which will receive all  mirror all packet ingress on the same port. Can we have pseudo-code for same ?

Our point is on port P0 you can have sampling/sflow enable and you can mirror sflow copy to session {S0,S1} and at the same time on same Port P0 enable ingress mirroring to mirror all incoming traffic to session {S2}

I don't that above use-case can be supported with current proposed SAI changes.

inc/saimirror.h
kcudnik: yes, we don't allow conditional attributes to be dependend on other conditional attributes

srikrishnagopu: Added a valid attribute now.

rck-innovium:
I was hoping to discuss this in today's meeting but we ran out of time. Please see inline.


> SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION



> > > @rck-innovium
> > > > With the proposed model, how do we realize the above example? Having a new attribute >avoids this limitation with the proposed model.
> > > 
> > > 
> > > Sorry, if I missed but which attribute are you proposing?
> > 
> > 
> > Pls refer to the last sentence of my first comment. SAI_PORT_ATTR_INGRESS_SFLOW_MIRROR_SESSION.
> > And I was wrong to use the term sFlow instead of sampling. I am rephrasing my original comment:
> > How do we support both mirroring and sampling on the same port? SAI_PORT_ATTR_INGRESS_MIRROR_SESSION points to a list of Mirror objects. A packet coming in on this port is mirrored to all the mirrors in the list. In the proposed model, how do we find which Mirror objects should be used for sampling, and which mirror objects should be used for plain mirroring without sampling? It would be better to separate it out, say something like SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION
> 
> In your model,
> 
> 1. does the new attribute belong to sample packet object or port object?


**SAI_PORT_ATTR**_INGRESS_SAMPLE_MIRROR_SESSION  => it is a SAI Port object attribute

> 2. is the new attribute going to be list of mirrors or just one mirror to receive samples?

A list of mirrors. An OID list same as SAI_PORT_ATTR_INGRESS_MIRROR_SESSION.

> 3. will the same mirror objects exist in both attributes - new attribute and port's list of mirror sessions?

It is upto the user but reasonable to see that it will be in only one list.

> 4. what do we gain with this model? The existing model is simpler in which sampled traffic goes to all mirrors but in the new model, sampled traffic goes to subset of mirrors (identified) by new attribute while un-sampled traffic goes to other mirrors. 


The new model is simple if you just use SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION instead of SAI_PORT_ATTR_INGRESS_MIRROR_SESSION. Whatever you had suggested will work.

The advantage with having this new attribute is for those devices that can support the case where traffic coming in on a port can be sampled for certain destinations, and sent out unsampled for other destinations. 
Also, it does not change the behavior for existing implementations.

> It also calls for differentiated treatment both in application and in adapter code among mirrors based on which should get sampled traffic vs unsampled traffic. @srikrishnagopu / @JaiOCP  your thoughts on this?

They are two different sessions, so they can go to two different applications. One application can be an sflow collector and the other application a mirror collector.

srikrishnagopu:
@abdosi @rck-innovium @phshaikh I have updated the proposal document and the changes as per the review comments.  I have added a new attribute in saiport to differentiate the sampled mirrored packets. 

Please let me know.

inc/saiport.h
```diff
@@ -912,6 +912,34 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE,
 
+    /**
+     * @brief Enable/Disable Samplepacket session
+     *
+     * Enable sample ingress mirroring by assigning mirror object id Disable
+     * sample ingress mirroring by assigning #SAI_NULL_OBJECT_ID as attribute value.
+     *
+     * @type sai_object_id_t
```
rck-innovium, line 921: * @type sai_object_list_t

inc/saiport.h
```diff
@@ -912,6 +912,34 @@ typedef enum _sai_port_attr_t
...
+     */
+    SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION,
+
+    /**
+     * @brief Enable/Disable Samplepacket session
+     *
+     * Enable sample egress mirroring by assigning mirror object id Disable
+     * sample egress mirroring by assigning #SAI_NULL_OBJECT_ID as attribute value.
+     *
+     * @type sai_object_id_t
```
rck-innovium, line 935: * @type sai_object_list_t

inc/saiport.h
srikrishnagopu: Thank you. Fixed them now.

rck-innovium:
+1

meta/saisanitycheck.c
srikrishnagopu: @kcudnik can you check now please ?

inc/saisamplepacket.h
```diff
@@ -41,6 +41,9 @@ typedef enum _sai_samplepacket_type_t
     /** Copy the sample packets to CPU */
     SAI_SAMPLEPACKET_TYPE_SLOW_PATH,
 
+    /** Copy the sample packets using mirror session */
+    SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION,
```
wendani, line 45: From reference guide I read, some chip is capable of copying a packet sample to CPU and, at the same time, sending the sample to 1 or more mirror sessions . Not sure of any practical use case of such a capability, but as `SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE` and `SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE` takes only one sample object id, not an objlist like the case of `SAI_PORT_ATTR_INGRESS_MIRROR_SESSION` and `SAI_PORT_ATTR_EGRESS_MIRROR_SESSION`, this limits a sample to be either sent to CPU or sent to mirror session(s) as stated in the first sentence of the proposal, but not both as some chip can be capable of.

srikrishnagopu: @wendani Currently, having a single sample object OID will help us configure the same sampling rate for all mirror sessions (which was the proposal) and making it a list will over complicate it. I would like to keep it as is and would like to take this proposal forward.

PR #1084, opened 2020-05-11
Add tunnel termination table entry attributes for specifying src and dst IP masks

dushyantarora:
1. Added MP2P and MP2MP tunnel termination table entry types.
2. Add SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_DST_IP_MASK and SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_SRC_IP_MASK attributes to allow specifying mask address for dst and src IP.


inc/saitunnel.h
```diff
@@ -822,17 +828,18 @@ typedef enum _sai_tunnel_term_table_entry_attr_t
     /**
      * @brief Tunnel termination IP address
      *
-     * @type sai_ip_address_t
+     * @type sai_ip_prefix_t
```
lguohan, line 831: this change is not backward compatible.

dushyantarora: Thanks for the comment. You are right I have changed the data type. However, you can still specify an IP address by providing a mask of all 1s, making it backwards compatible wrt the intent. In a nutshell, I am proposing to make the data type more generic to take advantage ASICs which support tunnel termination of prefixes.

lguohan: how, this still need all existing application to change their codes. can you introduce additional attribute to avoid the need to change existing applications?

dushyantarora: Thanks for the comment. I got the same feedback in today's SAI WG meeting. The consensus we reached in the meeting was to add new attributes for specifying src and dst IP mask. PTAL.

inc/saitunnel.h
```diff
@@ -824,18 +830,39 @@ typedef enum _sai_tunnel_term_table_entry_attr_t
...
+     * @condition SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TYPE == SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_P2P or SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TYPE == SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_P2MP
      */
     SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_DST_IP,
 
+    /**
+     * @brief Tunnel termination IP address mask
+     *
+     * @type sai_ip_address_t
+     * @flags CREATE_ONLY
+     * @default 0.0.0.0
```
lguohan, line 842: default to be 255.255.255.255 to enable backward compatiblity?

inc/saitunnel.h
```diff
@@ -824,18 +830,39 @@ typedef enum _sai_tunnel_term_table_entry_attr_t
...
 
+    /**
+     * @brief Tunnel source IP address mask
+     *
+     * @type sai_ip_address_t
+     * @flags CREATE_ONLY
+     * @default 0.0.0.0
+     * @validonly SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TYPE == SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_P2MP or SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TYPE == SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_MP2MP
+     */
+    SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_SRC_IP_MASK,
```
lguohan, line 864: default to be 255.255.255.255 to enable backward compatibility?

abdosi: same comment as above.

inc/saitunnel.h
abdosi: since it is not Mandatory-on Create it will be backward compatible.  But I feel default we can remove as based on v4/v6 it's value will be different . When this attribute is user is always expected to pass value.

dushyantarora: Ack. Removed default.

lguohan: what does this 0.0.0.0 mean here? I can see the default is here.

inc/saitunnel.h
dushyantarora: See above reply.

inc/saitunnel.h
dushyantarora: Sorry I hit reply on comment too soon. I have edited my prev comment. Please let me know if your question is not answered by the updated reply.

abdosi (commented):
LGTM. Has this been Verified it does not break existing SAI model with new default values ?

rlhui:
> LGTM. Has this been Verified it does not break existing SAI model with new default values ?

@dushyantarora  - could you comment? Thanks.

dushyantarora:
Sorry I did not get a notification for this until now. Thanks for tagging me Rita. The change is passing the existing checks. Is there more verification that needs to be done? If so, how should I do it?

rlhui:
@JaiOCP  - please review/approve if this is ok to you/Broadcom. Thanks.

rlhui:
@lguohan  - are we ok to remove the "breaking-change" label for this one?

rlhui (approved):

rlhui:
> Sorry I did not get a notification for this until now. Thanks for tagging me Rita. The change is passing the existing checks. Is there more verification that needs to be done? If so, how should I do it?

@abdosi - would you like to respond on this?  Thanks.

prsunny (approved):

rlhui:
@marian-pritsak  - please help reivew/signoff. Thanks.

marian-pritsak (approved):

PR #1086, opened 2020-05-14
Create and Set for Tunnel Attributes

JaiOCP:
Currently Tunnel attributes are create only. They need to create and set as they can change runtime.


rlhui:
@marian-pritsak , would you please help review/approve? Thanks

rlhui:
@kcudnik , please help approve, thanks.

PR #1088, opened 2020-05-22
Tunnel/RIF QoS Map Support

JaiOCP:
This PR brings in following support
- Ability to apply QoS Map on a RIF interface
- Ability to apply QoS Map on a Tunnel interface
  * Ingress packet (IP or non IP) will go through TC/Color -> DSCP QoS Map and resulting DSCP value or color is applied on outer header i.e. tunnel DSCP value
 * On a tunnel decap, tunnel DSCP goes through DSCP -> TC/Color Map and resulting DSCP (if IP packet) or .1p bits (if non IP packet) is applied to inner header i.e. customer packet
 * This scheme helps handle non IP to tunnel DSCP mapping and also helps fine grain QoS policy in a multi tenancy environment

### DIFF
--- a/inc/sairouterinterface.h
+++ b/inc/sairouterinterface.h
@@ -286,6 +286,39 @@ typedef enum _sai_router_interface_attr_t
     SAI_ROUTER_INTERFACE_ATTR_DECREMENT_TTL,
 
     /**
+     * @brief Enable DSCP -> TC MAP
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_QOS_MAP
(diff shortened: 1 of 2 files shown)


lguohan:
if there is map on port, rif and tunnel, which one take precedence?

JaiOCP:
> if there is qos map applying on port, rif and tunnel, which one take precedence?

L3 packet ingress on a RIF will use the qos map applied on RIF. If Port is configured as L3 interface (single RIF) then QoS map applied on port takes effect. If packet is ingress on a RIF or Vlan and egress on a tunnel then QoS Map applied on RIF/VLAN and tunnel both will take effect.

PR #1090, opened 2020-05-28
Generic Hostif Trap Exclude Object

JaiOCP:
This PR introduces generic exclude object for control packets.
Exclude object is part of the hostif attribute list and can be associated with a particular trap type, packet action.

Note that this is an exclusion object for control traffic only.

### DIFF
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -458,7 +458,7 @@ typedef enum _sai_hostif_trap_attr_t
     SAI_HOSTIF_TRAP_ATTR_TRAP_PRIORITY,
 
     /**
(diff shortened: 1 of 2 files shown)


JaiOCP:
Folks,

We were able to figure out a way to handle the use case of suppressing ARP to a flood domain using the hostif table.
1. Create hostif table entry for vlan type
2. Create hostif trap attribute for trap id ARP Request and packet action as TRAP (this will copy the packet
3. Create hostif object with priority and queue information

Host if table entries are per vlan and flood behavior can be achieved on a per vlan basis using this.

I am going to close this PR and open a new one just for adding new IPv6 trap types.

PR #1093, opened 2020-06-02
Fix linker error when compiling SAI RPC server.

smaheshm:
The variable was not part of the librpcserver.a library. Moved the variable to the source file that's added in librpcserver.a.


test/saithrift/src/saiserver.cpp
```diff
@@ -70,6 +69,7 @@ void on_fdb_event(_In_ uint32_t count,
            
     sai_fdb_entry_t fdb_m;
     sai_object_id_t b_id;
+    extern std::vector<std::pair<sai_fdb_entry_t, sai_object_id_t>> gFdbMap;
```
lguohan, line 72: move this out of function?

PR #1097, opened 2020-06-11
Added attribute to identify  vlan type.

rajkumar38:
Added attribute to identify  vlan type, which can be one of,
SAI_VLAN_TYPE_DATA : VLAN configured to carry user generated data traffic.
SAI_VLAN_TYPE_MANAGEMENT : VLAN configured to access management capabilities of the node.
SAI_VLAN_TYPE_CONTROL : VLAN configured to carry traffic between the devices on same node or between the nodes.

Default : SAI_VLAN_TYPE_DATA

Signed-off-by: Rajkumar Pennadam Ramamoorthy <rpennadamram@marvell.com>

### DIFF
--- a/inc/saivlan.h
+++ b/inc/saivlan.h
@@ -134,7 +134,8 @@ typedef enum _sai_vlan_attr_t
      * @brief Vlan Type
      *
      * @type sai_vlan_type_t
-     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @flags CREATE_ONLY
+     * @default SAI_VLAN_TYPE_DATA
      */
     SAI_VLAN_ATTR_VLAN_TYPE,
 
--- a/inc/saivlan.h
+++ b/inc/saivlan.h
@@ -134,8 +134,7 @@ typedef enum _sai_vlan_attr_t
      * @brief Vlan Type
      *
      * @type sai_vlan_type_t
-     * @flags MANDATORY_ON_CREATE | CREATE_ONLY | KEY
-     * @default SAI_VLAN_TYPE_DATA
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      */
     SAI_VLAN_ATTR_VLAN_TYPE,
 
--- a/inc/saivlan.h
+++ b/inc/saivlan.h
@@ -98,6 +98,19 @@ typedef enum _sai_vlan_flood_control_type_t
 } sai_vlan_flood_control_type_t;
 
 /**
+ * @brief Attribute data for vlan type parameter
+ */
+typedef enum _sai_vlan_type_t
+{
+    SAI_VLAN_TYPE_DATA,
+
+    SAI_VLAN_TYPE_MANAGEMENT,
+
+    SAI_VLAN_TYPE_CONTROL
+
+} sai_vlan_type_t;
+
+/**
  * @brief Attribute Id in sai_set_vlan_attribute() and
  * sai_get_vlan_attribute() calls
  */
@@ -116,6 +129,15 @@ typedef enum _sai_vlan_attr_t
      * @isvlan true
      */
     SAI_VLAN_ATTR_VLAN_ID = SAI_VLAN_ATTR_START,
+
+    /**
+     * @brief Vlan Type
+     *
+     * @type sai_vlan_type_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY | KEY
+     * @default SAI_VLAN_TYPE_DATA
+     */
+    SAI_VLAN_ATTR_VLAN_TYPE,
 
     /**
      * @brief List of vlan members in a VLAN


inc/saivlan.h
```diff
@@ -119,6 +132,14 @@ typedef enum _sai_vlan_attr_t
      */
     SAI_VLAN_ATTR_VLAN_ID = SAI_VLAN_ATTR_START,
 
+    /**
+     * @brief Vlan Type
+     *
+     * @type sai_vlan_type_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
```
kcudnik, line 139: making this mandatory on create breaks backward compatibility with creating vlans,
would it be possible to make this create_only and assign default value?

lguohan:
can you describe the motivation here? why do we need to differentiate the traffic type in the data plane?

inc/saivlan.h
rajkumar38: Agree. This attribute can be CREATE_ONLY optional param with default value, "SAI_VLAN_TYPE_DATA". I will push the changes.

rajkumar38:
> can you describe the motivation here? why do we need to differentiate the traffic type in the data plane?

We offer the following use-case.

A system that comprises a switch running SONiC with dedicated CPU that manages the entire system.
Other components in the systems are hidden from the outside network.
For such a use case, the network operator would like to assign one or more VLANs for CONTROL traffic between the different system components. These VLANs must not be exposed to the uplink ports.
Designating these VLANs as CONTROL facilitates an easy and consistent definition amounts the different system components as well as easing changing the CONTROL VLAN.
The outside network requires ability to control the switch and its CPU using one or more MANAGEMENT VLANs.

Currently, the network operator required to manually add members to each VLAN but nothing ensures that the same group of ports is member of all COTROL VLANs and all of a second group of ports is members in all MANAGEMENT VLANs.

bandaru-viswanath:
> > can you describe the motivation here? why do we need to differentiate the traffic type in the data plane?
> 
> We offer the following use-case.
> 
> A system that comprises a switch running SONiC with dedicated CPU that manages the entire system.
> Other components in the systems are hidden from the outside network.
> For such a use case, the network operator would like to assign one or more VLANs for CONTROL traffic between the different system components. These VLANs must not be exposed to the uplink ports.
> Designating these VLANs as CONTROL facilitates an easy and consistent definition amounts the different system components as well as easing changing the CONTROL VLAN.
> The outside network requires ability to control the switch and its CPU using one or more MANAGEMENT VLANs.
> 
> Currently, the network operator required to manually add members to each VLAN but nothing ensures that the same group of ports is member of all COTROL VLANs and all of a second group of ports is members in all MANAGEMENT VLANs.

It is not very clear on what the expectations are from a SAI implementation. From this description, it still seems that application (or even the user) is making the decision on which ports belong to each vlan. The notion of what is a 'control vlan' vs 'management vlan' seems to be belonging to Application or above, but not necessarily SAI. For example, in your comment you say -

> Currently, the network operator required to manually add members to each VLAN but nothing ensures that the same group of ports is member of all COTROL VLANs

Do you expect SAI to automatically add all ports that belong to one control vlan, to all other control vlans ? 

It would be best to describe the SAI implementation requirements when a vlan is specified as a control vlan, or a data vlan or a management vlan.

rajkumar38:
> > > can you describe the motivation here? why do we need to differentiate the traffic type in the data plane?
> > 
> > 
> > We offer the following use-case.
> > A system that comprises a switch running SONiC with dedicated CPU that manages the entire system.
> > Other components in the systems are hidden from the outside network.
> > For such a use case, the network operator would like to assign one or more VLANs for CONTROL traffic between the different system components. These VLANs must not be exposed to the uplink ports.
> > Designating these VLANs as CONTROL facilitates an easy and consistent definition amounts the different system components as well as easing changing the CONTROL VLAN.
> > The outside network requires ability to control the switch and its CPU using one or more MANAGEMENT VLANs.
> > Currently, the network operator required to manually add members to each VLAN but nothing ensures that the same group of ports is member of all COTROL VLANs and all of a second group of ports is members in all MANAGEMENT VLANs.
> 
> It is not very clear on what the expectations are from a SAI implementation. From this description, it still seems that application (or even the user) is making the decision on which ports belong to each vlan. The notion of what is a 'control vlan' vs 'management vlan' seems to be belonging to Application or above, but not necessarily SAI. For example, in your comment you say -
> 
> > Currently, the network operator required to manually add members to each VLAN but nothing ensures that the same group of ports is member of all COTROL VLANs
> 
> Do you expect SAI to automatically add all ports that belong to one control vlan, to all other control vlans ?
> 
> It would be best to describe the SAI implementation requirements when a vlan is specified as a control vlan, or a data vlan or a management vlan.

Indeed, SAI would keep all VLANs designated as Control or Management consistent.
So adding new Control VLAN would cause SAI to add it to all ports members in Control.
This way, there is a default, out of the box setting, but changing or adding a VLAN is transparent to the user/application as SAI maintains the consistency.

In case of in-band management, implantation must ensure that traffic is trapped to centralized host CPU.

Data-vlan is no different from current implementation. It carries the regular data-traffic.

Another use-case of multi-chassis architecture, wherein management and control-traffic from data-path (say on a network port from one of the line card in slave chassis) may switch via back-plane control switch (different from data-plane switch) to reach host CPU or other sub-system in the NE.
In such scenario, we have to add these vlan configurations in back-plane switch for switching the traffic. 
Hence, these special vlan needs to be differentiated and additional configurations has to be applied in NE that are not managed by SAI layer.

atitjain:
> SAI_VLAN_TYPE_MANAGEMENT : VLAN configured to access management capabilities of the node.
I think the management traffic that you refer here is still ssh/telnet etc, which has to go to CPU based on the routing/forwarding. I dont think a vanila trap to CPU on the management VLAN solves the purpose. VLAN might help in doing the QoS. Like giving management traffic some bw and priority in scheduling. Over the network, you still need to create the topology for the VLAN that you intend to use for this purpose. I would still treat this as a data vlan for most of the configuration purpose. In case you want to set it up out of the box, use SAI init config facility. I don't think you intend to route the traffic in this vlan in the control plane to default lifting to CPU on all nodes isn't really useful.

> SAI_VLAN_TYPE_CONTROL : VLAN configured to carry traffic between the devices on same node or between the nodes.
I think this is more of internal implementation and should be managed by using VLAN translation etc. Is the back-plane switch visible as an entity to the SAI application? How does this all fit into the SAI Bridge implementation.

inc/saivlan.h
```diff
@@ -117,6 +130,15 @@ typedef enum _sai_vlan_attr_t
...
     SAI_VLAN_ATTR_VLAN_ID = SAI_VLAN_ATTR_START,
 
+    /**
+     * @brief Vlan Type
+     *
+     * @type sai_vlan_type_t
+     * @flags CREATE_ONLY
+     * @default SAI_VLAN_TYPE_DATA
+     */
+    SAI_VLAN_ATTR_VLAN_TYPE,
```
kcudnik, line 140: can you move this as last attribute? in ongoing trend to be enum values backward compatible this would break

kcudnik (changes requested):
can you move this as last attribute? in ongoing trend to be enum values backward compatible this would break

PR #1098, opened 2020-06-13
Additional Port Interface Types

phshaikh:
Summary:

Providing more options for port interface types to allow more granular programmability of port settings. These interface types are part of  IEEE 802.3 standard

GMII - Gigabit
SFI - 10 Gigabit
XLAUI - 40 Gigabit
CAUI - 100 Gigabit

Each of these have different electrical characteristics and different from available port interface types.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -249,11 +249,11 @@ typedef enum _sai_port_interface_type_t
     /** Interface type XLAUI */
     SAI_PORT_INTERFACE_TYPE_XLAUI,
 
-    /** Interface type CUSTOM BEGIN */
-    SAI_PORT_INTERFACE_TYPE_CUSTOM_BEGIN,
+    /** Custom range base value */
+    SAI_PORT_INTERFACE_TYPE_CUSTOM_RANGE_START = 0x10000000,
 
-    /** Interface type CUSTOM END */
-    SAI_PORT_INTERFACE_TYPE_CUSTOM_END = SAI_PORT_INTERFACE_TYPE_CUSTOM_BEGIN,
+    /** End of custom range base */
+    SAI_PORT_INTERFACE_TYPE_CUSTOM_RANGE_END = SAI_PORT_INTERFACE_TYPE_CUSTOM_RANGE_START,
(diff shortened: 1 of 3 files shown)


inc/saiport.h
```diff
@@ -236,6 +236,24 @@ typedef enum _sai_port_interface_type_t
...
+    SAI_PORT_INTERFACE_TYPE_SFI,
+
+    /** Interface type XLAUI */
+    SAI_PORT_INTERFACE_TYPE_XLAUI,
+
+    /** Interface type CUSTOM BEGIN */
+    SAI_PORT_INTERFACE_TYPE_CUSTOM_BEGIN,
+
+    /** Interface type CUSTOM END */
+    SAI_PORT_INTERFACE_TYPE_CUSTOM_END = SAI_PORT_INTERFACE_TYPE_CUSTOM_BEGIN,
```
kcudnik, line 256: Can you change that to: SAI_PORT_INTERFACE_TYPE_CUSTOM_BASE =  0x10000000
this way it will follow scheme like in other enums

phshaikh: Yes, i will do it shortly!

phshaikh: @kcudnik done!

inc/saiport.h
```diff
@@ -236,6 +236,24 @@ typedef enum _sai_port_interface_type_t
...
+
+    /** Interface type GMII */
+    SAI_PORT_INTERFACE_TYPE_GMII,
+
+    /** Interface type SFI */
+    SAI_PORT_INTERFACE_TYPE_SFI,
+
+    /** Interface type XLAUI */
+    SAI_PORT_INTERFACE_TYPE_XLAUI,
+
```
JaiOCP, line 251: @kcudnik Do we need to add SAI_PORT_INTERFACE_TYPE_MAX or extensions can be added using CUSTOM_RANGE_START.

phshaikh: how do we plan to use SAI_PORT_INTERFACE_TYPE_MAX?

PR #1106, opened 2020-06-17
Voq System

Israel-Meilik:
Explain "add route" in a VOQ System

Define and explain changes to Router-Interface and Neighbor create

Show changes to the pipeline model


rlhui (commented):
There are two "SAI VoQ System.pdf" in this PR.  Should one of them be removed?  Also could we please move it to doc/VoQ folder?  Thanks.

rlhui (commented):
Do we need "SAI-Proposal-VoQ-Switch.docx" in this PR?  There's already a similar doc in .md format merged in earlier.

rlhui:
@Israel-Meilik

PR #1108, opened 2020-06-18
updated sailport.h

tarakreddy1984:
Fail over mode support

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -313,23 +313,23 @@ typedef enum _sai_port_fail_over_mode_t
 
 /**
  * @brief Attribute data for #SAI_PORT_ATTR_FAIL_OVER_CONFIG_MODE
- * Used for Fail over configuration mode on port w.r.t hitless mode
+ * Used for Fail over configuration mode on port w.r.t hit less mode
  */
 typedef enum _sai_port_fail_over_config_mode_t
 {
-    /** Fail over ports are configured but do not operate in hitless mode */
+    /** Fail over ports are configured but do not operate in hit less mode */
     SAI_PORT_FAIL_OVER_CONFIG_MODE_NO_HITLESS,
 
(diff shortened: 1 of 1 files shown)


rlhui:
Could we update the PR title to be more descriptive? Thanks.

tarakreddy1984:
Fail over mode is ability to switch the primary port to secondary port and vice versa without loosing data when any fail fault occur on primary port.   This feature make sure continuous data transmission.

inc/saiport.h
```diff
@@ -292,6 +292,57 @@ typedef enum _sai_port_prbs_config_t
...
+ * @brief Attribute data for #SAI_PORT_ATTR_FAIL_OVER_MODE
+ * Used for Fail over mode configuration on port
+ */
+typedef enum _sai_port_fail_over_mode_t
+{
+    /** Fail over mode disable */
+    SAI_PORT_FAIL_OVER_MODE_DISABLE,
+
+    /** Enable Fail over mode on the port */
+    SAI_PORT_FAIL_OVER_MODE_ENABLE,
```
rck-innovium, line 305: The enum value SAI_PORT_FAIL_OVER_MODE_ENABLE seems redundant. If fail-over is enabled on one port, then it has to be either primary or secondary.

lguohan:
this is fair fundamental change, @tarakreddy1984 ,  can you provide a design doc to describe the scenario here and also sample examples. you can refer to this pr. https://github.com/opencomputeproject/SAI/blob/master/doc/Hash/Policy-based-hash.md

lguohan:
as we also discuess, such switchover is better fitted into port connector model design, can you enhance the port connector model in order to introduce this function.

tarakreddy1984 (commented):
saiport.h updated

tarakreddy1984:
Will create new request

PR #1112, opened 2020-07-03
Fix max lane per port

stephenxs:
It assumes there are only 4 lanes per port, which is not correct because
there are 8 lanes per port on MSN4700 for 400G ports
Unnecessary assignment to thrift_attr_list.attr_count in function sai_thrift_get_port_attribute has been removed as well.

Signed-off-by: Stephen Sun <stephens@mellanox.com>

### DIFF
--- a/test/saithrift/src/switch_sai_rpc_server.cpp
+++ b/test/saithrift/src/switch_sai_rpc_server.cpp
@@ -3200,7 +3200,6 @@ public:
       queue_list_object_attribute.value.objlist.count = max_queues;
       port_api->get_port_attribute(port_id, 1, &queue_list_object_attribute);
 
-      thrift_attr_list.attr_count = 2;
       std::vector<sai_thrift_attribute_t>& attr_list = thrift_attr_list.attr_list;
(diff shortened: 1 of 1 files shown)


stephenxs:
Hi @lguohan 
could you please cherry-pick to v1.5 and v1.6? or do I need to create another PR for it?
thanks

lguohan:
done

stephenxs:
Thank you!

stephenxs:
Hi @lguohan,
I see the v1.5.2 is in SONiC 201911 but it includes a commit which is not included in v1.5
```
commit dda965adffdf27dea372dcc8790d811f04d66b2e
Author: Jai Kumar <jai.kumar@broadcom.com>
Date:   Thu Mar 5 18:30:03 2020 -0800

    adds nat type in sai_nat_entry_t to distinguish between dnat pool and dnat entry (#1042)
    
    This change adds nat type in sai_nat_entry_t to distinguish between dnat pool and dnat entry. Its possible that for dnat pool and dnat entry all the 
```
Can we cherry-pick the above commit into v1.5 and re-tag v1.5.2? After that, we can update sairedis and bring this PR into SONiC 201911.
Thank you and sorry for the inconvenience.

PR #1113, opened 2020-07-06
[saiport.h]Support max accumulative headroom size[saiport] Add SAI_PORT_QOS_ATTR_MAXIMUM_HEADROOM_SIZE

stephenxs:
Support max accumulative headroom size, which is a read-only attribute.
For example, if a port has 2 PGs and headroom size of each is N bytes, the accumulative headroom size should be 2*N.
The motivation is that:
- We're going to have a port's headroom size automatically calculated according to its cable length and speed. If a user configured a very long cable length the headroom size can be very large as well, which will exceed the limit of switch chip and thus failing the SDK and then the orchagent.
- To prevent that case, we need to know the limit in hardware and prevent a limit-exceeding value from programming to switch chip by SAI.
- This attribute stands for that limit.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -369,7 +369,8 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_QOS_SCHEDULER_GROUP_LIST,
 
     /**
-     * @brief The maximum accumulative headroom size of a port in bytes
+     * @brief The sum of the headroom size of the ingress priority groups belonging to this port
+     *        should not exceed the SAI_PORT_ATTR_QOS_MAXIMUM_HEADROOM_SIZE value.
      *
      * This attribute is applicable only for per-port, per-PG headroom model
      * (which means SAI_BUFFER_POOL_ATTR_XOFF_SIZE is zero)
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -371,10 +371,13 @@ typedef enum _sai_port_attr_t
     /**
      * @brief The maximum accumulative headroom size of a port in bytes
      *
+     * This attribute is applicable only for per-port, per-PG headroom model
+     * (which means SAI_BUFFER_POOL_ATTR_XOFF_SIZE is zero)
+     *
      * @type sai_uint32_t
      * @flags READ_ONLY
      */
-    SAI_PORT_ATTR_QOS_MAXIMUM_ACCUMULATIVE_HEADROOM_SIZE,
+    SAI_PORT_ATTR_QOS_MAXIMUM_HEADROOM_SIZE,
 
     /**
      * @brief Query list of supported port speed(full-duplex) in Mbps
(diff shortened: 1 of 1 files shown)


inc/saiport.h
```diff
@@ -368,6 +368,14 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_QOS_SCHEDULER_GROUP_LIST,
 
+    /**
+     * @brief The maximum accumulative headroom size of a port in bytes
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_QOS_MAXIMUM_ACCUMULATIVE_HEADROOM_SIZE,
```
rck-innovium, line 377: 1) As discussed, please add comments to say this is applicable only for per-port per-PG headroom model (i.e., SAI_BUFFER_POOL_ATTR_XOFF_SIZE = 0)

2) Consider renaming the attribute.
SAI_PORT_ATTR_BUFFER_MAX_HEADROOM_SIZE

stephenxs: Fixed.

rlhui:
@rck-innovium , @JaiOCP  - would you please help review/approve this PR? Thanks.

inc/saiport.h
```diff
@@ -368,6 +368,17 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_QOS_SCHEDULER_GROUP_LIST,
 
+    /**
+     * @brief The maximum accumulative headroom size of a port in bytes
```
rck-innovium, line 372: 1) This line should be reworded to make it easier to understand. Like:

```
The sum of the headroom size of the ingress priority groups belonging to this port should not exceed SAI_PORT_ATTR_QOS_MAXIMUM_HEADROOM_SIZE value.
```

inc/saiport.h
```diff
@@ -368,6 +368,17 @@ typedef enum _sai_port_attr_t
...
+     *
+     * This attribute is applicable only for per-port, per-PG headroom model
+     * (which means SAI_BUFFER_POOL_ATTR_XOFF_SIZE is zero)
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
```
rck-innovium, line 379: What should be the value returned by the platforms that do not have such a limit?

inc/saiport.h
stephenxs: ok. Fixed.

inc/saiport.h
stephenxs: They can just return 0.

rck-innovium: Thanks. Please update this in the comments as well so that the NOS/application handles the special case of 0.

stephenxs: Done.

stephenxs:
Hi @rlhui @JaiOCP ,
Could please you review/approve this PR?
Thanks.
Stephen

kcudnik (approved):

JaiOCP (approved):

stephenxs:
Hi @rlhui 
as all the reviewers approved the PR, can we merge it?
thanks

PR #1114, opened 2020-07-09
renaming decrement ttl enum to disable decrement ttl enum

srikrishnagopu:
In order to reflect the actual behavior, renaming SAI_NEXT_HOP_ATTR_DECREMENT_TTL to  SAI_NEXT_HOP_ATTR_DISABLE_DECREMENT_TTL.

SAI_*_ATTR_DISABLE_DECREMENT_TTL when set to true will disable decrementing ttl in the pipeline.

### DIFF
--- a/inc/sainexthop.h
+++ b/inc/sainexthop.h
@@ -227,7 +227,7 @@ typedef enum _sai_next_hop_attr_t
      * @flags CREATE_AND_SET
      * @default false
      */
-    SAI_NEXT_HOP_ATTR_DECREMENT_TTL,
+    SAI_NEXT_HOP_ATTR_DISABLE_DECREMENT_TTL,
 
     /**
      * @brief MPLS Outsegment type
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1466,7 +1466,7 @@ typedef enum _sai_port_attr_t
      * @flags CREATE_AND_SET
      * @default false
      */
-    SAI_PORT_ATTR_DECREMENT_TTL,
+    SAI_PORT_ATTR_DISABLE_DECREMENT_TTL,
 
     /**
      * @brief Enable EXP -> TC MAP on port
--- a/inc/sairouterinterface.h
+++ b/inc/sairouterinterface.h
@@ -283,7 +283,7 @@ typedef enum _sai_router_interface_attr_t
      * @flags CREATE_AND_SET
      * @default false
      */
-    SAI_ROUTER_INTERFACE_ATTR_DECREMENT_TTL,
+    SAI_ROUTER_INTERFACE_ATTR_DISABLE_DECREMENT_TTL,
 
     /**
      * @brief End of attributes


inc/sainexthop.h
```diff
@@ -227,7 +227,7 @@ typedef enum _sai_next_hop_attr_t
      * @flags CREATE_AND_SET
      * @default false
      */
-    SAI_NEXT_HOP_ATTR_DECREMENT_TTL,
+    SAI_NEXT_HOP_ATTR_DISABLE_DECREMENT_TTL,
```
atitjain, line 231: What is the issue here, why just changing the   `* @default false` to `true` helps resolve confusion.

srikrishnagopu: @atitjain I am not sure if I understood your question here. The original intention of this to attribute is to disable the ttl decrementing in the pipeline so that the packet loops in the dataplane and hence the attribute name change. 

By default, the attribute will be set to false and can be enabled when we run tests in the hardware.

atitjain: The original intention of this to attribute is to disable the ttl decrementing in the pipeline so that the packet loops in the dataplane and hence the attribute name change.

- Did not get the "packet loops in the data plane" part, you mean if all the hops in the network dont decrement TTL? What I mean is that it's just a convention and different people will find one or other more suitable. If I don't want to decrement TTL, its ok to say decrement_ttl = no/false which is same as disable_decrement=yes/true or even enable_decrement = no/false. So other than convention what changes both can convey the same.

srikrishnagopu: - Please read this for the packet loop part - https://github.com/opencomputeproject/SAI/pull/1050
- The attribute name does not clearly indicate that its trying to disable the ttl decrement as per the above PR and I prefer changing the attribute and setting it to false by default.

atitjain: - Thanks for the pointer and I understand that its one of the usecase.
- I am not sure why do you feel so. I don't see an issue with the existing name as well with the default value of false.

srikrishnagopu: For two reasons:
- As I mentioned, the attribute name is not clear based on the PR - https://github.com/opencomputeproject/SAI/pull/1050
- I think you meant DECREMENT_TTL to true as default ? Decrementing TTL is a default behavior in the pipeline. To override it, you can set DISABLE_DECREMENT_TTL to true.

salilpant-brcm: The current way of using decrement TTL attributes/ACL action is not very clear. Decrementing TTL is the default behavior even if the ACL action/NH attribute is not provided.  To disable decrementing TTL, we need to use TTL attribute with a value of false.  
Instead, its much clearer to change the name to indicate intent and use it with the value of true to indicate disabling needed. 
There is already a similarly named action in ACL - SAI_ACL_ENTRY_ATTR_ACTION_SET_DO_NOT_LEARN.

PR #1116, opened 2020-07-17
Update README.md

rajeevsharma1:
I have spoken with Xin about this change.


qiluo-msft:
There is no python code change in this PR, so LGTM check could be override. However I do not have permission to force merge.

kcudnik (approved):

PR #1118, opened 2020-07-29
Add DHCP L2 trap

itaibaz:
Currently SAI defines router DHCP trap :

     * @brief DHCP traffic (UDP ports 67, 68), either L3 broadcast or unicast
     * to local router IP address (default packet action is forward)
    SAI_HOSTIF_TRAP_TYPE_DHCP = 0x00002002,

There is a bootstrap problem with L3 trap in case of ZTP use case
To be able to receive an L3 trap, a RIF has to be set up (and that requires having IP address)
However, such IP address can only be obtained by receiving DHCP packets, which in turn is only possible currently when having a RIF set up

The proposed solution is introducing a L2 switch DHCP trap
Such trap can be received without a RIF
Therefor, DHCP OFFER and DHCP ACKNOWLEDGE can be received

### DIFF
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -232,9 +232,15 @@ typedef enum _sai_hostif_trap_type_t
 
     /**
      * @brief DHCP traffic (UDP ports 67, 68)
-     * (default packet action is drop)
+     * (default packet action is forward)
      */
     SAI_HOSTIF_TRAP_TYPE_DHCP_L2 = 0x00000012,
+
+    /**
+     * @brief DHCPV6 traffic (UDP ports 546, 547)
+     * (default packet action is forward)
+     */
+    SAI_HOSTIF_TRAP_TYPE_DHCPV6_L2 = 0x00000013,
 
     /** Switch traps custom range start */
     SAI_HOSTIF_TRAP_TYPE_SWITCH_CUSTOM_RANGE_BASE = 0x00001000,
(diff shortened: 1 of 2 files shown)


inc/saihostif.h
```diff
@@ -230,6 +230,12 @@ typedef enum _sai_hostif_trap_type_t
      */
     SAI_HOSTIF_TRAP_TYPE_PTP_TX_EVENT = 0x00000011,
 
+    /**
+     * @brief DHCP traffic (UDP ports 67, 68)
+     * (default packet action is drop)
```
rck-innovium, line 235: shouldn't the default be forward? The other switch traps are by default dropped since they are link local.

inc/saihostif.h
```diff
@@ -230,6 +230,12 @@ typedef enum _sai_hostif_trap_type_t
...
+     */
+    SAI_HOSTIF_TRAP_TYPE_DHCP_L2 = 0x00000012,
+
```
rck-innovium, line 238: Do we need something similar for DHCPv6

inc/saihostif.h
itaibaz: default changed to forward

inc/saihostif.h
itaibaz: Added DHCPv6_L2 trap

rlhui:
@rck-innovium , @JaiOCP  - would you approve? Thanks.

JaiOCP (approved):

PR #1122, opened 2020-08-12
UDF Field Qualifier in ACL

JaiOCP:
This PR brings in change to be able to specify field extracted by UDF group as a qualifier in ACL Table.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1456,6 +1456,16 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_PRBS_LOCK_LOSS_STATUS,
 
     /**
+     * @brief Attribute data for #SAI_PORT_ATTR_PRBS_RX_STATUS
+     *
+     * PRBS rx status: 1 for OK, 0 for error
+     *
+     * @type bool
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_PRBS_RX_STATUS,
+
+    /**
      * @brief Attribute data for #SAI_PORT_ATTR_AUTO_NEG_STATUS
      *
      * Auto negotiation (AN) done state: 0 for AN in progress, 0 for AN done
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1456,16 +1456,6 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_PRBS_LOCK_LOSS_STATUS,
 
     /**
-     * @brief Attribute data for #SAI_PORT_ATTR_PRBS_RX_STATUS
-     *
-     * PRBS rx status: 1 for OK, 0 for error
-     *
-     * @type bool
-     * @flags READ_ONLY
-     */
-    SAI_PORT_ATTR_PRBS_RX_STATUS,
-
-    /**
      * @brief Attribute data for #SAI_PORT_ATTR_AUTO_NEG_STATUS
      *
      * Auto negotiation (AN) done state: 0 for AN in progress, 0 for AN done
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -219,17 +219,11 @@ typedef enum _sai_port_interface_type_t
     /** Interface type CR */
     SAI_PORT_INTERFACE_TYPE_CR,
 
-    /** Interface type CR2 */
-    SAI_PORT_INTERFACE_TYPE_CR2,
-
     /** Interface type CR4 */
     SAI_PORT_INTERFACE_TYPE_CR4,
 
     /** Interface type SR */
     SAI_PORT_INTERFACE_TYPE_SR,
-
-    /** Interface type SR2 */
-    SAI_PORT_INTERFACE_TYPE_SR2,
 
     /** Interface type SR4 */
     SAI_PORT_INTERFACE_TYPE_SR4,
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -219,11 +219,17 @@ typedef enum _sai_port_interface_type_t
     /** Interface type CR */
     SAI_PORT_INTERFACE_TYPE_CR,
 
+    /** Interface type CR2 */
+    SAI_PORT_INTERFACE_TYPE_CR2,
+
     /** Interface type CR4 */
     SAI_PORT_INTERFACE_TYPE_CR4,
 
     /** Interface type SR */
     SAI_PORT_INTERFACE_TYPE_SR,
+
+    /** Interface type SR2 */
+    SAI_PORT_INTERFACE_TYPE_SR2,
 
     /** Interface type SR4 */
     SAI_PORT_INTERFACE_TYPE_SR4,
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -3960,7 +3960,9 @@ void check_acl_table_fields_and_acl_entry_fields()
          * check acl table flags and attr value type
          */
 
-        if (mtable->attrid == SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE)
+        if ((mtable->attrid == SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE) ||
+        ((mtable->attrid >= SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MIN) &&
+        (mtable->attrid <= SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MAX)))
         {
             /*
              * This field is exception, it's not bool, it's a list and it's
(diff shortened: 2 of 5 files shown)


JaiOCP:
> As discussed, please consider the approach followed by saidebugcounter.h
> Specifically, the attribute:
> 
> ```
> /* READ-ONLY */
> 
> /**
>  * @brief Object stat index
>  * Index is added to base start
>  *
>  * @type sai_uint32_t
>  * @flags READ_ONLY
>  */
> SAI_DEBUG_COUNTER_ATTR_INDEX
> ```
> 
> This solves backward compatibility issues as well as doesn't waste hardware resources.
> 
> The steps should be:
> 
> 1. Create one or more UDF groups.
>    Read back the SAI_UDF_GROUP_ATTR_INDEX
>    SDK driver ensures that it returns an unique index for each UDF group.
> 
> 1. Let us say that the UDF groups of our interest have  SAI_UDF_GROUP_ATTR_INDEX values as i and j.
> 2. create ACL table with attributes:
> 
> ```
> ...
> SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MIN+i;
>     acl_attr_list[index++].value.booldata = True;
> SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MIN+j;
>     acl_attr_list[index++].value.booldata = True;
> 
> ...
> ```

This is a good idea. I will incorporate this change. Only side effect of this is that there is a state maintained in SAI driver and need to be maintained and carried across during warm boot or ISSU

mickeyspiegel:
This can be simplified considerably by changing SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MIN through SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MAX to use @type sai_object_id_t with @objects SAI_OBJECT_TYPE_UDF_GROUP.

At the previous SAI meeting, we saw that not all SAI_ACL_TABLE_ATTR_FIELD_* attributes use @type bool.
Specifically, SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE uses @type sai_s32_list_t sai_acl_range_type_t.
While there is no current example of a SAI_ACL_TABLE_ATTR_FIELD_* attribute using @type sai_object_id_t, IMO the precedent to use something other than @type bool implies that this should work without causing too many problems.

If you make this change, then the addition of SAI_UDF_GROUP_ATTR_INDEX is no longer necessary.
The goal of being able to determine the ACL key at table definition time can be achieved by leveraging the rest of this proposal.

JaiOCP:
> This can be simplified considerably by changing SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MIN through SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MAX to use @type sai_object_id_t with @objects SAI_OBJECT_TYPE_UDF_GROUP.
> 
> At the previous SAI meeting, we saw that not all SAI_ACL_TABLE_ATTR_FIELD_* attributes use @type bool.
> Specifically, SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE uses @type sai_s32_list_t sai_acl_range_type_t.
> While there is no current example of a SAI_ACL_TABLE_ATTR_FIELD_* attribute using @type sai_object_id_t, IMO the precedent to use something other than @type bool implies that this should work without causing too many problems.
> 
> If you make this change, then the addition of SAI_UDF_GROUP_ATTR_INDEX is no longer necessary.
> The goal of being able to determine the ACL key at table definition time can be achieved by leveraging the rest of this proposal.

Hi Mickey,

Looks like the match fields in ACL table has to be a bool datatype. I tried the change and get following error. 
Other attributes we were looking are not the match fields (between SAI_ACL_TABLE_ATTR_FIELD_START and SAI_ACL_TABLE_ATTR_FIELD_END) that is why they can be of other datatypes. This also means we can not change the match fields to uint8 as we were discussing. They have to be bool.

Let me know if you can think of any other way to handle this. 
...
 ASSERT FAILED (on line 3980): SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MIN: acl table attr value type should be bool
...

mickeyspiegel:
@JaiOCP I am not sure how to handle this, but I wonder how SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE got around this.
It is clearly in between SAI_ACL_TABLE_ATTR_FIELD_START and SAI_ACL_TABLE_ATTR_FIELD_END.

JaiOCP:
> @JaiOCP I am not sure how to handle this, but I wonder how SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE got around this.
> It is clearly in between SAI_ACL_TABLE_ATTR_FIELD_START and SAI_ACL_TABLE_ATTR_FIELD_END.

There is an exception in saisanitycheck.c for this
....
       if (mtable->attrid == SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE)
        {
            /*
             * This field is exception, it's not bool, it's a list and it's
             * designed in this way to save resources on device to not support
             * all ranges on each acl table when it's not necessary.
             */
        }
....

If others agree we can create an exception for UDF group as well.
@lguohan @kcudnik Please suggest

rck-innovium (approved):
The below pending items are minor:
- Remove accidental changes in saiport.h
- Update the API sequence based on using UDF Group sai-object-id when creating the ACL Table .

inc/saiport.h
```diff
@@ -219,12 +219,18 @@ typedef enum _sai_port_interface_type_t
...
+    SAI_PORT_INTERFACE_TYPE_CR2,
+
     /** Interface type CR4 */
     SAI_PORT_INTERFACE_TYPE_CR4,
 
     /** Interface type SR */
     SAI_PORT_INTERFACE_TYPE_SR,
 
+    /** Interface type SR2 */
+    SAI_PORT_INTERFACE_TYPE_SR2,
```
lguohan, line 232: can you remove this from the pr, it is not related to udf.

inc/saiport.h
```diff
@@ -219,12 +219,18 @@ typedef enum _sai_port_interface_type_t
     /** Interface type CR */
     SAI_PORT_INTERFACE_TYPE_CR,
 
+    /** Interface type CR2 */
+    SAI_PORT_INTERFACE_TYPE_CR2,
```
lguohan, line 223: can you remove this from the pr, it is not related to udf.

inc/saiport.h
JaiOCP: There was an issue with signoff. Fixed it

inc/saiport.h
JaiOCP: There was an issue with signoff. Fixed it

rlhui:
@marian-pritsak , would you please help review? Thanks.

meta/saisanitycheck.c
```diff
@@ -3960,7 +3960,9 @@ void check_acl_table_fields_and_acl_entry_fields()
          * check acl table flags and attr value type
          */
 
-        if (mtable->attrid == SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE)
+        if ((mtable->attrid == SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE) ||
+        ((mtable->attrid >= SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MIN) &&
```
kcudnik, line 3964: why this condition is added? seems like not neeed, remove ?

JaiOCP: Attributes between SAI_ACL_TABLE_ATTR_FIELD_START and SAI_ACL_TABLE_ATTR_FIELD_END are expected to be bool.

RANGE_TYPE and USER_DEFINED_FIELD_GROUP are exception to this rule as the length can not be derived from these attributes directly.

......
Error message: If toolchain is not modified for exception

 ASSERT FAILED (on line 3980): SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MIN: acl table attr value type should be bool

JaiOCP:
@mickeyspiegel @lguohan @rlhui Please approve the PR

meta/saisanitycheck.c
kcudnik: why then those attributes are changed from bool to oid ?

JaiOCP: Kamil, more details are in the spec. There is no way to get length of the UDF field just by the bool, it is stored in the UDF group OID hence the change.

lguohan: @kcudnik, is it ok?

lguohan (approved):
pending on kamil's comments on the metadata

kcudnik (approved):

meta/saisanitycheck.c
JaiOCP: Thanks

JaiOCP:
@rlhui Rita, Can you please merge this PR as well

marian-pritsak (approved):

PR #1124, opened 2020-08-20
PRBS Rx Status

JaiOCP:
New attribute indicating PRBS Rx status is introduced

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -296,6 +296,25 @@ typedef enum _sai_port_link_training_rx_status_t
 } sai_port_link_training_rx_status_t;
 
 /**
+ * @brief Attribute data for #SAI_PORT_ATTR_PRBS_RX_STATUS
+ */
+typedef enum _sai_port_prbs_rx_status_t
+{
+    /** PRBS is locked and error_count is 0 */
+    SAI_PORT_PRBS_RX_STATUS_OK,
+
+    /** PRBS is locked, but there are errors */
+    SAI_PORT_PRBS_RX_STATUS_LOCK_WITH_ERRORS,
+
+    /** PRBS not locked */
+    SAI_PORT_PRBS_RX_STATUS_NOT_LOCKED,
+
+    /** PRBS locked but there is loss of lock since last call */
+    SAI_PORT_PRBS_RX_STATUS_LOST_LOCK,
+
+} sai_port_prbs_rx_status_t;
+
+/**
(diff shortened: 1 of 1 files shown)


inc/saiport.h
```diff
@@ -1481,6 +1481,16 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_PRBS_LOCK_LOSS_STATUS,
 
+    /**
+     * @brief Attribute data for #SAI_PORT_ATTR_PRBS_RX_STATUS
+     *
+     * PRBS rx status: 1 for OK, 0 for error
```
Ashish1805, line 1487: One question:
What does OK and error mean here? Does it give any hint about lock status (SAI_PORT_ATTR_PRBS_LOCK_STATUS), lock lost since last read (SAI_PORT_ATTR_PRBS_LOCK_LOSS_STATUS) or PRBS error count (SAI_PORT_STAT_PRBS_ERROR_COUNT)? Are this independent to other 3?

Thank you.

JaiOCP: Added enumeration for prbs rx status.


+typedef enum _sai_port_prbs_rx_status_t
+{
+    /** PRBS is locked and error_count is 0 */
+    SAI_PORT_PRBS_RX_STATUS_OK,
+
+    /** PRBS is locked, but there are errors */
+    SAI_PORT_PRBS_RX_STATUS_LOCK_WITH_ERRORS,
+
+    /** PRBS not locked */
+    SAI_PORT_PRBS_RX_STATUS_NOT_LOCKED,
+
+    /** PRBS locked but there is loss of lock since last call */
+    SAI_PORT_PRBS_RX_STATUS_LOST_LOCK,
+
+} sai_port_prbs_rx_status_t;

Ashish1805: Thank you.

So, if we get rx status SAI_PORT_PRBS_RX_STATUS_LOCK_WITH_ERRORS, is expectation here to use SAI_PORT_STAT_PRBS_ERROR_COUNT to get the PRBS error count? Thank you.

JaiOCP: yes, RX status is the aggregate status and based on the return value, NOS can query following attributes for information
- SAI_PORT_ATTR_PRBS_LOCK_LOSS_STATUS
- SAI_PORT_ATTR_PRBS_LOCK_STATUS
- SAI_PORT_STAT_PRBS_ERROR_COUNT

kcudnik (approved):

PR #1137, opened 2020-10-15
Add packet allocate, free function to allow for 0 copy packet TX

jasmeetbagga:
Sending a packet requires allocating it first. Typically,
this requires allocating from a special pool of DMAed memory.
For this a ASIC vendor SDK may support custom allocate and
free functions. So mirror those in the SAI spec.

Without such a ability we are left to allocate heap memory,
pass it to send_hostif_packet_fn, which then allocates from the
afore mentioned DMA memory, copies the contents of heap memory
and sends the packet along. This causes us to have a extra
allocation and copy. This can induce a substantial slowdown -
~30% in our measurements


kcudnik:
please resolve conflicts

kcudnik (approved):

rlhui:
@mattyk , @JaiOCP - please help review/approve, thanks.

jasmeetbagga:
@rlhui, @kcudnik approved already. Do we need more approvals before this can be merged in? What are the next steps? Thanks

rlhui:
> @rlhui, @kcudnik approved already. Do we need more approvals before this can be merged in? What are the next steps? Thank

@jasmeetbagga  - we'd need more approvals as well.

kcudnik:
i just approve from code perspective, whether this is good to go should be probably decided on weekly SAI meeting

lguohan:
the pr looks good to me. suggestion debuggability can be another pr, not necessarily to be couple with this pr. the point is that the debuggability improvement should not change the signature of current dma allocation api. @JaiOCP , can you have a quick sync. we plan to include this pr in the v1.7.0 release.

gechiang:
During the review meeting we discussed about if the Tx DMA can be improved in a future PR to allow caller to specify "priority".  I believe BRCM suggested that it could be achieved by pre-allocate some DMA buffer for those high priority senders so that it will not need to compete the buffer allocation with the low priority senders.  But I feel that having different Tx priority Queues will still benefit the case where let's say there are multiple low priority senders running in a multi-threaded environment where each one wants to send a lengthy packet.  If we do have different priority queue then SAI could let the high priority sender to "cut to the front of the line" and proceed to Tx that one first before other lower priority send requests.  Again,. this comment is not meant to ask this PR to implement it but just to solicit further comment over this idea that we briefly discussed over the meeting.
Thanks!

lguohan:
@gechiang , thanks for the comment. maybe it is better to create an issue for the topic you discussed above.

JaiOCP (approved):
Aproving with a followup action.
Please open a followup PR for debuggability.
Following things need to be considered
- Total DMA Memory Available: Either using CRM infra or a local read-only attribute

New Object with get API. Some of the attributes I can think of are
- free memory
- Client IDs consuming memory
- Alloc fail count

rlhui:
@marian-pritsak - reminder on this.

jasmeetbagga:
@JaiOCP #1155 created as followup

PR #1142, opened 2020-10-31
PFC RX/TX pause duration in micro seconds

itaibaz:
Add PFC RX/TX pause duration in micro seconds, in addition to existing quanta units


lguohan (approved):

kcudnik:
why not have original ones in microseconds at first place ?

lguohan:
probably should

PR #1148, opened 2020-10-31
Query statistics capability

itaibaz:
Add ability to query statistics values capability per object type - meaning which statistics enum values are implemented per object type
Currently, these is not exposed, and in order to check capabilities, the user has to retry each value, one by one

Signed-off-by: Itai Baz <itaib@mellanox.com>

### DIFF
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -237,13 +237,15 @@ sai_status_t sai_query_attribute_enum_values_capability(
  * @param[in] switch_id SAI Switch object id
  * @param[in] object_type SAI object type
  * @param[inout] stats_values_capability List of implemented enum values
+ * @param[inout] stats_modes_capability List of implemented sai_stats_mode_t values
  *
- * @return #SAI_STATUS_SUCCESS on success, #SAI_STATUS_BUFFER_OVERFLOW if list size insufficient, failure status code on error
+ * @return #SAI_STATUS_SUCCESS on success, #SAI_STATUS_BUFFER_OVERFLOW if lists size insufficient, failure status code on error
  */
 sai_status_t sai_query_stats_capability(
         _In_ sai_object_id_t switch_id,
         _In_ sai_object_type_t object_type,
-        _Inout_ sai_s32_list_t *stats_values_capability);
+        _Inout_ sai_s32_list_t *stats_values_capability,
+        _Inout_ sai_s32_list_t *stats_modes_capability);
 
 /**
  * @}
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -236,16 +236,14 @@ sai_status_t sai_query_attribute_enum_values_capability(
  *
  * @param[in] switch_id SAI Switch object id
  * @param[in] object_type SAI object type
- * @param[inout] stats_values_capability List of implemented enum values
- * @param[inout] stats_modes_capability List of implemented sai_stats_mode_t values
+ * @param[inout] stats_capability List of implemented enum values, and the statistics modes (bit mask) supported per value
  *
  * @return #SAI_STATUS_SUCCESS on success, #SAI_STATUS_BUFFER_OVERFLOW if lists size insufficient, failure status code on error
  */
 sai_status_t sai_query_stats_capability(
         _In_ sai_object_id_t switch_id,
         _In_ sai_object_type_t object_type,
-        _Inout_ sai_s32_list_t *stats_values_capability,
-        _Inout_ sai_s32_list_t *stats_modes_capability);
+        _Inout_ sai_stat_capability_list_t *stats_capability);
 
 /**
  * @}
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -232,6 +232,20 @@ sai_status_t sai_query_attribute_enum_values_capability(
         _Inout_ sai_s32_list_t *enum_values_capability);
 
 /**
+ * @brief Query statistics capability
+ *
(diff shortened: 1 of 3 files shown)


kcudnik (approved):

inc/saiobject.h
```diff
@@ -231,6 +231,20 @@ sai_status_t sai_query_attribute_enum_values_capability(
...
+ * @param[in] switch_id SAI Switch object id
+ * @param[in] object_type SAI object type
+ * @param[inout] stats_values_capability List of implemented enum values
+ *
+ * @return #SAI_STATUS_SUCCESS on success, #SAI_STATUS_BUFFER_OVERFLOW if list size insufficient, failure status code on error
+ */
+sai_status_t sai_query_stats_capability(
+        _In_ sai_object_id_t switch_id,
+        _In_ sai_object_type_t object_type,
+        _Inout_ sai_s32_list_t *stats_values_capability);
```
lguohan, line 246: do we know if the stats support read and clear mode or read mode?

kcudnik: thats good point, but anyway if vendor dont support read_and_clear in atomic way is there any approach that would eliminate potential read/clrear error ? also if read_and_clear mode is not supported then stats api should return SAI_STATUS_NOT_SUPPORTED, but to have that to know before api call, we would probably need to have another fieldarray param about supported modes

itaibaz: I added additional param, as requested, to query list of implemented stat modes (read, read and clear, and future extensions)
Please recheck

rlhui:
@JaiOCP please help review/approve

lguohan (approved):

itaibaz (commented):
@lguohan @kcudnik 
Can you please review the changes done based on your feedback in today's OCP meeting

lguohan:
lgtm

inc/saitypes.h
```diff
@@ -1355,19 +1355,49 @@ typedef sai_status_t (*sai_bulk_object_get_attribute_fn)(
...
      * @brief Read and clear after reading
      */
-    SAI_STATS_MODE_READ_AND_CLEAR,
+    SAI_STATS_MODE_READ_AND_CLEAR = 1 << 1,
 } sai_stats_mode_t;
 
+typedef struct _sai_stat_capability_t
+{
+    /** Stat enum value */
+    int32_t stat_enum;
```
kcudnik, line 1382: sai_stat_id_t ? its used in all stats api

kcudnik (approved):

kcudnik:
WARNING: found trailing spaces in serialize.pm:255:     }
thanks for serialization fix, i just started to work on that :P

kcudnik (approved):

inc/saitypes.h
itaibaz: done
yes, sai_stat_id_t used in all get stats / get extended stats / clear stats
so better to use it here, thanks

JaiOCP (changes requested):
Reduce stat_modes to 32bits given that there are only 2 modes and we do not expect to expand to a large value. 64bit will waste memory for large query sets

inc/saitypes.h
```diff
@@ -1355,19 +1355,49 @@ typedef sai_status_t (*sai_bulk_object_get_attribute_fn)(
...
+    /** Stat enum value */
+    sai_stat_id_t stat_enum;
+
+    /**
+     * @brief Bit mask of supported statistics modes (sai_stats_mode_t)
+     *
+     * For example, if read and read_and_clear are supported, value is
+     * SAI_STATS_MODE_READ | SAI_STATS_MODE_READ_AND_CLEAR
+     */
+    uint64_t stat_modes;
```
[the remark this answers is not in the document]
itaibaz: Right now the biggest statistics enum is ~150 for port
So we are talking about allocation of 150 * (4 + 8) = ~1.8KB with uint64, vs 150 * (4 + 4) = ~1.2KB with uint32
While I agree we probably won't have more than few (maybe 5?) modes, since perhaps there is a scenario we don't anticipate now, and trying to give maximum flexibility for future growth, while paying relatively small price (extra 0.5 KB allocation), I think 64 bits are preferable. Also, in terms of usage, uint32 and uint64 are the same, as both atomic types so it doesn't complex things

inc/saiobject.h
```diff
@@ -231,6 +231,20 @@ sai_status_t sai_query_attribute_enum_values_capability(
         _In_ sai_attr_id_t attr_id,
         _Inout_ sai_s32_list_t *enum_values_capability);
 
+/**
+ * @brief Query statistics capability
```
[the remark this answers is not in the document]
itaibaz: done

inc/saitypes.h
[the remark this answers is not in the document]
lguohan: considering the alignment, since sai_stat_id_t is 32 bits, if we can make stat_modes 32 bits, then the structure will be packed into 64-bit structure. if stat_mode is 64 bits, then it could be 128 bits depending on the structure alignment.

wonder if we can use 32-bit for now and extend to 64-bit if needed in the future. that is still backward compatible?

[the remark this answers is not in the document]
itaibaz: the function signature won't change if we modify in the future from uint32 to uint64. some code checking the modes capability may change, or may not. in any case, this is very minor change, and since it seems for the moment uint32 is prefered, i changed to using it

JaiOCP (approved):
Looks Good

abdosi (approved):

PR #1152, opened 2020-11-04
Packet loopback action for tunnels

JaiOCP:
Added loopback action as switch attribute: This will apply to all the tunnels configured
Added loopback action as tunnel attribute: This will apply as a tunnel specific action.

### DIFF
--- a/inc/saitunnel.h
+++ b/inc/saitunnel.h
@@ -644,6 +644,15 @@ typedef enum _sai_tunnel_attr_t
     SAI_TUNNEL_ATTR_TERM_TABLE_ENTRY_LIST,
 
     /**
+     * @brief Packet action when a packet ingress and gets routed back to same tunnel
+     *
+     * @type sai_packet_action_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PACKET_ACTION_FORWARD
+     */
+    SAI_TUNNEL_ATTR_LOOPBACK_PACKET_ACTION,
+
+    /**
      * @brief End of attributes
      */
     SAI_TUNNEL_ATTR_END,
--- a/inc/saidebugcounter.h
+++ b/inc/saidebugcounter.h
@@ -347,6 +347,13 @@ typedef enum _sai_out_drop_reason_t
      */
     SAI_OUT_DROP_REASON_L3_EGRESS_LINK_DOWN,
 
+    /* Tunnel reasons */
+
+    /**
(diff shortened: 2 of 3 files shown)


kcudnik (approved):

inc/saitunnel.h
```diff
@@ -643,6 +643,15 @@ typedef enum _sai_tunnel_attr_t
      */
     SAI_TUNNEL_ATTR_TERM_TABLE_ENTRY_LIST,
 
+    /**
+     * @brief Packet action when a packet ingress and gets routed back to same tunnel
+     *
+     * @type sai_packet_action_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PACKET_ACTION_FORWARD
```
prsunny, line 651: should the default be SAI_PACKET_ACTION_DROP?

JaiOCP: Most of the usecases I am familiar with are forward for eg MPLS, IPinIP and GRE with L3 payload.

rlhui:
What if the next-hop after decap is another tunnel, but not this same tunnel where pkt is received from?  Say if the next-hop for the dst is an ecmp?

rlhui:
Can we increment a drop reason counter when it's dropped due to this?

JaiOCP:
> What if the next-hop after decap is another tunnel, but not this same tunnel where pkt is received from? Say if the next-hop for the dst is an ecmp?

If packet is forwarded to some other tunnel instead of the ingress tunnel then this attribute will not be applied. Only when packet loops back into the ingress tunnel then this attribute is applicable.

JaiOCP:
> Can we increment a drop reason counter when it's dropped due to this?

yes, I am thinking to add switch level attribute: SAI_SWITCH_STAT_TUNNEL_LOOPBACK_PACKET_DROP
@itaibaz Is this OK?

itaibaz:
Hi, let me check internally and update in several days, thanks

JaiOCP:
> We prefer using
> saidebugcounter.h:sai_out_drop_reason_t:SAI_OUT_DROP_REASON_TUNNEL_LOOPBACK_PACKET_DROP
> As this is already existing mechanism used to debug drops, and in specific tunnel drops, in sonic

Done

JaiOCP:
@rlhui @itaibaz @prsunny please review and approve

itaibaz (approved):

rlhui (approved):

PR #1153, opened 2020-11-04
Switch attributes for tunnel ECN mode

JaiOCP:
Add switch attributes for switch level configuration of tunnel ECN modes.
Encap and Decap ecn modes can be configured individually with separately assigned ECN mappers for user defined ECN mode.

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -328,43 +328,46 @@ typedef enum _sai_switch_failover_config_mode_t
 /**
  * @brief Defines tunnel encap ECN mode
  */
-typedef enum _sai_switch_tunnel_encap_ecn_mode_t
+typedef enum _sai_tunnel_encap_ecn_mode_t
 {
     /**
      * @brief Normal mode behavior defined in RFC 6040
      * section 4.1 copy from inner
      */
-    SAI_SWITCH_TUNNEL_ENCAP_ECN_MODE_STANDARD,
+    SAI_TUNNEL_ENCAP_ECN_MODE_STANDARD,
 
     /**
-     * @brief User defined behavior
+     * @brief User defined behavior.
+     *
+     * Need to provide #SAI_TUNNEL_MAP_TYPE_OECN_TO_UECN in #SAI_TUNNEL_ATTR_ENCAP_MAPPERS.
      */
-    SAI_SWITCH_TUNNEL_ENCAP_ECN_MODE_USER_DEFINED
+    SAI_TUNNEL_ENCAP_ECN_MODE_USER_DEFINED
 
-} sai_switch_tunnel_encap_ecn_mode_t;
+} sai_tunnel_encap_ecn_mode_t;
 
 /**
  * @brief Defines tunnel decap ECN mode
  */
-typedef enum _sai_switch_tunnel_decap_ecn_mode_t
+typedef enum _sai_tunnel_decap_ecn_mode_t
 {
     /**
-     * @brief Normal mode behavior defined in RFC 6040
-     * section 4.1 copy from inner
+     * @brief Behavior defined in RFC 6040 section 4.2
      */
-    SAI_SWITCH_TUNNEL_DECAP_ECN_MODE_STANDARD,
+    SAI_TUNNEL_DECAP_ECN_MODE_STANDARD,
 
     /**
      * @brief Copy from outer ECN
      */
-    SAI_SWITCH_TUNNEL_DECAP_ECN_MODE_COPY_FROM_OUTER,
+    SAI_TUNNEL_DECAP_ECN_MODE_COPY_FROM_OUTER,
 
     /**
      * @brief User defined behavior
+     *
+     * Need to provide #SAI_TUNNEL_MAP_TYPE_UECN_OECN_TO_OECN in #SAI_TUNNEL_ATTR_DECAP_MAPPERS
      */
-    SAI_SWITCH_TUNNEL_DECAP_ECN_MODE_USER_DEFINED
+    SAI_TUNNEL_DECAP_ECN_MODE_USER_DEFINED
 
-} sai_switch_tunnel_decap_ecn_mode_t;
+} sai_tunnel_decap_ecn_mode_t;
 
 /**
  * @brief Attribute Id in sai_set_switch_attribute() and
@@ -2310,9 +2313,9 @@ typedef enum _sai_switch_attr_t
     /**
      * @brief Tunnel encap ECN Mode load method
      *
-     * @type sai_switch_tunnel_encap_ecn_mode_t
+     * @type sai_tunnel_encap_ecn_mode_t
      * @flags CREATE_AND_SET
-     * @default SAI_SWITCH_TUNNEL_ENCAP_ECN_MODE_STANDARD
+     * @default SAI_TUNNEL_ENCAP_ECN_MODE_STANDARD
      */
     SAI_SWITCH_ATTR_TUNNEL_ENCAP_ECN_MODE,
 
@@ -2329,9 +2332,9 @@ typedef enum _sai_switch_attr_t
     /**
      * @brief Tunnel decap ECN Mode load method
      *
-     * @type sai_switch_tunnel_decap_ecn_mode_t
+     * @type sai_tunnel_decap_ecn_mode_t
      * @flags CREATE_AND_SET
-     * @default SAI_SWITCH_TUNNEL_DECAP_ECN_MODE_STANDARD
+     * @default SAI_TUNNEL_DECAP_ECN_MODE_STANDARD
      */
     SAI_SWITCH_ATTR_TUNNEL_DECAP_ECN_MODE,
 
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -326,6 +326,29 @@ typedef enum _sai_switch_failover_config_mode_t
 } sai_switch_failover_config_mode_t;
 
 /**
+ * @brief Defines tunnel ECN mode
+ */
+typedef enum _sai_switch_tunnel_ecn_mode_t
+{
+    /**
+     * @brief Normal mode behavior defined in RFC 6040
+     * section 4.1 copy from inner
+     */
+    SAI_SWITCH_TUNNEL_ECN_MODE_STANDARD,
+
+    /**
+     * @brief Copy from outer ECN
+     */
+    SAI_SWITCH_TUNNEL_ECN_MODE_COPY_FROM_OUTER,
+
+    /**
+     * @brief User defined behavior
+     */
+    SAI_SWITCH_TUNNEL_ECN_MODE_USER_DEFINED
+
+} sai_switch_tunnel_ecn_mode_t;
+
+/**
  * @brief Attribute Id in sai_set_switch_attribute() and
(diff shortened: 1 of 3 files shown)


inc/saiswitch.h
```diff
@@ -325,6 +325,47 @@ typedef enum _sai_switch_failover_config_mode_t
...
+     * section 4.1 copy from inner
+     */
+    SAI_SWITCH_TUNNEL_ENCAP_ECN_MODE_STANDARD,
+
+    /**
+     * @brief User defined behavior
+     */
+    SAI_SWITCH_TUNNEL_ENCAP_ECN_MODE_USER_DEFINED
+
+} sai_switch_tunnel_encap_ecn_mode_t;
```
itaibaz, line 344: I suggest to move
sai_tunnel_encap_ecn_mode_t and _sai_tunnel_decap_ecn_mode_t
from saitunnel.h to saiswitch.h
and use a single enum (similar to sai_packet_action_t, sai_packet_vlan_t, which are defined in saiswitch.h and also used in other header files)

inc/saiswitch.h
```diff
@@ -2256,6 +2297,44 @@ typedef enum _sai_switch_attr_t
...
 
+    /**
+     * @brief Tunnel mappers for encap
+     *
+     * @type sai_object_list_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_TUNNEL_MAP
+     * @default empty
+     */
+    SAI_SWITCH_ATTR_TUNNEL_ENCAP_ECN_MAPPERS,
```
itaibaz, line 2308: The new API assumes one global configuration for all tunnel types. Meaning, if we have IP in IP tunnel and VXLAN tunnel, and we set this attribute, they will both use the same map.
Some ASICs have capability to set it per tunnel type. Meaning, having SAI_SWITCH_ATTR_TUNNEL_IPINIP_ENCAP_ECN_MAPPERS, SAI_SWITCH_ATTR_TUNNEL_VXLAN_ENCAP_ECN_MAPPERS,
SAI_SWITCH_ATTR_TUNNEL_MPLS_ECN_MAPPERS (similar to hash API where we have global configuration per traffic type SAI_SWITCH_ATTR_ECMP_HASH_IPV4, SAI_SWITCH_ATTR_ECMP_HASH_IPV4_IN_IPV4, etc.)

inc/saiswitch.h
```diff
@@ -2256,6 +2297,44 @@ typedef enum _sai_switch_attr_t
...
+    SAI_SWITCH_ATTR_TUNNEL_DECAP_ECN_MAPPERS,
+
+    /**
+     * @brief Tunnel decap ECN Mode load method
+     *
+     * @type sai_switch_tunnel_decap_ecn_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_SWITCH_TUNNEL_DECAP_ECN_MODE_STANDARD
+     */
+    SAI_SWITCH_ATTR_TUNNEL_DECAP_ECN_MODE,
```
itaibaz, line 2336: if we believe switch level configuration is needed for tunnel QoS, why do we expose it only for ECN ? and not for DSCP, TTL, etc. ?
seems incomplete model

inc/saiswitch.h
JaiOCP: Won't it make it backward incompatible change? We can add comment for deprecation and plan it move it out in net major release.

inc/saiswitch.h
JaiOCP: Thats correct. Idea is to use global switch level configuration if there is a uniform ecn mode for given tunnel types for encap or decap. 
This can always be overridden by the tunnel specific ECN mapper.

inc/saiswitch.h
JaiOCP: Current requirement is to enhance it for ECN mode only but if we all agree then yes other attributes like TTL/DSCP are good candidates as well.

inc/saiswitch.h
itaibaz: my suggestion is to move the enum as is, so it's backward compatible
not for example packet action, which is used in multiple headers -

    /** This is a combination of SAI packet action COPY_CANCEL and FORWARD */
    SAI_PACKET_ACTION_TRANSIT

} sai_packet_action_t;

the enum doesn't have to start with SAI_SWITCH, can be exactly as it is today

inc/saiswitch.h
itaibaz: Still we don't allow global per tunnel type
There can be several ways to configure :
1) Global per all tunnels
2) Global per tunnel type (meaning, all IP in IP tunnels, all VXLAN tunnels, etc.)
3) Per specific tunnel (what we have today)
This PR is adding #1, but not adding #2 which is more flexible. I suggest adding #2 instead of #1

inc/saiswitch.h
JaiOCP: I see what you are saying. Yes it makes sense. Committed the suggestion.

https://github.com/opencomputeproject/SAI/pull/1153/commits/f09a8fd1a14890e9d7ae126b47d171c8091fc2d9

inc/saiswitch.h
JaiOCP: Lets discuss this. Currently this PR is supporting (1)

inc/saiswitch.h
JaiOCP: As discussed in community meeting, current requirement is specific to ECN (silicon vendor limitations). Other attributes if needed will be addressed as a cleaner encapsulate approach.

inc/saiswitch.h
gechiang: I think having per tunnel type level of default setting as suggested by @itaibaz is good which gives more granularity option to the application user.  Perhaps we can still have an attribute that covers for ALL tunnel types as well such as SAI_SWITCH_ATTR_ALL_TUNNEL_ECN_MAPPERS that sets the default at switch level when there are no specific per Tunnel type nor at the per tunnel level.

JaiOCP: Thanks @gechiang @itaibaz  for comments.
Looks like ECN mappers are easier to extend as it is an object list.
But ECN mode is of type sai_tunnel_decap_ecn_mode_t. So either we create these enums per tunnel type (sai_tunnel_decap/encp/_vxlan/ipinip/mpls_ecn_mode_t
or just leave the ecn mode to support model 1. What do you think?

itaibaz: I would prefer per tunnel type attributes
Also, not sure I follow the usage of single ecn mappers attribute with object list. We can attach multiple maps to a single attrib, but how will we specify which map is valid for IP in IP and which one is valid for VXLAN ?

JaiOCP: @itaibaz can you confirm if this is what you are requesting. 
- sai_tunnel_decap_vxlan_ecn_mode_t
- sai_tunnel_encap_vxlan_ecn_mode_t
- sai_tunnel_decap_ipinip_ecn_mode_t
- sai_tunnel_encap_ipinip_ecn_mode_t
- sai_tunnel_decap_mpls_ecn_mode_t
- sai_tunnel_encap_mpls_ecn_mode_t

JaiOCP: @itaibaz It makes sense to keep it consistent with tunnel ecn mode and provide per tunnel type attributes.

Change Done !!

https://github.com/opencomputeproject/SAI/pull/1153/commits/7347e0b87b95f283d4e1d0311ffbc934a298de6c

inc/saiswitch.h
```diff
@@ -2256,6 +2432,158 @@ typedef enum _sai_switch_attr_t
...
+     * @brief MPLS Tunnel encap ECN Mode load method
+     *
+     * @type sai_tunnel_mpls_encap_ecn_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_TUNNEL_MPLS_ENCAP_ECN_MODE_STANDARD
+     */
+    SAI_SWITCH_ATTR_TUNNEL_MPLS_ENCAP_ECN_MODE,
+
+    /**
+     * @brief Tunnel decap ECN Mode load method
```
itaibaz, line 2552: Add IP in IP in beginning of comment

inc/saiswitch.h
```diff
@@ -2256,6 +2432,158 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * @type sai_object_list_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_TUNNEL_MAP
+     * @default empty
+     */
+    SAI_SWITCH_ATTR_TUNNEL_MPLS_DECAP_ECN_MAPPERS,
+
+    /**
+     * @brief Tunnel encap ECN Mode load method
```
itaibaz, line 2516: Add IP in IP in beginning of comment

inc/saiswitch.h
```diff
@@ -2256,6 +2432,158 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * @type sai_object_list_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_TUNNEL_MAP
+     * @default empty
+     */
+    SAI_SWITCH_ATTR_TUNNEL_MPLS_ENCAP_ECN_MAPPERS,
+
+    /**
+     * @brief Tunnel mappers for decap
```
itaibaz, line 2476: Add IP in IP in beginning of comment

inc/saiswitch.h
```diff
@@ -2256,6 +2432,158 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_SUPPORTED_FAILOVER_MODE,
 
+    /**
+     * @brief Tunnel mappers for encap
```
itaibaz, line 2436: Add IP in IP in beginning of comment

inc/saiswitch.h
```diff
@@ -325,6 +325,182 @@ typedef enum _sai_switch_failover_config_mode_t
 
 } sai_switch_failover_config_mode_t;
 
+/**
+ * @brief Defines tunnel encap ECN mode
+ */
+typedef enum _sai_tunnel_ipinip_encap_ecn_mode_t
```
itaibaz, line 331: Do we need to have enum per type, or can use a single enum ?
I think single enum would be good enough, but remember there is a prefix check, which maybe forces enum per type. Perhaps ask Kamil if a single enum can be used ?

meta/saisanitycheck.c
```diff
@@ -372,7 +372,7 @@ void check_attr_by_object_type()
             sai_object_type_t current = ot[index]->objecttype;
 
             META_ASSERT_TRUE(current == i, "object type must be equal on object type list");
-            META_ASSERT_TRUE(index < 200, "object defines > 200 attributes, metadata bug?");
+            META_ASSERT_TRUE(index < 300, "object defines > 300 attributes, metadata bug?");
```
kcudnik, line 375: how come this is triggered? i dont see any @brief tag to be longer than 200 chars ?

kcudnik (approved):

kcudnik:
please fix conflicts

JaiOCP:
Please refer to https://github.com/opencomputeproject/SAI/pull/1173 for updated design

PR #1155, opened 2020-11-05
Add attribute to query available packet DMA pool size

jasmeetbagga:
Summary:
Add more attribute for querying available packet DMA
pool size. This can be used for accounting for and
debugging allocations from packet DMA pool.

Test Plan: make in meta dir

Reviewers:

Subscribers:

Tasks:

Tags:


jasmeetbagga:
@JaiOCP @rlhui  follow up PR from #1137 I had already added an attribute in #1137 to query the total packet DMA pool size. This now adds another attribute to query available packet DMA pool size. Taken together this would be sufficient to debug requests failing due to exhaustion of DMA pool.

I choose not to add a notion of client ID to allocations, since at least in our use case there is a single client allocating/freeing this memory. And when going through the kernel (say knet in BRCM parlance) allocations would be through the kernel (which is again a single client). So it did not seem worthwhile cluttering the interface till we have a motivating use case. Once we have such a use case, its also debatable whether such accounting belongs to the SAI-adaptor or in the application itself. For e.g. for other resources that are "allocated" from the ASIC, we don't account for which client allocated it. Why should this resource be any different. 

Thanks,
Jasmeet

inc/saiswitch.h
```diff
@@ -2256,6 +2256,16 @@ typedef enum _sai_switch_attr_t
...
 
+    /**
+     * @brief The size of the available packet DMA pool memory in bytes
+     * This can be used in conjunction with total packet DMA pool
+     * size to account/debug % of memory available.
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_PACKET_AVAILABLE_DMA_MEMORY_POOL_SIZE,
```
kcudnik, line 2267: is this like a single pool size? or there could be multiple pools and each of them could have different sizes ?

jasmeetbagga:
@kcudnik yes single pool. This is consistent with the already present overall pool size - https://github.com/opencomputeproject/SAI/blob/master/inc/saiswitch.h#L2240 and the allocation functions - https://github.com/opencomputeproject/SAI/blob/master/inc/saihostif.h#L1300-L1319, neither of which have notion of multiple pools. 
Multiple pools is a theoretical possibility, but in the ASICs I am familiar with, there is a single pool.

kcudnik (approved):

jasmeetbagga:
@rlhui @JaiOCP any comments?

rlhui:
@gechiang - would you please help review?

gechiang:
@jasmeetbagga Can you please resolve the conflict?
Thanks!

rlhui:
Not sure why there were so many commits in this PR now.  Maybe it's better to create a new PR for this?

jasmeetbagga:
@rlhui created a new one at #1198

rlhui:
Please close this one.

jasmeetbagga:
Duped and lande #1198

PR #1156, opened 2020-11-06
Updates for Linux MACsec driver and 802.1AE-2018

qbdwlr:
===================================================================================
This commit contains following non-backward compatible changes

1.            Replace attribute SAI_MACSEC_SC_ATTR_MACSEC_SSCI with SAI_MACSEC_SA_ATTR_MACSEC_SSCI.
Reason:
IEEE standards 802.1AE-2018 and 802.1X-2020 both state that SSCI is a property of the SA not the SC.  SSCI can only be assigned once the full Live Peer List has been exchanged via EAPOL between peers, which will occur after the KaY has already created the transmit SC for the actor.

    /**
     * @brief SSCI value for this Secure Association
     *
     * Valid when SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_128 or SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_256.
     *
     * @type sai_uint32_t
     */
    SAI_MACSEC_SA_ATTR_MACSEC_SSCI,

2.            Replace attribute SAI_MACSEC_SA_ATTR_ENCRYPTION_ENABLE with SAI_MACSEC_SC_ATTR_ENCRYPTION_ENABLE
Reason:
As seen from both the open source KaY (wpa-supplicant) and SecY (Linux MACsec driver), the encryption attribute is typically an attribute of the SC, not the SA.  This alignment can be retained by shifting the SAI MACsec definition of the encryption attribute from the SA to the SC.  This should allow for a cleaner, more direct translation between the SAI MACsec layer and the underlying platform driver.

    /**
     * @brief True means encryption is enabled.  False means encryption is disabled.
     *
     * @type bool
     * @flags CREATE_AND_SET
     * @default true
     */
    SAI_MACSEC_SC_ATTR_ENCRYPTION_ENABLE,

3.            Replace attribute SAI_MACSEC_SC_ATTR_MACSEC_XPN64_ENABLE and SAI_MACSEC_SA_ATTR_SAK_256_BITS with SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE
Reason:
As seen from both the open source KaY (wpa-supplicant) and SecY (Linux MACsec driver), the MACsec cipher suite is typically a single attribute of the SC.  This alignment can be retained by combining the two properties of the MACsec cipher suite (SAK length and XPN enable) of the SAI MACsec definition into a single attribute of the SC.  This should allow for a cleaner, more direct translation between the SAI MACsec layer and the underlying platform driver. 

    /**
     * @brief Cipher suite for this Secure Channel.
     *
     * @type sai_macsec_cipher_suite_t
     */
    SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE,
===================================================================================

### DIFF
--- a/inc/saimacsec.h
+++ b/inc/saimacsec.h
@@ -46,13 +46,13 @@ typedef enum _sai_macsec_direction_t
 /**
  * @brief MACsec Cipher Suites.
  */
-typedef enum _sai_macsec_direction_t
+typedef enum _sai_macsec_cipher_suite_t
 {
     SAI_MACSEC_CIPHER_SUITE_GCM_AES_128,
     SAI_MACSEC_CIPHER_SUITE_GCM_AES_256,
     SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_128,
     SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_256
-} sai_macsec_direction_t;
+} sai_macsec_cipher_suite_t;
 
 /**
  * @brief Attribute Id for sai_macsec
--- a/inc/saimacsec.h
+++ b/inc/saimacsec.h
@@ -773,6 +773,7 @@ typedef enum _sai_macsec_sa_attr_t
 
     /**
      * @brief SSCI value for this Secure Association
+     *
      * Valid when SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_128 or SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_256.
      *
      * @type sai_uint32_t
--- a/inc/saimacsec.h
+++ b/inc/saimacsec.h
@@ -734,7 +734,7 @@ typedef enum _sai_macsec_sa_attr_t
      * @brief MACsec Salt used for encryption/decryption.
      * Network Byte order.
      *
-     * Valid when SAI_MACSEC_SC_ATTR_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_128 or SAI_MACSEC_SC_ATTR_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_256.
+     * Valid when SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_128 or SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_256.
      *
      * @type sai_macsec_salt_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
@@ -773,7 +773,7 @@ typedef enum _sai_macsec_sa_attr_t
 
     /**
      * @brief SSCI value for this Secure Association
-     * Valid when SAI_MACSEC_SC_ATTR_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_128 or SAI_MACSEC_SC_ATTR_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_256.
+     * Valid when SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_128 or SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_256.
      *
      * @type sai_uint32_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
--- a/inc/saimacsec.h
+++ b/inc/saimacsec.h
@@ -646,6 +646,23 @@ typedef enum _sai_macsec_sc_attr_t
     SAI_MACSEC_SC_ATTR_SA_LIST,
 
     /**
+     * @brief Cipher suite for this Secure Channel.
+     *
+     * @type sai_uint64_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE,
+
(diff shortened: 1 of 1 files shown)


kcudnik:
please change phrase "Used only" to "Valid only"

inc/saimacsec.h
```diff
@@ -645,6 +645,23 @@ typedef enum _sai_macsec_sc_attr_t
      */
     SAI_MACSEC_SC_ATTR_SA_LIST,
 
+    /**
+     * @brief Cipher suite for this Secure Channel.
+     *
+     * @type sai_uint64_t
```
lguohan, line 651: if it is cipher suite, then it is better to an enum instead of uint64

inc/saimacsec.h
```diff
@@ -778,6 +795,15 @@ typedef enum _sai_macsec_sa_attr_t
      */
     SAI_MACSEC_SA_ATTR_MINIMUM_XPN,
 
+    /**
+     * @brief SSCI value for this Secure Association
+     * Used only if SAI_MACSEC_SC_ATTR_MACSEC_XPN64_ENABLE == true.
```
lguohan, line 800: please check @kcudnik 's comment

inc/saimacsec.h
```diff
@@ -645,6 +645,23 @@ typedef enum _sai_macsec_sc_attr_t
...
+    SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE,
+
+    /**
+     * @brief True means encryption is enabled.  False means encryption is disabled.
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default true
+     */
+    SAI_MACSEC_SC_ATTR_ENCRYPTION_ENABLE,
```
dipankar-ba, line 663: This is definitely more optimal.  However, this change should be accompanied by removal of SAI_MACSEC_SA_ATR_ENCRYPTION_ENABLE, if the associate driver and NOS code churn is acceptable.

inc/saimacsec.h
```diff
@@ -778,6 +795,15 @@ typedef enum _sai_macsec_sa_attr_t
...
     SAI_MACSEC_SA_ATTR_MINIMUM_XPN,
 
+    /**
+     * @brief SSCI value for this Secure Association
+     * Used only if SAI_MACSEC_SC_ATTR_MACSEC_XPN64_ENABLE == true.
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_MACSEC_SA_ATTR_MACSEC_SSCI,
```
dipankar-ba, line 805: For point-to-point SC, the SC does not survive CA membership change and there is no need for per-SA SSCI.  The only membership change in an established CA is when it goes from 2 members to 1 and the SC should be dismantled.
For point-to-multipoint SC, the only reason SSCI has to change is because SSCI is not part of the key agreement protocol and is derived from SCI in a kludgy way.  Since we designed saimacsec to only support p2p, we decided to avoid the kludge.
If we need to support point-to-multipoint SC, then this change is needed.

inc/saimacsec.h
```diff
@@ -645,6 +645,23 @@ typedef enum _sai_macsec_sc_attr_t
      */
     SAI_MACSEC_SC_ATTR_SA_LIST,
 
+    /**
+     * @brief Cipher suite for this Secure Channel.
+     *
+     * @type sai_uint64_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE,
```
dipankar-ba, line 654: In saimacsec, the ciphersuite parameters are all separately defined (e.g. SAI_MACSEC_SA_ATTR_SAK_256_BITS).  What extra information will this parameter provide the driver?

inc/saimacsec.h
qbdwlr: TX SC is created prior to knowledge of peer SCI even in p2p scenario.  It is not possible to assign SSCI value to SC without knowledge of the peer SCI.  Standard states that SSCI is a parameter of the SA.

inc/saimacsec.h
qbdwlr: I would propose to additionally eliminate the existing SAI_MACSEC_SA_ATTR_SAK_256_BITS and SAI_MACSEC_SC_ATTR_XPN64_ENABLE.  I am not aware of the history behind the current saimacsec definition, but there doesn't seem to be a benefit to separating the cipher suite into two attributes and putting one in the SC and the other in the SA when both are determined at the same time by the cipher suite.

inc/saimacsec.h
qbdwlr: Yes, I would propose to eliminate the existing SAI_MACSEC_SA_ATTR_ENCRYPTION_ENABLE in favor of SC attr.

inc/saimacsec.h
dipankar-ba: The sai_create_macsec_sc_fn can be called after the NOS has established SC in both directions.  So it is still a choice.  I have no objection a change to SAI_MACSEC_SA_ATTR_MACSEC_SSCI, but it is would require change to some code that has already been written.

inc/saimacsec.h
dipankar-ba: Key-length and sequence numbers are independent parameters and names clearly describe the functionality.  Both are part of cipher-suite definition for MACsec, but not for IPsec.  In future we may use the same API for other protocols.

inc/saimacsec.h
```diff
@@ -778,6 +771,15 @@ typedef enum _sai_macsec_sa_attr_t
      */
     SAI_MACSEC_SA_ATTR_MINIMUM_XPN,
 
+    /**
+     * @brief SSCI value for this Secure Association
+     * Valid when SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_128 or SAI_MACSEC_SC_ATTR_MACSEC_CIPHER_SUITE == SAI_MACSEC_CIPHER_SUITE_GCM_AES_XPN_256.
```
kcudnik, line 776: please add empty line before brief and Valid ..

kcudnik (approved):

lguohan (approved):

JaiOCP:
Looks Good

JaiOCP (approved):

dipankar-ba (approved):
Looks good

PR #1158, opened 2020-11-12
Add label attribute for LAG and virtual router

kcudnik:
Since LAG and virtual router can be created without any unique mandatory attributes, anchor attribute can be used to uniquely identify specified objects. This will be very useful in SONiC warm boot scenario, for example to identify 2 empty LAGs which are present on the device after reboot.

This attribute doesn't correspond to any internal SAI vendor device resources, and at this point it don't need to be implemented by any vendor, it implementation at create function internals **CAN** be skipped without posting error. This attribute is considered as user data attached to specific object.  
If vendor wants to support this attribute it should also be persistent over warm boot scenario

**Cons**: extra attribute, that has no actual resource meaning and it would need to be added to objects that need to be uniquely identified on user side

**Pros**: convenient for SONiC, since metadata for all attributes is auto generated, and it will make very easy to to use existing attribute, instead of making convoluted way to pass extra information from sairedis/syncd path (this is because of current SONiC architecture design in sairedis/syncd model)

Related to: 
https://github.com/opencomputeproject/SAI/pull/1080
https://github.com/opencomputeproject/SAI/pull/1132

### DIFF
--- a/inc/sailag.h
+++ b/inc/sailag.h
@@ -168,13 +168,13 @@ typedef enum _sai_lag_attr_t
     SAI_LAG_ATTR_SYSTEM_PORT_AGGREGATE_ID,
 
     /**
-     * @brief Anchor attribute used to unique identify empty LAG.
+     * @brief Label attribute used to unique identify empty LAG.
      *
      * @type char
      * @flags CREATE_AND_SET
      * @default ""
      */
-    SAI_LAG_ATTR_ANCHOR,
+    SAI_LAG_ATTR_LABEL,
 
     /**
      * @brief End of attributes
--- a/inc/sailag.h
+++ b/inc/sailag.h
@@ -170,9 +170,9 @@ typedef enum _sai_lag_attr_t
     /**
      * @brief Anchor attribute used to unique identify empty LAG.
      *
-     * @type sai_uint64_t
-     * @flags CREATE_ONLY
-     * @default 0
+     * @type char
+     * @flags CREATE_AND_SET
+     * @default ""
      */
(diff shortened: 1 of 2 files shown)


inc/sailag.h
```diff
@@ -167,6 +167,15 @@ typedef enum _sai_lag_attr_t
...
     SAI_LAG_ATTR_SYSTEM_PORT_AGGREGATE_ID,
 
+    /**
+     * @brief Anchor attribute used to unique identify empty LAG.
+     *
+     * @type sai_uint64_t
+     * @flags CREATE_ONLY
+     * @default 0
+     */
+    SAI_LAG_ATTR_ANCHOR,
```
marian-pritsak, line 177: Can we make it more meaningful (string instead of u64)?

The reason for that is that SAI can be able to pass this metadata to vendor SDK, which will be displayed in the debug dump and other tools. Otherwise, SAI will just ignore it anyway.

kcudnik: i actually wanted to make this simple as possible, previous approach #1080 was adding string here, i wanted this be simpler to store for sonic and vendors

kcudnik: addressed meeting notes, a changed to create_and_set for warm boot scenario, and also changed to char type which corresponds to chardata in sai_attribute_value_t which can be up 32 bytes. This we would also assume that this will be some ascii user readable string

marian-pritsak (approved):

lguohan:
apart from warm reboot use case, what are other use cases? can we get more concrete? @marian-pritsak

lguohan (approved):

PR #1169, opened 2020-12-07
Separate XPN configuration attribute from read-only attribute

dipankar-ba:
Added a new read-only current XPN (available for both ingress and egress) and separate from the egress attribute for setting initial XPN value.
Renamed ingress set-and-clear attribute MINIMUM_XPN to MINIMUM_INGRESS_XPN.

### DIFF
--- a/inc/saimacsec.h
+++ b/inc/saimacsec.h
@@ -751,14 +751,14 @@ typedef enum _sai_macsec_sa_attr_t
     SAI_MACSEC_SA_ATTR_AUTH_KEY,
 
     /**
-     * @brief MACsec egress packet number (PN/XPN).  At most 1 less than the next PN/XPN.
+     * @brief Configured value of next MACsec egress packet number (PN/XPN).
      *
      * @type sai_uint64_t
      * @flags CREATE_AND_SET
      * @default 0
      * @validonly SAI_MACSEC_SA_ATTR_MACSEC_DIRECTION == SAI_MACSEC_DIRECTION_EGRESS
      */
-    SAI_MACSEC_SA_ATTR_INITIAL_EGRESS_XPN,
+    SAI_MACSEC_SA_ATTR_CONFIGURED_EGRESS_XPN,
 
     /**
      * @brief MACsec current packet number (PN/XPN). For ingress, largest
--- a/inc/saimacsec.h
+++ b/inc/saimacsec.h
@@ -118,12 +118,36 @@ typedef enum _sai_macsec_attr_t
     SAI_MACSEC_ATTR_SCI_IN_INGRESS_MACSEC_ACL,
 
     /**
-     * @brief List of supported cipher-suites
+     * @brief Indicates if 32-bit Packer Number (PN) is supported.
      *
-     * @type sai_s32_list_t sai_macsec_cipher_suite_t
+     * @type bool
      * @flags READ_ONLY
      */
-    SAI_MACSEC_ATTR_SUPPORTED_CIPHER_SUITE_LIST,
+    SAI_MACSEC_ATTR_PN_32BIT_SUPPORTED,
+
+    /**
+     * @brief Indicates if 64-bit Extended Packer Number (PN) is supported.
+     *
+     * @type bool
+     * @flags READ_ONLY
+     */
+    SAI_MACSEC_ATTR_XPN_64BIT_SUPPORTED,
+
+    /**
+     * @brief Indicates if GCM-AES128 cipher-suite is supported.
+     *
+     * @type bool
+     * @flags READ_ONLY
+     */
+    SAI_MACSEC_ATTR_GCM_AES128_SUPPORTED,
+
+    /**
+     * @brief Indicates if GCM-AES256 cipher-suite is supported.
+     *
+     * @type bool
+     * @flags READ_ONLY
+     */
+    SAI_MACSEC_ATTR_GCM_AES256_SUPPORTED,
 
     /**
      * @brief List of supported SecTAG offset values for both ingress parsing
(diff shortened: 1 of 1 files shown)


kcudnik:
We discussed about this whether we want to separate attributes, or add a flag "@dynamic" for existing attribute. There are pros and cons for both of them, we didn't decided yet for that. Will let you know.

@lguohan yfi

qbdwlr:
I think it makes sense to have a separate operational vs configuration attribute to avoid multiple writers (SAI and MKA) to the same attribute.
There is no precedent in this SAI MACsec definition for including the keywords "INGRESS" and "EGRESS" even though there are other attributes that are specific to a single direction.  I don't think changing the naming is warranted as the comments make the usage clear and this will keep the naming convention consistent.
Since there is no longer a scenario for their values to change post-creation, I would suggest updating the configuration attributes to CREATE_ONLY.

inc/saimacsec.h
```diff
@@ -758,18 +758,27 @@ typedef enum _sai_macsec_sa_attr_t
      * @default 0
      * @validonly SAI_MACSEC_SA_ATTR_MACSEC_DIRECTION == SAI_MACSEC_DIRECTION_EGRESS
      */
-    SAI_MACSEC_SA_ATTR_XPN,
+    SAI_MACSEC_SA_ATTR_INITIAL_EGRESS_XPN,
```
lguohan, line 761: i think we can do alias here so that we can maintain backward compatible here.

inc/saimacsec.h
```diff
@@ -758,18 +758,27 @@ typedef enum _sai_macsec_sa_attr_t
...
+     * @brief Configured minimum acceptable ingress MACsec packet number (PN/XPN).
      * Updated by value from MACsec peer by Key Agreement protocol.
      *
      * @type sai_uint64_t
      * @flags CREATE_AND_SET
      * @default 1
      * @validonly SAI_MACSEC_SA_ATTR_MACSEC_DIRECTION == SAI_MACSEC_DIRECTION_INGRESS
      */
-    SAI_MACSEC_SA_ATTR_MINIMUM_XPN,
+    SAI_MACSEC_SA_ATTR_MINIMUM_INGRESS_XPN,
```
lguohan, line 781: i think we can do alias here so that we can maintain backward compatible here.

dipankar-ba: I would rather change the 3 attribute names to:
SAI_MACSEC_SA_ATTR_CONFIGURED_EGRESS_XPN (new attribute)
SAI_MACSEC_SA_ATTR_XPN (changed from CREATE_AND_SET to READ_ONLY) **
SAI_MACSEC_SA_ATTR_MINIMUM_XPN (no change)

** There is no need to change the default value of this attribute for normal MACsec operation.  If existing NOS code never set this attribute, it would work and there would be no backward compatibility issue.

I prefer this approach to attribute name alias,  But the other deterrent is header compile (meta/make) fails with a bunch of parsing warnings when I try ATTR2 = ATTR1, preceded by @ignore flag.

kcudnik: What errors you get ?

dipankar-ba: The parse warning & resulting compile failure goes away if we put the ignore in 1 line likes this:
    /** @ignore - for backward compatibility */
I will use this and do another commit.

Actually the parse errors occurred if I did it in the usual style the compile failure was preceded by the following parse error:
WARNING: empty line required between each below elements:                                        
WARNING: desired elemen order: @brief @warning? description? @attributes? @params? @return?      WARNING: invalid spacing (@) on saimacsec.h:                                                     
/**                                                                                                   * @ignore - for backward compatibility                                                      
     */

kcudnik: seems like you get wrong spaces, checkout other "@ignore" attributes and try to follow them

kcudnik (approved):

prafull-brcm (approved):
Changes looks good to me.

lguohan (approved):

inc/saimacsec.h
```diff
@@ -751,25 +751,40 @@ typedef enum _sai_macsec_sa_attr_t
...
+     * @brief MACsec current packet number (PN/XPN). For ingress, largest
+     * received packet number. For egress, 1 less than the next packet number.
+     *
+     * @type sai_uint64_t
+     * @flags READ_ONLY
+     */
+    SAI_MACSEC_SA_ATTR_CURRENT_XPN,
+
+    /** @ignore - for backward compatibility */
+    SAI_MACSEC_SA_ATTR_XPN = SAI_MACSEC_SA_ATTR_CURRENT_XPN,
```
abdosi, line 773: @dipankar-ba this does not look correct. SAI_MACSEC_SA_ATTR_XPN should be alias SAI_MACSEC_SA_ATTR_CONFIGURED_EGRESS_XPN ? This has broken backward compatibility as SAI_MACSEC_SA_ATTR_XPN was Create and Set and now it became Read only.  

Here is the error when we are moving v1.8.0 
https://dev.azure.com/mssonic/build/_build/results?buildId=9116&view=logs&jobId=83516c17-6666-5250-abde-63983ce72a49&j=83516c17-6666-5250-abde-63983ce72a49&t=6177235f-d4f1-5f72-835a-90ebb93a1784

can you please check again

cc @rlhui @kcudnik

dipankar-ba: The underlying hardware has a counter whose value has to be periodically read by NOS to run MKA protocol through SAI_MACSEC_SA_ATTR_CURRENT_XPN.  Since this read is mandatory for MACsec operation, I tried to keep this backward compatible.
For counter should be initialized to 0 by the driver and normally never be modified by NOS for normal operation.  So I do not expect SAI_MACSEC_SA_ATTR_CONFIGURED_EGRESS_XPN to be read or written except perhaps for synthetic tests for testing counter saturation.

PR #1171, opened 2020-12-08
Add SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO format for GB MDIO sysfs access  

sujinmkang:
SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO is the right attribute to be used to provide the each Phy information for mdio sysfs driver support

- SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO shall contain the interface_name and phy_id with “/” as the separator

- The data should be formatted as {intf_name}/{phy_id} : Ex) eth0/1

- GB Phy SAI is expected to parse the SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO to determine the phy_id along with interface name for each phySAI


kcudnik:
Fix aspell "sysfs", and also this seems like a information related only to gearbox not generic, other vendors can put any information they want there.

lguohan (approved):

lguohan:
@sujinmkang can you add the keyword into dictionary.

kcudnik:
```WARNING: Word 'SysFS' is misspelled saiswitch.h 1447:     * Example: Like PCI location, I2C address, MDIO bus SysFS information etc.
```
just add SysFS to aspell dictionary in meta dir

kcudnik (approved):

rlhui:
@prafull-brcm , @apogedik - would you please help review/approve?  Thanks.

inc/saiswitch.h
```diff
@@ -1444,8 +1444,10 @@ typedef enum _sai_switch_attr_t
      *
      * Hardware information format is based on SAI implementations by vendors.
      * String is NULL terminated. Format is vendor specific.
-     * Example: Like PCI location, I2C address etc.
+     * Example: Like PCI location, I2C address, MDIO bus SysFS information etc.
      * In case of NULL, First NPU attached to CPU will be initialized.
+     * For the MDIO SysFS driver support, the interface name and phy_id should be
```
prafull-brcm, line 1449: For non sysfs MDIO, I think this statement should also be added as it is not always we are using mdio sysfs.
For non Sysfs MDIO, only phy_id should be provided.

prafull-brcm (changes requested):
For non sysfs MDIO, I think this statement should also be added as it is not always we are using mdio sysfs.
For non Sysfs MDIO, only phy_id should be provided.

prafull-brcm (approved):
Changes looks good to me.

PR #1172, opened 2020-12-09
[macsec]: add SAI_MACSEC_ATTR_SUPPORTED_CIPHER_SUITE_LIST

dipankar-ba:
…nsistent with the PR1156 style change in SAI_MACSEC_SC_ATTR

Signed-off-by: dipankar-ba <dipankar@arista.com>

### DIFF
--- a/inc/saimacsec.h
+++ b/inc/saimacsec.h
@@ -126,6 +126,42 @@ typedef enum _sai_macsec_attr_t
     SAI_MACSEC_ATTR_SUPPORTED_CIPHER_SUITE_LIST,
 
     /**
+     * @brief Indicates if 32-bit Packer Number (PN) is supported.  This is To be deprecated
+     * in a future release, replaced by SAI_MACSEC_ATTR_SUPPORTED_CIPHER_SUITE_LIST.
+     *
+     * @type bool
+     * @flags READ_ONLY
+     */
+    SAI_MACSEC_ATTR_PN_32BIT_SUPPORTED,
+
+    /**
+     * @brief Indicates if 64-bit Extended Packer Number (PN) is supported. This is To be deprecated
+     * in a future release, replaced by SAI_MACSEC_ATTR_SUPPORTED_CIPHER_SUITE_LIST.
+     *
+     * @type bool
+     * @flags READ_ONLY
+     */
+    SAI_MACSEC_ATTR_XPN_64BIT_SUPPORTED,
+
+    /**
+     * @brief Indicates if GCM-AES128 cipher-suite is supported. This is To be deprecated
+     * in a future release, replaced by SAI_MACSEC_ATTR_SUPPORTED_CIPHER_SUITE_LIST.
+     *
+     * @type bool
+     * @flags READ_ONLY
+     */
+    SAI_MACSEC_ATTR_GCM_AES128_SUPPORTED,
+
+    /**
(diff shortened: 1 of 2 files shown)


inc/saimacsec.h
```diff
@@ -118,36 +118,12 @@ typedef enum _sai_macsec_attr_t
...
-    SAI_MACSEC_ATTR_GCM_AES128_SUPPORTED,
-
-    /**
-     * @brief Indicates if GCM-AES256 cipher-suite is supported.
-     *
-     * @type bool
+     * @type sai_s32_list_t sai_macsec_cipher_suite_t
      * @flags READ_ONLY
      */
-    SAI_MACSEC_ATTR_GCM_AES256_SUPPORTED,
```
lguohan, line 150: I think you can mark these attribute deprecated, not removing the immediately. we can remove it in 3~6 months.

inc/saimacsec.h
```diff
@@ -118,31 +118,43 @@ typedef enum _sai_macsec_attr_t
...
-     * @brief Indicates if 32-bit Packer Number (PN) is supported.
+     * @brief List of supported cipher-suites
+     *
+     * @type sai_s32_list_t sai_macsec_cipher_suite_t
+     * @flags READ_ONLY
+     */
+    SAI_MACSEC_ATTR_SUPPORTED_CIPHER_SUITE_LIST,
+
+    /**
+     * @brief Indicates if 32-bit Packer Number (PN) is supported.  This is To be deprecated
```
kcudnik, line 129: there is a flag "@depreacated true" which can be added to flags

meta/parse.pl
```diff
@@ -265,12 +265,13 @@ sub ProcessTagIsRecourceType
 
 sub ProcessTagDeprecated
 {
-    # just return true if defined
+    # just return true if not defined
```
kcudnik, line 268: previous statement was correct

meta/parse.pl
```diff
@@ -265,12 +265,13 @@ sub ProcessTagIsRecourceType
...
 
     my ($type, $value, $val) = @_;
 
-    LogError "deprecated tag should not have value '$val'" if not $val =~ /^$/i;
+    return $val if $val =~ /^(true|false)$/i;
 
+    LogError "deprecated tag value '$val', expected true/false";
```
kcudnik, line 274: this is similar to isresourcetype, return "undef" maybe here?

meta/parse.pl
kcudnik: actually this comment could be removed

kcudnik (changes requested):

inc/saimacsec.h
dipankar-ba: ok

inc/saimacsec.h
dipankar-ba: Done

meta/parse.pl
dipankar-ba: Done

meta/parse.pl
dipankar-ba: Changed to undef

kcudnik (approved):

prafull-brcm (commented):
Changes looks good to me.

lguohan (approved):

PR #1173, opened 2020-12-09
Switch scoped tunnel attributes

JaiOCP:
This PR is the closure of discussion in
https://github.com/opencomputeproject/SAI/pull/1153

There is switch level tunnel object which can be set for various attributes per tunnel type.

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -341,6 +341,22 @@ typedef enum _sai_tunnel_type_t
 } sai_tunnel_type_t;
 
 /**
+ * @brief Defines VXLAN tunnel UDP source port mode
+ */
+typedef enum _sai_tunnel_vxlan_udp_sport_mode_t
+{
+    /**
+     * @brief User define value
+     */
+    SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_USER_DEFINED,
+
+    /**
+     * @brief RFC6335 Computed hash value in range 49152-65535
+     */
+    SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_EPHEMERAL,
+} sai_tunnel_vxlan_udp_sport_mode_t;
+
+/**
  * @brief Defines tunnel encap ECN mode
  */
 typedef enum _sai_tunnel_encap_ecn_mode_t
@@ -424,7 +440,7 @@ typedef enum _sai_switch_tunnel_attr_t
     SAI_SWITCH_TUNNEL_ATTR_TUNNEL_ENCAP_ECN_MODE,
 
     /**
-     * @brief Tunnel encap mappers
+     * @brief Tunnel encap ECN mappers only
      *
      * @type sai_object_list_t
      * @flags CREATE_ONLY
@@ -445,7 +461,7 @@ typedef enum _sai_switch_tunnel_attr_t
     SAI_SWITCH_TUNNEL_ATTR_TUNNEL_DECAP_ECN_MODE,
 
     /**
-     * @brief Tunnel decap mappers
+     * @brief Tunnel decap ECN mappers only
      *
      * @type sai_object_list_t
      * @flags CREATE_ONLY
@@ -453,6 +469,26 @@ typedef enum _sai_switch_tunnel_attr_t
      * @default empty
      */
     SAI_SWITCH_TUNNEL_ATTR_DECAP_MAPPERS,
+
+    /**
+     * @brief Tunnel VXLAN UDP source port mode
+     *
+     * @type sai_tunnel_vxlan_udp_sport_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_EPHEMERAL
+     */
+    SAI_SWITCH_TUNNEL_ATTR_TUNNEL_VXLAN_UDP_SPORT_MODE,
+
+    /**
+     * @brief Tunnel UDP source port
+     *
+     * @type sai_uint16_t
+     * @flags CREATE_AND_SET
+     * @isvlan false
+     * @default 0
+     * @validonly SAI_SWITCH_TUNNEL_ATTR_TUNNEL_TYPE == SAI_TUNNEL_TYPE_VXLAN and SAI_SWITCH_TUNNEL_ATTR_TUNNEL_VXLAN_UDP_SPORT_MODE == SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_USER_DEFINED
+     */
+    SAI_SWITCH_TUNNEL_ATTR_VXLAN_UDP_SPORT,
 
     /**
      * @brief End of attributes
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -326,6 +326,135 @@ typedef enum _sai_switch_failover_config_mode_t
 } sai_switch_failover_config_mode_t;
 
 /**
+ * @brief Defines tunnel type
+ */
+typedef enum _sai_tunnel_type_t
+{
+    SAI_TUNNEL_TYPE_IPINIP,
+
+    SAI_TUNNEL_TYPE_IPINIP_GRE,
+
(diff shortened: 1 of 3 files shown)


inc/saiswitch.h
```diff
@@ -325,6 +325,135 @@ typedef enum _sai_switch_failover_config_mode_t
...
+    /**
+     * @brief Start of attributes
+     */
+    SAI_SWITCH_TUNNEL_ATTR_START,
+
+    /**
+     * @brief Tunnel type
+     *
+     * @type sai_tunnel_type_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
```
itaibaz, line 397: should this also be marked as KEY ?
as we allow up to one object instance per switch ID per tunnel ID, meaning there can't be 2 objects for example for same switch ID for same tunnel type

itaibaz (commented):
I think we need to add a short comment about precedence of specific tunnel ECN settings over switch tunnel type settings

inc/saitypes.h
```diff
@@ -277,7 +277,9 @@ typedef enum _sai_object_type_t
     SAI_OBJECT_TYPE_MACSEC_SA                = 92,
     SAI_OBJECT_TYPE_SYSTEM_PORT              = 93,
     SAI_OBJECT_TYPE_FINE_GRAINED_HASH_FIELD  = 94,
-    SAI_OBJECT_TYPE_MAX,  /* Must remain in last position */
+    SAI_OBJECT_TYPE_SWITCH_TUNNEL            = 95,
+
+SAI_OBJECT_TYPE_MAX,  /* Must remain in last position */
```
kcudnik: Fox spacing

inc/saitypes.h
```diff
@@ -277,6 +277,8 @@ typedef enum _sai_object_type_t
     SAI_OBJECT_TYPE_MACSEC_SA                = 92,
     SAI_OBJECT_TYPE_SYSTEM_PORT              = 93,
     SAI_OBJECT_TYPE_FINE_GRAINED_HASH_FIELD  = 94,
+    SAI_OBJECT_TYPE_SWITCH_TUNNEL            = 95,
+
```
kcudnik, line 281: remove empty line

inc/saiswitch.h
```diff
@@ -2265,6 +2394,16 @@ typedef enum _sai_switch_attr_t
...
 
+    /**
+     * @brief Switch scoped Tunnel objects
+     *
+     * @type sai_object_list_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_SWITCH_TUNNEL
+     * @default empty
+     */
+    SAI_SWITCH_ATTR_TUNNEL_OBJECTS_LIST,
```
dgsudharsan, line 2405: Can we have this as read only and have switch object ID as attribute in the sai_switch_tunnel_object? This will make it inline with other objects

kcudnik: yes, switch_id is passed when create object function is called, so you already have it.
i don't see a point to have object_type_switch on any attribute, and even if you want to get switch id from tunnel object id, then use sai_switch_id_query()

JaiOCP: @kcudnik Sudharsan was thinking if we can create a binding of switch object in SAI_OBJECT_TYPE_SWITCH_TUNNEL. I think thats implicit.
@dgsudharsan Sudharsan, what do you think?

kcudnik: yes, thats implicit

inc/saitypes.h
JaiOCP: Done
https://github.com/opencomputeproject/SAI/pull/1173/commits/1ed0f2e544818c598b455b3046e159689e5fbfa4

inc/saiswitch.h
JaiOCP: Done
https://github.com/opencomputeproject/SAI/pull/1173/commits/1ed0f2e544818c598b455b3046e159689e5fbfa4

JaiOCP:
> I think we need to add a short comment about precedence of specific tunnel ECN settings over switch tunnel type settings

Done
https://github.com/opencomputeproject/SAI/pull/1173/commits/1ed0f2e544818c598b455b3046e159689e5fbfa4

inc/saitunnel.h
```diff
@@ -385,50 +370,6 @@ typedef enum _sai_tunnel_dscp_mode_t
 
 } sai_tunnel_dscp_mode_t;
 
-/**
- * @brief Defines tunnel encap ECN mode
- */
-typedef enum _sai_tunnel_encap_ecn_mode_t
```
kcudnik, line 391: why to move those enums to saiswitch.h instead of #include "saiswitch.h" here?

dgsudharsan (approved):

rlhui:
@kcudnik , @itaibaz , would you please sign-off? Thanks.

kcudnik (approved):

JaiOCP:
@itaibaz Please approve this PR. I have another set of changes on top this coming.

inc/saiswitch.h
```diff
@@ -325,9 +325,151 @@ typedef enum _sai_switch_failover_config_mode_t
...
+
+    /**
+     * @brief Tunnel encap mappers
+     *
+     * @type sai_object_list_t
+     * @flags CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_TUNNEL_MAP
+     * @default empty
+     */
+    SAI_SWITCH_TUNNEL_ATTR_ENCAP_MAPPERS,
```
marian-pritsak, line 434: This will allow for any kind of mapper to be defined at a switch level (VLAN2VNI, VRF2VNI etc.). I assume it was not intended and we need to mention it at least in the comment, or better to allow only for specific types of mappers that are switch scoped.

inc/saiswitch.h
```diff
@@ -325,9 +325,151 @@ typedef enum _sai_switch_failover_config_mode_t
...
+    SAI_SWITCH_TUNNEL_ATTR_TUNNEL_TYPE = SAI_SWITCH_TUNNEL_ATTR_START,
+
+    /**
+     * @brief Packet action when a packet ingress and gets routed back to same tunnel
+     *
+     * @type sai_packet_action_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PACKET_ACTION_FORWARD
+     */
+    SAI_SWITCH_TUNNEL_ATTR_LOOPBACK_PACKET_ACTION,
```
marian-pritsak, line 413: Why is this at switch scope and not per tunnel?

marian-pritsak (changes requested):

inc/saiswitch.h
itaibaz: Other than Marian's comments, I have no further comments. So once they are resolved, I will approve

itaibaz: There is already per tunnel loopback action, SAI_TUNNEL_ATTR_LOOPBACK_PACKET_ACTION
So here the intention was to add also switch attribute per tunnel type, for those ASIC who don't support per tunnel

rlhui:
@itaibaz  - reminder if there's anything needed from your end on this PR, or you'd sign off?  Thanks.

itaibaz:
> @itaibaz - reminder if there's anything needed from your end on this PR, or you'd sign off? Thanks.

There are 2 open comments from Marian, that needs to be addressed before we can sign off

inc/saiswitch.h
JaiOCP: It is as explained by Itai.

inc/saiswitch.h
JaiOCP: Added comments to intent that switch scoped mappers are for ECN mappers only
https://github.com/opencomputeproject/SAI/pull/1173/commits/ff7b05fb5c25a5d609fa5c0039d30f0e349596a3

marian-pritsak (approved):

JaiOCP:
@rlhui Please merge this PR

PR #1119, opened 2020-07-30
SAI Advanced Telemetry

JaiOCP:
This PR brings in advanced telemetry feature set in SAI
- Drop Monitoring
   Packet and Switch level drop monitoring with metadata
   Intelligent drop monitoring using flow and flow’s drop state
- Flow Monitoring
- Flow Latency Monitoring
   Static Thresholding
   Hysteresis based High/Low Watermark Thresholding

PR #1178, opened 2020-12-20
Support for ACL extensions in metadata

bandaru-viswanath:
SAI sanitychecker doesn't expect any extensions to be developed for ACL attributes. This PR aims to add support by adding checks whether the 'under-check-attribute' is an extension attribute. The value range for an attribute between SAI_ACL_ENTRY_ATTR_ACTION_START and SAI_ACL_ENTRY_ATTR_ACTION_END is valid only for non-extensions. The extensions are to have attributes beyond SAI_ACL_ENTRY_ATTR_ACTION_END. This PR allows the range check to be conditional for extension attributes.

If there is a better way to enhance the metachecker, this PR can be changed accordingly.

Signed-off-by: bandaru.viswanath@broadcom.com

### DIFF
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -435,6 +435,11 @@ bool sai_metadata_is_acl_field_or_action(
         {
             return true;
         }
+
+        if (metadata->isextensionattr == true)
+        {
+            return true;
+        }
     }
 
     return false;
@@ -1856,9 +1861,10 @@ void check_attr_acl_fields(
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_ID:
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_LIST:
 
-            if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY ||
-                    md->attrid < SAI_ACL_ENTRY_ATTR_ACTION_START ||
-                    md->attrid > SAI_ACL_ENTRY_ATTR_ACTION_END)
+            if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY  ||
+                    ((md->isextensionattr == false) &&
+                    (md->attrid < SAI_ACL_ENTRY_ATTR_ACTION_START  ||
+                    md->attrid > SAI_ACL_ENTRY_ATTR_ACTION_END)))
             {
                 META_MD_ASSERT_FAIL(md, "acl action may only be set on acl action");
             }
@@ -4078,7 +4084,7 @@ void check_acl_entry_actions()
             break;
         }
 
-        if (meta->attrid > SAI_ACL_ENTRY_ATTR_ACTION_END)
+        if ((meta->isextensionattr == false) && (meta->attrid > SAI_ACL_ENTRY_ATTR_ACTION_END))
         {
             break;
         }
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -1861,10 +1861,10 @@ void check_attr_acl_fields(
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_ID:
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_LIST:
 
-			if (md->objecttype == SAI_OBJECT_TYPE_ACL_ENTRY && md->isextensionattr)
-			{
-				break;
-			}
+            if (md->objecttype == SAI_OBJECT_TYPE_ACL_ENTRY && md->isextensionattr)
+            {
+                break;
+            }
 
             if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY  ||
                     md->attrid < SAI_ACL_ENTRY_ATTR_ACTION_START  ||
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -1861,8 +1861,7 @@ void check_attr_acl_fields(
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_ID:
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_LIST:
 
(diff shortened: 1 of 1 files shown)


meta/saisanitycheck.c
```diff
@@ -1856,9 +1861,10 @@ void check_attr_acl_fields(
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_ID:
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_LIST:
 
-            if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY ||
-                    md->attrid < SAI_ACL_ENTRY_ATTR_ACTION_START ||
-                    md->attrid > SAI_ACL_ENTRY_ATTR_ACTION_END)
+            if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY  ||
```
kcudnik, line 1864: Those kind od changes needs to be introduced with changes that demonstrates need for relaxing this conditions

kcudnik (changes requested):

meta/saisanitycheck.c
bandaru-viswanath: @kcudnik I am trying to basically use the ACL changes from Jai's PR #1119 as an extension since that PR is not approved as yet.  I am copy pasting the relevant extension enum here.  

```
typedef enum _sai_acl_entry_attr_extensions_t
{
    SAI_ACL_ENTRY_ATTR_EXTENSIONS_RANGE_START = SAI_ACL_ENTRY_ATTR_END,

    /**
     * @brief ACL bind point for TAM object
     *
     * Bind (or unbind) a TAM object.
     *
     * @type sai_acl_action_data_t sai_object_id_t
     * @flags CREATE_AND_SET
     * @objects SAI_OBJECT_TYPE_TAM
     * @allownull true
     * @default disabled
     */
    SAI_ACL_ENTRY_ATTR_ACTION_TAM_OBJECT,

    SAI_ACL_ENTRY_ATTR_EXTENSIONS_RANGE_END

} sai_acl_entry_attr_extensions_t;
```

As you can see the sanity checker doesn't allow any extensions for ACL module since it explicitly checks for the value to be between ACTION_START and ACTION_END.  So, it prevents any extension since the extensions range starts after ACTION_END.

Let me know if this helps.

kcudnik: my question here is why this change is separate from adding that extension acl entry ?

bandaru-viswanath: Basically, this specific extension is a short term measure since Jai's PR is not merged in. The proposed change (to metachecker) on the other hand is intended to be generic.

meta/saisanitycheck.c
```diff
@@ -435,6 +435,11 @@ bool sai_metadata_is_acl_field_or_action(
         {
             return true;
         }
+
+        if (metadata->isextensionattr == true)
+        {
+            return true;
```
kcudnik, line 441: problem with this is that, this function checkes whether attr id is acl field or acl action, but if attribute is some other attribute and is extension, then this function will also return true :( and i think there is no good way to annotate that

kcudnik: also change "if (metadata->isextensionattr == true)" to "if (metadata->isextensionattr)"

meta/saisanitycheck.c
kcudnik: the problem im having (as in second added comment) is that not all acl entry extension attributes will be action/field, it could be regular acl_entry attribute like SAI_ACL_ENTRY_ATTR_ADMIN_STATE
lucky enough currently there are only 2 non field/action attributes and possibly extension attributes will be in that range

meta/saisanitycheck.c
```diff
@@ -1856,9 +1861,10 @@ void check_attr_acl_fields(
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_ID:
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_LIST:
 
-            if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY ||
-                    md->attrid < SAI_ACL_ENTRY_ATTR_ACTION_START ||
-                    md->attrid > SAI_ACL_ENTRY_ATTR_ACTION_END)
+            if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY  ||
+                    ((md->isextensionattr == false) &&
```
kcudnik, line 1865: can you bring back previous condition and add this one before it:

```
if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY && metadata->isextensionattr)
{
    /* allow acl action data on excetnsion acl etnry attributes */
    break;
}
```

i think this will make code more readable

kcudnik (changes requested):
please address latest comments

meta/saisanitycheck.c
bandaru-viswanath: Done

meta/saisanitycheck.c
bandaru-viswanath: Done

meta/saisanitycheck.c
```diff
@@ -1856,8 +1861,13 @@ void check_attr_acl_fields(
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_ID:
         case SAI_ATTR_VALUE_TYPE_ACL_ACTION_DATA_OBJECT_LIST:
 
-            if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY ||
-                    md->attrid < SAI_ACL_ENTRY_ATTR_ACTION_START ||
+			if (md->objecttype == SAI_OBJECT_TYPE_ACL_ENTRY && md->isextensionattr)
```
kcudnik, line 1864: can you fix ident ?

bandaru-viswanath: Done! @kcudnik could you please check now ?

kcudnik (approved):

bandaru-viswanath:
@kcudnik Can this be merged in ?

PR #1179, opened 2021-01-06
PRBS RX State Data Type

JaiOCP:
Read Only attribute for getting Rx error count


kcudnik (approved):

inc/saiport.h
```diff
@@ -1501,6 +1501,14 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_PRBS_POLYNOMIAL,
 
+    /**
+     * @brief Port PRBS receive error count
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_PRBS_RX_ERROR_COUNT,
```
lguohan, line 1510: can this be in the port stat, rather than port attribute?

rlhui:
@marian-pritsak - would you like to review this for MLNX?

inc/saiport.h
JaiOCP: move to port stats
https://github.com/opencomputeproject/SAI/pull/1179/commits/bed90dd32bff2db5c88a0821a7ac2f050c199b79

inc/saiport.h
```diff
@@ -2290,6 +2290,9 @@ typedef enum _sai_port_stat_t
     /** PRBS Error Count */
     SAI_PORT_STAT_PRBS_ERROR_COUNT,
 
+    /** PRBS Receive Error Count */
+    SAI_PORT_STAT_PRBS_RX_ERROR_COUNT,
```
lguohan, line 2294: what is the existing SAI_PORT_STAT_PRBS_ERROR_COUNT?

JaiOCP: Removed this attribute
https://github.com/opencomputeproject/SAI/pull/1179/commits/ce4f8429862738e2707537cd3459a30a08912321

Also introduced the complex data type for handling atomic state as discussed in the meeting

inc/saitypes.h
```diff
@@ -391,6 +391,64 @@ typedef struct _sai_ip_prefix_t
...
+{
+    /** @validonly type == SAI_FLEX_DATA_TYPE_PRBS */
+    sai_port_prbs_rx_status_t rx_status;
+
+} sai_flex_status_t;
+
+/**
+ * @extraparam sai_flex_data_type_t type
+ */
+typedef union _sai_flex_data_t
```
kcudnik, line 434: why this is union ? and contains only 1 element ? are you planning to extend that ?

JaiOCP: Yes, thats the idea. This is a generic data type for atomic I/O and more fields will be added later as needed

kcudnik (approved):

rlhui:
@rck-innovium please help review/approve, thanks.

inc/saitypes.h
```diff
@@ -391,6 +391,64 @@ typedef struct _sai_ip_prefix_t
...
+{
+    /** @validonly type == SAI_FLEX_DATA_TYPE_PRBS */
+    uint32_t count;
+
+} sai_flex_data_t;
```
lguohan, line 439: we can define a specific attribute value type to combine prbs status and prbs error count in the new attribute value type.

JaiOCP: https://github.com/opencomputeproject/SAI/pull/1179/commits/65aec0b9d7be45f449561d25f1a8277bfff14bd3

Changed as per review comments

inc/saiport.h
```diff
@@ -1566,6 +1547,14 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_PRBS_RX_STATUS,
 
+    /**
+     * @brief Attribute data for #SAI_PORT_ATTR_PRBS_RX_STATE
```
lguohan, line 1551: add some comments here to explain why we introduce this attribute and not reuse existing attribute.

a simplified version of following para is good.

There are many “clear on read” states in hardware. Sometimes these are individual standalone fields. "Many a times HW combines multiple fields in a single atomic read. 
Given that SAI API attributes are defined at sub field granularity, SAI adapter has to emulate “clear on read” for sub field queries. Flex state complex data type allows read operation on the combination of such attributes. This makes SAI adapter implementation easy by just making a single query to hardware. Hardware natively takes care of clear on read for all the subfields with in the field.

JaiOCP: https://github.com/opencomputeproject/SAI/pull/1179/commits/97ed2bb8e433f7ce80e9e898c6c75c4d8eff9d06

Done

lguohan (approved):

mikeberesford (approved):

JaiOCP:
@rlhui Please merge

inc/saiport.h
```diff
@@ -1566,6 +1547,16 @@ typedef enum _sai_port_attr_t
...
+     * Used for clear on read status/count register.
+     * Adapter should return return SAI_STATUS_NOT_SUPPORTED if not supported.
```
rck-innovium, line 1553: typo: return return

rck-innovium (changes requested):
Hi Jai,

We need the below understanding to be documented:

----
There are multiple SAI attributes/stat-id (viz., SAI_PORT_ATTR_PRBS_RX_STATE,  SAI_PORT_ATTR_PRBS_RX_STATUS, SAI_PORT_PRBS_RX_STATUS_LOCK_WITH_ERRORS) for PORT PRBS status. It depends on the underlying ASIC as to which of these attributes are applicable. The suggestion was that the application would first query  SAI_PORT_ATTR_PRBS_RX_STATE. If the SAI implementation returns that the attribute is unsupported, the application would then fallback to query the existing SAI_PORT_ATTR_PRBS_RX_STATUS and SAI_PORT_PRBS_RX_STATUS_LOCK_WITH_ERRORS.

----

inc/saiport.h
JaiOCP: > Hi Jai,
> 
> We need the below understanding to be documented:
> 
> There are multiple SAI attributes/stat-id (viz., SAI_PORT_ATTR_PRBS_RX_STATE, SAI_PORT_ATTR_PRBS_RX_STATUS, SAI_PORT_PRBS_RX_STATUS_LOCK_WITH_ERRORS) for PORT PRBS status. It depends on the underlying ASIC as to which of these attributes are applicable. The suggestion was that the application would first query SAI_PORT_ATTR_PRBS_RX_STATE. If the SAI implementation returns that the attribute is unsupported, the application would then fallback to query the existing SAI_PORT_ATTR_PRBS_RX_STATUS and SAI_PORT_PRBS_RX_STATUS_LOCK_WITH_ERRORS.

Comments in header file are limited in characters.
We will add this detail in release notes

JaiOCP: > typo: return return
Fixed
https://github.com/opencomputeproject/SAI/pull/1179/commits/e276ffafbc70c59d7171dfe4ce5e22b90d25ddfa

rlhui:
> Hi Jai,
> 
> We need the below understanding to be documented:
> 
> There are multiple SAI attributes/stat-id (viz., SAI_PORT_ATTR_PRBS_RX_STATE, SAI_PORT_ATTR_PRBS_RX_STATUS, SAI_PORT_PRBS_RX_STATUS_LOCK_WITH_ERRORS) for PORT PRBS status. It depends on the underlying ASIC as to which of these attributes are applicable. The suggestion was that the application would first query SAI_PORT_ATTR_PRBS_RX_STATE. If the SAI implementation returns that the attribute is unsupported, the application would then fallback to query the existing SAI_PORT_ATTR_PRBS_RX_STATUS and SAI_PORT_PRBS_RX_STATUS_LOCK_WITH_ERRORS.

@rck-innovium  - do you agree/approve this PR?

rck-innovium:
> @rck-innovium - do you agree/approve this PR?

Yes. Jai has said that he will document the NOS expectation in the release nots.

rck-innovium (approved):

PR #1182, opened 2021-01-19
Do not call sai_metadata_sai get APIs if they are not allocated

ngoc-do:
New asic - fabric - doesn't have attributes like VLAN, HOSTIF... So Broadcom SAI will not allocate these APIs or function pointers of those API objects are not allocated. This causes  attempts that try to get/read these attributes to fail. So sai discovery for fabric will crash.

Solution: Modify `parse.pl` to verify sai_metadata_sai if they are allocated or not before trying to  attributes.

This PR will change only for getting attributes. Create, Set, Remove may also need to be changed, but this is out of the scope of this PR.

Signed-off-by: ngocdo <ngocdo@arista.com>

### DIFF
--- a/meta/parse.pl
+++ b/meta/parse.pl
@@ -2684,10 +2684,18 @@ sub ProcessGet
     }
     elsif (not defined $struct)
     {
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

ngoc-do:
All seem fine. Could you please merge it Kamil?

PR #1184, opened 2021-01-24
[saithrift]: add correct libs to build saithrift on libsaivs

lguohan:
Signed-off-by: Guohan Lu <lguohan@gmail.com>

### DIFF
--- a/test/saithrift/Makefile
+++ b/test/saithrift/Makefile
@@ -32,7 +32,12 @@ CTYPESGEN = /usr/local/bin/ctypesgen
 else
 CTYPESGEN = /usr/local/bin/ctypesgen.py
 endif
-LIBS = -lthrift -lpthread -lsai
+LIBS = -lthrift -lpthread
+ifeq ($(platform),vs)
+LIBS += -lsaivs -lsaimeta -lsaimetadata
+else
+LIBS += -lsai
+endif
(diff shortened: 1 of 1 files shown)


test/saithrift/Makefile
```diff
@@ -32,7 +32,12 @@ CTYPESGEN = /usr/local/bin/ctypesgen
 else
 CTYPESGEN = /usr/local/bin/ctypesgen.py
 endif
-LIBS = -lthrift -lpthread -lsai
+LIBS = -lthrift -lpthread
+ifeq ($(platform),vs)
+LIBS += -lsaivs -lsaimeta -lsaimetadata
```
kcudnik, line 37: this sounds wrong, why SAI would be dependent on saivs, saimeta and saimetadata ? those libs are generated ins sairedis repo

lguohan: this is for compile saiserver, saithrift.

kcudnik: yes, but it still seems wired to me that those libs are here as dependencies, because sairedis is directly dependent on SAI repo, seems to me like this saithrift should be separate repo with dependency on SAI and sairedis and not be located in SAI repo itself

kcudnik (approved):

test/saithrift/Makefile
lguohan: sai repo also host ptf test which requires saiserver,

PR #1186, opened 2021-01-27
Override VRF

JaiOCP:
New virtual router boolean attribute is introduced to indicate that this is an override VRF and overrides other VRs

### DIFF
--- a/doc/Override-VRF.md
+++ b/doc/Override-VRF.md
@@ -1,0 +1,89 @@
+Override VRF
+-------------------------------------------------------------------------------
+ Title       | SAI Override VRF
+-------------|-----------------------------------------------------------------
+ Authors     | Jai Kumar, Broadcom Inc.
+ Status      | In review
+ Type        | Standards track
+ Created     | 02/18/2021: Initial Draft
+ SAI-Version | TBD
+ 
+ 
+-------------------------------------------------------------------------------
+
+This spec talks about the override VRF usecase.
+
+SAI pipleine define L3 lookup block as an abstracted stage where route lookup is abstracted as a single operation. 
+SAI pipeline do not specify how many parallel L3 lookups a device can/may support.
+
+Typically devices support multiple parallel L3 lookup to provide policy based forwarding.
+Following usecase 1 is already covered in SAI spec. This spec adds support for usecase 2 and 3.
+
+New switch level default override VRF is introduced. 
+```
+    /**
+     * @brief Default SAI Override Virtual Router ID
+     *
+     * Must return #SAI_STATUS_OBJECT_IN_USE when try to delete this VR ID.
+     *
+     * @type sai_object_id_t
+     * @flags READ_ONLY
+     * @objects SAI_OBJECT_TYPE_VIRTUAL_ROUTER
+     * @default internal
+     */
+    SAI_SWITCH_ATTR_DEFAULT_OVERRIDE_VIRTUAL_ROUTER_ID,
+```
+
+
+### Usecase 1: VRF Faillback
+In this case 2 parrallel lookups are performed
+> RIF/ACL -> VRF -> [VRF, DIP] -> Result_VRF
+RIF/ACL -> Deafault_VRF -> [Default_VRF, DIP] -> Result_Default_VRF
+If Result_VRF == HIT then it takes precedence over Result_Default_VRF.
+
+
+### Usecase 2: Override VRF
+In this case 2 parallel lookups are performed
+>RIF/ACL->VRF->[VRF, DIP] -> Result_VRF
+Switch->Override_VRF->[Override_VRF, DIP] -> Result_Override_VRF
+If Result_Override_VRF == HIT then it takes precedence over Result_VRF.
+
+
+### Usecase 3: Override VRF and VRF Fallback
+In this case 3 parallel lookups are performed
+>Switch->Override_VRF->[Override_VRF, DIP] -> Result_Override_VRF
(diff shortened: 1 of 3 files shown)


kcudnik (approved):

inc/saiswitch.h
```diff
@@ -683,6 +683,18 @@ typedef enum _sai_switch_attr_t
...
+     * @brief Default SAI Override Virtual Router ID
+     *
+     * Must return #SAI_STATUS_OBJECT_IN_USE when try to delete this VR ID.
+     *
+     * @type sai_object_id_t
+     * @flags READ_ONLY
+     * @objects SAI_OBJECT_TYPE_VIRTUAL_ROUTER
+     * @default internal
+     */
+    SAI_SWITCH_ATTR_DEFAULT_OVERRIDE_VIRTUAL_ROUTER_ID,
```
kcudnik, line 696: does this mean that this is equal to default virtual router ? or this is new virtual router and by default there will be 2 of them ?

JaiOCP: By default there will be two VR

lguohan: what is the motivation to introduce this new override virtual router id?

reshmaintel: Is it possible to add a capability check to see if the underlying ASIC/SDK supports the feature - using the sai_acl_stage_t SAI_ACL_STAGE_PRE_INGRESS during staging

JaiOCP: > what is the motivation to introduce this new override virtual router id?

I have added the use case document explaining
https://github.com/opencomputeproject/SAI/pull/1186/commits/bb633a2421a63cb95c7e3d34296c3f422bc642c9

JaiOCP: > Is it possible to add a capability check to see if the underlying ASIC/SDK supports the feature - using the sai_acl_stage_t SAI_ACL_STAGE_PRE_INGRESS during staging

There are no ACL stages for this change. Override VRF is not supported by SAI adapter will return "ERROR" on the read of this attribute by application.

JaiOCP:
@marian-pritsak @dgsudharsan Please help review

doc/Override-VRF.md
```diff
@@ -0,0 +1,89 @@
...
+     * @type sai_object_id_t
+     * @flags READ_ONLY
+     * @objects SAI_OBJECT_TYPE_VIRTUAL_ROUTER
+     * @default internal
+     */
+    SAI_SWITCH_ATTR_DEFAULT_OVERRIDE_VIRTUAL_ROUTER_ID,
+```
+
+
+### Usecase 1: VRF Faillback
```
mikeberesford, line 38: typo: Faillback should be Fallback

mikeberesford (approved):
Thanks Jai. This looks like what what discussed in the meeting.

inc/saiswitch.h
reshmaintel: /**
--
  | * @brief List of   actions in sai_acl_action_type_t
  | *
  | * Based on the   ACL capability per stage obtained from the switch
  | * attributes   #SAI_SWITCH_ATTR_ACL_STAGE_INGRESS and #SAI_SWITCH_ATTR_ACL_STAGE_EGRESS
  | * application   should pass the action list if its mandatory per stage.
  | * Pass the   action list if its mandatory per stage.
  | * If its not   mandatory application can either pass the action list
  | * or ignore it.
  | *
  | * @type   sai_s32_list_t sai_acl_action_type_t
  | * @flags   CREATE_ONLY
  | * @default empty
  | */
  | SAI_ACL_TABLE_ATTR_ACL_ACTION_TYPE_LIST,
  |  
This one is for list of actions supported in each stage. You could make a query here for STAGE_PRE_INGRESS and get SET_VRF as action or a NULL list in which case it cannot support this feature. This is for CREATE only, should be extended to get the value to check capability exists.

doc/Override-VRF.md
```diff
@@ -0,0 +1,89 @@
...
+If Result_Override_VRF == HIT then it takes precedence over Result_VRF.
+
+
+### Usecase 3: Override VRF and VRF Fallback
+In this case 3 parallel lookups are performed
+>Switch->Override_VRF->[Override_VRF, DIP] -> Result_Override_VRF
+RIF/ACL -> VRF -> [VRF, DIP] -> Result_VRF
+RIF/ACL -> Deafault_VRF -> [Default_VRF, DIP] -> Result_Default_VRF
+If Result_Override_VRF == HIT then it takes precedence over Result_VRF and Result_Default_VRF.
+If Result_Override_VRF == HIT then it takes precedence over Result_VRF.
```
lguohan, line 58: can you check these two lines? it seems they are the same if condition.

doc/Override-VRF.md
JaiOCP: https://github.com/opencomputeproject/SAI/pull/1186/commits/9d9b853826af813555ed435ec6af1c761b3c3c45

doc/Override-VRF.md
JaiOCP: https://github.com/opencomputeproject/SAI/pull/1186/commits/9d9b853826af813555ed435ec6af1c761b3c3c45

JaiOCP:
@lguohan @reshmaintel @rlhui Please approve

lguohan (approved):

marian-pritsak (approved):

reshmaintel (approved):

PR #1187, opened 2021-01-29
New attribute to start fw broadcast

SidharajU:
A new attribute is added to start the FW broadcast. This is necessary to start FW broadcast on all PHY chips

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -2266,6 +2266,18 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_TUNNEL_LOOPBACK_PACKET_ACTION,
 
     /**
+     * @brief Start Firmware Broadcast
+     *
+     * Configure PHY to start Firmware Broadcast.
+     *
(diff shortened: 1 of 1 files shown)


inc/saiswitch.h
```diff
@@ -2265,6 +2265,18 @@ typedef enum _sai_switch_attr_t
...
+     * @brief Start Firmware Broadcast
+     *
+     * Configure PHY to start Firmware Broadcast.
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     * @validonly SAI_SWITCH_ATTR_FIRMWARE_DOWNLOAD_BROADCAST == true
+     */
+    SAI_SWITCH_ATTR_FIRMWARE_BROADCAST_START,
```
kcudnik, line 2278: _START and _END ending names for attributes is reserved for ranges, current logic is expecting same attribute with _END naming, maybe change attr name ?

SidharajU:
Will create a new one

PR #1190, opened 2021-02-03
[meta] Add support for allow empty list tag

kcudnik:


itaibaz:
Hi, the current commit for allowempty infrastructure looks fine
The question is, whether we also want to go over all SAI header files, and mark needed list with allowemptry true
For example, SAI_PORT_ATTR_INGRESS_MIRROR_SESSION
This will require going over each list in header files, and considering the use case, and probably might require feedback from more community members about use case
This can be done in this commit, or as a separate PR, which you can lead or I can lead, let me know what you think

kcudnik:
that depends should be default value, currently when "default empty" is present automatically allowempty will be set to true, otherwise it will require explicit tag

itaibaz:
I'm not sure it's correct to allow all lists to be empty
I quickly went over SAI headers, and saw several examples

For example, 
    /**
     * @brief Hardware Lane list
     *
     * @type sai_u32_list_t
     */
    SAI_PORT_ATTR_HW_LANE_LIST,
Seems when you create a port you always have to supply at least one HW lane

Another example :
    /**
     * @brief Destination/Analyzer/Monitor Port List.
     *
     * @type sai_object_list_t
     * @objects SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_LAG, SAI_OBJECT_TYPE_SYSTEM_PORT
     * @condition SAI_MIRROR_SESSION_ATTR_MONITOR_PORTLIST_VALID == true
     */
    SAI_MIRROR_SESSION_ATTR_MONITOR_PORTLIST,
if port list is valid, I think list should contain at least one object

Another example : 
    /**
     * @brief UDF Mask
     *
     * Default to 2 bytes, value 0xFF, 0xFF
     *
     * The count in the list must be equal to the UDF byte length.
     * The mask only applies to extracted UDF when it is used for hash,
     * it does not apply to the extracted UDF when it is used for ACL.
     *
     * @type sai_u8_list_t
     * @flags CREATE_AND_SET
     * @default const
     */
    SAI_UDF_ATTR_HASH_MASK,
Seems the mask always is not empty (as empty mask would mean zero length UDF which is just not functional)

I think for backward compatibility, we can use default true, if allow empty is not specified, so we don't "lock" any existing behavior
But still, we do need to "lock" some specific lists where allow empty just doesn't make sense (or at least, to me...)

kcudnik:
yes, but those attributes by default have allowemtpty=false

itaibaz:
so my point was -
either we have a default of false if allownull is not present (same as today), and explicitly open it for several attributes where it's needed. This can be done in a separate PR. Even without PR, it's today's situation which is not correct but seems nobody was using it up until now, so wouldn't break any SAI vendor.
or we have a default of true if not present, and block very few specific attribs
There are more attribs we will need to open in first approach than attribs we will need to lock in second approach, but I am fine with both, and again, I can lead that in separate PR

kcudnik:
ok agree

itaibaz (approved):

itaibaz:
Thanks, so all look fine
After this is merged, I will review all list metadata and suggest any PR if I think something is needed

PR #1193, opened 2021-02-18
Class-Based Forwarding

j-bos:
This PR defines class-based forwarding. It contains two aspects:

- Assignment of a Forwarding Class to a packet, via QOS map or ACL

- New next-hop group type: SAI_NEXT_HOP_GROUP_TYPE_CLASS_BASED

Where member selection is based on the forwarding-class of the packet.

Signed-off-by: j-bos <jbos@cisco.com>


rlhui:
@JaiOCP , @marian-pritsak , @ashutosh-agrawal - would you like to review/sign-off this?  Thanks.

rlhui:
@j-bos  - thanks. possible to also upload the ppt to https://github.com/opencomputeproject/SAI/tree/master/doc/ECMP?

rlhui:
@j-bos  - please check the DCO error if this can be resolved.

inc/saiqosmap.h
```diff
@@ -77,6 +77,12 @@ typedef enum _sai_qos_map_type_t
     /** QOS Map to set traffic class and color to EXP */
     SAI_QOS_MAP_TYPE_TC_AND_COLOR_TO_MPLS_EXP = 0x0000000c,
 
+    /** QOS Map to set DSCP to Forwarding class */
```
rck-innovium, line 80: Where are the bindpoints for these maps defined? We need it at a switch, port, RIF or tunnel etc. Today, we need something like SAI_PORT_ATTR_QOS_DSCP_TO_TC_MAP and SAI_SWITCH_ATTR_QOS_MPLS_EXP_TO_TC_MAP

rck-innovium (changes requested):

rck-innovium:
Since we do not have an explicit capability check,  can you spell out the mechanism to know whether a SAI implementation supports this capability. Say, if SAI_SWITCH_ATTR_MAX_NUMBER_OF_FORWARDING_CLASSES is 0?

JaiOCP (commented):
We need to enhance the SAI behavioral model to accommodate this change.

Concept of Internal Class as an index in ECMP Group need to come from ACL table.
Today Next Hop Group block has a match only on next hop group hash value. This need to be explicitly enhanced to include class id.

JaiOCP (commented):
Raising this question to capture in review.

Given the concept of class id based ECMP group indexing is a specific implementation  to silicon and is not RFC based: Should it be not part of SAI extensions?

inc/saiqosmap.h
j-bos: Yes, should be switch and port. I will add it.

j-bos:
> Since we do not have an explicit capability check, can you spell out the mechanism to know whether a SAI implementation supports this capability. Say, if SAI_SWITCH_ATTR_MAX_NUMBER_OF_FORWARDING_CLASSES is 0?

The most direct would be an enum-capability for next-hop group type (class-based). Though, I think 0 for forwarding-class would be reasonable as well.

rlhui:
@lguohan  - would you like to review? Thanks.

inc/sainexthopgroup.h
```diff
@@ -263,7 +276,8 @@ typedef enum _sai_next_hop_group_member_attr_t
      *
      * Index specifying the strict member's order.
      * Allowed value range for is from 0 to SAI_NEXT_HOP_GROUP_ATTR_REAL_SIZE - 1.
-     * Should only be used if the type of owning group is SAI_NEXT_HOP_GROUP_TYPE_FINE_GRAIN_ECMP.
+     * Should only be used if the type of owning group is SAI_NEXT_HOP_GROUP_TYPE_FINE_GRAIN_ECMP
```
TACappleman, line 279: Should also extend SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_ID to support pointing to objects of type SAI_OBJECT_TYPE_NEXT_HOP_GROUP (currently can only point to a single next hop)

inc/saiqosmap.h
```diff
@@ -77,6 +77,12 @@ typedef enum _sai_qos_map_type_t
...
+    SAI_QOS_MAP_TYPE_DSCP_TO_FORWARDING_CLASS = 0x0000000d,
```
TACappleman, line 81: Also need FC added to sai_qos_map_params_t in saitypes.h

j-bos:
Signed-off-by: j-bos <jbos@cisco.com>

inc/saiswitch.h
```diff
@@ -2531,6 +2531,42 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_QOS_MAP
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_SWITCH_ATTR_QOS_MPLS_EXP_TO_FORWARDING_CLASS_MAP,
+
     /**
```
JaiOCP, line 2570: Please add CRM details and example in the doc

rlhui:
@marian-pritsak , please also help review, thanks.

j-bos:
> @j-bos - please check the DCO error if this can be resolved.

Resolved

kcudnik:
please resolve conflicts

inc/saiport.h
```diff
@@ -1965,6 +1965,34 @@ typedef enum _sai_port_attr_t
...
+     * Map id = #SAI_NULL_OBJECT_ID to disable map on port.
+     * Default no map.
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_QOS_MAP
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_PORT_ATTR_QOS_MPLS_EXP_TO_FORWARDING_CLASS_MAP,
```
kcudnik, line 1994: attributes not added at the end of attribute list

j-bos: Fixed

doc/ECMP/Class-Based-Forwarding.md
```diff
@@ -0,0 +1,177 @@
...
+     *
+     *  @type  sai_acl_action_data_t  sai_uint8_t
+     *  @flags  CREATE_AND_SET
+     *  @default  disabled
+     */
+    SAI_ACL_ENTRY_ATTR_ACTION_SET_FORWARDING_CLASS,
+...
+}  sai_acl_entry_attr_t;
+```
+
```
j-bos, line 109: From OCP call: Add more detail on how FC is derived

doc/ECMP/Class-Based-Forwarding.md
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * @brief Forwarding-class to index map
+     *
+     * @type sai_map_list_t
+     * @flags CREATE_AND_SET
+     * @default empty
+     * @validonly SAI_NEXT_HOP_GROUP_ATTR_TYPE == SAI_NEXT_HOP_GROUP_TYPE_CLASS_BASED
+     */
+    SAI_NEXT_HOP_GROUP_ATTR_FORWARDING_CLASS_TO_INDEX_MAP,
+```
```
j-bos, line 39: From OCP call: Use separate map object to enable CRM

ashutosh-agrawal: Also, it would be a good idea to define capability attributes for the number of forwarding classes and number of indices supported by the SAI implementation.

abdosi:
@j-bos Are you working on update using new object ?

j-bos:
> @j-bos Are you working on update using new object ?

I have added some changes for that and can accept review comments. I am working on updating the documentation now.

TACappleman (approved):

doc/ECMP/Class-Based-Forwarding.md
```diff
@@ -0,0 +1,204 @@
...
+ attrs.clear();
+
+ sai_object_id_t nh_group;
+
+ attr.id = SAI_NEXT_HOP_GROUP_ATTR_TYPE;
+ attr.value.u32 = SAI_NEXT_HOP_GROUP_TYPE_CLASS_BASED;
+ attrs.push_back(attr);
+
+ attr.id = SAI_NEXT_HOP_GROUP_ATTR_CONFIGURED_SIZE;
+ attr.value.u32 = num_forwarding_classes;
```
abanu-ms, line 167: Could we change the `num_forwarding_classes` to something else that would make it more obvious that is rather the number of members that we have to fill in, not the number of entries in the map?

At a first read I was confused as why that SAI attribute was reused like that before I remembered the example has a 1-1 mapping. Other first time readers might be confused by this as well.

Also, because of this, I think a better example would contain a mapping between `X` FCs and `Y` NHs, where `X > Y` just so we make it obvious that that is possible.

doc/ECMP/Class-Based-Forwarding.md
```diff
@@ -0,0 +1,204 @@
...
+ for (int i = 0; i < num_forwarding_classes; ++i) {
+     fc_map[i].key = i;
+     fc_map[i].value = i;
+ }
+
+ sai_map_list_t fc_map_list;
+ fc_map_list.key.count = num_forwarding_classes;
+ fc_map_list.key.list = fc_map;
+
+ attr.type = SAI_NEXT_HOP_GROUP_MAP_ATTR_TYPE;
```
abanu-ms, line 141: It's `attr.id`.

doc/ECMP/Class-Based-Forwarding.md
abanu-ms: Also, the `@validonly` flag of `SAI_NEXT_HOP_GROUP_ATTR_CONFIGURED_SIZE` states `SAI_NEXT_HOP_GROUP_ATTR_TYPE == SAI_NEXT_HOP_GROUP_TYPE_FINE_GRAIN_ECMP`. So either we need to update the attribute's comment, or it was miss-used here.

doc/ECMP/Class-Based-Forwarding.md
j-bos: Fixed.

doc/ECMP/Class-Based-Forwarding.md
j-bos: Added documentation of switch attribute for max forwarding classes.

j-bos: Implemented separate map object.

abanu-ms (approved):

inc/sainexthopgroup.h
```diff
@@ -311,6 +327,50 @@ typedef enum _sai_next_hop_group_member_attr_t
...
+     * @brief Start of attributes
+     */
+    SAI_NEXT_HOP_GROUP_MAP_ATTR_START,
+
+    /**
+     * @brief Next hop group map type
+     *
+     * @type sai_next_hop_group_map_type_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
```
abanu-ms, line 349: `@isresourcetype true` markdown needs to be added in order to query the switch's object availability.

abanu-ms (changes requested):

inc/sainexthopgroup.h
j-bos: I think it can be a resource regardless, as the ``@isresourcetype`` flag is to flag a specific attribute as separating the object into multiple resource pools.

I'm not sure what types could be added in the future, but I can add that tag if we think it is probable to have different resource pools for those on some implementations.

abanu-ms: I'm not 100% sure, but while writing some swss code which makes use of the NHG maps, I couldn't query the switch object availability for NHG maps until I've added that flag. The SAI metadata validation was throwing an error saying that the queried object was not a resource type.

PR #1194, opened 2021-02-19
Add source/dest/double NAPT entry available attributes

kperumalbfn:
Add attributes for Source/Destination/Double NAPT entries in saiswitch

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -2266,6 +2266,30 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_TUNNEL_LOOPBACK_PACKET_ACTION,
 
     /**
+     * @brief Available SNAPT entries
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_AVAILABLE_SNAPT_ENTRY,
+
+    /**
+     * @brief Available DNAPT entries
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_AVAILABLE_DNAPT_ENTRY,
+
+    /**
+     * @brief Available Double NAPT entries
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_AVAILABLE_DOUBLE_NAPT_ENTRY,
+
+    /**
      * @brief End of attributes
      */
     SAI_SWITCH_ATTR_END,
--- a/meta/acronyms.txt
+++ b/meta/acronyms.txt
@@ -143,3 +143,7 @@ I2C - I squared C
 CPLD - Complex programmable logic device
 EEPROM - Electrically erasable programmable read-only memory
 PRBS - Pseudorandom binary sequence
+NAPT - Network Address Port Translation
+SNAPT - Source Network Address Port Translation
+DNAPT - Destination Network Address Port Trasnlation
+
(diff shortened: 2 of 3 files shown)


kcudnik:
```
WARNING: Word 'DNAPT' is misspelled or is acronym, add to acronyms.txt? saiswitch.h 2277:     * @brief Available DNAPT entries
WARNING: Word 'NAPT' is misspelled or is acronym, add to acronyms.txt? saiswitch.h 2285:     * @brief Available Double NAPT entries
WARNING: Word 'SNAPT' is misspelled or is acronym, add to acronyms.txt? saiswitch.h 2269:     * @brief Available SNAPT entries
```
please add words to acronyms.txt

kcudnik (approved):

rlhui:
@JaiOCP , @marian-pritsak or @mattyk  - please start reviewing this at earliest. Thanks.

rck-innovium (approved):

JaiOCP (approved):

rlhui:
We might need a similar attribute for SAI_NAT_TYPE_DESTINATION_NAT_POOL as well?

inc/saiswitch.h
```diff
@@ -2265,6 +2265,30 @@ typedef enum _sai_switch_attr_t
...
+
+    /**
+     * @brief Available Double NAPT entries
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_AVAILABLE_DOUBLE_NAPT_ENTRY,
+
     /**
```
rlhui, line 2292: Since in sainat.h or nat pipeline model, it's not using the term SNAPT, DNAPT, or DOUBLE_NAPT, but just SNAT, DNAT, or DOUBLE_NAT, would be more clear to use the same terms.

kperumalbfn: @rlhui Attributes to query Available SNAT, DNAT, DOUBLE_NAT entries are already part of saiswitch. These new attributes are for port-based NAT entries. Some hardware use separate table for NAPT entries and these new attributes can be used to query available entries for port-based SNAT/DNAT/Double_nat entries.

kperumalbfn:
> We might need a similar attribute for SAI_NAT_TYPE_DESTINATION_NAT_POOL as well?

@rlhui We can add separate attribute for DNAT_POOL as well.

inc/saiswitch.h
rlhui: In the NAT behavioral model, the "NAT" table is actually NAPT table.  Based on the model, even if the hardware has two separate tables, one for NAT, one for NAPT, then if one queries the size of the table, it should return the "NAPT" table size.
Do we have any existing example today, that we have two attributes to query different HW portions of one logical table in the SAI behavioral model?

kperumalbfn: @rlhui In most of the cases, ASIC will have symmetric entry sized tables across set of hardware tables for a single logical table.

For NAT,
- For simple address-only translation(Basic NAT), most applications might need only smaller table with minimum number of translations.
- For port based translation(NAPT), applications might need larger entries in hardware.

These new set of switch attributes will query different counters and that will be useful for applications to determine maximum entries for each of the NAT type.

I see one other similar case for V4/V6 routes for which SAI doesn't have separate attributes. Currently SAI has one attribute(SAI_SWITCH_ATTR_AVAILABLE_IPV4_ROUTE_ENTRY) for available V4 route entries. But in most of the ASICs, there will be separate tables for host and LPM entries. We might need separate SAI counter for this. Same for V6 as well.

rlhui (approved):

PR #1195, opened 2021-02-19
Add switch attributes for hash offset configuration.

mikeberesford:
Add ability to specify the hash offset (rotation) in addition to the
hash seed.

Signed-off-by: Mike Beresford <mberes@google.com>

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -1169,7 +1169,12 @@ typedef enum _sai_switch_attr_t
     /**
      * @brief SAI ECMP default hash offset
      *
-     * @type sai_uint32_t
+     * When set, the output of the ECMP hash calculation will be shifted or
+     * rotated by the specified number of bits. The direction and exact behavior
+     * (shift vs offset) as well as the maximum rotation supported may be
+     * ASIC-specific.
+     *
+     * @type sai_uint8_t
      * @flags CREATE_AND_SET
      * @default 0
      */
@@ -1243,7 +1248,12 @@ typedef enum _sai_switch_attr_t
     /**
      * @brief SAI LAG default hash offset
      *
-     * @type sai_uint32_t
+     * When set, the output of the LAG hash calculation will be shifted or
+     * rotated by the specified number of bits. The direction and exact behavior
+     * (shift vs offset) as well as the maximum rotation supported may be
+     * ASIC-specific.
+     *
+     * @type sai_uint8_t
      * @flags CREATE_AND_SET
      * @default 0
      */
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -1169,10 +1169,8 @@ typedef enum _sai_switch_attr_t
     /**
      * @brief SAI ECMP default hash offset
      *
-     * When set, the output of the ECMP hash calculation will be shifted or
-     * rotated by the specified number of bits. The direction and exact behavior
-     * (shift vs offset) as well as the maximum rotation supported may be
-     * ASIC-specific.
+     * When set, the output of the ECMP hash calculation will be rotated right
+     * by the specified number of bits.
      *
      * @type sai_uint8_t
      * @flags CREATE_AND_SET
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

mikeberesford:
@JaiOCP @lguohan would you take a look and see if this covers what we discussed today (and help tag others that should review?)

mikeberesford:
@rlhui are there others I should be tagging for review here?

rlhui:
> @rlhui are there others I should be tagging for review here?

@marian-pritsak , @ashutosh-agrawal - please help review as well? Thanks.

lguohan (approved):

JaiOCP (approved):

inc/saiswitch.h
```diff
@@ -1166,6 +1166,15 @@ typedef enum _sai_switch_attr_t
...
     SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED,
 
+    /**
+     * @brief SAI ECMP default hash offset
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_OFFSET,
```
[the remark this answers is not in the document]
mikeberesford: I believe "default" in this context is mean to imply that it's a global setting, where the algorithm and seed are configured per-switch but the extracted fields may be different for the configurable hashes.

http://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Proposal-13-Hash-UDF.md#hash
> Global hash algorithm and seed are used to calculate the hash value based on the extracted fields, i.e, SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_ALGORITHM, SAI_SWITCH_ATTR_LAG_DEFAULT_HASH_ALGORITHM, SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED, SAI_SWITCH_ATTR_LAG_DEFAULT_HASH_SEED.

inc/saiswitch.h
```diff
@@ -1166,6 +1166,20 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED,
 
+    /**
+     * @brief SAI ECMP default hash offset
+     *
+     * When set, the output of the ECMP hash calculation will be shifted or
+     * rotated by the specified number of bits. The direction and exact behavior
+     * (shift vs offset) as well as the maximum rotation supported may be
+     * ASIC-specific.
```
konne88, line 1175: This flexibility in the vendor implementation makes it hard to use SAI implementations from different vendors. Why can't we be more prescriptive? 

If there are truly vendors that implements this as a shift vs rotation, can we have two fields SHIFT_OFFSET and ROTATION_OFFSET and an API to query which one is supported? 

Rotation direction shouldn't matter how it's implemented in hardware, because a SAI implementation can always do field_width - shift_length to flip the rotation.

mikeberesford: We discussed this a bit in the community meeting, and there was some reluctance to specify the exact ASIC behavior. I think there are probably some good reasons to specify the behavior here (particularly to prevent polarization in mixed-ASIC networks). I'd personally be on board with defining this as "rotate right by N bits" but would need the ASIC vendors to weigh in. @JaiOCP @marian-pritsak @ashutosh-agrawal @lguohan any comments?

rlhui (approved):

inc/saiswitch.h
ashutosh-agrawal: I agree with @konne88 here. We can define the exact behavior for the "offset" attribute here and other implementations like "shift" etc can be added later as additional attributes.

mikeberesford: Updated to specify right rotation.

inc/saiswitch.h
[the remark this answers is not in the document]
mikeberesford: @JaiOCP any more questions here?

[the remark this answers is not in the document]
JaiOCP: No more, thanks

ashutosh-agrawal (approved):

mikeberesford:
@lguohan any objections to the revised wording? If not, I think this can be merged.

Issue #1203, opened 2021-03-12
Add new port type for Recycle port


SarathBug:
It is helpful in certain scenarios for packets that have gone through a full switch pipeline to be re-injected into the beginning (Ingress) of the pipeline. SONiC on a disaggregated VOQ chassis uses this capability. There are other examples. Currently this can only be done via SAI by putting a port in loopback configuration. But this takes a port away from regular network traffic. What is needed is a dedicated Recycle port. Such a port must be treated like a regular network port for all ASIC programming purposes (routing, bridging, ACL, queuing, tunnels, mirroring …). The only exception would be that it does not need to support some MAC layer functions.

PR #1204, opened 2021-03-15
Added new port type for Recycle Port

SarathBug:
For https://github.com/opencomputeproject/SAI/issues/1203

It is helpful in certain scenarios for packets that have gone through a full switch pipeline to be re-injected into the beginning (Ingress) of the pipeline. SONiC on a disaggregated VOQ chassis uses this capability. There are other examples. Currently this can only be done via SAI by putting a port in loopback configuration. But this takes a port away from regular network traffic. What is needed is a dedicated Recycle port. Such a port must be treated like a regular network port for all ASIC programming purposes (routing, bridging, ACL, queuing, tunnels, mirroring …). The only exception would be that it does not need to support some MAC layer functions.

Signed-off-by: SarathBug <sarath.gadagottu@broadcom.com>


rlhui:
@JaiOCP , @marian-pritsak , @lguohan , @ashutosh-agrawal  - please feel free to review, thanks.

rlhui:
@SarathBug - would we please also provide a doc for this as discussed, thanks?

rlhui:
Adding more reviewers for this, thanks. - @eswaran, @skeesara-nokia , @michaelli10

inc/saiport.h
```diff
@@ -47,6 +47,9 @@ typedef enum _sai_port_type_t
     /** Fabric Port */
     SAI_PORT_TYPE_FABRIC,
 
+    /** Recycle Port */
+    SAI_PORT_TYPE_RECYCLE,
+
```
skeesara-nokia, line 52: Is it more appropriate to call the SAI_PORT_TYPE_ETHERNET_RECYCLE ? This would make it clear that we are not passing meta data from the previous pass of the packet through the ASIC pipeline.

If a capability to pass meta data was to be standardized in the future we could call it a different type of recycle port.

skeesara-nokia (approved):

inc/saiport.h
SarathBug: SAI_PORT_TYPE_RECYCLE is chosen to be generic and we could a read only attribute to get the current headers on recycle port, ETH being the default.

skeesara-nokia: This sounds good to me

michaelli10: I agree with keeping the recycle port type generic and adding a separate "header mode" attribute.

michaelli10 (approved):

rlhui (approved):

inc/saiport.h
```diff
@@ -364,6 +367,16 @@ typedef enum _sai_port_connector_failover_mode_t
     SAI_PORT_CONNECTOR_FAILOVER_MODE_SECONDARY
 } sai_port_connector_failover_mode_t;
 
+/**
+ * @brief Attribute data for #SAI_PORT_ATTR_RECYCLE_PORT_HEADER_MODE
+ */
+typedef enum _sai_port_recycle_port_header_mode_t
+{
+    /** Ethernet header mode */
+    SAI_PORT_RECYCLE_PORT_HEADER_MODE_ETH,
```
rlhui, line 376: This additional header_mode for recycle port probably needs more reviews/discussion.  I'm not sure if it's needed now.

inc/saiport.h
```diff
@@ -1707,6 +1720,14 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_SYSTEM_PORT,
 
+    /**
+     * @brief Header mode for recycle port
+     *
+     * @type sai_port_recycle_port_header_mode_t
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_RECYCLE_PORT_HEADER_MODE,
```
rck-innovium, line 1729: @validonly SAI_PORT_ATTR_TYPE==SAI_PORT_TYPE_RECYCLE

kcudnik:
@default flag is required on craete_only attrbiute

JaiOCP (commented):
@SarathBug 
How would we differentiate if there are multiple types of recycle port?

SarathBug:
@rlhui , @skeesara-nokia , @JaiOCP 

Removed the attribute SAI_PORT_ATTR_RECYCLE_PORT_HEADER_MODE which was added initially.  

Reasons:
1. The attribute is readonly.  HW lane is not applicable for recycle port, so CREATE_ONLY or  MANDATORY_ON_CREATE | CREATE_AND_SET cannot be used here.  Does not really help if it is not writeable.
2. Could not enforce check through metadata e.g valid only for SAI_PORT_TYPE_RECYCLE because attribute SAI_PORT_ATTR_TYPE is READ_ONLY and \@validonly check is not applicable for READ_ONLY attributes.

The PR just adds SAI_PORT_TYPE_RECYCLE in sai_port_type_t

inc/saiport.h
```diff
@@ -1707,6 +1720,14 @@ typedef enum _sai_port_attr_t
...
+
```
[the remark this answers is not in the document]
SarathBug: This attribute SAI_PORT_ATTR_RECYCLE_PORT_HEADER_MODE is removed from the change.  Comment is no longer valid.

PR #1205, opened 2021-03-16
User defined VxLAN UDP source port calculation

marian-pritsak:


inc/saiswitch.h
```diff
@@ -490,6 +495,32 @@ typedef enum _sai_switch_tunnel_attr_t
      */
     SAI_SWITCH_TUNNEL_ATTR_VXLAN_UDP_SPORT,
 
+    /**
+     * @brief Tunnel UDP source port base for a calculated hash range
+     *
+     * Value of the fixed portion of the UDP source port's MSB
```
itaibaz, line 501: perhaps would be more clear to write most significant bits in plural as MSB is most significant bit
same for LSB below

inc/saiswitch.h
```diff
@@ -490,6 +495,32 @@ typedef enum _sai_switch_tunnel_attr_t
...
+     * @validonly SAI_SWITCH_TUNNEL_ATTR_TUNNEL_TYPE == SAI_TUNNEL_TYPE_VXLAN and SAI_SWITCH_TUNNEL_ATTR_TUNNEL_VXLAN_UDP_SPORT_MODE == SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_USER_DEFINED_HASH
+     */
+    SAI_SWITCH_TUNNEL_ATTR_VXLAN_UDP_SPORT_BASE,
+
+    /**
+     * @brief Tunnel UDP source port hash bit length
+     *
+     * Number of LSB reserved for the calculated hash value
+     *
+     * @type sai_uint16_t
```
itaibaz, line 516: probably can be uint8_t instead of uint16_t, as we need to describe a value between 0 to 16

itaibaz (commented):
should we also add per tunnel attrib ?
as original attributes are both per {switch, tunnel type} and {tunnel}
otherwise, we may need to specify this mode is not per tunnel

JaiOCP:
Have you looked at PR#1173. It added following attributes
/**
 * @brief Defines VXLAN tunnel UDP source port mode
 */
typedef enum _sai_tunnel_vxlan_udp_sport_mode_t
{
    /**
     * @brief User define value
     */
    SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_USER_DEFINED,

    /**
     * @brief RFC6335 Computed hash value in range 49152-65535
     */
    SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_EPHEMERAL,
} sai_tunnel_vxlan_udp_sport_mode_t;

marian-pritsak:
> should we also add per tunnel attrib ?
> as original attributes are both per {switch, tunnel type} and {tunnel}
> otherwise, we may need to specify this mode is not per tunnel

It doesn't make sense at the moment to have it per tunnel. If anyone will come up with the use case, it can be added.

marian-pritsak:
> Have you looked at PR#1173. It added following attributes
> /**
> 
> * @brief Defines VXLAN tunnel UDP source port mode
>   _/
>   typedef enum _sai_tunnel_vxlan_udp_sport_mode_t
>   {
>   /_*
>   
>   * @brief User define value
>     */
>     SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_USER_DEFINED,
>   
>   /**
>   
>   * @brief RFC6335 Computed hash value in range 49152-65535
>     */
>     SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_EPHEMERAL,
>     } sai_tunnel_vxlan_udp_sport_mode_t;

"Ephemeral" is not flexible enough

marian-pritsak:
> > should we also add per tunnel attrib ?
> > as original attributes are both per {switch, tunnel type} and {tunnel}
> > otherwise, we may need to specify this mode is not per tunnel
> 
> It doesn't make sense at the moment to have it per tunnel. If anyone will come up with the use case, it can be added.

I misunderstood a comment - yes, it should be in both since the enum is shared

inc/saiswitch.h
```diff
@@ -354,6 +354,11 @@ typedef enum _sai_tunnel_vxlan_udp_sport_mode_t
      * @brief RFC6335 Computed hash value in range 49152-65535
      */
     SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_EPHEMERAL,
+
+    /**
+     * @brief User defined bit range of computed hash
+     */
+    SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_USER_DEFINED_HASH,
```
lguohan, line 361: can be merged into user_defined mode

marian-pritsak: Done

rlhui:
@JaiOCP  - would you like to review/approve?

lguohan (approved):

kcudnik (approved):

JaiOCP:
> @JaiOCP - would you like to review/approve
Changes looks good

JaiOCP (approved):

PR #1207, opened 2021-03-22
Update SAI version to v1.8.0

rlhui:


rlhui:
Update SAI version to v1.8.0

PR #1215, opened 2021-04-02
Fix the Backward compatibility changes in MACSEC attribute

abdosi:
Why I did:

PR https://github.com/opencomputeproject/SAI/pull/1169  the backward compatibility of attribute `SAI_MACSEC_SA_ATTR_XPN`  is not handled correctly. Needed for v1.8.0

How I did

Assign `SAI_MACSEC_SA_ATTR_XPN` to `SAI_MACSEC_SA_ATTR_CONFIGURED_EGRESS_XPN` .

### DIFF
--- a/inc/saimacsec.h
+++ b/inc/saimacsec.h
@@ -776,6 +776,9 @@ typedef enum _sai_macsec_sa_attr_t
      */
     SAI_MACSEC_SA_ATTR_CONFIGURED_EGRESS_XPN,
 
+    /** @ignore - for backward compatibility */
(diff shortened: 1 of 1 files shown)


abdosi:
cc @dipankar-ba / @prafull-brcm please check.

abdosi:
There is still discussion where swss need to updated to use read-only attribute going forward and there should not be use-case of doing write.

Make PR as draft mode.

kcudnik (approved):

abdosi:
Closing the PR. SONiC will need to align with new changes wr.t MACSEC attributes to use now read-only attribute `SAI_MACSEC_SA_ATTR_XPN`

cc @lguohan and @Pterosaur

PR #1216, opened 2021-04-04
Fix error when compiling Broadcom SAI with v1.8.0

abdosi:
What I did:

With SAI v1.8.0 sanity check for having max attribute for any object to be < 200 is not valid for switch attribute.
Commenting out check as of  now.

Why I did:

Fix Broadcom build with SAI v1.8.0

How I verify:
After fix `make` is fine.

### DIFF
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -373,8 +373,8 @@ void check_attr_by_object_type()
 
             META_ASSERT_TRUE(current == i, "object type must be equal on object type list");
             /* For Switch Attribute we have crossed > 200 with Vendor extention for SAIv1.8.0
-              so commenting it for now to avoid error when doing make */
-            /* META_ASSERT_TRUE(index < 200, "object defines > 200 attributes, metadata bug?"); */
+               so increasing threshold  */
+            META_ASSERT_TRUE(index < 300, "object defines > 300 attributes, metadata bug?");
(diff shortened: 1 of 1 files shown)


meta/saisanitycheck.c
```diff
@@ -372,7 +372,9 @@ void check_attr_by_object_type()
             sai_object_type_t current = ot[index]->objecttype;
 
             META_ASSERT_TRUE(current == i, "object type must be equal on object type list");
-            META_ASSERT_TRUE(index < 200, "object defines > 200 attributes, metadata bug?");
+            /* For Switch Attribute we have crossed > 200 with Vendor extention for SAIv1.8.0
+              so commenting it for now to avoid error when doing make */
+            /* META_ASSERT_TRUE(index < 200, "object defines > 200 attributes, metadata bug?"); */
```
kcudnik, line 377: don't comment this, just make this 300, i just put this sanity check in case of some wrong metadata generation

kcudnik (changes requested):

meta/saisanitycheck.c
abdosi: @kcudnik  Updated to check for max 300 attributes.

kcudnik (approved):

kcudnik:
With SAI v1.8.0 sanity check for having max attribute for any object to be < 200 is not valid for switch attribute.
Commenting out check as of now.

PR #1219, opened 2021-04-08
read only port attribute SAI_PORT_ATTR_LAG

phshaikh:
If a port is a member of LAG, return LAG object id in this read only port attribute.

If port is not a member of LAG, return SAI_NULL_OBJECT_ID

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1699,6 +1699,15 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_SYSTEM_PORT,
 
     /**
+     * @brief LAG if port is member of that LAG, else SAI_NULL_OBJECT_ID
+     *
+     * @type sai_object_id_t
+     * @flags READ_ONLY
+     * @objects SAI_OBJECT_TYPE_LAG
+     */
+    SAI_PORT_ATTR_LAG,
+
+    /**
      * @brief End of attributes
      */
     SAI_PORT_ATTR_END,


inc/saiport.h
```diff
@@ -1698,6 +1698,15 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_SYSTEM_PORT,
 
+    /**
+     * @brief LAG if port is member of that LAG, else SAI_NULL_OBJECT_ID
```
kcudnik, line 1702: just curious if this is redundant, if user created lag and lag member, user knows if port is in particular lag, just indirectly

phshaikh: During warm boot reconstructing port state typically should depend on sdk state or the state in ASIC, this is attribute would be particularly useful during warm boot, minimally relying on state saved by the application.

Ports must exist before non-empty lags but then when port joins lag, application may need to handle it differently, especially on warm-boot. For example is to not create bridge port for a port (since  the bridge port is created with lag). that's one example. There may be other such instances. To simplify application programming, I proposed this attribute.

kcudnik: which warm boot scenario are you talking ? internal vendor SAI perspective or sonic/syncd/OA perspective? from internal SAI, if you able to return this attribute value, internally sdk already know the state that this port is a lag member

phshaikh: From the perspective of SAI application. But I do get your point, to find if port is part of a LAG, it can be done this way
- get the list of lag objects using sai_get_object_key
- for each of that lag object, iterate over each lag member
- for each of that lag member, check if its SAI_LAG_MEMBER_ATTR_PORT_ID matches object id of this port
- if yes, return that lag object id
- if no lag found, return null object id
This attribute does the same but in SDK.

Not saying this can't be absolutely handled in application, it can be. But this attribute abstracts away above detail and let SAI SDK itself provide the information which is already present in SDK.

kcudnik (approved):

inc/saiport.h
rlhui: How to differentiating not yet supporting this vs. this is a non-lag port?

phshaikh: > How to differentiate not yet supporting this vs. this is a non-lag port?

Doesn't this question exist for almost every attributes?  SDK can always return _NOT_SUPPORTED or _NOT_IMPLEMENTED if they support LAG but do not implement this attribute or _NULL_OBJECT_ID if they implement this and support LAG?

rlhui:
@JaiOCP ,  @marian-pritsak  - would you like to review this? Thanks.

inc/saiport.h
itaibaz: You mentioned the motivation is
During warm boot reconstructing port state typically should depend on sdk state or the state in ASIC
But this attribute returns particular OID. SDK/ASIC state has internal object representation and not the SAI/Sonic OID.

As for 
Ports must exist before non-empty lags but then when port joins lag, application may need to handle it differently, especially on warm-boot.
So should this attribute be OID for specific LAG membership, or maybe a boolean is_lag_member
In such case, there wouldn't be a need to correlate to specific OID, just rely on SDK/ASIC state to check if port is member in any LAG

phshaikh: > You mentioned the motivation is
> During warm boot reconstructing port state typically should depend on sdk state or the state in ASIC
> But this attribute returns particular OID. SDK/ASIC state has internal object representation and not the SAI/Sonic OID.
> 
> As for
> Ports must exist before non-empty lags but then when port joins lag, application may need to handle it differently, especially on warm-boot.
> So should this attribute be OID for specific LAG membership, or maybe a boolean is_lag_member
> In such case, there wouldn't be a need to correlate to specific OID, just rely on SDK/ASIC state to check if port is member in any LAG

Not really sure I understood. But the attribute returns OID of LAG which holds the port if port is member of LAG (after port is used to create LAG member). Or it returns SAI_NULL_OBJECT_ID if port is not part of LAG. I don't mind if it returns OID of LAG member but then that really doesn't reflect as 'attribute of port'. It is more intuitive to find which LAG this port belongs to, rather than which lag member this port is. But either way is fine with me.

@rlhui can you approve this? This is not a major change in API.

itaibaz: My point is - at the described warm boot scenario, we rely on SDK/ASIC DB. This DB contains LAG membership, with SDK representation of ports. This DB doesn't contain SAI/Sonic Object ID, not is it straightforward at that point to derive the LAG OID from SDK OID. Therefor returning LAG OID is a problem.
My proposition was - it seemed that you described a scenario when the application wants to know if the port is a member in any LAG, or not a member. This can also be described by a boolean. In which case we can add a new bool attribute is_lag_member. Will this be suitable for you ?

phshaikh: @itaibaz isn't this internal implementation detail of SDK/ASIC DB, I don't think API really be defined based on caveats or limitations of SDK. There're places where we use OID already as an object attribute. SDK implementation may chose not to support it if there is no SDK customer using that attribute. SDK implementation in future my extend its implementation if there is a strong need of it from SDK customer.

itaibaz: if the use case is to know of whether a port is lag member, wouldn't boolean is_lag_member be good enough ?

phshaikh: Ya sure, I don't mind changing to boolean "is_lab_member" but I don't think we should let SDK implementation dictate SAI spec :) it should be other way round :) and I do feel not every vendor support every attribute and when they decide to, SDK implementation may have to adjust to absorb new attribute.

PR #1224, opened 2021-04-21
Add additional FEC modes for PAM4 ports

mikeberesford:
Add additional RS544 FEC modes for PAM4 ports. The existing FEC configurations are not sufficient to unambiguously select the FEC mode used at the ASIC level.

Cases where the existing configuration may be insufficient:
- 200G PAM4 may use either RS544 or RS422 with 2x interleave
- gearbox optics mean that some links may be run with different FEC than the default for that speed.
  - 2x25G NRZ to 1x50G PAM4 - need to use RS528 on the 1x50G side instead of RS544
  - 2x50G PAM4 to 1x100G PAM4 - need to use RS544 on the 1x100G side instead of interleaved


dgsudharsan:
Can we add the comprehensive list of FEC modes? This would avoid ambiguity and also simplify the implementation.

dgsudharsan:
@AshokDaparthi @SathishKumarTR-Dell @mikeberesford @JaiOCP  Can we revive this PR? https://github.com/opencomputeproject/SAI/pull/920

inc/saiport.h
```diff
@@ -171,11 +171,17 @@ typedef enum _sai_port_fec_mode_t
     /** No FEC */
     SAI_PORT_FEC_MODE_NONE,
 
-    /** Enable RS-FEC - 25G, 50G, 100G ports */
+    /** Enable RS-528 FEC - 25G, 50G, 100G ports */
     SAI_PORT_FEC_MODE_RS,
```
apogedik, line 175: Instead of overloading SAI_PORT_FEC_MODE_RS with RS-528 FEC, can we add SAI_PORT_FEC_MODE_RS528 instead so that the backward compatibility is not broken?

mikeberesford: To avoid ambiguity in the readback (if application sets SAI_PORT_FEC_MODE_RS and SAI adapter actually configures RS-544, what is the correct readback?), I think we either need to redefine this or define a new port attribute that explicitly defines the FEC mode.

Since breaking backward compatibility is a concern here, I'm leaning toward the latter based on the conversation today - creating SAI_PORT_ATTR_DETAILED_FEC_MODE which allows for detailed control while leaving the existing enum alone. The problems I see with doing this are:
1) we need a way to enforce that only one or the other is set (or specify which is used)
2) we probably need to create corresponding attributes for all uses of sai_port_fec_mode_t (SAI_PORT_ATTR_SUPPORTED_FEC_MODE, SAI_PORT_ATTR_REMOTE_ADVERTISED_FEC_MODE, SAI_PORT_ATTR_ADVERTISED_FEC_MODE, SAI_PORT_ATTR_FEC_MODE)

Thoughts on this? @lguohan @JaiOCP @SathishKumarTR-Dell @dgsudharsan @rlhui

tarakreddy1984: SAI_PORT_FEC_MODE_RS has been using for RS528, RS544, RS272 and RS544_2XN.

mikeberesford: Yes. The problem I'm trying to solve is that the mapping of SAI_PORT_FEC_MODE_RS is to the actual FEC type is ambiguous in at least one case (200G PAM4 can use either RS544 or RS544_2XN), and there are cases when using optics with built-in gearbox that we may need to use different FEC configuration (I believe 100G/lane DR8 communicating with 50G/lane DR4 will be an example of this).

apogedik: The comment should be changed to the original one (i.e. remove 528 from the comment)

mikeberesford: Are you suggesting that we have all of the RS-FEC types in addition to the legacy "FEC_MODE_RS"?
SAI_PORT_FEC_MODE_RS
SAI_PORT_FEC_MODE_RS528
SAI_PORT_FEC_MODE_RS544
SAI_PORT_FEC_MODE_RS544_2X_INTERLEAVED

I think I'd be OK with that, but we would need to document that if FEC_MODE_RS is set then the actually selected FEC mode (one of RS528,544,etc) would be returned? This is similar to what's being proposed by @jleveque in https://github.com/opencomputeproject/SAI/pull/1233.

apogedik: Yes.  This sounds OK.  Thanks.

apogedik: Please make sure to remove change in line 174.

mikeberesford: Are there any concerns around warmboot reconciliation or SYNCD_INSPECT_ASIC at the SONiC (or other application) level if we start introducing attributes that read back differently from what is set? It appears to me that this type of setting (write one value, read back a different one) would require special handling in those cases.

mikeberesford: Based on the conversations above (needing to preserve the existing behavior) and my concerns about reading back a different value than was programmed, I made the change to a new enum for detailed FEC control. Please take a look and let me know if this is more acceptable.

mikeberesford:
Consensus from today's community meeting is that we should add a new boolean to control which of the FEC settings is used, as well as documentation of the behavior around the FEC controls. I'll work on getting the changes for both of these done in the next day or so.

mikeberesford:
Updated documentation and added additional boolean to control extended/non-extended FEC use. Please take a look at this latest revision when you have a chance. @JaiOCP @apogedik @tarakreddy1984 @dgsudharsan

inc/saiport.h
```diff
@@ -513,6 +542,14 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_REMOTE_ADVERTISED_FEC_MODE,
 
+    /**
+     * @brief Query extended list of Advertised remote port FEC control
+     *
+     * @type sai_s32_list_t sai_port_fec_mode_extended_t
+     * @flags READ_ONLY
```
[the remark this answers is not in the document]
kcudnik: i dont know internals to answer this, if you feel it should be there, then add it

inc/saiport.h
```diff
@@ -788,6 +847,16 @@ typedef enum _sai_port_attr_t
      */
```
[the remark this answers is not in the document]
kcudnik: i dont know internals to answer this, if you feel it should be there, then add it

JaiOCP (approved):
This looks good.

inc/saiport.h
```diff
@@ -174,13 +174,34 @@ typedef enum _sai_port_fec_mode_t
...
     SAI_PORT_FEC_MODE_RS,
 
     /** Enable FC-FEC - 10G, 25G, 40G, 50G ports */
     SAI_PORT_FEC_MODE_FC,
 } sai_port_fec_mode_t;
 
+/**
+ * @brief Attribute data for #SAI_PORT_ATTR_FEC_MODE_EXTENDED
+ */
+typedef enum _sai_port_fec_mode_extended_t
```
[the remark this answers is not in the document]
motinisim: RS-FEC has 2 variants - RS(514,528) for NRZ and RS(514,544) for PAM4. As per IEEE802.3 the 514 value is used to calculate the number of parity symbols carried to protect the data so no need to mention it in specific.

inc/saiport.h
```diff
@@ -174,13 +174,34 @@ typedef enum _sai_port_fec_mode_t
...
+    SAI_PORT_FEC_MODE_EXTENDED_NONE,
+
+    /** Enable RS-528 FEC - 25G, 50G, 100G ports */
+    SAI_PORT_FEC_MODE_EXTENDED_RS528,
+
+    /** Enable RS544-FEC - 100G PAM4, 200G ports */
+    SAI_PORT_FEC_MODE_EXTENDED_RS544,
+
+    /** Enable RS544-FEC (2x interleaved) - 100G, 200G, 400G ports */
+    SAI_PORT_FEC_MODE_EXTENDED_RS544_2X_INTERLEAVED,
```
motinisim, line 199: Maybe RS544_INTERLEAVED instead of RS544_2X_INTERLEAVED? The interleaved FEC sublayer does 2:1 FEC codeword interleaving but I wonder if it should be mentioned in specific (maybe if we expect other types like that to be developed though the latency may become an issue).

inc/saiport.h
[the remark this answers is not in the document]
mikeberesford: Thanks, so we can skip the 514. CL74 is FC FEC, so that one doesn't need to be added. I'm not finding good details on CL108's Reed-Solomon length/symbols parameters, so I'm reluctant to add it unless it's done as RS-CL108 (which doesn't match the naming of the other FEC types).

inc/saiport.h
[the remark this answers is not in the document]
mikeberesford: Ack, added.

inc/saiport.h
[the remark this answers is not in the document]
mikeberesford: Ack, added. Thanks.

inc/saiport.h
mikeberesford: Done.

inc/saiport.h
mikeberesford: I believe that this is resolved at this point.

inc/saiport.h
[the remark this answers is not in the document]
motinisim: I think we can also skip CL108. As per IEEE802.3 the 25GBASE-R RS-FEC sublayer employs RS(528,514).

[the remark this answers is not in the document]
mikeberesford: Ack, thanks for helping out with this!

rlhui (approved):

PR #1231, opened 2021-04-30
Updates to SRv6 programming model and related objects/attributes

ashutosh-agrawal:
Changes are listed in the attached md file.

Following items were deferred and will be addressed in subsequent PRs:
- Additional capability attributes to specify number of SIDs supported by the platform
- Need for shift-only behavior in case of uSID
- Consider specifying SID structure in a separate container (object/struct) and reuse as necessary
- Split the headend behavior from SID list and add it as a separate attribute for nexthop so SID lists can be shared across different behaviors


meta/parse.pl
```diff
@@ -2260,6 +2260,8 @@ sub ProcessStructValueType
     return "SAI_ATTR_VALUE_TYPE_MACSEC_AUTH_KEY" if $type eq "sai_macsec_auth_key_t";
     return "SAI_ATTR_VALUE_TYPE_MACSEC_SALT"    if $type eq "sai_macsec_salt_t";
     return "SAI_ATTR_VALUE_TYPE_BOOL"           if $type eq "bool";
+    return "SAI_ATTR_VALUE_TYPE_IPV6"           if $type eq "sai_ip6_t";
+    return "SAI_ATTR_VALUE_TYPE_UINT8"           if $type eq "sai_uint8_t";
```
kcudnik, line 2264: could you fix spacing before if?

ashutosh-agrawal: Fixed.

kcudnik (approved):

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,483 @@
...
+        local_sid_attr[1].id = SAI_LOCAL_SID_ENTRY_ATTR_VRF
+        local_sid_attr[1].value.oid = vr_id_1001 // overlay vrf, created elsewhere
+        saistatus = saiv6sr_api->create_local_sid(&local_sid_entry, 2, local_sid_attr)
+
+
+## References ##
+1. https://tools.ietf.org/html/rfc8754
+2. https://tools.ietf.org/html/rfc8986
+3. https://tools.ietf.org/html/draft-ietf-spring-segment-routing-policy-09
+
```
abdosi, line 483: we have latest version https://tools.ietf.org/html/draft-ietf-spring-segment-routing-15. can we update this ?

also can we specify segment header draft https://tools.ietf.org/html/draft-ietf-6man-segment-routing-header-08

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,483 @@
...
+        1. Create a SID list object with 3 segments
+
+            sidlist_entry_attrs[0].id = SAI_SEGMENTROUTE_SIDLIST_ATTR_TYPE
+            sidlist_entry_attrs01].value.s32 = SAI_SEGMENTROUTE_SIDLIST_TYPE_ENCAPS_RED
+            sidlist_entry_attrs[1].id = SAI_SEGMENTROUTE_SIDLIST_ATTR_SEGMENT_LIST
+            sidlist_entry_attrs[1].value.objlist.count = 3;
+            CONVERT_STR_TO_IPV6(sidlist_entry_attrs[1].value.objlist.list[0], "2001:db8:85a3::8a2e:370:7334");
+            CONVERT_STR_TO_IPV6(sidlist_entry_attrs[1].value.objlist.list[1], "2001:db8:85a3::8a2e:370:2345");
+            CONVERT_STR_TO_IPV6(sidlist_entry_attrs[1].value.objlist.list[2], "2001:db8:85a3::8a2e:370:3456");
+            saistatus = sai_v6sr_api->create_segmentroute_sidlist(&sidlist_id, switch_id, 2, sidlist_entry_attrs);
```
abdosi, line 427: do we have/need query to find the capability of how many v6 sid can be encapsulated ?

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,483 @@
...
+        local_sid_entry.vr_id = vr_id_1 // underlay VRF
+        local_sid_entry.locator_len = 64
+        local_sid_entry.function_len = 8
+        CONVERT_STR_TO_IPV6(local_sid_entry.sid, "2001:db8:0:1::1000:0:0:0");
+    
+        local_sid_attr[0].id = SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_TYPE
+        local_sid_attr[0].value = SAI_LOCAL_SID_ENTRY_ENDPOINT_TYPE_DT46
+        local_sid_attr[1].id = SAI_LOCAL_SID_ENTRY_ATTR_VRF
+        local_sid_attr[1].value.oid = vr_id_1001 // overlay vrf, created elsewhere
+        saistatus = saiv6sr_api->create_local_sid(&local_sid_entry, 2, local_sid_attr)
```
abdosi, line 476: is this exact match or we can have lpm/route type lookup for my-local sid ?

ashutosh-agrawal: Changed SID from ip6_t to ip_prefix_it based on discussions in SAI review meeting.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -477,7 +477,7 @@ typedef struct _sai_local_sid_entry_t
 
 
 ## References ##
```
pcamaril, line 479: draft-filsfils-spring-srv6-net-pgm-insertion-04 is missing (H.Insert, H.Insert.Red, End.B6.Insert, End.B6.Insert.Red)

inc/saisegmentroute.h
```diff
@@ -149,6 +235,297 @@ typedef sai_status_t (*sai_get_segmentroute_sidlist_attribute_fn)(
...
+
+    /**
+     * @brief Length of the Args part of the SID
+     */
+    sai_uint8_t args_len;
+
+    /**
+     * @brief IPv6 prefix for Local SID
+     */
+    sai_ip_prefix_t sid;
```
pcamaril, line 383: Couple of comments:
- sai_ip_prefix_t may be used for an IPv4 or IPv6 prefix. While in SRv6 we only have IPv6. Hence, wouldn't it make sense to define a sai_ipv6_prefix_t ?
- Additionally: given that we already have the locator_len, function_len, args_len I'm not really sure you really need a prefix here...
- The definition of the prefix should be clarified. For example, I have Locator (64b) + Fun (16b) + Arg (16b). What is the prefix? the /80 or the /96 ?  This could be easily clarified in the comment right before it.

inc/saisegmentroute.h
```diff
@@ -149,6 +235,297 @@ typedef sai_status_t (*sai_get_segmentroute_sidlist_attribute_fn)(
...
+     * @brief Virtual Router ID
+     *
+     * @objects SAI_OBJECT_TYPE_VIRTUAL_ROUTER
+     */
+    sai_object_id_t vr_id;
+
+    /**
+     * @brief Length of the Locator part of the SID
+     */
+    sai_uint8_t locator_len;
```
pcamaril, line 368: The control plane (e.g. ISIS) advertises the Locator length splitted into two pieces: locator block and locator node. Might be useful to split this locator_len into those two pieces...

From SRv6-ISIS;

>       LB Length: 1 octet.  SRv6 SID Locator Block length in bits.
>       LN Length: 1 octet.  SRv6 SID Locator Node length in bits.
>       Fun. Length: 1 octet.  SRv6 SID Function length in bits.
>       Arg. Length: 1 octet.  SRv6 SID Arguments length in bits.

In RFC8986:
> A locator may be represented as B:N where B is the SRv6 SID block
> (IPv6 prefix allocated for SRv6 SIDs by the operator) and N is the
> identifier of the parent node instantiating the SID.

inc/saisegmentroute.h
ashutosh-agrawal: @pcamaril All very good points. 1) Earlier, I was thinking about defining two separate attributes - sid and sid_mask of type ip6_t instead of sai_ipv6_prefix_t. Defining a sai_ipv6_prefix_t is also a fine alternative.
2) Yes, prefix_len = locator_len + func_len + args_len and can be considered redundant.

We can discuss it further in the next meeting. If community agrees, I'll revert it back to sai_ipv6_t.

inc/saisegmentroute.h
ashutosh-agrawal: @pcamaril Are you thinking of using B:N split to also specify Fbbuu notation for uSID? I'm wondering if this split needs to be specified on a per-SID basis or can it be a per-switch configuration for practical use cases?

pcamaril: > Are you thinking of using B:N split to also specify Fbbuu notation for uSID?
I don't follow the Fbbuu part for uSID. Take the compression part out of the equation. Normally an operator would allocate a block B (e.g. /48) to the entire network for SR. Then it would allocate B:N::/64 to each SRv6-enabled router for instantiating SIDs. That is what I was thinking about. 

> I'm wondering if this split needs to be specified on a per-SID basis or can it be a per-switch configuration for practical use cases?
That actually a very interesting point... I would tend to think its safer to specify it on a per-SID basis. For example I think of the case where you support compressed and non-compressed SIDs in one single node. Specifying on a per-sid basis provides more freedom to the operator...no?

rlhui:
@rck-innovium , @marian-pritsak  - would you like to review this? Thanks.

rlhui:
@Israel-Meilik - please help review/approve.  Thanks.

rlhui:
@skraza - please help review, thanks.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+![SRv6 Endpoint Behavioral Model](figures/SRv6_Endpoint_behavioral_model.png "Figure 1: Endpoint Behavior ")
+__Figure 1: Endpoint Behavior.__
+
+### Source/Headend Behavior ###
+- This proposal will remove "Endpoint" from the Nexthop table attributes and leave other attributes like SIDLIST in order to implement the Headend Behavior.
+
+![SRv6 Headend Behavioral Model](figures/SRv6_Headend_behavioral_model.png "Figure 1: Headend Behavior ")
+__Figure 1: Headend Behavior.__
+
+### Transit Behavior ###
```
skraza, line 37: Why is this same as Headend? Which exact behavior are your refering here ?

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+1. sai_next_hop_endpoint_type_t and sai_nexthop_endpoint_pop_type_t is deprecated.
+
+```
+typedef enum _sai_next_hop_type_t
+{
+    ....
+    ....
+
+    /** IPv6 Segment Route SID List */
+    SAI_NEXT_HOP_TYPE_SEGMENTROUTE_SIDLIST,
```
skraza, line 53: As a general comment: SR has two dataplane types: SR-mpls and SRv6 - Using SEGMENTROUTE to mean srv6 is misleading -> I suggest to use SEGMENTROUTE_SRV6 or SRV6 only.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+ */
+
+typedef enum _sai_next_hop_endpoint_pop_type_t
+{
+    /** Penultimate segment pop */
+    SAI_NEXT_HOP_ENDPOINT_POP_TYPE_PSP,
+
+    /** Ultimate Segment pop */
+    SAI_NEXT_HOP_ENDPOINT_POP_TYPE_USP,
+
```
skraza, line 74: There are more combos.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+} sai_segmentroute_sidlist_type_t;
+``` 
+
+2. Added enum sai_local_sid_entry_endpoint_type_t to provide the initial list of endpoint behaviors.
+
+```
+/**
+ * @brief Enum defining Endpoint Action types
+ */
+typedef enum _sai_local_sid_entry_endpoint_type_t
```
skraza, line 178: As per net-pgm RFC, these are called behaviors - let's call them behavior_t

skraza, line 178: Moreover: I am afraid that use of "local_sid" term going forward may lead to confusion - recall that in SRv6 micro-segment (uSID) draft-filsfils-spring-net-pgm-extension-srv6-usid, there are segments are global (uN) and other as local (uA, uD). A endpoint node terminates both its global and local segments.

Shall we just call it "sid_entry" instead of "local_sid_entry" here and other places?

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+} sai_local_sid_entry_endpoint_type_t;
+```
+3. Added enum sai_local_sid_entry_endpoint_flavor_t to list all the combinations of endpoint flavors 
+   as defined in RFC-8986
+
+```
+/**
+ * @brief Enum defining Endpoint Segment flavors for End, End.X and End.T functions
+ */
+typedef enum _sai_local_sid_entry_endpoint_flavor_t
```
skraza, line 231: behavior_flavour_t

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+    SAI_LOCAL_SID_ENTRY_ENDPOINT_FLAVOR_USD_AND_USP,
+
+    /** PSP and USD */
+    SAI_LOCAL_SID_ENTRY_ENDPOINT_FLAVOR_PSP_AND_USD,
+
+    /** PSP, USP and USD */
+    SAI_LOCAL_SID_ENTRY_ENDPOINT_FLAVOR_PSP_AND_USP_AND_USD
+
+    /** None */
+    SAI_LOCAL_SID_ENTRY_ENDPOINT_FLAVOR_NONE
```
skraza, line 255: I suggest making this as the first (value 0) rather than the last in the list.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+} sai_local_sid_entry_endpoint_flavor_t;
+```
+
+4. Added a new object sai_local_sid_entry_t configure Endpoint behavior.
+
+```
+/**
+ * @brief Attribute list for Local SID
+ */
+typedef enum _sai_local_sid_entry_attr_t
```
skraza, line 266: > sid_entry_attr_t

see my earlier comment

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+
+    /**
+     * @brief Flavor for End, End.X and End.T functions
+     *
+     * @type sai_local_sid_entry_endpoint_flavor_t
+     * @flags CREATE_AND_SET
+     * @default SAI_LOCAL_SID_ENTRY_ENDPOINT_FLAVOR_PSP
+     * @validonly SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_TYPE == SAI_LOCAL_SID_ENTRY_ENDPOINT_TYPE_E or SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_TYPE == SAI_LOCAL_SID_ENTRY_ENDPOINT_TYPE_X or SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_TYPE == SAI_LOCAL_SID_ENTRY_ENDPOINT_TYPE_T
+     */
+    SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_FLAVOR,
```
skraza, line 289: What will be default flavour if this is not specified ? 

Conversely, if were to specify "base" End or EndX behavior, this attr need not set.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+
+    /** PSP and USD */
+    SAI_LOCAL_SID_ENTRY_ENDPOINT_FLAVOR_PSP_AND_USD,
+
+    /** PSP, USP and USD */
+    SAI_LOCAL_SID_ENTRY_ENDPOINT_FLAVOR_PSP_AND_USP_AND_USD
+
+    /** None */
+    SAI_LOCAL_SID_ENTRY_ENDPOINT_FLAVOR_NONE
+
```
skraza, line 256: For uSID, will need "Shift-only" flavour as well.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+    /**
+     * @brief Next hop for cross-connect function
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_NEXT_HOP, SAI_OBJECT_TYPE_NEXT_HOP_GROUP, SAI_OBJECT_TYPE_ROUTER_INTERFACE
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     * @validonly SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_TYPE == SAI_LOCAL_SID_ENTRY_ENDPOINT_TYPE_X
```
skraza, line 317: Also valid for B6*

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+     * @brief VRF for decapsulation and specific table lookup functions
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_VIRTUAL_ROUTER
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     * @validonly SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_TYPE == SAI_LOCAL_SID_ENTRY_ENDPOINT_TYPE_T or SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_TYPE == SAI_LOCAL_SID_ENTRY_ENDPOINT_TYPE_DT4 or SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_TYPE == SAI_LOCAL_SID_ENTRY_ENDPOINT_TYPE_DT6 or SAI_LOCAL_SID_ENTRY_ATTR_ENDPOINT_TYPE == SAI_LOCAL_SID_ENTRY_ENDPOINT_TYPE_DT46
+     */
+    SAI_LOCAL_SID_ENTRY_ATTR_VRF,
```
skraza, line 331: While this attr alone works for DT46, how does this work for DT4 or DT6. Don't we need an AFI attr (ipv4/ipv6) along with VRF attr for DT4, DT6?

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+     * @brief Virtual Router ID
+     *
+     * @objects SAI_OBJECT_TYPE_VIRTUAL_ROUTER
+     */
+    sai_object_id_t vr_id;
+
+    /**
+     * @brief Length of the Locator part of the SID
+     */
+    sai_uint8_t locator_len;
```
skraza, line 381: Can we define anything about SID stcuture in a separate object/container (sid_struct_t) -> locator_len, func_leb, args_len.

This way, this common struct could be re-used at places as/when required.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+            CONVERT_STR_TO_IPV6(sidlist_entry_attrs[1].value.objlist.list[0], "2001:db8:85a3::8a2e:370:7334");
+            CONVERT_STR_TO_IPV6(sidlist_entry_attrs[1].value.objlist.list[1], "2001:db8:85a3::8a2e:370:2345");
+            CONVERT_STR_TO_IPV6(sidlist_entry_attrs[1].value.objlist.list[2], "2001:db8:85a3::8a2e:370:3456");
+            saistatus = sai_v6sr_api->create_segmentroute_sidlist(&sidlist_id, switch_id, 2, sidlist_entry_attrs);    
+            
+        2. Create a tunnel object with source IP used for tunnel encapsulation
+
+            tunnel_entry_attrs[0].id = SAI_TUNNEL_ATTR_TYPE
+            tunnel_entry_attrs[0].value = SAI_TUNNEL_TYPE_SRV6
+            tunnel_entry_attrs[1].id = SAI_TUNNEL_ATTR_ENCAP_SRC_IP
```
skraza, line 433: How about other encap attributes: Traffic-class, Hop-limit, Flow-label values?

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,484 @@
...
+
+
+## Examples ##
+- SR Headend
+    Example configuration for H.Encaps.Red behavior
+       
+        1. Create a SID list object with 3 segments
+
+            sidlist_entry_attrs[0].id = SAI_SEGMENTROUTE_SIDLIST_ATTR_TYPE
+            sidlist_entry_attrs01].value.s32 = SAI_SEGMENTROUTE_SIDLIST_TYPE_ENCAPS_RED
```
skraza, line 421: Note: Adding Headend flavour within a SID-list will mean that same SID-list can not be used across different flavours.

inc/saisegmentroute.h
```diff
@@ -149,6 +235,302 @@ typedef sai_status_t (*sai_get_segmentroute_sidlist_attribute_fn)(
...
+ *
+ * @param[in] local_sid_entry Local SID entry
+ *
+ * @return #SAI_STATUS_SUCCESS on success, failure status code on error
+ */
+typedef sai_status_t (*sai_remove_local_sid_entry_fn)(
+        _In_ const sai_local_sid_entry_t *local_sid_entry);
+
+/**
+ * @brief Set Local SID attribute value
```
skraza, line 417: Q: Is this how a sid entry modified?

inc/saiswitch.h
```diff
@@ -338,6 +338,8 @@ typedef enum _sai_tunnel_type_t
 
     SAI_TUNNEL_TYPE_MPLS,
 
+    SAI_TUNNEL_TYPE_SRV6,
```
skraza, line 341: Right. So if we are using SRV6 here, why are we not calling it SRV6 other places where we are using SEGMENTROUTE

ashutosh-agrawal: Good point. When SRv6 support was added to SAI a few years ago, authors decided to use "SEGMENTROUTE" for some reason. If backward compatibility is not a concern, I can replace all instances of SEGMENTROUTE with SRV6.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: This section of the document just lists the attribute which has been deprecated. All the combinations are defined for the new local_sid table/attributes in saisegmentroute.h.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: Fair point. I used the term "local_sid" because it's being used in other implementations like VPP and Linux. Also, net-pgm document used to have a MyLocalSID table but I don't see it there anymore. Do you think "my_sid_entry" will be a more appropriate name since this table is supposed to have all the SIDs advertised by this node?

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: Sorry, I don't understand the issue here. For DT4 or DT6, SAI implementation will check if inner packet type (ipv4 or ipv6) matches with the endpoint_behavior before applying the endpoint functionality. Isn't the AFI implicit with DT4 or DT6?

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: @skraza Do you have any references or examples (VPP, Linux, FRR etc) of specifying the SID structure for anything other than the endpoint functionality? I'm trying to understand the implications of it on the hardware and also on proposed SONiC implementation of SRv6.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: I didn't list explicitly list the attributes related to Hop-Limit, Traffic-Class, ECN marking and Hashing. It's assumed that existing attributes for other IPv6 tunnels like IP-in-IPv6 will apply to the SRv6 as well. I'll note that in the document.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: Agree. I wanted to split the Headend flavor from the SID list and add it as a separate attribute but left it there for backward compatibility reasons. I'm open to making this change now.

inc/saisegmentroute.h
ashutosh-agrawal: Yes, this API can be used to modify an existing SID entry.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: It was defined in previous SRv6 SAI spec and I meant to say that transit behavior will be same as headend but I agree that it's doesn't make much sense here. I've removed this section.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: Updated. s/endpoint_type/endpoint_behavior/

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: Updated. s/endpoint_flavor/endpoint_behavior_flavor/

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: Updated as suggested.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: Default is set to "None" and that will correspond to the "base" behavior. So, if this attribute is not set, base behavior will be assumed.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: Fixed.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: If backward compatibility is not a concern, I can replace all instances of SEGMENTROUTE with SRV6.

ashutosh-agrawal:
> Tunnel id is for encapsulation not decapsulation

@caizhenghui-juniper local_sid entry has tunnel id for decap and nexthop entry will have tunnel id for encap. SAI application can use appropriate tunnel_id to set both encap/decap attributes for features like ttl/qos modes.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: @rlhui @abdosi Any objections to replacing SEGMENTROUTE with SRV6?

caizhenghui-juniper:
@ashutosh-agrawal ok, got it.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: For uSID, I was thinking about add uN and uA behaviors in a future PR. For uN/uA, the shift-and-lookup/shift-and-forward functionality can be implicit. Can you please explain where shift-only flavor will be required?

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,502 @@
...
+### Endpoint Behavior ###
+In the earlier proposal, In order to program the hardware for an Endpoint SID, first a route entry is created for the local SID and then a nexthop entry 
+is created to specify the endpoint behavior and parameters. This will require either the application or the operating system responsible for making the SAI 
+API calls to create two additional objects for every LocalSID entry - a route object and a nexthop object. This model is different from the programming model 
+used for IP tunnels and MPLS. For IP and MPLS tunnel terminations, a separate tunnel_term or Insegment object is created without a need to decompose the tunnel 
+termination entry into multiple SAI objects. 
+This proposal will follow the programming model of MPLS segment routing and IP tunnel termination by adding a LocalSID object for SRv6 similar to InSegment
+for MPLS.
+  
+![SRv6 Endpoint Behavioral Model](figures/SRv6_Endpoint_behavioral_model.png "Figure 1: Endpoint Behavior ")
```
Israel-Meilik, line 30: The figure view is very narrow to the exact new table.
It would help to see a broader perspective, i.e. add few stages before and after to get the context in the processing pipe.
E.g. in the MPLS-SR there is a similar change with a drawing that shows large portion of the processing pipe.

pcamaril (approved):
Looks good to me. Thanks.

doc/SAI-IPv6-Segment-Routing-Update.md
reshmaintel: @rlhui @abdosi  @Israel-Meilik @caizhenghui-juniper are we ok to replace SEGMENTROUTE with SRv6 ?
Thanks

caizhenghui-juniper: I don't see any issue with it.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: IIRC, MPLS-SR document had a drawing which showed the pipeline for L3VPN use case only. Since MPLS inseg table doesn't have an explicit operation (like DT4 etc), an example was needed there to show the intent and also other attributes are needed. With SRv6, there is an explicit behavior specified and also other attributes like tunnel/vrf etc have a condition attached to them to show when they are needed.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: I've added uN and uA behaviors for uSID. My understanding is that shift-only behavior needs to be advertised in the control plane by the legacy devices but I haven't found a need for it at the SAI level.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,503 @@
...
+
+...
+...
+
+} sai_next_hop_attr_t;
+```
+
+### sairouterinterface.h ###
+
+1. Added attribute to enable/disable SRv6 on a RIF.
```
skraza, line 126: As provided earlier, unlike MPLS, per-intf enabling for SRv6 does not make sense. SRv6 should be enabled /diabled globally instead.

doc/SAI-IPv6-Segment-Routing-Update.md
```diff
@@ -0,0 +1,503 @@
...
+    - get_local_sid_entry_attribute
+    - create_local_sid_entries
+    - remove_local_sid_entries
+    - set_local_sid_entries_attribute
+    - get_local_sid_entries_attribute
+
+
+## Examples ##
+- SR Headend
+    Example configuration for H.Encaps.Red behavior
```
skraza, line 434: SID structure (a la mySID) also be allowed per SID-list (if not per SID)

Israel-Meilik (approved):

skraza (approved):
PR#1 approved. Some comments to be discussed/addressed in PR#2.

Israel-Meilik (approved):
PR#1 reviewed by Broadcom team and approved.
Thanks.

inc/saiswitch.h
ashutosh-agrawal: This topic was discussed in SRv6 workgroup and everyone agreed that backward compatibility is not a concern. I'll rename segmentroute to srv6 everywhere.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: In SRv6 workgroup meeting, we agreed to implement additional attributes for length of SID list in a subsequent PR. Marking this as resolved for this PR.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: This topic is discussed in another section as well. In SRv6 workgroup, we agreed that application will program lengths for Locator Node, Locator block, function, arg fields and SAI adapter will calculate the appropriate prefix length from that information. Application doesn't need to explicitly set the prefix length.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: Decided to rename in SRv6 workgroup meeting.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: If required, will be implemented in a subsequent PR.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: will split it in the next PR.

doc/SAI-IPv6-Segment-Routing-Update.md
ashutosh-agrawal: In SRv6 meeting, we agreed to postpone this for a subsequent PR.

reshmaintel (approved):

abdosi (approved):

abdosi:
Based on community meeting discussion there were 2 major points:

1> No need to address Backward Compatibility for SRv6 since feature has evolve lot when SAI api were design intially.
2> All the TODO enhancements will be done as part of subsequent PR's 

There were agreement to approve/merge this PR as there was no outstanding issue.

abdosi:
@ashutosh-agrawal Can you please add details in PR description above. We can merge after that.

ashutosh-agrawal:
> @ashutosh-agrawal Can you please add details in PR description above. We can merge after that.

@abdosi In the PR description, I've added a list of items which were deferred (as discussed in the SRv6 WG).

ashutosh-agrawal:
@abdosi Please let me know if anything else needs to be done to get this PR to merge.

abdosi:
thanks . will merge this when PR check passes

caizhenghui-juniper:
@ashutosh-agrawal do we have vslib code change which uses old segmentroute sid list in the code before moving the pointer of SAI in sonic to avoid build break?

zhangyanzhao:
@abdosi  are all the code PRs merged? Can you please update the code PRs by following example https://github.com/Azure/SONiC/pull/806?

PR #1233, opened 2021-05-07
[saiport.h] Add 'internal' FEC mode and ability to query operational mode

jleveque:
Add new FEC mode, `SAI_PORT_FEC_MODE_INTERNAL`, which allows the port to be configured without an explicit FEC mode, rather allowing the driver/firmware to determine the FEC mode (e.g., via autonegotiation).

With the addition of the\is new FEC mode attribute, we also need a way to query the current operational FEC mode for a port. This is accomplished by the addition of a new attribute, `SAI_PORT_ATTR_OPER_FEC_MODE`. This is expected to return the current operational FEC mode (i.e., `SAI_PORT_FEC_MODE_NONE`, `SAI_PORT_FEC_MODE_RS`, or `SAI_PORT_FEC_MODE_FC`). It should never return `SAI_PORT_FEC_MODE_INTERNAL`.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1709,7 +1709,7 @@ typedef enum _sai_port_attr_t
      * it was set explicitly or determined internally. Thus
      * it should never return SAI_PORT_FEC_MODE_INTERNAL.
      *
-     * @type sai_s32_list_t sai_port_fec_mode_t
+     * @type sai_port_fec_mode_t
      * @flags READ_ONLY
      */
     SAI_PORT_ATTR_OPER_FEC_MODE,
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

inc/saiport.h
```diff
@@ -176,6 +176,10 @@ typedef enum _sai_port_fec_mode_t
 
     /** Enable FC-FEC - 10G, 25G, 40G, 50G ports */
     SAI_PORT_FEC_MODE_FC,
+
+    /** FEC mode is chosen by driver/firmware, e.g., via auto-negotiation */
+    SAI_PORT_FEC_MODE_INTERNAL,
```
lguohan, line 181: as we have discussed, this breaks the backward compatibility.

jleveque:
Closing in favor of https://github.com/opencomputeproject/SAI/pull/1242.

Note that a future PR should still be opened to add `SAI_PORT_ATTR_OPER_FEC_MODE`

PR #1234, opened 2021-05-11
Adding remote loopback option

SidharajU:
This allows packets to be sent to the PHY twisted-pair interface to the PHY PCS layer and back to PHY twisted-pair interface.


kcudnik (approved):

SidharajU:
Can we push these changes?

kcudnik:
must be consulted with SAI community?

kcudnik (approved):

inc/saiport.h
```diff
@@ -126,6 +127,24 @@ typedef enum _sai_port_internal_loopback_mode_t
...
+    SAI_PORT_LOOPBACK_MODE_NONE,
+
+    /** Port loopback at PHY module */
+    SAI_PORT_LOOPBACK_MODE_PHY,
+
+    /** Port loopback at MAC module */
+    SAI_PORT_LOOPBACK_MODE_MAC,
+
+    /** Port loopback at remote end */
+    SAI_PORT_LOOPBACK_MODE_REMOTE
```
ashutosh-agrawal, line 145: Should we explicitly state where remote loopback is done - phy, pcs or mac?

SidharajU: Done

ashutosh-agrawal: Thanks. Since we are defining a new type, we should add enums for all 6 combinations - MAC/PCS/PHY x NEAR/FAR. I understand that most of the devices will not support all 6 modes but some devices do support loopback modes not listed here like - near PCS loopback or remote MAC loopback.

rlhui: Thanks.  @ashutosh-agrawal - would you like to approve this one?  Can add more in a separate PR if needed.

ashutosh-agrawal: In that case, can you please rename SAI_PORT_LOOPBACK_MODE_REMOTE so it's possible to add other remote loopback options in the future.

SidharajU: If we rename it and in future if someone wants to add more options then we will have to change the name again and it wont be backward compatible, so we'll keep the name as it is.

ashutosh-agrawal: Sorry, I didn't understand your reply. If you rename it to SAI_PORT_LOOPBACK_MODE_REMOTE_MAC, then other enums for PCS or PHY loopbacks can be easily added in the feature. There will be no need to change the name for remote mac loopback.

SidharajU: How about we keep SAI_PORT_LOOPBACK_MODE_PHY_REMOTE as we have for now and anyone who wants to support other Loopbacks can  extend it to SAI_PORT_LOOPBACK_MODE_MAC_REMOTE or SAI_PORT_LOOPBACK_MODE_PCS_REMOTE or name it according to their hardware support.

ashutosh-agrawal (approved):

PR #1243, opened 2021-05-26
Add support for programming My MAC table separately from RIF

erohsik:
RIF Source MAC is used to set the Source MAC address in transmitted packet.
It can also be used to match the Destination MAC address in received packet.
This proposal provides the capability to program the
MAC address used in the receive path.

### DIFF
--- a/inc/saimymac.h
+++ b/inc/saimymac.h
@@ -1,0 +1,347 @@
+/**
+ * Copyright (c) 2014 Microsoft Open Technologies, Inc.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *    not use this file except in compliance with the License. You may obtain
+ *    a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    THIS CODE IS PROVIDED ON AN *AS IS* BASIS, WITHOUT WARRANTIES OR
+ *    CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT
+ *    LIMITATION ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS
+ *    FOR A PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
+ *
+ *    See the Apache Version 2.0 License for specific language governing
+ *    permissions and limitations under the License.
+ *
+ *    Microsoft would like to thank the following companies for their review and
+ *    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
+ *    Dell Products, L.P., Facebook, Inc., Marvell International Ltd.
+ *
+ * @file    saimymac.h
+ *
+ * @brief   This module defines SAI My MAC
+ */
+
+#if !defined (__SAIMYMAC_H_)
+#define __SAIMYMAC_H_
+
+#include <saitypes.h>
+
+/**
+ * @brief My MAC entry attribute IDs
+ */
+typedef enum _sai_my_mac_attr_t
+{
+    /**
+     * @brief Start of attributes
+     */
+    SAI_MY_MAC_ATTR_START,
+
+    /**
+     * @brief Priority
+     *
+     * Value must be in the range defined in
+     * \[#SAI_MY_MAC_TABLE_ATTR_MINIMUM_PRIORITY,
+     * #SAI_MY_MAC_TABLE_ATTR_MAXIMUM_PRIORITY\]
+     * (default = #SAI_MY_MAC_TABLE_ATTR_MINIMUM_PRIORITY)
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_MY_MAC_ATTR_PRIORITY = SAI_MY_MAC_ATTR_START,
+
+    /**
+     * @brief Object id of the My MAC table
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_MY_MAC_TABLE
+     */
+    SAI_MY_MAC_ATTR_TABLE_ID,
+
+    /**
+     * @brief Associated Port, LAG object id,
+     * if not specified any port will match
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_LAG
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_MY_MAC_ATTR_PORT_ID,
+
+    /**
+     * @brief Associated Vlan,
+     * if not specified any vlan will match
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_VLAN
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_MY_MAC_ATTR_VLAN_ID,
+
+    /**
+     * @brief MAC Address
+     *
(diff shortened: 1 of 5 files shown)


kcudnik:
why this needs separate object type instead of having a table/list of mac's in rif object ?

lguohan:
i have the same question as @kcudnik, can you explain?

inc/saimymac.h
```diff
@@ -0,0 +1,319 @@
...
+ */
+typedef sai_status_t (*sai_get_my_mac_attribute_fn)(
+        _In_ sai_object_id_t my_mac_id,
+        _In_ uint32_t attr_count,
+        _Inout_ sai_attribute_t *attr_list);
+
+/**
+ * @brief My MAC table attribute IDs
+ */
+typedef enum _sai_my_mac_table_attr_t
```
rck-innovium, line 190: Can we simplify this table to:

```
/**
 * @brief My MAC entry attribute IDs
 */
typedef enum _sai_my_mac_attr_t
{
    /**
     * @brief Start of attributes
     */
    SAI_MY_MAC_ATTR_START,

    
    /**
     * @brief Object id of the My MAC table
     *
     * @type sai_object_id_t
     * @objects SAI_OBJECT_TYPE_MY_MAC_TABLE
     */
    SAI_MY_MAC_ATTR_TABLE_ID =  SAI_MY_MAC_ATTR_START,

    /**
     * @brief Associated Port, LAG object id,
     * if not specified any port will match
     *
     * @type sai_object_id_t
     * @flags CREATE_ONLY
     * @objects SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_LAG
     * @allownull true
     * @default SAI_NULL_OBJECT_ID
     */
    SAI_MY_MAC_ATTR_PORT_ID,

    /**
     * @brief Associated Vlan Id,
     * if not specified any vlan id will match
     *
     * @type sai_uint16_t
     * @flags CREATE_ONLY
     * @isvlan true
     * @default 0
     */
    SAI_MY_MAC_ATTR_VLAN_ID,

    /**
     * @brief MAC Address
     *
     * @type sai_mac_t
     * @flags CREATE_ONLY
     * @default vendor
     */
    SAI_MY_MAC_ATTR_MAC_ADDRESS,

    /**
     * @brief MAC Address Mask
     *
     * @type sai_mac_t
     * @flags CREATE_ONLY
     * @default vendor
     */
    SAI_MY_MAC_ATTR_MAC_ADDRESS_MASK,
    
    /**
     * @brief End of attributes
     */
    SAI_MY_MAC_ATTR_END,

    /** Custom range base value */
    SAI_MY_MAC_ATTR_CUSTOM_RANGE_START = 0x10000000,

    SAI_MY_MAC_ATTR_CUSTOM_RANGE_END

} sai_my_mac_attr_t;
```

rck-innovium (changes requested):

inc/saimymac.h
ashutosh-agrawal: I agree with the proposed simplification.

inc/saimymac.h
```diff
@@ -0,0 +1,319 @@
...
+    SAI_MY_MAC_ATTR_MAC_ADDRESS_MASK,
+
+    /**
+     * @brief Enable IPv4 processing
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default true
+     */
+    SAI_MY_MAC_ATTR_V4_ENABLE,
```
ashutosh-agrawal, line 111: Since V4 and V6 enable flags are in the RIF as well, please clarify what would be the expected behavior in case of a conflict between my_mac and rif flags.

inc/saimymac.h
```diff
@@ -0,0 +1,319 @@
...
+     */
+    SAI_MY_MAC_TABLE_ATTR_MAXIMUM_PRIORITY,
+
+    /**
+     * @brief Number of My MAC entries installed on the switch
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_MY_MAC_TABLE_ATTR_INSTALLED_ENTRIES,
```
ashutosh-agrawal, line 219: Number of installed/available entries is a switch-level attribute. Please take a look at other SAI_SWITCH_ATTR_AVAILABLE_*** attributes in saiswitch.h

SathishKumarTR-Dell:
With this proposed new table, can a application bypass creation of the VR and RIF objects but still expect L3 unicast route termination for packets from a SAI pipeline?

inc/saimymac.h
erohsik: My MAC entry determines if packet gets L3 forwarded or L2 forwarded.
Once the packet is in the L3 forwarding pipeline, RIF L3 admin state functionality
will take effect as it does currently.

erohsik:
> 
> 
> i have the same question as @kcudnik, can you explain?

The MAC address that can be programmed in this table need not belong to a particular port.

erohsik:
> 
> 
> With this proposed new table, can a application bypass creation of the VR and RIF objects but still expect L3 unicast route termination for packets from a SAI pipeline?

Entries in this table only determine L2 vs L3 processing. Once in the L3 pipeline, RIF and VR objects would still be needed.

inc/saimymac.h
erohsik: Would it be ok to pass in the flags, and let the SAI implementation then program the entry (or not) based on the flags.

inc/saimymac.h
```diff
@@ -0,0 +1,303 @@
...
+     */
+    SAI_MY_MAC_TABLE_ATTR_END,
+
+    /** Custom range base value */
+    SAI_MY_MAC_TABLE_ATTR_CUSTOM_RANGE_START = 0x10000000,
+
+    /** End of custom range base */
+    SAI_MY_MAC_TABLE_ATTR_CUSTOM_RANGE_END
+
+} sai_my_mac_table_attr_t;
```
kcudnik, line 233: all the attributes on sai_my_mac_table_attr_t are READ_ONLY, why then this object is needed at all, if nothing can be actually SET or CREATE here ?

inc/saimymac.h
```diff
@@ -0,0 +1,303 @@
...
+    /**
+     * @brief Start of attributes
+     */
+    SAI_MY_MAC_ATTR_START,
+
+    /**
+     * @brief Priority
+     *
+     * Value must be in the range defined in
+     * \[#SAI_MY_MAC_TABLE_ATTR_MINIMUM_PRIORITY,
```
kcudnik, line 44: will that priority range vary between different vendor/asics ? if not maybe it could be enum ?

inc/saitypes.h
```diff
@@ -279,6 +279,8 @@ typedef enum _sai_object_type_t
     SAI_OBJECT_TYPE_FINE_GRAINED_HASH_FIELD  = 94,
     SAI_OBJECT_TYPE_SWITCH_TUNNEL            = 95,
     SAI_OBJECT_TYPE_MY_SID_ENTRY          = 96,
```
kcudnik, line 281: please also fix spacing on SAI_OBJECT_TYPE_MY_SID_ENTRY          = 96

kcudnik (changes requested):

rlhui:
As discussed in meeting, a md doc will be helpful to understand the use cases and call sequences.  Also it's not yet clear if additional ipv4/ipv6 enable flags are needed here.

inc/saimymac.h
erohsik: Original intent was to group these attributes under a table object.. Moved these attributes to the switch..

erohsik:
@JaiOCP, @rck-innovium , @ashutosh-agrawal  please review..

rck-innovium (approved):

ashutosh-agrawal (approved):

tushar-ty (approved):

inc/saimymac.h
erohsik: It may vary between vendors. Leaving it as is..

kcudnik (approved):

PR #1217, opened 2021-04-05
Port attributes for MDIX (Medium Dependent Interface Crossover), Media config and for Latency

tarakreddy1984:

PR #1244, opened 2021-06-01
Copper PHY Attributes for MDIX, Module type, Autoneg mode, Latency mode, Dual Media mode

SidharajU:
The original PR for this change was https://github.com/opencomputeproject/SAI/pull/1217. I have created a new PR based on the feedback received from SAI community and will abandon the original PR.


inc/saiport.h
```diff
@@ -1713,6 +1796,74 @@ typedef enum _sai_port_attr_t
...
+     */
+    SAI_PORT_ATTR_EEE_LATENCY_MODE,
+
+    /**
+     * @brief Configure Fiber module type
+     *
+     * @type sai_port_module_type_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PORT_MODULE_TYPE_1000BASE_X
+     * @condition SAI_PORT_ATTR_SPEED == 1000 or SAI_PORT_ATTR_SPEED == 100 or SAI_PORT_ATTR_MEDIA_TYPE == SAI_PORT_ATTR_MEDIA_TYPE_FIBER
```
kcudnik, line 1854: value for SAI_PORT_ATTR_MEDIA_TYPE  must be one of SAI_PORT_MEDIA_TYPE_*

kcudnik: also condition attribute must be mandatory on create, or you can change this to validonly

SidharajU: So there is no way to have conditions with multiple attributes. As discussed last week, we would like to allow this attribute when media type is set to fiber and speed is either 1000 or 100. How do I do that?

kcudnik (approved):

lguohan:
@jleveque , can you check this pr?

jleveque (changes requested):
Proposal should be written in Markdown language. Also please fix conflicts.

SidharajU:
There were some issues with the merge so I am creating a new PR. I don't know how to recover this PR now.

SidharajU:
This PR has merge issues so created another PR
https://github.com/opencomputeproject/SAI/pull/1248

PR #1247, opened 2021-06-03
Allow modification of nexthop member sequence id.

mikeberesford:
Allow application to modify the sequence ID of the nexthop group member. This allows the application layer to use stateless incrementing indices to specify the ordering while allowing for group modification.


kcudnik (approved):

marian-pritsak (approved):

marian-pritsak (approved):

mikeberesford:
If there aren't any pending objections, can we merge this PR? (are any other reviews needed?)

PR #1248, opened 2021-06-04
Copper PHY attributes

SidharajU:


kcudnik (approved):

inc/saiport.h
```diff
@@ -1813,6 +1896,74 @@ typedef enum _sai_port_attr_t
...
+     * @brief Configure EEE latency mode
+     *
+     * False: Fixed latency
+     * True: Variable latency
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_PORT_ATTR_EEE_LATENCY_MODE,
```
jleveque, line 1946: Naming is a bit confusing for a boolean value. `SAI_PORT_ATTR_EEE_LATENCY_MODE = True` isn't clear what it does. With a name like `SAI_PORT_ATTR_EEE_LATENCY_MODE` it seems more suited for an enum. If a boolean is desired, maybe a more understandable name would be `SAI_PORT_ATTR_EEE_LATENCY_VARIABLE`

doc/macsec-gearbox/Copper_PHY_Attributes_Proposal.md
```diff
@@ -0,0 +1,305 @@
...
+    /**  Cable state no faults */
+    SAI_PORT_CABLE_PAIR_STATE_OK,
+
+    /**  Cable state open */
+    SAI_PORT_CABLE_PAIR_STATE_OPEN,
+
+    /**  Cable state short */
+    SAI_PORT_CABLE_PAIR_STATE_SHORT,
+
+    /**  Cable state cross talk */
```
ds952811, line 231: Is it expected that OPENSHORT will not be available in the PAI?
i.e.
#define BCM_PORT_CABLE_STATE_OK         _SHR_PORT_CABLE_STATE_OK
#define BCM_PORT_CABLE_STATE_OPEN       _SHR_PORT_CABLE_STATE_OPEN
#define BCM_PORT_CABLE_STATE_SHORT      _SHR_PORT_CABLE_STATE_SHORT
#define BCM_PORT_CABLE_STATE_OPENSHORT  _SHR_PORT_CABLE_STATE_OPENSHORT
#define BCM_PORT_CABLE_STATE_CROSSTALK  _SHR_PORT_CABLE_STATE_CROSSTALK
#define BCM_PORT_CABLE_STATE_UNKNOWN    _SHR_PORT_CABLE_STATE_UNKNOWN
#define BCM_PORT_CABLE_STATE_COUNT      _SHR_PORT_CABLE_STATE_COUNT

/* bcm_port_cable_diag_t */
typedef _shr_port_cable_diag_t bcm_port_cable_diag_t;

meta/aspell.en.pws
```diff
@@ -38,6 +38,7 @@ ffff
 FPGAs
 fprintf
 functionalities
+FX
```
kcudnik, line 41: could be in acronym as full duplex ?

kcudnik (approved):

inc/saiport.h
```diff
@@ -388,6 +388,136 @@ typedef enum _sai_port_connector_failover_mode_t
...
+
+    /**  Cable type CAT6 */
+    SAI_PORT_CABLE_TYPE_CAT6,
+
+    /**  Cable type CAT6A */
+    SAI_PORT_CABLE_TYPE_CAT6A,
+
+    /**  Cable type CAT7 */
+    SAI_PORT_CABLE_TYPE_CAT7,
+} sai_port_cable_type_t;
```
ds952811, line 519: In the case of legacy Broadcom PHY, the cable-length is only valid when the state is OK, and by having state and cable-length tested in 2 separated calls could lead to a race condition, maybe we should deal with them in one single call, agree?

SidharajU: No. I don't think we can do it in single call. I was thinking if we should add a @validonly for the cable length but then we can't support the other PHYs which do not require, cable state to be OK to get the cable length. I think user can check the cable state before querying the cable length on the legacy PHYs.

doc/macsec-gearbox/Copper_PHY_Attributes_Proposal.md
SidharajU: No. We don't support OPENSHORT. You can extend the enum if you want to support it once this is approved.

meta/aspell.en.pws
SidharajU: Looks like you don't need this change as you approved it.

doc/macsec-gearbox/Copper_PHY_Attributes_Proposal.md
```diff
@@ -0,0 +1,308 @@
...
+ * @brief Configure EEE latency mode
+ *
+ * False: Fixed latency
+ * True: Variable latency
+ *
+ * @type bool
+ * @flags CREATE_AND_SET
+ * @default false
+ */
+SAI_PORT_ATTR_EEE_LATENCY_MODE,
```
lguohan, line 136: there is already SAI_PORT_ATTR_EEE_ENABLE attribute. How is this new attribute related to the EEE_ENABLED attribute?

inc/saiport.h
```diff
@@ -1813,6 +1946,102 @@ typedef enum _sai_port_attr_t
...
+    SAI_PORT_ATTR_CABLE_PAIR_STATE,
+
+    /**
+     * @brief Get cable pair length
+     * Returns cable pair length sequentially from list index 0 to n (n = number of pairs - 1)
+     *
+     * @type sai_s32_list_t
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_CABLE_PAIR_LENGTH,
```
lguohan, line 2033: what is the unit of the cable pair length?

lguohan: how will user know how many pairs a cable has?

inc/saiport.h
```diff
@@ -1813,6 +1946,102 @@ typedef enum _sai_port_attr_t
...
+
+    /**
+     * @brief Configure cable type to check the cable status
+     *
+     * @type sai_port_cable_type_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PORT_CABLE_TYPE_UNKNOWN
+     * @validonly SAI_PORT_ATTR_MEDIA_TYPE == SAI_PORT_MEDIA_TYPE_COPPER
```
lguohan, line 2041: is it true, this attribute is only applicable to 1000Base-T?

inc/saiport.h
lguohan: is it true, this attribute is only applicable to 1000Base-T? If yes, can you add comment?

inc/saiport.h
```diff
@@ -1813,6 +1946,102 @@ typedef enum _sai_port_attr_t
...
+    SAI_PORT_ATTR_DUAL_MEDIA,
+
+    /**
+     * @brief Read cable pair state.
+     * Returns pair states sequentially from list index 0 to n (n = number of pairs - 1)
+     *
+     * @type sai_s32_list_t sai_port_cable_pair_state_t
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_CABLE_PAIR_STATE,
```
lguohan, line 2024: is it true, this attribute is only applicable to 1000Base-T? If yes, can you add comment?

doc/macsec-gearbox/Copper_PHY_Attributes_Proposal.md
SidharajU: I need to update the documentation. This attribute has been changed to SAI_PORT_ATTR_EEE_LATENCY_VARIABLE. Its updated in saiport.h. Some PHYs support variable latency mode. This attribute will be used to enable variable latency mode vs fixed mode. SAI_PORT_ATTR_EEE_ENABLE will be used to enable/disable EEE.

inc/saiport.h
SidharajU: I will update the documentation. It will be in meters.
list count will provide the number of pairs.
This is not only applicable for 1000BaseT. This can be applicable to speeds up to 10G or can be extended to 25G

inc/saiport.h
SidharajU: This is not only applicable for 1000BaseT. This can be applicable to speeds up to 10G or can be extended to 25G.

rlhui:
The PR will be revised to reduce a few attributes.

doc/macsec-gearbox/Copper_PHY_Attributes_Proposal.md
SidharajU: Will add a condition to check SAI_PORT_ATTR_EEE_ENABLE

jleveque (approved):

PR #1249, opened 2021-06-09
Enum reordering

shri-khare:


kcudnik:
what's wrong with current order? why this change?

tushar-ty:
> what's wrong with current order? why this change?

enum type values like SAI_NATIVE_HASH_FIELD_SRC_IPV4 should not change from one spec version to another as that can break any warm upgrade (ISSU).

B/w v1.6.x and v1.8.x some enums are inserted in the middle vs at the end.

subtype enums should always be added to the end of the list.

rlhui:
will you be able to add description of this PR.  Currently it says "No description provided."

shri-khare:
While trying to fix the sanity checks failing here, I accidentally ended up posting another pull request https://github.com/opencomputeproject/SAI/pull/1250. That has the latest changes with all the sanity checks passing. Am closing this request thus in favor of pull request 1250.

PR #1250, opened 2021-06-15
2 enum reordering

shri-khare:
Consider a scenario where an application (adapter host) using SAI spec X
(say) warmboots to an application using new SAI spec Y (say).

If an enum is extended between X and Y with fields inserted "in between"
rather than "at the end" of the enum, the old fields can get "renumbered".
However, the SAI implementation (adapter) continues to return old enums
for previously returned objects thereby breaking the warmboot.

Fix this by moving  the newly inserted fields "to the end"
(this could mean at the end of _END block) to keep the enum field values
the same.

### DIFF
--- a/inc/saihash.h
+++ b/inc/saihash.h
@@ -62,50 +62,6 @@ typedef enum _sai_native_hash_field_t
     /** Native hash field inner destination IP */
     SAI_NATIVE_HASH_FIELD_INNER_DST_IP,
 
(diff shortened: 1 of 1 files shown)


kcudnik (changes requested):
you need to discuss this on weekly SAI meeting, this seems like a lot of shuffling and seems not binary backward compatible

PR #1258, opened 2021-06-28
Trap type changes

dkourkouzelis:
- Added UDP port 6784 for BFD and BFDv6 trap types
- Added new LDP trap type

### DIFF
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -359,25 +359,35 @@ typedef enum _sai_hostif_trap_type_t
     SAI_HOSTIF_TRAP_TYPE_BGPV6 = 0x00004004,
 
     /**
-     * @brief BFD traffic (UDP dst port == 3784 or UDP dst port == 4784 or
-     * UDP dst port == 6784) to local router IP address (default packet
-     * action is drop)
+     * @brief BFD traffic (UDP dst port == 3784 or UDP dst port == 4784) to local
+     * router IP address (default packet action is drop)
      */
     SAI_HOSTIF_TRAP_TYPE_BFD = 0x00004005,
 
     /**
-     * @brief BFD traffic (UDP dst port == 3784 or UDP dst port == 4784 or
-     * UDP dst port == 6784) to local router IP address (default packet
-     * action is drop)
+     * @brief BFDV6 traffic (UDP dst port == 3784 or UDP dst port == 4784) to
+     * local router IP address (default packet action is drop)
      */
     SAI_HOSTIF_TRAP_TYPE_BFDV6 = 0x00004006,
+
+    /**
+     * @brief Micro BFD traffic (UDP dst port == 6784) to local
+     * router IP address (default packet action is drop)
+     */
+    SAI_HOSTIF_TRAP_TYPE_BFD_MICRO = 0x00004007,
+
+    /**
+     * @brief Micro BFDV6 traffic (UDP dst port == 6784) to local
+     * router IP address (default packet action is drop)
+     */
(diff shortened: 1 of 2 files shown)


kcudnik:
please fix error: WARNING: Word 'LDP' is misspelled or is acronym, add to acronyms.txt? saihostif.h 376:     * @brief LDP traffic (TCP src port == 646 or TCP dst port == 646) to local

kcudnik (approved):

rlhui:
@rck-innovium - please help review/sign off.  Thanks.

inc/saihostif.h
```diff
@@ -359,17 +359,26 @@ typedef enum _sai_hostif_trap_type_t
     SAI_HOSTIF_TRAP_TYPE_BGPV6 = 0x00004004,
 
     /**
-     * @brief BFD traffic (UDP dst port == 3784 or UDP dst port == 4784) to local
-     * router IP address (default packet action is drop)
+     * @brief BFD traffic (UDP dst port == 3784 or UDP dst port == 4784 or
+     * UDP dst port == 6784) to local router IP address (default packet
+     * action is drop)
```
rck-innovium, line 364: Can we add a new trap type like below instead of modifying the existing traps.

Advantages:

* Does not change behavior for existing implementations

* Allows optimization of h/w resources when end user needs only BFD and not micro-BFD.


```
/* 
* @brief BFD traffic (UDP dst port == 6784) to local
* router IP address (default packet action is drop)
*/
SAI_HOSTIF_TRAP_TYPE_BFD_MICRO,
```

Similarly for BFDv6

rck-innovium (changes requested):

inc/saihostif.h
dkourkouzelis: I wish you had brought that up during the meeting today. I feel like the rest of the community should be aware of this request given that they all know we decided on using the same trap type.

In regards to your points:
a) I see it differently, given that someone will get BFD over LAG for free pretty much and it's a common use case rather than not.
b)If someone doesn't need BFD over LAG then they don't need to implement it, correct? An entry with type+port will have to be added to the hardware anyway, it's not easy to call out what is being optimized.

Perhaps we need someone to decide whether we need another meeting for this to a) rediscuss with the community, b) go with your proposal or c) go with the original one.

dkourkouzelis: @rhui Could you please share your thoughts on this?

rck-innovium: It would be good if we can discuss this in the next community meeting. 

In most of the TCAM based Host Tables, we would need one ACL entry per L4 port match.

dkourkouzelis: Made changes per your recommendation.

anshuv-mfst:
Hi @kcudnik - could you please help with merge, thanks.

kcudnik:
> Hi @kcudnik - could you please help with merge, thanks.

was that already approved on weekly SAI meeting?

anshuv-mfst:
> > Hi @kcudnik - could you please help with merge, thanks.
> 
> was that already approved on weekly SAI meeting?

Hi @kcudnik - yes, the PR is reviewed in SAI subgroup meeting. 
Hi @dkourkouzelis - could you please confirm if all review comments are addressed.

rck-innovium (approved):
lgtm

JafarSeyedi:
Hi, 

with new changes, we have two problems:

1. when changing copp config file to add isis, the sonic will crash and dockers will stop.
2, eigrp does not work after these changes. we checked eigrp in sonic with activating it in frr and it was working, but currently, the messages did not pass to frr.

PR #1259, opened 2021-06-29
Assign values to enum

shri-khare:
Problem
========

Consider a scenario where an application (adapter host) using SAI spec X
(say) warmboots to an application using new SAI spec Y (say).

If an enum is extended between X and Y with fields inserted "in between"
rather than "at the end" of the enum, the old fields can get "renumbered".
However, the SAI implementation (adapter) continues to return old enums
for previously returned objects thereby breaking the warmboot.

Possible approaches
===================

We discussed this during June 24, 2021 weekly SAI meeting and observed:

- This can e avoided by mandating that any new inserts to an enum will be
"at the end" (this could mean at the end of _END block). This will keep
the enum field values the same.
- However, that will come at the cost of readability: newly added fields
may be related to existing fields and thus for readability, those fields
belong togeter. Adding those to the end scatters the related fields
across the enum.

Solution
=========

Thus, we agreed to go with the below approach:
- Every enum field will have an explicit integer value assigned to it.
- Insertions can be "in between" (maintains readability) but the next
available unused integer value would be used for the new insertions.
- Thus, once an integer is assigned to an enum field, it does not change
with spec revision thereby not breaking warmboot.
- Between SAI spec 1.6.5 to 1.8.1, saihash.h and saiacl.h had new enms
inserted "in between".

Those will be "fixed" by assigning integer values to enum fields.

This fix will be applied to top of the sai.git tree.
This fix will then be ported to 1.8.1 and 1.8.2 will be released.

Towards that end, these patches assigns integer values to enums in 
saihash.h and saiacl.h that were changed from 1.6.5 to 1.8.1


shri-khare:
The meta checks fail here: ./meta/saisanitycheck.c

 META_ENUM_ASSERT_FAIL(emd, "values are not increasing by 1: last: %d current: %d, should be marked as @flags?", last, value);

This is expected as we are now introducing consecutive enum fields that are not increasing by 1 by design.
Perhaps, we should remove this check from saisanitycheck.c?

kcudnik (changes requested):
This problem is addressed already here: https://github.com/opencomputeproject/SAI/pull/1256
This change will make sure enum numbers don't change over time, and no need to explicit numbering

There is no need for gaps in enums like SAI_NATIVE_HASH_FIELD_IPV6_FLOW_LABEL vs SAI_NATIVE_HASH_FIELD_NONE. Old enums should exists, and should be marked as depreacakted.

kcudnik:
> This is expected as we are now introducing consecutive enum fields that are not increasing by 1 by design.
> Perhaps, we should remove this check from saisanitycheck.c?

no, this check must stay, to make sure enums are consistent, there is no need to introduce gaps in enums.

shri-khare:
Thanks Kamil. 

Posted few questions about this here: https://github.com/opencomputeproject/SAI/pull/1256#issuecomment-870701601

kcudnik:
> Posted few questions about this here: [#1256 (comment)](https://github.com/opencomputeproject/SAI/pull/1256#issuecomment-870701601)

done

inc/saiacl.h
```diff
@@ -2370,12 +2370,12 @@ typedef enum _sai_acl_entry_attr_t
...
      * @default disabled
      */
-    SAI_ACL_ENTRY_ATTR_FIELD_TAM_INT_TYPE,
+    SAI_ACL_ENTRY_ATTR_FIELD_TAM_INT_TYPE = SAI_ACL_ENTRY_ATTR_FIELD_START + 0x152,
 
     /**
      * @brief End of Rule Match Fields
      */
-    SAI_ACL_ENTRY_ATTR_FIELD_END = SAI_ACL_ENTRY_ATTR_FIELD_TAM_INT_TYPE,
+    SAI_ACL_ENTRY_ATTR_FIELD_END = SAI_ACL_ENTRY_ATTR_FIELD_DST_IPV6_WORD0,
```
kcudnik, line 2378: previous was correct, i assume that SAI_ACL_ENTRY_ATTR_FIELD_END  will shift (can only increase from SAI to SAI)

inc/saiacl.h
```diff
@@ -2920,12 +2920,12 @@ typedef enum _sai_acl_entry_attr_t
...
      * @default disabled
      */
-    SAI_ACL_ENTRY_ATTR_ACTION_SET_VRF,
+    SAI_ACL_ENTRY_ATTR_ACTION_SET_VRF = SAI_ACL_ENTRY_ATTR_ACTION_START + 0x31,
 
     /**
      * @brief End of Rule Actions
      */
-    SAI_ACL_ENTRY_ATTR_ACTION_END = SAI_ACL_ENTRY_ATTR_ACTION_SET_VRF,
+    SAI_ACL_ENTRY_ATTR_ACTION_END = SAI_ACL_ENTRY_ATTR_ACTION_ADD_VLAN_PRI,
```
kcudnik, line 2928: previous was correct, i assume that SAI_ACL_ENTRY_ATTR_ACTION_END will shift (can only increase from SAI to SAI)

inc/saiacl.h
```diff
@@ -1383,12 +1383,12 @@ typedef enum _sai_acl_table_attr_t
...
      * @default false
      */
-    SAI_ACL_TABLE_ATTR_FIELD_TAM_INT_TYPE,
+    SAI_ACL_TABLE_ATTR_FIELD_TAM_INT_TYPE = SAI_ACL_TABLE_ATTR_FIELD_START + 0x152,
 
     /**
      * @brief End of ACL Table Match Field
      */
-    SAI_ACL_TABLE_ATTR_FIELD_END = SAI_ACL_TABLE_ATTR_FIELD_TAM_INT_TYPE,
+    SAI_ACL_TABLE_ATTR_FIELD_END = SAI_ACL_TABLE_ATTR_FIELD_DST_IPV6_WORD0,
```
kcudnik, line 1391: previous was correct, i assume that SAI_ACL_TABLE_ATTR_FIELD_END will shift (can only increase from SAI to SAI)

inc/saiacl.h
kcudnik: addressed here: https://github.com/opencomputeproject/SAI/pull/1256/files#diff-799f04554f369b93c06b1c438c55c5a9f64252d1a2defb1cbaf1b8bf5222ecacR190

inc/saiacl.h
kcudnik: addressed here: https://github.com/opencomputeproject/SAI/pull/1256/files#diff-799f04554f369b93c06b1c438c55c5a9f64252d1a2defb1cbaf1b8bf5222ecacR190

inc/saiacl.h
kcudnik: addressed here: https://github.com/opencomputeproject/SAI/pull/1256/files#diff-799f04554f369b93c06b1c438c55c5a9f64252d1a2defb1cbaf1b8bf5222ecacR190

kcudnik:
Actually this PR is not needed, i addressed this in comment https://github.com/opencomputeproject/SAI/pull/1259#pullrequestreview-694659449, after checking #1256 those checks will be enforced

kcudnik:
if you want to keep those values, 
```
AI_ACL_ENTRY_ATTR_FIELD_IP_PROTOCOL = SAI_ACL_ENTRY_ATTR_FIELD_START + 0x1b,
```
it would be better to reorder those enums to achieve desired values, it would have better looking, and after enum lock PR merged, you never have to assign them explicitly, since they always will be locked

inc/saihash.h
```diff
@@ -54,123 +54,123 @@ typedef enum _sai_native_hash_field_t
...
+    SAI_NATIVE_HASH_FIELD_INNER_DST_IP = 0x00000003,
 
     /**
      * @brief Native hash field source IPv4.
      *
      * Also, refers to the outer source IPv4
      * in case for encapsulated packets
      */
-    SAI_NATIVE_HASH_FIELD_SRC_IPV4,
+    SAI_NATIVE_HASH_FIELD_SRC_IPV4 = 0x00000019,
```
kcudnik, line 71: why this enum must bat a this place and not at the end when it was previously there? SRC/DST enums are grouped together even in this configuration so what the point to not make those enums not continuous ?

kcudnik:
This change https://github.com/opencomputeproject/SAI/pull/1274 allows to not ordered enum values and should accept your changes

kcudnik (approved):

rlhui:
@JaiOCP , @tushar-ty , @marian-pritsak , please help to review/approve as well.  Thanks.

rlhui:
> @JaiOCP , @tushar-ty , @marian-pritsak , please help to review/approve as well. Thanks.

@itaibaz , @ashutosh-agrawal , @rck-innovium - please feel free to review/approve as well.  Thanks.

itaibaz (approved):

JaiOCP (commented):
How do we know what's the last enum value to be used for new enums

kcudnik:
> How do we know what's the last enum value to be used for new enums

you guess XD if you guess wrong, metadata validator will show you previous and current value

ashutosh-agrawal (approved):

JaiOCP:
> > How do we know what's the last enum value to be used for new enums
> 
> you guess XD if you guess wrong, metadata validator will show you previous and current value

Thanks Kamil,
Would it make sense to have a "XXX_ATTR_LAST_ENUM" always updated with last value so that programmer do not have to hunt or guess the value. 
How would MD checker catch a value if it is indeed the last but not contagious to the previous last enum value.

kcudnik:
> How would MD checker catch a value if it is indeed the last but not contagious to the previous last enum value.

magic ! :P all enums must be contagious, even in respective ranges (there are some exceptions like sai_status_t or bit field based enums, and if enum value is duplicate then metadata will throw error on that, also not always you need to hunt last value, if you are adding enum at the end of list, you dont need to specify it explicitly, only when you are inserting enum in the middle

im not sure about concept with XXX_ATTR_LAST_ENUM, if you mean that this enum is artificially generated and explicitly set to a last value, then it will not work, since not always you want to add enum at the end, there are ranges based on 0x1000 index for example in hostif and acl, and you want to add specific value at the end of that specific range

JaiOCP:
> > How would MD checker catch a value if it is indeed the last but not contagious to the previous last enum value.
> 
> magic ! :P all enums must be contagious, even in respective ranges (there are some exceptions like sai_status_t or bit field based enums, and if enum value is duplicate then metadata will throw error on that, also not always you need to hunt last value, if you are adding enum at the end of list, you dont need to specify it explicitly, only when you are inserting enum in the middle
> check here https://github.com/opencomputeproject/SAI/blob/master/meta/saisanitycheck.c#L5090
> 
> im not sure about concept with XXX_ATTR_LAST_ENUM, if you mean that this enum is artificially generated and explicitly set to a last value, then it will not work, since not always you want to add enum at the end, there are ranges based on 0x1000 index for example in hostif and acl, and you want to add specific value at the end of that specific range

I like magic :)
Yes, I was thinking of autogenerated last enum value but I see your point. It will indeed get tricky for ranges.

JaiOCP (approved):

kcudnik:
@JaiOCP @lguohan @shri-khare i make  git diff 2cc5ba0 origin/master where that commit is branch (tag: v1.6.5) and i found other attributes added in the middle besides acl:

attributes between 1.6.5 and origin/master that caused shift in enums:
- many ACL attributes an HASH (addressed in this PR)
- many SAI_SWITCH_ATTR_* added in the middle
- many shifts in SAI_MACSEC_SC_ATTR_* 
- many shifts in SAI_MACSEC_SA_ATTR_*
- segmentroute api was rename to SRV6, cause to shift many SAI_NEXT_HOP_ATTR_* 
- many SAI_PORT_INTERFACE_TYPE_* added in the middle 
- many SAI_PORT_ATTR_* added in the middle 
- many  SAI_PORT_STAT_* added in the middle
- many  SAI_TUNNEL_ATTR_* added in the middle

many other attributes were added but at the end of list, so does not break anything

Should this be fixed before merging this? or you just want acl and hash, ignoring other changes? we can actually make fix for that even after marge this, but i **will need a base commit ID from which we want to be binary backward compatible**, but that fix would need to be on master, and from master we would check compatibility. There is also some issue, that entire segment route object was renamed to srv6 object in the meantime

JaiOCP:
> @JaiOCP @lguohan @shri-khare i make git diff [2cc5ba0](https://github.com/opencomputeproject/SAI/commit/2cc5ba070a126dab9d29dca60dcb3ea01e9e77be) origin/master where that commit is branch (tag: v1.6.5) and i found other attributes added in the middle besides acl:
> 
> attributes between 1.6.5 and origin/master that caused shift in enums:
> 
> * many ACL attributes an HASH (addressed in this PR)
> * many SAI_SWITCH_ATTR_* added in the middle
> * many shifts in SAI_MACSEC_SC_ATTR_*
> * many shifts in SAI_MACSEC_SA_ATTR_*
> * segmentroute api was rename to SRV6, cause to shift many SAI_NEXT_HOP_ATTR_*
> * many SAI_PORT_INTERFACE_TYPE_* added in the middle
> * many SAI_PORT_ATTR_* added in the middle
> * many  SAI_PORT_STAT_* added in the middle
> * many  SAI_TUNNEL_ATTR_* added in the middle
> 
> many other attributes were added but at the end of list, so does not break anything
> 
> Should this be fixed before merging this? or you just want acl and hash, ignoring other changes? we can actually make fix for that even after marge this, but i **will need a base commit ID from which we want to be binary backward compatible**, but that fix would need to be on master, and from master we would check compatibility. There is also some issue, that entire segment route object was renamed to srv6 object in the meantime

Given that no one has complained for these attributes and warm boot, I would say its safe to fix these attributes while we are at it.

@lguohan What do you think?

kcudnik:
I wrote a tool to list those attributes, and this also will be later doing validation on each PR. couple infos:
1.6.5 is on 2cc5ba0, diverges from master on ec7e055 (1.6.0)
1.8 is on 5254200, diverges from master on 5a98bc3 (1.8.0)

so diff ec7e055 to 2cc5ba0 (master to 1.6.5) (excluding acl_table, acl_entry, tam_int and macsec:
```
ERROR: check ! SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE value is 0x00000014, but on was 0x10000001 on commit ec7e055
ERROR: check ! SAI_VLAN_ATTR_TAM_OBJECT value is 0x00000015, but on was 0x10000002 on commit ec7e055
```
this vlan can be skipped, i remember bringing it back from custom range to attr range

5a98bc3 to 5254200 has no changes 

and ec7e055 to origin/master (total 347 enum values affected, too much to list here but those are numbers:
```
     31 SAI_ACL_ACTION_TYPE_
      1 SAI_LAG_ATTR_SYSTEM_PORT_AGGREGATE_ID
     21 SAI_NATIVE_HASH_FIELD_
      9 SAI_NEXT_HOP_ATTR_
    114 SAI_PORT_
     12 SAI_PORT_INTERFACE_TYPE_
      1 SAI_STATS_MODE_READ
      1 SAI_STATS_MODE_READ_AND_CLEAR
    152 SAI_SWITCH_
      1 SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_ACTION_TUNNEL_ID
      1 SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_SRC_IP
      1 SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TUNNEL_TYPE
      1 SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE
      1 SAI_VLAN_ATTR_TAM_OBJECT
```

PR #1262, opened 2021-06-30
Add enable/disable option for forwarding pause frame to SAI port

Pterosaur:
Enhance the flow control APIs including enable/disable forwarding flow control frame.

Signed-off-by: Ze Gan <ganze718@gmail.com>


inc/saimacsec.h
```diff
@@ -292,6 +292,31 @@ typedef enum _sai_macsec_attr_t
      */
     SAI_MACSEC_ATTR_AVAILABLE_MACSEC_SA,
 
+    /**
+     * @brief Terminate/Forward Pause Frames.
+     *
+     * If enabled (true) pause frames are forwarded to the user application.
+     * In normal mode (false) pause frames are terminated and discarded within the MAC
```
kcudnik, line 299: missing dot (.) at the end of sentence

Pterosaur: Thanks for reminding.

Pterosaur:
This PR need to be discussed with Guohan further, so just draft it again.

kcudnik (approved):

inc/saimacsec.h
```diff
@@ -292,6 +292,31 @@ typedef enum _sai_macsec_attr_t
...
+     * @brief Terminate/Forward Pause Frames.
+     *
+     * If enabled (true) pause frames are forwarded to the user application.
+     * In normal mode (false) pause frames are terminated and discarded within the MAC.
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_MACSEC_ATTR_PAUSE_FRAME_FORWARD_ENABLE,
```
lguohan, line 305: this is not specific to macsec, can you move these two new attributes to SAI_PORT_ATTR ?

inc/saimacsec.h
```diff
@@ -364,6 +377,15 @@ typedef enum _sai_macsec_port_attr_t
...
     SAI_MACSEC_PORT_ATTR_SWITCH_SWITCHING_MODE,
 
+    /**
+     * @brief Control the pause frame
+     *
+     * @type sai_macsec_pause_frame_ctrl_t
+     * @flags CREATE_AND_SET
+     * @default SAI_MACSEC_PAUSE_FRAME_CTRL_ACCEPT
+     */
+    SAI_MACSEC_PORT_ATTR_MACSEC_PAUSE_FRAME_CTRL,
```
lguohan, line 387: it is not related to macsec at all, it should be normal port attribute.

inc/saimacsec.h
```diff
@@ -308,6 +308,19 @@ typedef enum _sai_macsec_attr_t
...
+ * @brief The behavior of pause frame to MACsec chip
+ * Accept means the MACsec chip will realize the pause frame and be controlled by PFC, and Ignore is the opposite,
+ * Forward means the MACsec chip will forward the pause frame to system.
+ */
+typedef enum _sai_macsec_pause_frame_ctrl_t
+{
+    SAI_MACSEC_PAUSE_FRAME_CTRL_ACCEPT,
+    SAI_MACSEC_PAUSE_FRAME_CTRL_ACCEPT_FORWARD,
+    SAI_MACSEC_PAUSE_FRAME_CTRL_INGORE,
+    SAI_MACSEC_PAUSE_FRAME_CTRL_IGNORE_FORWARD,
```
kcudnik, line 321: maybe add empty line between enums just for style and readability

inc/saimacsec.h
Pterosaur: Fixed.

inc/saiport.h
```diff
@@ -1434,6 +1443,26 @@ typedef enum _sai_port_attr_t
...
+    /**
+     * @brief Enable/disable port PFC to all bits
+     *
+     * True means enable PFC and false is opposite
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default true
+     */
+    SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL_ALL_BITS,
```
lguohan, line 1455: no need this attribute.

just add a new enum.

SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_DISABLED around line 192.

Pterosaur: Done

inc/saiport.h
```diff
@@ -226,6 +226,9 @@ typedef enum _sai_port_fec_mode_extended_t
  */
 typedef enum _sai_port_priority_flow_control_mode_t
 {
+    /** Disable PFC for both rx and tx */
+    SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_DISABLED,
```
kcudnik, line 230: FYI, attributes are added in the middle of the enum will cause to shift other attributes and not be binary backward compatible

lguohan: move to last

lguohan: for the attribute, i think we need to explicitly add the enum value as we have discussed @kcudnik it helps to group enum together meanwhile maintain backward compatibility.

Pterosaur: Make sense, should I also reorder other attributes to the end of the enum. But I think it will break a kind of readability.

inc/saiport.h
```diff
@@ -232,6 +232,8 @@ typedef enum _sai_port_priority_flow_control_mode_t
     /** Separate values for RX/TX */
     SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_SEPARATE,
 
+    /** Disable PFC for both rx and tx */
+    SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_DISABLED,
```
dipankar-ba, line 236: Is this attribute necessary?  Without this, PFC could still be disabled using SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL or (SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL_RX and SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL_TX).

inc/saiport.h
```diff
@@ -1032,6 +1034,15 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_GLOBAL_FLOW_CONTROL_MODE,
 
+    /**
+     * @brief Forward or not forward the global flow control frame
```
dipankar-ba, line 1038: When not forwarded, I suppose the flow control frame is terminated in the MAC.  Perhaps better to make that clear.

lguohan: yes, @Pterosaur , can you add comments to make this clear.

inc/saiport.h
```diff
@@ -1434,6 +1445,15 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL_TX,
 
+    /**
+     * @brief Forward or not forward the PFC frame
```
dipankar-ba, line 1449: When not forwarded, I suppose the PFC frames are honored by the switch queues.  Perhaps it is better to make that clear.

inc/saiport.h
lguohan: the problem here is for gearbox, they do not control per bitmap. they can do disable for all priority or none. the per bit control is not supported on the gearbox.

dipankar-ba: Makes sense.  Thanks.

inc/saiport.h
lguohan: the forward means, when it is gearbox, it is forward the flow control frame from one side of the port to the other side of the port, it could be lineside to the systemside or vice versa.

when it is asic, it means the flow control frame will be asic to the packet core from the serdes block and asic can process the flow control frame.

dipankar-ba: Agree - it would be less confusing to simply say the ASIC or the port terminates and honors the flow control frame.

rlhui:
Discussed in community meeting that https://github.com/opencomputeproject/SAI/pull/1256 should be merged in before this one.

lguohan (approved):

kcudnik:
attributes inserted in the middle and not in the end, maybe dont merge this until we add enum values enforcement

dipankar-ba:
Hi Kamil,

I tried moving all attributes to the end of enums.  If you notice anything
else, please let me know.

- Dipankar

kcudnik (changes requested):
> Hi Kamil, I tried moving all attributes to the end of enums. If you notice anything else, please let me know.

both attributes are in this PR are still added int the middle of PORT enum, please add them at the end. We will allow explicit enum values to place them in the middle, probably in this week, since that change hangs for over a month, so you can ether wait or move them and the end, and after that, bringthem back to where you want

Pterosaur:
> > Hi Kamil, I tried moving all attributes to the end of enums. If you notice anything else, please let me know.
> 
> both attributes are in this PR are still added int the middle of PORT enum, please add them at the end. We will allow explicit enum values to place them in the middle, probably in this week, since that change hangs for over a month, so you can ether wait or move them and the end, and after that, bringthem back to where you want

Hi Kamil, I moved them to the end of enums. Please review it.

kcudnik (approved):

inc/saiport.h
rlhui: @dipankar-ba - do you approve this pr or anything to be revisited?

PR #1265, opened 2021-06-30
[meta] Add enum base range check to start from beginning

kcudnik:
This change is to make sure that when range is introduced then the first enum of the range will start with multiplicity 0x1000.


kcudnik:
Requires https://github.com/opencomputeproject/SAI/pull/1264 to pass

PR #1266, opened 2021-06-30
Update SAI_ACL_TABLE_ATTR_ENTRY_LIST value

kcudnik:
Setting to 0x2000 will allow future expansion of acl table fields without changing attr values after SAI_ACL_TABLE_ATTR_FIELD_END.


kcudnik:
This is not critical change, since all attributes after SAI_ACL_TABLE_ATTR_FIELD_END are readonly.

lguohan (approved):

PR #1267, opened 2021-07-01
Ported Marvell armhf build on amd64 for debian buster to use cross-comp…

gregshpit:
…ilation instead of qemu emulation

…ilation instead of qemu emulation

Motivation:
Current armhf Sonic build on amd64 host uses qemu emulation. Due to the nature of the emulation it takes a very long time, about 22-24 hours to complete the build. The change I did to improve the building time ports Sonic armhf build on amd64 host for Marvell platform for debian buster to use cross-compilation on arm64 host for armhf target. The overall Sonic armhf building time using cross-compilation is about 6 hours.

The Sonic configure and build for the armhf cross-compilation is as following:
NOJESSIE=1 NOSTRETCH=1 BLDENV=buster CROSS_BLDENV=1 make configure PLATFORM=marvell-armhf PLATFORM_ARCH=armhf
NOJESSIE=1 NOSTRETCH=1 BLDENV=buster CROSS_BLDENV=1 make target/sonic-marvell-armhf.bin

Sonic module should check if $CROSS_BUILD_ENVIRON is 'y' to make sure that it is cross-compilation build.

### DIFF
--- a/meta/Makefile
+++ b/meta/Makefile
@@ -68,6 +68,9 @@ $(foreach bin,$(BINS),$(if $(shell which $(bin)),,$(error "Missing $(bin) in PAT
 
 CFLAGS += -I../inc -I../experimental $(WARNINGS)
 
+CC = $(CROSS_COMPILE)gcc
+CXX = $(CROSS_COMPILE)g++
+
 DEPS = $(wildcard ../inc/*.h) $(wildcard ../experimental/*.h)
 XMLDEPS = $(wildcard xml/*.xml)
 
@@ -83,7 +86,7 @@ all: toolsversions saisanitycheck saimetadatatest saiserializetest saidepgraph.s
 
 toolsversions:
 	@make -v | grep -i make
-	@gcc --version | grep gcc
+	@$(CC) --version | grep gcc
 	@perl --version | grep version
 	@aspell --version
 	@dot -V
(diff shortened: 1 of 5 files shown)


meta/Makefile
```diff
@@ -68,6 +68,14 @@ $(foreach bin,$(BINS),$(if $(shell which $(bin)),,$(error "Missing $(bin) in PAT
 
 CFLAGS += -I../inc -I../experimental $(WARNINGS)
 
+ifeq ($(CROSS_BUILD_ENVIRON),y)
```
kcudnik, line 71: CROSS_BUILD_ENVIRON is that your internal variable name?

gregshpit: This environment variable is set in Sonic slave.mk when Sonic build uses cross-complation.

kcudnik: i grepped sonic-buildimage and i could not find it, can you give a link in github where this is set up?

kcudnik: also, im just wondering, normally when we have configure.ac and ./configure, to specify cross compiler paramst are used  from --host=HOST, this is standalone Makefile, so maybe it make more sense to make it similar like here: https://lists.gnu.org/archive/html/help-make/2005-05/msg00029.html and then passing 
```
$ CROSS_COMPILE=xxx make
```
or whoever would set CROSS_COMPILE, or it could be even set CC and CXX before make, instead of doing all those if/else across all makefiles on the entire sonic

kcudnik (changes requested):

meta/Makefile
gregshpit: This sonic-buildimage PR https://github.com/Azure/sonic-buildimage/pull/8035 is part of of the overall PR. You can find CROSS_BUILD_ENVIRON environment variable set in slave.mk that I've modified. The naming convention is taken from another variable MULTIARCH_QEMU_ENVIRON defined in this file. CROSS_BUILD_ENVIRON env. variable is used in hundreds files in the PR. I'll check the rest of your request tomorrow. Thanks for your comments.

gregshpit:
Hi,

I think do pass CROSS_COMPILE variable to the Makefile. If it is true (I'll check it) then instead of:
+ifeq ($(CROSS_BUILD_ENVIRON),y)
+    CC=$(CROSS_CC)
+    CXX=$(CROSS_CXX)
+else
+    CC=gcc
+    CXX=g++
+endif

I can just write as you suggested in all modified files:
CC = $(CROSS_COMPILE)gcc
CXX = $(CROSS_COMPILE)g++

Is it OK with you.

Gregory

kcudnik:
> I can just write as you suggested in all modified files:
> CC = $(CROSS_COMPILE)gcc
> CXX = $(CROSS_COMPILE)g++
> 
> Is it OK with you.

yes, this is exactly what i had in mind!

kcudnik (approved):

lguohan (approved):

PR #1278, opened 2021-07-20
Unset SAI_PORT_ATTR_HW_LANE_LIST as key

byu343:
The reason is that we are using one syncd process to support multiple phy chips and for the ports from different phys, they may
have the same value on SAI_PORT_ATTR_HW_LANE_LIST.

This change is depended by https://github.com/Azure/sonic-buildimage/pull/8146. Specifically, multiple switches are defined under guid 1 as below, however, ports from different switches may have the value on SAI_PORT_ATTR_HW_LANE_LIST.
        {
            "guid" : 1,
            "name" : "phys",                                                                                                                                                                                                                                                                "dbAsic" : "GB_ASIC_DB",
            "dbCounters" : "GB_COUNTERS_DB",
            "dbFlex": "GB_FLEX_COUNTER_DB",
            "dbState" : "STATE_DB",
            "zmq_enable" : false,
            "zmq_endpoint": "tcp://127.0.0.1:5565",                                                                                                                                                                                                                                         
            "zmq_ntf_endpoint": "tcp://127.0.0.1:5566",
            "switches": [
                {                                                                                                                       
                    "index" : 0,
                    "hwinfo" : "mdio0_0_0/0"
                },
                {                                                                                                                       
                    "index" : 1,
                    "hwinfo" : "mdio1_0_0/0"
                }
        }

Signed-off-by: Boyang Yu <byu@arista.com>

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -745,7 +745,7 @@ typedef enum _sai_port_attr_t
      * @brief Hardware Lane list
      *
      * @type sai_u32_list_t
-     * @flags MANDATORY_ON_CREATE | CREATE_ONLY | KEY
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      */
     SAI_PORT_ATTR_HW_LANE_LIST,
 


lguohan:
i do not understand this, they are different switches, the key should only matter on one switch. where do you see the issue?

inc/saiport.h
```diff
@@ -745,7 +745,7 @@ typedef enum _sai_port_attr_t
      * @brief Hardware Lane list
      *
      * @type sai_u32_list_t
-     * @flags MANDATORY_ON_CREATE | CREATE_ONLY | KEY
```
kcudnik, line 748: hw lane list identifies uniquely ports on a single switch, it must be a KEY, please revert

kcudnik (changes requested):

inc/saiport.h
kcudnik: is the error you getting from metadata in OA or syncd? can you past error?

kcudnik: this fix will resolve that: https://github.com/Azure/sonic-sairedis/pull/865

byu343:
> i do not understand this, they are different switches, the key should only matter on one switch. where do you see the issue?

Yes, the ports having the same SAI_PORT_ATTR_HW_LANE_LIST are from different switches.  

The error is 
Jul 20 16:03:41.998421 smm211 DEBUG swss#orchagent: :- constructKey: constructed key: SAI_PORT_ATTR_HW_LANE_LIST:2,3;
Jul 20 16:03:41.998421 smm211 ERR swss#orchagent: :- meta_generic_validation_create: attribute key SAI_PORT_ATTR_HW_LANE_LIST:2,3; already exists, can't create_

raised from https://github.com/Azure/sonic-sairedis/blob/216e5497bfb23e697c269fdac0d6965902ebe644/meta/Meta.cpp#L5060.

It is happening  under the setup of using one syncd process and one context to serve 8 phys. I'll investigate further why it happens.

inc/saiport.h
byu343: Sure. I'll the try your fix and hopefully close this PR thereafter. Thanks.

kcudnik:
my fix on sairedis meta addresses this issue

byu343:
The issue is fixed by https://github.com/Azure/sonic-sairedis/pull/865.

PR #1279, opened 2021-07-21
Define new port attributes to configure PFC DLR intervals

miheot:
New attributes:

SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL_RANGE, SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL, SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL_RANGE, SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL
 - This matches the same set of attributes defined at the switch level (saiswitch.h)
 - Provides more granularity to have different intervals per port.

SAI_QUEUE_ATTR_PFC_DLR_PACKET_ACTION
 - This matches SAI_SWITCH_ATTR_PFC_DLR_PACKET_ACTION at the queue level granularity. 
 - For HW that can perform drops at queue level without the need of an ACL
 
SAI_QUEUE_ATTR_PFC_DLR_STATE
 - This attribute represents the queue internal hardware PFC deadlock state. 
 - It is an aggregation of all HW state used to determine if a queue is in PFC deadlock.
 - This attribute should only be queried as part of the PFC deadlock and recovery detection processing.
 - True indicates the queue is in a PFC deadlock state and each successive query, a True indicates the queue has stayed in that state.
- This is different from the SAI_QUEUE_ATTR_PAUSE_STATUS which only provides an instantaneous status of PFC on a queue.

Signed-off-by: Michael H Otto <motto@cisco.com>


inc/saiqueue.h
```diff
@@ -194,6 +194,34 @@ typedef enum _sai_queue_attr_t
...
+     * This attribute represents the queue internal hardware PFC
+     * deadlock state. It is an aggregation of all HW state used to
+     * determine if a queue is in PFC deadlock. This attribute should
+     * only be queried as part of the PFC deadlock and recovery detection
+     * processing.
+     * True indicates the queue is in a PFC deadlock state and
+     * on each successive query, a True indicates the queue has stayed
+     * in that state.
+     *
+     * @type boolean
```
kcudnik, line 220: ERROR: invalid type tag value 'boolean' expected sai type or enum
please fix to bool

miheot: I will fix this.

kcudnik:
you didnt signed last commit thats why DCO is failing

lguohan:
can you provide example code, how those attributes are going to be used?

kcudnik:
please resolve conflicts

lguohan:
please pass the build

miheot:
Opened a new PR 1332.

PR #1293, opened 2021-08-18
Add New SAI P4 Extension API and HLD

mkbalani:


meta/saisanitycheck.c
```diff
@@ -4732,6 +4732,12 @@ void check_object_ro_list(
         return;
     }
 
+    if (SAI_OBJECT_TYPE_P4EXT_ENTRY == oi->objecttype)
+    {
+        META_LOG_WARN("p4ext entry object %s not present on any object list (eg. VLAN_MEMBER is present on SAI_VLAN_ATTR_MEMBER_LIST)", oi->objecttypename);
```
kcudnik, line 4737: instead of adding exception, please add make hits proper attribute

meta/saisanitycheck.c
```diff
@@ -4951,6 +4957,20 @@ void check_graph_connected()
...
 
+        if (SAI_OBJECT_TYPE_P4EXT_ENTRY == i)
+        {
+            /*
+             * Allow P4EXT Entry to be disconnected from main graph.
+             * This may change in future iterations but for now p4ext entry objects are
+             * independent of other SAI objects.
+             */
+
+            META_LOG_WARN("p4ext entry  object %s is disconnected from graph",
```
kcudnik, line 4968: instead of adding exception, make this object connected to graph

kcudnik (changes requested):

inc/saip4ext.h
```diff
@@ -0,0 +1,154 @@
...
+{
+    /**
+     * @brief Start of attributes
+     */
+    SAI_P4EXT_ENTRY_ATTR_START,
+
+    /**
+     * @brief SAI P4 EXT table id
+     *
+     * @type sai_s8_list_t
```
kcudnik, line 43: is this string table name ? where did it come frome?

mkbalani: Yes this is the table name. Added an explanation for this in the HLD [here](https://github.com/opencomputeproject/SAI/pull/1293/files#diff-f10288ab3ba3d4cfacfc71aeeb21c72dfd6ed605f50c7f1db4ca10115c425c71R75-R79)

meta/saisanitycheck.c
```diff
@@ -1821,6 +1821,20 @@ void check_attr_key(
 
                 META_MD_ASSERT_FAIL(md, "marked as key, but have invalid attr value type (list)");
 
+            case SAI_ATTR_VALUE_TYPE_INT8_LIST:
+
+                if (md->objecttype == SAI_OBJECT_TYPE_P4EXT_ENTRY)
+                {
+                    /*
+                     * This is special case when TABLE_ID and MATCH_FIELD_ID are actual KEYs for P4 Extension object.
+					 * Also these attributes are no longer simply a list of int8_t but rather a string
```
kcudnik, line 1830: fix spacin here

inc/saip4ext.h
kcudnik: what expected lengths are of this string? maybe it could fit to chardata[32] which is default in sai attribute?

mkbalani: Table name might fit into the 32byte chardata however for other attributes of p4 extension objects we would still need an arbitrary length string as there can be variable number of action parameters or match fields. Using the same data type for all the attributes here keeps things more consistent. Also having arbitrary length for table name allows vendors to use fqn for the table names if they need to, which can be longer than 32 bytes

kcudnik: ok, let leave it as is

kcudnik (approved):
please address other 2 comments

kcudnik (changes requested):
Please address left 2 comments

mkbalani:
Closing this PR as I am unable to sign off my previous commits on the branch.
Created a new PR - [Add SAI P4 Extension API and HLD](https://github.com/opencomputeproject/SAI/pull/1303) with the same changes (2 comments yet to be addressed)

PR #1307, opened 2021-09-14
Patch 1

JaiOCP:
Memory corruption can be classified into two buckets. One is hard failure and the other is soft failure. Soft failure/error is tend to cause by a single event upset (SEU) to the data memory. Hard failure is tend to be the result of raw defects in silicon wafer, natural aging or electrical over stress causing the P/N junction to break down permanently. Soft error is opposite from hard failure, that is, the memory corruption will go away after the device is reset or re- initialized. 

This PR introduces a callback for Soft Error and Recovery mechanism and associated recovery type and log information.


inc/saiswitch.h
```diff
@@ -400,6 +400,108 @@ typedef enum _sai_tunnel_decap_ecn_mode_t
...
+    SAI_SWITCH_CORRECTION_TYPE_SPECIAL = 5,
+} sai_switch_correction_type_t;
+
+/* soft error recovery log info flags */
+#define SAI_SWITCH_SER_LOG_MEM          0x00000001 /* Error happens on memory */
+#define SAI_SWITCH_SER_LOG_REG          0x00000002 /* Error happens on register */
+#define SAI_SWITCH_SER_LOG_MULTI        0x00000004 /* Parity errors detected more than once */
+#define SAI_SWITCH_SER_LOG_CORRECTED    0x00000008 /* Error be corrected by S/W */
+#define SAI_SWITCH_SER_LOG_ENTRY        0x00000010 /* Corrupt memory entry data is read in logfile */
+#define SAI_SWITCH_SER_LOG_CACHE        0x00000020 /* Cache data is valid */
```
kcudnik, line 477: this seems to be flags, why this is not enum ?

kcudnik: take a look at sai_stats_mode_t

JaiOCP: > take a look at sai_stats_mode_t

Good suggestion. Committed the review suggestion.

https://github.com/opencomputeproject/SAI/pull/1307/commits/ddf25999d43ec08e779565759450a3694e52e8d2

inc/saiswitch.h
```diff
@@ -406,6 +406,140 @@ typedef enum _sai_tunnel_decap_ecn_mode_t
...
+     */
+    SAI_SWITCH_SER_LOG_REG = 1 << 1,
+
+    /**
+     * @brief Parity errors detected more than once
+     */
+    SAI_SWITCH_SER_LOG_MULTI = 1 << 2,
+
+    /**
+     * @brief Error be corrected by S/W
```
mikeberesford, line 502: nit: be -> was?

inc/saiswitch.h
```diff
@@ -406,6 +406,140 @@ typedef enum _sai_tunnel_decap_ecn_mode_t
...
+     * @brief ECC single bit error
+     */
+    SAI_SWITCH_SOFT_ERROR_TYPE_ECC_SINGLE_BIT = 2,
+
+    /**
+     * @brief ECC double bit error
+     */
+    SAI_SWITCH_SOFT_ERROR_TYPE_ECC_DOUBLE_BIT = 3,
+
+    /**
```
mikeberesford, line 434: Is this meant to be "all other soft error types"? If so, how does it differ from UNKNOWN? (if not, what is it?)

inc/saiswitch.h
```diff
@@ -406,6 +406,140 @@ typedef enum _sai_tunnel_decap_ecn_mode_t
...
 
+/**
+ * @brief SAI switch soft error recovery error type.
+ */
+typedef enum _sai_switch_soft_error_type_t
+{
+    /**
+     * @brief Unknown error type
+     */
+    SAI_SWITCH_SOFT_ERROR_TYPE_UNKNOWN = 0,
```
kcudnik, line 417: note: you dont need to explicitly assign values to enums if they are contiguous

inc/saiswitch.h
```diff
@@ -406,6 +406,140 @@ typedef enum _sai_tunnel_decap_ecn_mode_t
...
+/**
+ * @brief SAI switch soft error recovery error correction type.
+ */
+typedef enum _sai_switch_correction_type_t
+{
+    /**
+     * @brief S/W takes no action when error
+     * happens (Like some working memories)
+     */
+    SAI_SWITCH_CORRECTION_TYPE_NO_ACTION = 0,
```
kcudnik, line 449: note: you dont need to explicitly assign values to enums if they are contiguous

inc/saiswitch.h
```diff
@@ -406,6 +406,140 @@ typedef enum _sai_tunnel_decap_ecn_mode_t
...
+    SAI_SWITCH_SER_LOG_CACHE = 1 << 5,
+} sai_switch_ser_log_t;
+
+/**
+ * @brief SAI switch soft error recovery info.
+ */
+typedef struct _sai_switch_ser_info_t
+{
+    /**
+     * @brief Error detected time
```
kcudnik, line 523: maybe a little bit more information is this seconds or milliseconds or usec, and if usec, maybe this should be uint64 ?

kcudnik:
please fix errors

inc/saiswitch.h
```diff
@@ -406,6 +406,140 @@ typedef enum _sai_tunnel_decap_ecn_mode_t
...
+     */
+    uint32_t time;
+
+    /**
+     * @brief Soft error recovery log info flags
+     */
+    sai_switch_ser_log_t flags;
```
kcudnik, line 530: should this be uint32_t if this will be combination of flags ? currently we have similar thing in stats_mode flags

inc/saiswitch.h
```diff
@@ -406,6 +406,140 @@ typedef enum _sai_tunnel_decap_ecn_mode_t
...
+
+    /**
+     * @brief Restore entry from another pipe
+     */
+    SAI_SWITCH_CORRECTION_TYPE_HW_CACHE_RESTORE = 4,
+
+    /**
+     * @brief Memory needs special correction handling
+     */
+    SAI_SWITCH_CORRECTION_TYPE_SPECIAL = 5,
```
lguohan, line 474: can you clarify if error is corrected or not?

inc/saiswitch.h
kcudnik: checkout here: https://github.com/opencomputeproject/SAI/blob/master/inc/saitypes.h#L1446

meta/saisanitycheck.c
```diff
@@ -5124,6 +5124,7 @@ void check_enum_range_base(
 
     SKIP_ENUM(sai_attr_flags_t);
     SKIP_ENUM(sai_stats_mode_t);
+    SKIP_ENUM(sai_switch_ser_log_t);
```
kcudnik, line 5127: no need for this change, i updated sainitycheck to use flags enum type

inc/saiswitch.h
```diff
@@ -406,6 +406,140 @@ typedef enum _sai_tunnel_decap_ecn_mode_t
...
+} sai_switch_correction_type_t;
+
+/**
+ * @brief Soft error recovery flag information
+ *
+ * Note enum values must be powers of 2 to be used as bit mask
+ *
+ * @flags Contains flags
```
kcudnik, line 482: you will need to change this to "@flags strict"

inc/saiswitch.h
kcudnik: checkout here: https://github.com/opencomputeproject/SAI/blob/master/inc/saitypes.h#L1446

inc/saiswitch.h
```diff
@@ -2929,6 +3074,18 @@ typedef void (*sai_switch_state_change_notification_fn)(
...
 
+/**
+ * @brief Switch soft error recovery event callback
+ *
+ * @count data[count]
+ *
+ * @param[in] count Number of notifications
+ * @param[in] data Array of soft error recovery event types
+ */
+typedef void (*sai_switch_ser_event_notification_fn)(
```
mikeberesford, line 3085: Should this include the switch_id in the callback?

rlhui:
Intel: no way for NOS to subscribe to which ones they're interested in

rlhui:
Mike: this seems to take away control from NOS.

PR #1308, opened 2021-09-15
Add attribute to support setting port speed and FEC together

mikeberesford:
There are cases where we technically cannot change from one port
speed/FEC combination to another using the existing single-attribute
write.

As an example: 40G/FC-FEC to 100G/RS-FEC - it is not valid to change to
100G with FC-FEC enabled, and likewise it is not valid to enable RS-FEC
on a 40G link.

Solve this by defining a new struct to handle setting FEC and speed at
the same time.


kcudnik:
please fix compilation errors

kcudnik:
enum sai_port_speed_fec_t not defined anywhere

mikeberesford:
Yep, thanks. I think I've got it now - the restrictions on how the fields are named/generated weren't clear to me. My local build still fails, but that appears to be due to the new enum numbering checks (which hopefully won't happen here).

mikeberesford:
@kcudnik any advice on how to handle the checkancestry.sh failure? It looks like this is because I've got multiple commits in this PR which effectively changes the enum values in the history. Easiest might be to just discard this PR and start a new commit with the latest set of changes.

kcudnik:
for now you could just squash all and force push

kcudnik:
after deeper though, this attribute or separate api will break SPEED and FEC attributes, since SET attributes should not change independently without user action, and adding new attribute or separate api it will violate that assumption

mikeberesford:
Ack, I agree with that. If we were to make this attribute, we would probably need to deprecate the extant SPEED and FEC attributes (or make them read only) and use this exclusively instead. This would be a (possibly non-trival) change from the NOS perspective, particularly since FEC is not a required configuration attribute today.

We're stuck with a problem here though - as stated in the meeting for a "40G FC-FEC -> 100G RS-FEC" transition, we technically can't set either the FEC or the SPEED attribute. Today the vendor layer has the following options:
- reject both of the configuration options since neither is valid (making the speed change impossible even though the final state is allowed) or
- accept the speed change and internally default the FEC to something. This breaks the "SET attributes should not change independently without user action" invariant, and effectively introduces a NOS requirement to always set FEC after SPEED if the NOS cares what the FEC is.

kcudnik:
i think there could be a workaround for this issue, we could use bulk set api, and specify same object twice each time with speed and fec combination, internally sai could pick that up and allow change

mikeberesford:
Hm, that seems to me like we're just backdooring a mutli-attribute set API (with more complexity). What's the advantage of doing this over just allowing a multi-set API with similar semantics.

I'm inclined to close out this pull request and change it to be an issue instead - thoughts?

kcudnik:
> Hm, that seems to me like we're just backdooring a mutli-attribute set API (with more complexity). What's the advantage of doing this over just allowing a multi-set API with similar semantics.
because is very similar, and is not changing current interface, so is not requiring new api, and don't require 2x set operation for both those attributes speed and fec
but this is just my proposal

> I'm inclined to close out this pull request and change it to be an issue instead - thoughts?
sure go ahead

mikeberesford:
Closing out this PR, will continue discussion in https://github.com/opencomputeproject/SAI/issues/1322

PR #1310, opened 2021-09-21
Update sai version

rlhui:


rlhui:
Update SAI version to v1.9.0

PR #1317, opened 2021-09-24
[meta] Disable enum conversion warning on metadata source

kcudnik:
because we are merging extension attributes into existing enums, new
versions of gcc can warn when 2 different enums are mixed, so lets
ignore this warning using pragmas

### DIFF
--- a/meta/parse.pl
+++ b/meta/parse.pl
@@ -1223,14 +1223,34 @@ sub ProcessExtraRangeDefines
     }
 }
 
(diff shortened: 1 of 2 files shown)


kcudnik:
@itaibaz fyi

itaibaz:
@kcudnik 
I tested the PR and compilation now passes fine
Thanks

itaibaz (approved):

PR #1325, opened 2021-10-13
SAI PTF 1: Proposal

pkochmanska:


meta/gensairpc.pl
```diff
@@ -0,0 +1,605 @@
+#!/usr/bin/env perl
+#
+# Copyright (c) 2014 Microsoft Open Technologies, Inc.
```
aljer, line 3: 2021

doc/SAI-Proposal-SAI-PTF.md
```diff
@@ -0,0 +1,636 @@
...
+
+![autogeneration gensairpc](figures/autogeneration_gensairpc.png "Figure 3: Autogeneration - gensairpc")
+__Figure 3: Autogeneration - parsing__
+
+Autogeneration when upgrading SAI:
+
+```
+cd <root>/meta
+make
+./gensairpc.pl --clean --no-meta-build
```
richardyu-ms, line 70: seems this parameter doesn't existing in perl

meta/templates/sai_adapter.py.tt
```diff
@@ -0,0 +1,487 @@
...
+        [%- PROCESS catch_exception %]
+[%- END -%]
+
+[%- ######################################################################## -%]
+
+[%- ######################################################################## -%]
+
+[%- BLOCK catch_exception -%]
+    except sai_thrift_exception as e:
+        status = e.status
```
richardyu-ms, line 165: how about generating those exception handlers into a utils method that can be used by the user?

meta/gensairpc.README
```diff
@@ -0,0 +1,34 @@
...
+
+    In order to generate debug files and avoid cleanup, use --dbg or --dump
+    flags. Use --experimental to generate extensions and experimental code.
+
+    For more information, use --help:
+
+      ./gensairpc.pl --help
+
+DEPENDENCIES
+    Ubuntu packages:
```
richardyu-ms, line 27: is there any place that mentioned 'graphviz' as a dependency for 'dot' which required in Makefile 'BINS = doxygen perl dot'
Not sure, if libmoose-perl needed

meta/templates/sai_adapter.py.tt
aljer: Not sure I understand your point of view. This block simply surround each RPC server function call in sai_adapter with try-except clause. How would you like to turn it into a method? Secondly, do you mean to put this into sai_utils file that we want to add here: https://github.com/opencomputeproject/SAI/pull/1335/files#diff-7a25e711c6b68d0df0743a168d6efcdb1258b18e7062942eefd507eec767d9cb or some other? sai_utils file is not intended to be autogenerated. It is written manually.

doc/SAI-Proposal-SAI-PTF.md
```diff
@@ -0,0 +1,636 @@
...
+- default_1q_bridge
+- acl_stage_ingress
+- acl_stage_egress
+- number_of_active_ports
+- port numbers stored into variables portX, where X is the number of port
+
+SaiHelper inherits from SaiHelperBase but provides additional configuration.
+The base configuration created by SaiHelper looks like this:
+
+| Name       | Vlan ID | Ports | Tagging |
```
aljer, line 104: Please update the table according to the new one we have in sai_base_test file that is to be upstreamed (https://github.com/aljer/SAI/blob/sai_ptf_infra/ptf/sai_base_test.py#L490)

kcudnik (approved):

meta/templates/sai_adapter.py.tt
richardyu-ms: I see, there is a condition 'if CATCH_EXCEPTIONS:'.

richardyu-ms: but still, have some concerns about this try-except, first, it is not thread-safe. secondly, an exception should be happened as unexpected. why do we need to wrap that, and also request a manual check for the status, but not throw it directly and let users to handle them as their needs?

meta/templates/sai.thrift.tt
```diff
@@ -0,0 +1,248 @@
...
+
+    [%- END -%]
+[% END -%]
+
+[%- ######################################################################## -%]
+
+[%- # The body of the file: -%]
+// AUTOGENERATED FILE! DO NOT EDIT
+
+namespace py sai
```
richardyu-ms, line 233: current namespace is switch_sai. 
saiservee.cpp depends on those generated files, like switch_sai_rpc, do we plan to use a new saiserver.cpp?
saiserver is referenced by syncd-rpc, then we need to upgrade the syncd-rpc to use the new saiserver if we want to use the new structure?

ravi861: Yes. The 2nd PR for SAI PTF has a new sai_rpc_frontend.cpp which uses the new namespace. I am going to create a 3rd PR which will include the Makefile.

meta/gensairpc.README
aljer: 'BINS = doxygen perl dot' is an old part of Makefile. We don't introduce 'dot' with the generator. That's why its dependencies are not listed here.

aljer: \+ IMO we could add libmoose-perl to the list, but it is a dependency of libmoosex-aliases-perl so they should be installed together.

meta/templates/sai_adapter.py.tt
pkochmanska: This is the design we established and according to that we have written all the tests, this change would mean we have to modify all the test cases we have to handle exceptions.

richardyu-ms: ok, i see

richardyu-ms (approved):

ravi861:
@lguohan @rlhui Should we give more time for reviews or can we go ahead with merge?

KrisNey-MSFT:
Hi team, per our SONiC-DASH-Test-Workgroup community meeting this morning, may we ask to accommodate SONiC-DASH (NIC) as we move forward?  @chrispsommers @rlhui @ravi861 
We are hoping to leverage SAI-Thrift for test purposes.
Thank you, @Krisney-MSFT

ravi861:
@KrisNey-MSFT Can you please reach out to me over email ravi.vantipalli@intel.com for what you are looking for?

KrisNey-MSFT:
Yes, will do - thank you.
Kristina



Kristina Moore, MBA, M.S., CISSP

Principal Program Manager

Azure Networking – Hardware/Switching/Performance

400G New Tech

Office:   40/4J00

Mobile:   (425) 876- 2040<tel:(425)%20876-%202040>

***@***.******@***.***>



***@***.***





KrisNey-MSFT:
Thanks Ravi, would you want to chat tomorrow at 1pm or so?  My #'s are below, or you could give me yours..?

***@***.***

Kristina Moore MBA, M.S., CISSP
Azure Core Principal PM
DASH

Office: 425-722-7720
Mobile: 425-876-2040
***@***.***

***@***.***

reshmaintel:
@KrisNey-MSFT Hi Kristina, would you please include me in the SAI PTF call with Ravi/Intel. Alternately I can arrange a meeting for all of us to discuss. Thanks,

richardyu-ms:
> @KrisNey-MSFT Hi Kristina, would you please include me in the SAI PTF call with Ravi/Intel. Alternately I can arrange a meeting for all of us to discuss. Thanks,

Me too, thanks.

rlhui (approved):

PR #1326, opened 2021-10-14
SAI Proposal for Stateful Tables

marian-pritsak:


inc/saicallback.h
```diff
@@ -0,0 +1,85 @@
...
+ *    FOR A PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
+ *
+ *    See the Apache Version 2.0 License for specific language governing
+ *    permissions and limitations under the License.
+ *
+ *    Microsoft would like to thank the following companies for their review and
+ *    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
+ *    Dell Products, L.P., Facebook, Inc., Marvell International Ltd.
+ *
+ * @file    saiacl.h
```
kcudnik, line 20: name is wrong, should be saicallback
PS. saicallback does not fit current parser/style for saiXXX.h that contains SAI api, so some havy changes to the prarser will be required to accept all those new features introduced here

inc/saicallback.h
```diff
@@ -0,0 +1,85 @@
...
+    sai_ipv4_t inner_ipv4;
+    
+    sai_tcp_t inner_tcp;
+
+} sai_parsed_headers_t;
+
+/**
+ * @brief State callback signature
+ */
+void handler(void *flow_ctx, void *global_ctx, const struct sai_parsed_headers_t *parsed_headers, sai_u32_t *packet_metatata);
```
kcudnik, line 72: missint In, Inout, Out and one param per line

kcudnik: also should this be typedef and sai_handler_fn type ?

inc/saicallback.h
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * @brief Set the next state
+ *
+ * Set the state callback to be executed when the stateful
+ * table is applied to the next packet classified to
+ * this flow.
+ *
+ * @param[in] state_handler state handler callback
+ */
+void sai_set_state(char *state);
```
kcudnik, line 83: missint In, Inout, Out and one param per line

PS.  this seems like setting global state to SAI, why this is char* and not handler* or  void*?

kcudnik (changes requested):

reshmaintel:
Marian, the restrictions section need to be discussed further, if we need to generalize this as applicable for all HW.
Adding @anjalisinghai1 as reviewer of this PR.

PR #1332, opened 2021-10-27
PFC Deadlock SAI attribute additions

miheot:
Signed-off-by: Michael H Otto <motto@cisco.com>

Adding new attributes:

SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL_RANGE, SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL, SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL_RANGE, SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL

This matches the same set of attributes defined at the switch level (saiswitch.h)
Provides more granularity to have different intervals per port.

SAI_QUEUE_ATTR_PFC_DLR_PACKET_ACTION

This matches SAI_SWITCH_ATTR_PFC_DLR_PACKET_ACTION at the queue level granularity.
For HW that can perform drops at queue level without the need of an ACL

SAI_QUEUE_ATTR_PFC_CONTINUOUS_DEADLOCK_STATE

This attribute represents the queue internal hardware PFC deadlock state.
It is an aggregation of all HW state used to determine if a queue is in PFC deadlock.
This attribute should only be queried as part of the PFC deadlock and recovery detection processing.
This attribute returns an enum indicating that the queue is not PFC paused, is PFC pause or is PFC paused but not for the whole polling interval time.
This is different from the SAI_QUEUE_ATTR_PAUSE_STATUS which only provides an instantaneous status of PFC on a queue.


kcudnik (approved):

sachinv-msft:
@kcudnik could you please merge this PR?

kcudnik:
was this approved by community sai meeting?

lguohan (approved):

PR #1335, opened 2021-11-02
SAI PTF 2: sai_base_test infra

aljer:
The changes introduce SAI PTF framework base tests infrastructure. Following files have been added:
- sai_rpc_frontend.cpp - RPC server frontend file - there is a set of manually written helper functions that converts/parses SAI and thrift attributes and also here we have autogenerated SAI RPC server functions included
- sai_base_test.py - python module with base test classes with common tests configuration
- sai_utils.py - python module with some helper function used in SAI PTF tests
- saitest.py - first test module with simple framework test


ravi861:
@rlhui @richardyu-ms fyi

meta/sai_rpc_frontend.cpp
```diff
@@ -0,0 +1,870 @@
+/**
+ * Copyright (c) 2021 Microsoft Open Technologies, Inc.
```
kcudnik, line 2: can this file be located in ptf directory ?

aljer: This file make use of autogenerated RPC code which also is to be placed in meta directory and its subdirectories. Please see: https://github.com/opencomputeproject/SAI/pull/1325/files

ravi861: We can maybe allow the gensairpc script to allow to generate the files in a different directory than meta as an argument. Would it help if we create a top-level "thriftrpc" directory to place this and other related files?

richardyu-ms (approved):

rlhui:
Adding @smaheshm  to review as well.

meta/sai_rpc_frontend.cpp
rlhui: @kcudnik - please confirm?

kcudnik: what should i confirm here?

kcudnik:
> **new alerts:**
> 
> * 32 for Imprecise assert
> * 3 for 'import *' may pollute namespace
> * 1 for Module is imported with 'import' and 'import from'

please address LGTM issues

aljer:
> This pull request **introduces 2 alerts** when merging [acd7f69](https://github.com/opencomputeproject/SAI/commit/acd7f6997896a44160cea8c3ad70c1f7f8eacaa7) into [05648bc](https://github.com/opencomputeproject/SAI/commit/05648bc62fbdf9c456c0b8d93d6ee53b5edd5df9) - [view on LGTM.com](https://lgtm.com/projects/g/opencomputeproject/SAI/rev/pr-2e244003b5d60bd6e5c3b38cd7dad687e1209d73)
> 
> **new alerts:**
> 
> * 2 for 'import *' may pollute namespace

@kcudnik I addressed almost all issues. Those two which remain are required from the perspective of tests infrastructure.

smaheshm:
Regarding "sai_rpc_frontend.cpp - RPC server frontend file - there is a set of manually written helper functions that converts/parses SAI and thrift attributes and also here we have autogenerated SAI RPC server functions included"

Why not keep these two separate? why mix auto generated file with manual addition?

ravi861:
> Regarding "sai_rpc_frontend.cpp - RPC server frontend file - there is a set of manually written helper functions that converts/parses SAI and thrift attributes and also here we have autogenerated SAI RPC server functions included"
> 
> Why not keep these two separate? why mix auto generated file with manual addition?

The helper functions are defined in sai_rpc_frontend.cpp and used in the generated sai_rpc_server.cpp.
The sai_rpc_server.cpp is a large class definition with all the member functions declared and defined inline and use the handwritten helper functions.
So there is a circular dependency on who includes the other. The sai_rpc_server.cpp is included in sai_rpc_frontend.cpp with a #include right after the helper function definitions, which, I agree is a little unconventional.

meta/sai_rpc_frontend.cpp
```diff
@@ -0,0 +1,870 @@
...
+      j++;
+    } else if (let >= 'A' && let <= 'F') {
+      m[j / 2] = (m[j / 2] << 4) + (let - 'A' + 10);
+      j++;
+    }
+  }
+  return (j == 12);
+}
+
+void sai_thrift_ip4_t_parse(const std::string s, unsigned int *m) {
```
smaheshm, line 61: How about adding unit tests for the APIs performing SAI thrift to SAI and vice versa.

meta/sai_rpc_frontend.cpp
```diff
@@ -0,0 +1,870 @@
...
+    default:
+      break;
+  }
+}
+
+// including it here we never have to modify the generated file
+#include "sai_rpc_server.cpp"
+
+class sai_rpcHandlerFrontend : virtual public sai_rpcHandler {
+  int64_t sai_thrift_object_type_get_availability(
```
smaheshm, line 767: curious why this (and others) had to be written manually.

ptf/sai_base_test.py
```diff
@@ -0,0 +1,945 @@
...
+
+        Args:
+            timeout (int): Timeout value in seconds
+        """
+        print("Waiting for fdb entry to age")
+        aging_interval_buffer = 10
+        time.sleep(timeout + aging_interval_buffer)
+
+
+class SaiHelper(SaiHelperBase):
```
smaheshm, line 491: Does it make sense to rename this class since it sets up a specific configuration that all tests may not be able to use.

smaheshm:
Rest looks good. I think adding unit tests (gtests) to the conversion APIs will be useful to catch regression when someone changes those functions.

meta/sai_rpc_frontend.cpp
rlhui: @kcudnik  - wanted to make sure if we're in agreement on "create a top-level "thriftrpc" directory to place this and other related files"

kcudnik: that should be fine

ptf/sai_base_test.py
aljer: Do you mean this class should not be named "SaiHelper"? Actually this class "helps" to set designed configuration which we use in most of the tests. This is even more common than using SaiHelperBace as a parent class.

ravi861: Most tests which are being upstreamed use this as the base class and use the configuration defined in this class. New tests may choose to use SaiHelperBase instead to define their own configuration.

meta/sai_rpc_frontend.cpp
aljer: These two manually written 'sai_thrift_' functions are specific and they don't match the patterns the generator is working with. There are only two of such functions related to CRM verification.

meta/sai_rpc_frontend.cpp
```diff
@@ -0,0 +1,870 @@
...
+#include "sai.h"
+#include "saitypes.h"
+#include "saimetadatatypes.h"
+#include "saimetadatautils.h"
+#include "saimetadata.h"
+}
+
+using namespace ::sai;
+
+unsigned int sai_thrift_mac_t_parse(const std::string s, void *data) {
```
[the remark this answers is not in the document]
aljer: Okay, I will add them

ptf/lpm.py
```diff
@@ -0,0 +1,110 @@
...
+sub-class IpInterval then could be used to get the first/last/random IP within
+this range. It could also check the length of the range and if an IP is within
+this range.
+
+To achieve the LPM functionality, use the LpmDict as a dictionary and use
+[] operator to get the corresponding value using the key (IP).
+
+Please check the test_lpm.py file to see the details of how this class works.
+'''
+class LpmDict():
```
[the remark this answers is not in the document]
aljer: This module is to be used in saiswitch tests that will be upstreamed soon. But in general it is a part of the infrastructure and may be used in any other test modules.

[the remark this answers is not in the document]
smaheshm: Code shouldn't be added with the assumption it will be used, this way we may end up with unused code lying around. Can you remove this file and add it when it's being used, i.e. along with saiswitch tests.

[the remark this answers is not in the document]
aljer: Okay, as you wish

smaheshm:
thanks for removing the file. can you address the LGTM alerts.

aljer:
> thanks for removing the file. can you address the LGTM alerts.

@smaheshm As I explained here https://github.com/opencomputeproject/SAI/pull/1335#issuecomment-989732708 these two imports causing the alerts are required by all other test modules. This is one of a framework assumptions.

smaheshm (approved):
:shipit:

ptf/sai_base_test.py
```diff
@@ -0,0 +1,945 @@
...
+            available_ipv6_neighbor_entry=True,
+            available_next_hop_group_entry=True,
+            available_next_hop_group_member_entry=True,
+            available_fdb_entry=True,
+            available_ipmc_entry=True,
+            available_snat_entry=True,
+            available_dnat_entry=True,
+            available_double_nat_entry=True,
+            number_of_ecmp_groups=True,
+            ecmp_members=True)
```
kcudnik, line 423: this is actually wrong approach, since not a lot of vendors has those implemented, for example current DASH implementation, and this function will throw sai thrift exception if any of those values will not be success, and this will then generate SkipTest exception if error is SAI_STATUS_NOT_SUPPORTED, which is what happens, and many test will/can be skipped if those values are not present, but they not prevent actually for test to be executed, each test should handle this independently or this function here should handle this in a way that if not supported, then set this resource to zero

kcudnik: @lguohan @chrispsommers fyi

chrispsommers: @kcudnik  It does seem like the PTF infrastructure makes some assumptions about base capabilities of any SAI target which has caused some issues in DASH and forced some workarounds. It's arguable whether this is a flaw in the PTF framework and its assumptions, or that the DASH behavioral model falls too short of the mark for emulating a target in all its SAI support. Recent work by some vendors in SAI made many allowances for DASH such as supporting only 2-port devices, etc. but it didn't try to accommodate all the shortcomings if the current DASH behavioral model.

If there is opportunity to establish the capabilities/resources of the target and use that information to skip tests, that would be preferred.

PR #1339, opened 2021-11-11
[PoC][Dont Merge]Add platform support and a sanity test sample

richardyu-ms:
Add a middle layer (strategy and factory) under the test cases which can
1. no more if-else for platform selecting
1. no more code injection for different platforms, just need to abstract to a method on the differences
1. auto select the platform by the parameters in the test starter (shell)
1. easy-distinct structure for the difference from platforms
1. prepared for future statistic 

Add a sanity sample, it can
1. configure across all the ports (base on port's configuration file)
1. simple FDB and VLAN naming according to port's number
1. make a basic check on the status of the devices
1. use as a sample for how to use the new middle layer

testing done:
barefoot and broadcom platforms

### DIFF
--- a/meta/sai_rpc_frontend.cpp
+++ b/meta/sai_rpc_frontend.cpp
@@ -1,14 +1,16 @@
-/*
-Copyright 2013-present Barefoot Networks, Inc.
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


richardyu-ms:
@ravi861 @aljer for a review
@lolyu I partially reached the goal on the structure I wanted (as you quote as '奇技淫巧', LOL ), but not satisfied yet ~ LOL

richardyu-ms:
close this one, the target branch is not right

PR #1349, opened 2021-12-01
Add DONOTDROP packet action.

Ashish1805:
This will help resolve the action conflict between ACLs in different priority groups where lower priority group ACL action is DROP and higher priority group ACL action is conflicting with DROP action, say REDIRECT, and requirement is to override the DROP action.

### DIFF
--- a/doc/Packet-action-DONOTDROP.md
+++ b/doc/Packet-action-DONOTDROP.md
@@ -1,0 +1,97 @@
+# SAI_PACKET_ACTION_DONOTDROP packet action
+-------------------------------------------------------------------------------
+ Title       | SAI_PACKET_ACTION_DONOTDROP packet action
+-------------|-----------------------------------------------------------------
+ Authors     | Ashish Singh, Google LLC
+ Status      | In review
+ Type        | Standards track
+ Created     | 01/25/2022
+
+-------------------------------------------------------------------------------
+
+This spec discusses the SAI_PACKET_ACTION_DONOTDROP packet action use-cases.
+
+Goal of this packet action attribute (SAI_PACKET_ACTION_DONOTDROP) is to resolve the conflicts of ACL actions when a packet hits more than one ACLs in an ACL stage such that when this packet action is used with an ACL, it serves as an action that prevents the possible drop from another ACL that is in a lower priority table.
+
+This new packet action attribute is in line with the current SAI architecture - actions in higher priority groups take precedence. If an ACL with DROP action is in a higher priority group than this ACL, DROP action is expected to be honored. Table priority group of a table can be set using SAI_ACL_TABLE_GROUP_MEMBER_ATTR_PRIORITY.
+
+Scope of the SAI_PACKET_ACTION_DONOTDROP packet action is limited to the ACL stage where it is used in the ACL as packet action - it would not override the DROP action in subsequent ACL stage(s) tables.
+
+Few cases are explained with examples in next section.
+
+# Usage Examples
+### Case 1
+
+| ACL in table T1                     | ACL in table T2                       |
+|-------------------------------------|---------------------------------------|
+| Table priority group:  P1           | Table priority group:  P2             |
+|Action: SAI_PACKET_ACTION_DONOTDROP  | Action: SAI_PACKET_ACTION_DROP        |
+
+
+
+If P1 > P2: action is “no DROP”
+
+
+If P2 > P1: action is “DROP”
+
+
+
+### Case 2
+
+| ACL in table T1                     | ACL in table T2                       |
+|-------------------------------------|---------------------------------------|
+| Table priority group:  P1           | Table priority group:  P2             |
+|Action: SAI_PACKET_ACTION_DONOTDROP  | Action: neither SAI_PACKET_ACTION_DROP nor SAI_PACKET_ACTION_DONOTDROP          |
+
+
+If P1 > P2: action is “no DROP”
+
+
+If P2 > P1: action is “no DROP”
+
+
+### Case 3
+
+| ACL in table T1                     | ACL in table T2                       |
+|-------------------------------------|---------------------------------------|
+| Table priority group:  P1           | Table priority group:  P2             |
+|Action: neither SAI_PACKET_ACTION_DROP conflicting action nor SAI_PACKET_ACTION_DONOTDROP  | Action: SAI_PACKET_ACTION_DROP     |
+
+
+If P1 > P2: action is “DROP”
+
+
+If P2 > P1: action is “DROP”
+
+### Case 4
+
+For SAI actions that are implemented as a copy of other actions including a DROP, the SAI_PACKET_ACTION_DONOTDROP action acts on only the DROP portion of the action.
+
+TRAP action may be implemented as COPY + DROP. A higher-priority DONOTDROP action will cancel the DROP action only, resulting in the packet being copied to CPU but not dropped.
+
+| ACL in table T1                     | ACL in table T2                       |
+|-------------------------------------|---------------------------------------|
+| Table priority group:  P1           | Table priority group:  P2             |
+|Action: SAI_PACKET_ACTION_DONOTDROP  | Action: SAI_PACKET_ACTION_TRAP        |
+
+
+If P1 > P2: action is "copy to CPU"
+
+
+If P2 > P1: action is "TRAP"
+
+### Case 5
+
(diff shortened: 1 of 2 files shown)


inc/saiswitch.h
```diff
@@ -131,7 +131,10 @@ typedef enum _sai_packet_action_t
     SAI_PACKET_ACTION_DENY,
 
     /** This is a combination of SAI packet action COPY_CANCEL and FORWARD */
-    SAI_PACKET_ACTION_TRANSIT
+    SAI_PACKET_ACTION_TRANSIT,
+
+    /** Do not drop the packet. */
+    SAI_PACKET_ACTION_DONOTDROP
```
kcudnik, line 137: whats the difference with SAI_PACKET_ACTION_FORWARD?

Ashish1805: Our intention here is to get an acl packet action which can help override ACL DROP action of another ACL based on acl priority groups of these ACLs. 

For example: 
Lets say we have an ACL in a table T1 (with ACL group priority = X) that has ACL action = DROP and we have another ACL in table Table T2 (With ACL group priority = Y) with ACL action = REDIRECT and a packet is hitting both the ACLs.

Since, ACL actions REDIRECT and DROP are conflicting actions on a packet, action resolution will happen. Irrespective of acl group priority of Table T1 and T2, DROP action will take precedence and packet will be dropped.

We want packet to not get dropped rather be redirected (which means we want to override the ACL DROP action of ACL in Table T1 by ACL REDIRECT action of ACL in Table T2) when Table T2's ACL group priority is higher than Table T1's ACL group priority, i.e. Y > X. To achieve this we want to add packet action SAI_PACKET_ACTION_DONOTDROP.

Using this (ACL action = REDIRECT + packet action = SAI_PACKET_ACTION_DONOTDROP), Acl in Table2 can override the ACL DROP action of ACL action in Table 1 given acl group priority of Table2 > acl group priority of Table1.

As per my understanding, SAI_PACKET_ACTION_FORWARD packet action helps forward the packets using forwarding info from L2/L3 table if packets are not being dropped - but it cannot help override ACL DROP action itself irrespective of acl group priorities of ACL tables.

rlhui:
@JaiOCP , @marian-pritsak  - would you please help review this? Thanks.

Ashish1805:
I updated the PR with a markup doc to explain how the DONOTDROP packet action will be used. 

Thank you.

JaiOCP (approved):
Thanks Ashish. These changes looks good to me

inc/saiswitch.h
mikeberesford: @kcudnik would you take a look at whether the documentation addresses your question?

Ashish1805:
@marian-pritsak Can you please take a look if changes look good to you? Thank you.

doc/Packet-action-DONOTDROP.md
```diff
@@ -0,0 +1,63 @@
...
+Goal of this packet action attribute (SAI_PACKET_ACTION_DONOTDROP) is to resolve the conflicts of ACL actions when a packet hits more than one ACLs such that when this packet action is used with an ACL, it serves as an action that prevents the possible drop from another ACL that is in a lower priority table.
+
+This new packet action attribute is in line with the current SAI architecture - actions in higher priority groups take precedence. If an ACL with DROP action is in a higher priority group than this ACL then, DROP action is expected to be honored. Few cases are explained with examples in next section.
+
+# Usage Examples
+### Case 1
+
+| ACL in table T1                     | ACL in table T2                       |
+|-------------------------------------|---------------------------------------|
+| Table priority group:  P1           | Table priority group:  P2             |
```
ashutosh-agrawal, line 25: Can you please elaborate on "Table priority group"? Currently, there is no priority attribute for ACL Group table in SAI.

doc/Packet-action-DONOTDROP.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+### Case 3
+
+| ACL in table T1                     | ACL in table T2                       |
+|-------------------------------------|---------------------------------------|
+| Table priority group:  P1           | Table priority group:  P2             |
+|Action: neither DROP nor SAI_PACKET_ACTION_DONOTDROP  | Action: DROP                         |
+
+
```
rck-innovium, line 58: We need to define the behavior for enums that are "Combination of Packet Actions" like  SAI_PACKET_ACTION_TRAP. For example, can you explain the below:

| ACL in table T1                     | ACL in table T2                       |
|-------------------------------------|---------------------------------------|
| Table priority group:  P1           | Table priority group:  P2             |
|Action: SAI_PACKET_ACTION_DONOTDROP  | Action: TRAP                         |

If P1 > P2: action is “Copy to CPU?”


If P2 > P1: action is “Trap?”

doc/Packet-action-DONOTDROP.md
rck-innovium: Yes. It is good to clarify if you meant SAI_ACL_TABLE_GROUP_MEMBER_ATTR_PRIORITY.

doc/Packet-action-DONOTDROP.md
```diff
@@ -0,0 +1,63 @@
...
+-------------------------------------------------------------------------------
+
+This spec discusses the SAI_PACKET_ACTION_DONOTDROP packet action use-cases.
+
+Goal of this packet action attribute (SAI_PACKET_ACTION_DONOTDROP) is to resolve the conflicts of ACL actions when a packet hits more than one ACLs such that when this packet action is used with an ACL, it serves as an action that prevents the possible drop from another ACL that is in a lower priority table.
+
+This new packet action attribute is in line with the current SAI architecture - actions in higher priority groups take precedence. If an ACL with DROP action is in a higher priority group than this ACL then, DROP action is expected to be honored. Few cases are explained with examples in next section.
+
```
rck-innovium, line 19: What is the behavior if  the highest priority entry in ACL stage 1 (say ingress) marks as DO_NOT_DROP, and another entry in ACL Stage 2 (say egress) marks the packet as drop? 

In order to avoid ambiguities, we should spell out that DO NOT DROP overrides the DROP action set by:

1) ACL tables with lower priority in this ACL Table Group, AND

2) ACL tables in preceding stages


In other words, it would not override the drop action set in subsequent ACL Table stages.

rck-innovium (changes requested):

doc/Packet-action-DONOTDROP.md
ashutosh-agrawal: But SAI_ACL_TABLE_GROUP_MEMBER_ATTR_PRIORITY is applicable only for the sequential ACL groups. For a sequential ACL group, only the highest priority table is applied so if the highest priority table has a redirect action, it will apply regardless of any conflicting actions in the lower priority tables. So, I'm bit unclear how this new packet action type will be used.

inc/saiswitch.h
kcudnik: explanation is very informative, thanks

kcudnik (approved):

doc/Packet-action-DONOTDROP.md
Ashish1805: As per my understanding, ACL action resolution is done per-stage and an ingress stage DONOTDROP action should not have have impact on VFP or EFP actions.

doc/Packet-action-DONOTDROP.md
Ashish1805: done. Thank you.

doc/Packet-action-DONOTDROP.md
Ashish1805: Added behavior of DONOTDROP with TRAP and DENY. Thank you.

Ashish1805: @rck-innovium would you take a look at whether the documentation addresses your question? Thank you.

ashutosh-agrawal:
LGTM with the assumption that Ashish will take care of a few typos which were identified in the meeting today.

ashutosh-agrawal (approved):

doc/Packet-action-DONOTDROP.md
Ashish1805: Updated the doc with comment on DONOTDROP action in one stage will not override the drop action is subsequent ACL stages. Thank you.

Ashish1805:
Updated the doc to incorporate the changes suggested. PTAL when you get a chance. Thank you.

Ashish1805:
Hi,

All the comments have been addressed - not sure anyone has any other comments/feedback. Can we go ahead and merge/commit it?

Thank you.

PR #1354, opened 2021-12-10
Add SAI_MIRROR_SESSION_ATTR_COUNTER_ID

mthatty:
Signed-off-by: Michael T Hatty <mhatty@cisco.com>

Adding New Attribute SAI_MIRROR_SESSION_ATTR_COUNTER_ID

This attribute allows us to use the existing SAI_OBJECT_TYPE_COUNTER with mirror sessions. This is the same functionality that exists for various other SAI types - saineighbor, sainexthop, etc.


itaibaz:
what is the counter counting in case of sampled mirroring ?
how many packet in the original flow, or how many packets were sampled ?

mthatty:
> what is the counter counting in case of sampled mirroring ?
> how many packet in the original flow, or how many packets were sampled ?

Number of packets sampled

rlhui:
@itaibaz , @rck-innovium - would you please help review/approve?  Thanks.

inc/saimirror.h
```diff
@@ -346,6 +346,19 @@ typedef enum _sai_mirror_session_attr_t
      */
     SAI_MIRROR_SESSION_ATTR_UDP_DST_PORT,
 
+    /**
+     * @brief Attach a counter
+     *
+     * When it is empty, then packet hits won't be counted
```
rck-innovium, line 352: Can you reword to say something like: 

The number of packets that have been mirrored

rck-innovium: We need to specify the behavior of SAI_COUNTER_STAT_BYTES value. For example, there is ambiguity whether it is:

-   the mirrored packet payload bytes (after accounting truncation) or 
-   the incoming packet bytes

rck-innovium (changes requested):

inc/saimirror.h
mthatty: Updated with your suggestion. Please let me know if this is sufficient or requires further revising.

rlhui:
@rck-innovium - please re-review/approve

rck-innovium (approved):

kcudnik (approved):

PR #1355, opened 2021-12-13
Add Dash extensions

marian-pritsak:
This pull request is missing typedefs like sai_ip_addr_list_t that should be added to master first


experimental/saiexperimentaldash.h
```diff
@@ -0,0 +1,2189 @@
...
+    sai_get_direction_lookup_entry_attribute_fn               get_direction_lookup_entry_attribute;
+    sai_create_slb_decap_entry_fn                             create_slb_decap_entry;
+    sai_remove_slb_decap_entry_fn                             remove_slb_decap_entry;
+    sai_set_slb_decap_entry_attribute_fn                      set_slb_decap_entry_attribute;
+    sai_get_slb_decap_entry_attribute_fn                      get_slb_decap_entry_attribute;
+    sai_create_eni_lookup_to_vm_entry_fn                      create_eni_lookup_to_vm_entry;
+    sai_remove_eni_lookup_to_vm_entry_fn                      remove_eni_lookup_to_vm_entry;
+    sai_set_eni_lookup_to_vm_entry_attribute_fn               set_eni_lookup_to_vm_entry_attribute;
+    sai_get_eni_lookup_to_vm_entry_attribute_fn               get_eni_lookup_to_vm_entry_attribute;
+    sai_create_inbound_routing_entry_fn                       create_inbound_routing_entry;
```
nanteby, line 2124: we can have inbound / outbound as parameters instead of 2 different APIs

experimental/saiexperimentaldash.h
```diff
@@ -0,0 +1,2189 @@
...
+ * @param[inout] attr_list Array of attributes
+ *
+ * @return #SAI_STATUS_SUCCESS on success Failure status code on error
+ */
+typedef sai_status_t (*sai_get_eni_meter_entry_attribute_fn)(
+        _In_ const sai_eni_meter_entry_t *eni_meter_entry,
+        _In_ uint32_t attr_count,
+        _Inout_ sai_attribute_t *attr_list);
+
+typedef struct _sai_dash_api_t
```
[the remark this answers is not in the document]
nanteby: In order to support atomic operation we need to get list of ACLs.
Need to extend the api to allow to remove an acl lists, add acl lists in the same api. 
In the sirius api - updating of the ACL Group (ACLStage) must be an atomic operation. No partial updates are allowed, as it might lead to security issues in the case of only partial rules being applied.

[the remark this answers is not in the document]
nanteby: It will be also much more efficient to allow to set list of eni / routes  and all the other configuration in a list and not one by one.
If we are talking about ca2pa for example , if we want to insert few thousands rules, it will be much better to get them in a list and not one by one

nanteby (commented):
Hi Marian,
Let us please change the API

experimental/saitypesextensions.h
```diff
@@ -42,6 +42,42 @@ typedef enum _sai_object_type_extensions_t
 
     SAI_OBJECT_TYPE_TABLE_META_TUNNEL_ENTRY,
 
+    SAI_OBJECT_TYPE_DIRECTION_LOOKUP_ENTRY,
+
+    SAI_OBJECT_TYPE_SLB_DECAP_ENTRY,
```
elibritstein, line 47: what is slb_decap?

experimental/saitypesextensions.h
```diff
@@ -42,6 +42,42 @@ typedef enum _sai_object_type_extensions_t
...
+
+    SAI_OBJECT_TYPE_ENI_LOOKUP_TO_VM_ENTRY,
+
+    SAI_OBJECT_TYPE_INBOUND_ROUTING_ENTRY,
+
+    SAI_OBJECT_TYPE_PA_VALIDATION_ENTRY,
```
elibritstein, line 53: according to the p4, there is a validation for the PA for inbound.
currently our "PA" is determined by "ip link/ifconfig of the ports" before the application starts, and the application only reads this value upon init. do you mean we need to set it from within the application and not externally?
please check if the way we did it is acceptable, because if not and need to change on the fly, it will be large effort to do.

prsunny (approved):

experimental/saiexperimentaldash.h
```diff
@@ -0,0 +1,1765 @@
...
+
+    /** End of custom range base */
+    SAI_OUTBOUND_ENI_TO_VNI_ENTRY_ATTR_CUSTOM_RANGE_END,
+
+} sai_outbound_eni_to_vni_entry_attr_t;
+
+/**
+ * @brief Attribute ID for dash_acl
+ */
+typedef enum _sai_dash_acl_attr_t
```
prsunny, line 415: I think we may want to have ACL rule and group separated and this would be `sai_dash_acl_rule_attr_t`

experimental/saiexperimentaldash.h
```diff
@@ -0,0 +1,1765 @@
...
+    SAI_DASH_ACL_ATTR_DST_PORT,
+
+    /**
+     * @brief Stage
+     *
+     * @type sai_dash_acl_stage_t
+     * @flags CREATE_AND_SET
+     * @default SAI_DASH_ACL_STAGE_OUTBOUND_ACL_STAGE1
+     */
+    SAI_DASH_ACL_ATTR_STAGE,
```
prsunny, line 486: I don't think stage belongs here. It should be outside of Rule attributes

prsunny (approved):
need to add metering id as well.

PR #1356, opened 2021-12-13
Bulk Set for NH Group Object

JaiOCP:
Bulk set API support for NH group object. This API will support ability to modify multiple member attributes in a NH group.


itaibaz (commented):
If we are adding bulk set for next hop group member, I think it would be good to add as well bulk create, bulk remove, bulk get
With most other objects that have bulk API, such as routes, NAT, FDB, we did the entire 4 CRUD functions

Some use cases -
Bulk set can be used with fine group, when one link fails and need to update all next hop members pointing to it
Bulk create can be used with big fine group (e.g. 4K group) to optimize creation time instead of one by one creation
Same for bulk remove

JaiOCP:
@marian-pritsak Please approve

kcudnik (approved):

rlhui:
@itaibaz , @marian-pritsak  - would you please help review/approve this?

itaibaz (approved):

PR #1361, opened 2021-12-17
SAI PTF 13: switch attributes tests and lpm module

pkochmanska:


rlhui:
Can we update the title and the description of this change.  It is not about saiswitch.

rlhui (approved):

PR #1365, opened 2021-12-20
SAI NAT aging notification

arvbb:
In the Hit Bit query mechanism described in [https://github.com/opencomputeproject/SAI/blob/master/doc/NAT/SAI-NAT-API.md]( https://github.com/opencomputeproject/SAI/blob/master/doc/NAT/SAI-NAT-API.md), the NOS periodically polls the NAT entries for aging out unused entries. In a highly scaled environment with tens of thousands of NAT entries programmed, the frequent polling for aging out is not performant. An alternative is to use a callback from SAI to notify NOS about the NAT entries that are aged out after a certain time. 

The NOS registers a callback named sai_nat_event_notification_fn through the switch attribute SAI_SWITCH_ATTR_NAT_EVENT_NOTIFY. 

The NOS then create/sets the NAT entry with the optional attribute SAI_NAT_ENTRY_ATTR_AGING_TIME. 

nat_entry_attr[0].id = SAI_NAT_ENTRY_ATTR_AGING_TIME; 
nat_entry_attr[0].value.u32 = aging_time; // in seconds 

If the attribute is not added the NAT entry does not age out. If the aging time is specified as 0, the NAT entry does not age out. The aging time configured can be queried using a GET request of this attribute. 

Once the NAT entry ages out, SAI notifies NOS using sai_nat_event_notification_fn callback. This callback provides events of type sai_nat_event_notification_data_t which has information about the specific NAT entries that got aged out. The NOS uses the NAT entry information to delete these NAT entries.

### DIFF
--- a/inc/sainat.h
+++ b/inc/sainat.h
@@ -210,6 +210,17 @@ typedef enum _sai_nat_entry_attr_t
     SAI_NAT_ENTRY_ATTR_HIT_BIT,
 
     /**
+     * @brief NAT entry aging time in seconds
+     *
+     * Zero means aging is disabled.
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_NAT_ENTRY_ATTR_AGING_TIME,
+
+    /**
      * @brief End of NAT Entry attributes
      */
     SAI_NAT_ENTRY_ATTR_END,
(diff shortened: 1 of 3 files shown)


kcudnik (approved):

arvbb:
Updating after review in SAI meeting on Jan 20, 2022:

1. A query mechanism can be used to check if a specific SAI implementation supports the NAT aging mechanism. The NOS can then query using sai_query_attribute_capability() API as follows:

    sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_NAT_ENTRY, SAI_NAT_ENTRY_ATTR_AGING_TIME, &capability);
 
  The returned 'capability' would indicate if SAI supports NAT aging



2. NOS can use an optional switch attribute SAI_SWITCH_ATTR_NAT_EVENT_NOTIFY_INTERVAL to inform SAI on frequency of invoking the sai_nat_event_notification_fn callback.

  If this attribute is not specified or if attribute set to 0 (i.e. default), SAI notifies NOS as soon as NAT entry is aged out.
  SAI can still do bulking of notifications even when this interval is 0.
 
 
    /**
     * @brief NAT event notification callback timeout in seconds
     *
     * SAI to notify NOS through the sai_nat_event_notification_fn callback at specified duration
     *
     * @type sai_uint32_t
     * @flags CREATE_AND_SET
     * @default 0
     */
    SAI_SWITCH_ATTR_NAT_EVENT_NOTIFY_INTERVAL,

rlhui:
@JaiOCP , @marian-pritsak  - please help review/approve?

arlakshm (approved):
lgtm

rlhui:
@JaiOCP  - reminder to please review/approve this. Thanks.

JaiOCP (commented):
Looks Good

PR #1386, opened 2022-01-05
Proposal for init switch and destroy switch from PRC

richardyu-ms:
keep the switch id in creat_switch and remove it from remove_switch

1. support to create the create_switch only once before the test run and use the same switch id within multi-round of testing
2. support remove the switch id


richardyu-ms:
@pkochmanska @aljer @kcudnik for a review

PR #1388, opened 2022-01-05
[SAIServer]Add new SaiServer to support ptf-SAI test structure

richardyu-ms:
support build saiserver in two versions

Add build script to generate the necessary package as saiserverv2
Add new saiserver to support new structure
Add makefile
add permission for copy_saithriftv2.sh
copy saiserver at dh builddeb as v2 (new version sai ptf)
install sai_adapter in sai_thrift package

test done:
brcm and bfn platform

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,11 @@
 #    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
 #    Dell Products, L.P., Facebook, Inc
 #
-
+ifdef SAITHRIFTV2
+SAITHRIFT_PATH=test/saithriftv2
+else
+SAITHRIFT_PATH=test/saithrift
+endif
 
 .PHONY: test doc clean
 
@@ -26,10 +30,10 @@ test:
 	make -C test
 
 saithrift-build:
-	make -C test/saithrift
+	make -C $(SAITHRIFT_PATH)
 
 saithrift-install: saithrift-build
-	make -C test/saithrift install
+	make -C $(SAITHRIFT_PATH) install
 
 clean:
 	make -C test clean    
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,13 @@
 #    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
 #    Dell Products, L.P., Facebook, Inc
 #
+
+# There are two versions of saithrift, in order not to break the origin tests, use ENVIRONMENT to build them respectively.
+# By default, it will make the saiserver and the related artifacts under folder test/saithrift.
+# For example, build saithrift v2
+# - export  SAITHRIFTV2=y
+# - make saithrift-build
+
 ifdef SAITHRIFTV2
 SAITHRIFT_PATH=test/saithriftv2
 else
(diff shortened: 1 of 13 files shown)


richardyu-ms:
@pkochmanska @aljer @kcudnik, could help make a review. thanks.

ptf/sai_base_test.py
```diff
@@ -34,7 +34,7 @@
 from sai_thrift import sai_rpc
 
 from sai_utils import *
-import sai_adapter as adapter
+import sai_thrift.sai_adapter as adapter
```
richardyu-ms, line 37: @aljer @pkochmanska 
In this new sai-ptf structure, looks like there are two kinds of files, handwriting code, and auto-generated code.
in order to make the test cases related files to import the generated files from an installed package, try to organize those generated files into a package.
Please help make a review, not sure if this will impact the design of this sai-ptf structure.

thanks.

test/saithriftv2/Makefile
```diff
@@ -0,0 +1,118 @@
...
+LDFLAGS = -L$(SAI_LIBRARY_DIR) -Wl,-rpath=$(SAI_LIBRARY_DIR)
+CPPFLAGS = -I$(SAI_HEADER_DIR) -I. -std=c++11
+ifeq ($(DEBUG),1)
+CPPFLAGS += -O0 -ggdb
+endif
+
+ifeq ($(platform),MLNX)
+CDEFS = -DMLNXSAI
+else
+ifeq ($(platform),BFT)
```
ravi861, line 24: BFT is a new vendor or a typo for BFN?

test/saithriftv2/Makefile
```diff
@@ -0,0 +1,118 @@
...
+				gen-cpp/sai_constants.h \
+				gen-cpp/sai_rpc.cpp \
+				gen-cpp/sai_rpc.h \
+				gen-cpp/sai_types.cpp \
+				gen-cpp/sai_types.h
+
+PY_SOURCES = 	gen-py/__init__.py \
+				gen-py/sai/constants.py \
+				gen-py/sai/__init__.py \
+				gen-py/sai/switch_sai_rpc.py \
```
ravi861, line 58: This is the old file. New file is gen-py/sai/sai_rpc.py and gen-py/sai/sai_rpc-remote

Makefile
```diff
@@ -15,7 +15,11 @@
 #    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
 #    Dell Products, L.P., Facebook, Inc
 #
-
+ifdef SAITHRIFTV2
```
ravi861, line 18: A comment here explaining why this is needed may help

ravi861: In addition, a CLI command on how to invoke this can be added here.

ptf/sai_base_test.py
ravi861: You mean this has to be incorporated into all test files?

richardyu-ms: Yes, this will impact all the test files.
Cause we import the sai_adapter in the sai_base_test, we don't need to change every test case.
the purpose of this change is to install all the dependent files into an installation package,(certainly, we can install those files under dist-package directly), and give it a naming space.

ravi861: I think we can update the name space. It will be more explicit

Makefile
richardyu-ms: thx. added more comment

RyoYang (approved):

PR #1393, opened 2022-01-10
Src fdb meta data 3

shri-khare:


inc/saiswitch.h
```diff
@@ -739,111 +739,119 @@ typedef enum _sai_switch_attr_t
...
+    SAI_SWITCH_ATTR_ACL_ENTRY_MINIMUM_PRIORITY = SAI_SWITCH_ATTR_START + 0x19,
 
     /**
      * @brief Maximum priority for ACL entry
      *
      * @type sai_uint32_t
      * @flags READ_ONLY
      */
-    SAI_SWITCH_ATTR_ACL_ENTRY_MAXIMUM_PRIORITY,
+    SAI_SWITCH_ATTR_ACL_ENTRY_MAXIMUM_PRIORITY = SAI_SWITCH_ATTR_START + 0x1a,
```
kcudnik, line 774: you dont need to use SAI_SWITCH_ATTR_START , you can just use explicit value, ps why you need to reorder everything ?

inc/saiswitch.h
```diff
@@ -1463,32 +1479,32 @@ typedef enum _sai_switch_attr_t
...
      */
-    SAI_SWITCH_ATTR_LAG_DEFAULT_HASH_OFFSET,
+    SAI_SWITCH_ATTR_LAG_DEFAULT_HASH_OFFSET = SAI_SWITCH_ATTR_START + 0x62,
 
     /**
      * @brief SAI LAG default symmetric hash
      *
-     * When set, the hash calculation will result in the same value as when the source and
-     * destination addresses (L2 src/dst MAC,L3 src/dst IP,L4 src/dst port) were swapped,
+     * When set  = SAI_SWITCH_ATTR_START + 0x,the hash calculation will result in the same value as when the source and
```
kcudnik, line 1499: some search/replace issue when replacing comma ","

shri-khare: Thanks @kcudnik. Am unable to run meta checks locally, and thus have to rely on checks that auto run when I create a draft PR. That is making it very hard to iterate and fix these editorial mistakes.

I have a fresh CentOS 8 install with

sudo dnf install perl-XML-Simple aspell aspell-en graphviz doxygen

git clone https://github.com/opencomputeproject/SAI.git sai.git
cd sai.git/meta
make

So, this does not carry my changes, but running with top of the tree sai.git. And yet, it fails with below:

ERROR: invalid type '' for SAI_ACL_COUNTER_ATTR_TABLE_ID
ERROR: flags are not defined for SAI_ACL_COUNTER_ATTR_TABLE_ID
ERROR: FATAL ERROR === MUST FIX === : Can't use an undefined value as an ARRAY reference at parse.pl line 1676.


any pointer on what might be missing there? Thanks

shri-khare: To add, the first run of make from meta directory gives me an error of form:

/home/skhareadmin/sai.git/experimental/saiexperimentalbmtor.h:284: warning: Illegal command par as part of a title section
/home/skhareadmin/sai.git/experimental/saiexperimentalbmtor.h:284: warning: Illegal command n as part of a title section
/home/skhareadmin/sai.git/experimental/saiexperimentalbmtor.h:285: warning: Illegal command par as part of a title section
/home/skhareadmin/sai.git/experimental/saiexperimentalbmtor.h:285: warning: Illegal command n as part of a title section
/home/skhareadmin/sai.git/meta/saimetadatatypes.h:452: warning: Illegal command n as part of a title section
<unknown>:1: warning: Illegal command n as part of a title section
make: *** [Makefile:101: xml] Error 1

and subsequent run fails with the error I originally reported above.

kcudnik:
you probably have doxygen error ralated that \n was replaced by ^^ in Doxygen file

inc/saiswitch.h
shri-khare: Thanks @kcudnik.

It helped me debug that the issue occurs on doxygen later than 1.8.14.
I compiled and used doxygen 1.8.13, and could locally run make in meta directory.

Should we update the below README to list that requirement?
https://github.com/opencomputeproject/SAI/blob/master/meta/README.md

I could send a PR for that. Pls let me know.

Also, I could create PR #1398 for this, its a draft and I have posted few questions on it.

kcudnik: Sure, readme can be updated for this description of different doxygen versions

PR #1398, opened 2022-01-18
Src fdb meta data

shri-khare:


shri-khare:
[Failures](https://dev.azure.com/mssonic/build/_build/results?buildId=65811&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=411) are of the following form:

ERROR: check ! SAI_ACL_TABLE_ATTR_FIELD_ROUTE_NPU_META_DST_HIT value is 0x0000115b, but on was 0x00001044 on commit 3132018

However, even before this change, value of SAI_ACL_TABLE_ATTR_FIELD_ROUTE_NPU_META_DST_HIT is 0x104c and not 0x1044 as the log above suggests. Reference: https://github.com/opencomputeproject/SAI/blob/master/inc/saiacl.h#L1307

Thus, am not sure if the check is honoring the enum value set in the sources. @kcudnik thoughts on what might be missing?

inc/saiswitch.h
```diff
@@ -555,7 +555,7 @@ typedef enum _sai_switch_attr_t
      * @type sai_uint32_t
      * @flags READ_ONLY
      */
-    SAI_SWITCH_ATTR_MAX_NUMBER_OF_SUPPORTED_PORTS,
+    SAI_SWITCH_ATTR_MAX_NUMBER_OF_SUPPORTED_PORTS = SAI_SWITCH_ATTR_START + 0x1,
```
kcudnik, line 558: maybe not use SAI_SWITCH_ATTR_START , just direct value 0x1, and on other

inc/saifdb.h
```diff
@@ -114,7 +114,7 @@ typedef enum _sai_fdb_entry_attr_t
      * @flags CREATE_AND_SET
      * @default SAI_PACKET_ACTION_FORWARD
      */
-    SAI_FDB_ENTRY_ATTR_PACKET_ACTION,
+    SAI_FDB_ENTRY_ATTR_PACKET_ACTION = SAI_FDB_ENTRY_ATTR_START + 0x1,
```
kcudnik, line 117: maybe not use SAI_FDB_ENTRY_ATTR_START , just direct value 0x1, and on other

shri-khare: @kcudnik I could fix it and repost the PR, but please see my question above on the errors which would persist even without that: https://github.com/opencomputeproject/SAI/pull/1398#issuecomment-1014988193

kcudnik:
> [Failures](https://dev.azure.com/mssonic/build/_build/results?buildId=65811&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=411) are of the following form:
> 
> ERROR: check ! SAI_ACL_TABLE_ATTR_FIELD_ROUTE_NPU_META_DST_HIT value is 0x0000115b, but on was 0x00001044 on commit [3132018](https://github.com/opencomputeproject/SAI/commit/31320180d4b1ff9087474f219092326df06b1e00)
> 
> However, even before this change, value of SAI_ACL_TABLE_ATTR_FIELD_ROUTE_NPU_META_DST_HIT is 0x104c and not 0x1044 as the log above suggests. Reference: https://github.com/opencomputeproject/SAI/blob/master/inc/saiacl.h#L1307
> 
> Thus, am not sure if the check is honoring the enum value set in the sources. @kcudnik thoughts on what might be missing?

yea, seems like value is 0x104c, probably there is a bug in parsing logic that incorrectly calculates value for SAI_ACL_TABLE_ATTR_FIELD_ROUTE_NPU_META_DST_HIT, i will take a look into that

kcudnik:
ok, i got this, error is related to this commit: https://github.com/opencomputeproject/SAI/pull/1398/commits/97fb4df895e0c33bd955c6dd5117d8c97c76e345

which is first commit you made, and it's specifies this :
```
SAI_ACL_TABLE_ATTR_FIELD_FDB_SRC_USER_META = SAI_ACL_TABLE_ATTR_FIELD_START + 0x15b,
```

and for this logic value will be 0x0000115b, and logs are also specifying this error  for this specific commit:
```
processing commit 97fb4df
ERROR: check ! SAI_ACL_TABLE_ATTR_FIELD_ROUTE_NPU_META_DST_HIT value is 0x0000115b, but on was 0x00001044 on commit 3132018
```

unfortunately current ancestry logic is checking each commit separately (in case you want to rebase instead of squash commits in PR) and everything must be consistent and values can't change between commits, so to fix this you will need to force push this PR with squashed commits to final form, or the one that don't change 

plus, you will need to fix SAI_ACL_TABLE_ATTR_FIELD_SRC_IPV6_ shift, and SWITCH_ATTR shift

PR #560, opened 2017-05-11
Update saitypes.h

tushar-ty:
Modified bulk operation type to error mode type.

PR #1400, opened 2022-01-20
Add support for VLAN Stacking/Translation API

kuanyu99:
- Define SAI layer VLAN translation and VLAN Stacking function API in saivlan.h

Use cases: https://github.com/Azure/SONiC/blob/f7920604b77525bd8a13432214fd8d4a5423933c/doc/vlan_stacking/vlan_stacking_HLD.md#vlan-stacking-deployment-use-cases

Examples:
https://github.com/Azure/SONiC/blob/f7920604b77525bd8a13432214fd8d4a5423933c/doc/vlan_stacking/vlan_stacking_HLD.md#sai-api


itaibaz:
Should we also extend saiacl.h to allow pop / swap ?
We have SAI_ACL_ACTION_TYPE_ADD_VLAN_ID (push) and set outer/inner vlan ID (swap use case)
But I don't see remove_vlan (pop)

kcudnik (approved):

pttch:
Add for CLI commands, will commit later

CLI
Add VLAN stacking rule on a physical interface or PortChannel

config vlan-stacking add <interface_name> <stage> <match_vlan> <action> <apply_vlan>
  * interface_name: The name of physical interface or PortChannel, e.g. "Ethernet0" or "PortChannel0001".
  * stage         : The traffic direction of the packet, e.g. "ingress" or "egress".
  * match_vlan    : The VLAN ID on the packet which expected to match the rule(The range is from 1 to 4094).
  * action        : The action applied to the packet when it matched the rule, e.g. "push", "pop" or "swap".
  * apply_vlan    : The VLAN ID applied to the packet according to the action field(The range is from 1 to 4094).
The format is in the following

admin@sonic:~$ sudo config vlan-stacking add -h
Usage: config vlan-stacking add [OPTIONS] <interface_name> <stage> <match_vlan> <action> <apply_vlan>

  Add vlan stacking rule, examples:
  - config vlan-stacking add Ethernet0 ingress 100 push 200
  - config vlan-stacking add Ethernet0 egress 100 pop
  - config vlan-stacking add Ethernet0 ingress 100 swap 200
  - config vlan-stacking add Ethernet0 egress 100 swap 200

Options:
  -?, -h, --help  Show this message and exit.
Delete VLAN stacking rule from a physical interface or PortChannel

config vlan-stacking del <interface_name> <stage> <match_vlan>
  * interface_name: The name of physical interface or PortChannel, e.g. "Ethernet0" or "PortChannel0001".
  * stage         : The traffic direction of the packet, e.g. "ingress" or "egress".
  * match_vlan    : The VLAN ID on the packet which expected to match the rule(The range is from 1 to 4094).
The format is in the following

admin@sonic:~$ sudo config vlan-stacking del -h
Usage: config vlan-stacking del [OPTIONS] <interface_name> <stage> <match_vlan>

  Delete vlan stacking rule, examples:
  - config vlan-stacking del Ethernet0 ingress 100
  - config vlan-stacking del Ethernet0 egress 100

Options:
  -?, -h, --help  Show this message and exit.
Display VLAN stacking rule configuration

show vlan-stacking
The display format is in the following

admin@root:~# show vlan-stacking
Interface    Stage      Match VLAN  Action      Apply VLAN
-----------  -------  ------------  --------  ------------
Ethernet0    egress            200  pop
Ethernet0    ingress           100  push               200
Ethernet4    egress            200  swap               100
Ethernet4    ingress           100  swap               200

rlhui:
Need to have a SAI specific doc for this.  And also describe how this fits to SAI behavioral model.

kuanyu99:
> Need to have a SAI specific doc for this. And also describe how this fits to SAI behavioral model.

I add a document about this new PR and also add a vlan stack stage into the pipeline.
I have some problem editing the vsdx file. So I use a PDF to represent the pipeline model about the vlan stacking.

kcudnik:
please fix errors:
```
WARNING: not expected number of spaces after * (1,4,8) saivlan.h 561:     *   outer vlan tag, the original one is an inner vlan tag.
WARNING: Long brief > 200 on SAI_VLAN_STACK_ATTR_MATCH_TYPE:
 - Match type The original Vlan ID should match to inner ta...
```

kcudnik (approved):

doc/SAI-Proposal-VlanStack-QinQ.md
```diff
@@ -0,0 +1,380 @@
...
+attr.id = SAI_VLAN_STACK_ATTR_STAGE;
+attr.value.s32 = SAI_VLAN_STACK_STAGE_INGRESS;
+vlan_stacking_entry_attrs.push(attr);
+
+attr.id = SAI_VLAN_STACK_ATTR_ACTION;
+attr.value.s32 = SAI_VLAN_STACK_ACTION_PUSH;
+vlan_stacking_entry_attrs.push(attr);
+
+attr.id = SAI_VLAN_STACK_ATTR_MATCH_TYPE;
+attr.value.s32 = SAI_VLAN_STACK_MATCH_TYPE_INNER;
```
ashutosh-agrawal, line 73: Should the match type be outer for the ingress stage? For example, in the 2nd row of the table below, we are matching on the outer tag (svid=10) of the packet.

kuanyu99: I know this point will confuse us easily. So, I added some notes at #560 of inc/saivlan.h
```
    /**
     * @brief Match type
     * The original Vlan ID should match to inner tag or outer tag
     * For PUSH action, this should always set to INNER. Since compare to the new
     * outer tag, the original one is an inner tag.
     *
     * @type sai_vlan_stack_match_type_t
     */
    SAI_VLAN_STACK_ATTR_MATCH_TYPE,
```

This design also make us simple to implement. But I think this can open for discussion.

ashutosh-agrawal: So, you are saying that match_type is based on how the packet looks like **after** all the stacking operations have been performed? Do we have a case where match_type will be outer?

kuanyu99: Not really at all.
This is what I think.
The PUSH action usually be used in QinQ case. When a packet come in, a new vlan tag will be added based on the inner tag. And when a packet go out, the vlan tag will be removed based on the outer tag. Doesn't this sounds more reasonable?

And for SWAP action. I can swap the inner vlan tag or the outer vlan tag.

doc/SAI-Proposal-VlanStack-QinQ.md
```diff
@@ -0,0 +1,380 @@
...
+![vlan_stack-pipeline](figures/vlan_stack/vlan_stack_pipeline.png "Figure 4: Vlan Stack SAI Pipeline")
+__Figure 4: Vlan Stack SAI Pipeline__
+
+## SAI Attribute
+
+| SAI Component | SAI attribute                                  | Description                     |
+| ------------- | ---------------------------------------------- | ------------------------------- |
+| VLAN Stacking | SAI_VLAN_STACK_ATTR_STAGE                      | At which packet flow direction the action should take effect. Ingress or Egress    |
+|               | SAI_VLAN_STACK_ATTR_ACTION                     | Include SWAP, PUSH and POP operation    |
+|               | SAI_VLAN_STACK_ATTR_MATCH_TYPE                 | At which dot1q header position should be used to match original vlan ID. Inner or Outer. Please note that for a PUSH action, this should always set to Inner.    |
```
prvattem, line 45: Could you please enhance this to support matching multi-tag ?
Example : `Match (OVID-100/IVID-10) translate to SVID-200 and vice versa`

doc/SAI-Proposal-VlanStack-QinQ.md
prvattem: This might be applicable, when we match a packet with double-tag. Otherwise, we should always be matching the outer-tag?

doc/SAI-Proposal-VlanStack-QinQ.md
```diff
@@ -0,0 +1,380 @@
...
+sai_vlan_api->create_vlan_stack(&vlan_stacking_oid, gSwitchId, (uint32_t)vlan_stacking_entry_attrs.size(), vlan_stacking_entry_attrs.data());
+```
+
+#### Behavior
+
+![vlan_stack-egress-pop-outer](figures/vlan_stack/egress_pop_outer.png "Table 2: Egress Pop Outer")
+__Table 2: Vlan Stack Egress Pop Outer__
+
+### Egress/Remove outer tag based on inner tag
+Remove outer dot1q header for those packets with original inner dot1q header VID=100 at Ethernet20
```
prvattem, line 147: Isn't this contradicting the SAI Egress pipeline? `Filter the packets based on first VID or PVID`
In this example we are matching "Inner" and popping "outer"

doc/SAI-Proposal-VlanStack-QinQ.md
```diff
@@ -0,0 +1,380 @@
...
+attr.value.u8 = 0;
+vlan_stacking_entry_attrs.push(attr);
+
+sai_vlan_api->create_vlan_stack(&vlan_stacking_oid, gSwitchId, (uint32_t)vlan_stacking_entry_attrs.size(), vlan_stacking_entry_attrs.data());
+```
+
+#### Behavior
+
+![vlan_stack-ingress-swap-outer](figures/vlan_stack/ingress_swap_outer.png "Table 5: Ingress Swap Outer")
+__Table 5: Vlan Stack Ingress Swap Outer__
```
prvattem, line 284: Typo??
image
Here outer matches with VID 10 and gets swapped to VID100

inc/saivlan.h
```diff
@@ -484,6 +484,193 @@ typedef enum _sai_vlan_stat_t
...
+
+/**
+ * @brief VLAN Stack match type
+ */
+typedef enum _sai_vlan_stack_match_type_t
+{
+    SAI_VLAN_STACK_MATCH_TYPE_INNER,
+
+    SAI_VLAN_STACK_MATCH_TYPE_OUTER,
+
```
prvattem, line 526: Please enhance this to match multi-tag also. To start with a double-tag matching. 
ex: SAI_VLAN_STACK_MATCH_TYPE_MULTI with stack depth 2?

doc/SAI-Proposal-VlanStack-QinQ.md
kuanyu99: This case might not be so useful. Since no people will want to send packets out of a port with different VID.
But in the behavior pipeline, this is a possible case. The packet will first be filtered by the port based on its VID, and do the vlan stacking afterwards.

The vlan stacking stage will be after the egress vlan filtering.

doc/SAI-Proposal-VlanStack-QinQ.md
kuanyu99: Yes, thank you. I will fix this in next commit.

doc/SAI-Proposal-VlanStack-QinQ.md
```diff
@@ -0,0 +1,380 @@
...
+attr.value.u8 = 0;
+vlan_stacking_entry_attrs.push(attr);
+
+sai_vlan_api->create_vlan_stack(&vlan_stacking_oid, gSwitchId, (uint32_t)vlan_stacking_entry_attrs.size(), vlan_stacking_entry_attrs.data());
+```
+
+#### Behavior
+
+![vlan_stack-ingress-push-inner](figures/vlan_stack/ingress_push_inner.png "Table 1: Ingress Push")
+__Table 1: Vlan Stack Ingress Push__
```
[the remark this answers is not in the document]
kuanyu99: Yes, it should be a trunk port.
A packet come into the port will first do the PUSH SVID action. After that, the ingress vlan filtering stage will check if the SVID match the port setting.

doc/SAI-Proposal-VlanStack-QinQ.md
kuanyu99: I will enhance this in the next commit ASAP.

doc/SAI-Proposal-VlanStack-QinQ.md
kuanyu99: You two persuade me. I will change this case to match outer in the next commit.

doc/SAI-Proposal-VlanStack-QinQ.md
kuanyu99: I think I will remove this case in the next commit. Due to it is not so useful and also confuse people easily.

doc/SAI-Proposal-VlanStack-QinQ.md
kuanyu99: I remove the attribute, SAI_VLAN_STACK_ATTR_MATCH_TYPE. And add a new data type, sai_vlan_stacking_vid_t, to match header position and vid in the same time.
For your example, please refer "Ingress/Swap on double tag from single tag" and "Egress/Swap on single tag from double tag" to achieve the goal.

doc/SAI-Proposal-VlanStack-QinQ.md
```diff
@@ -0,0 +1,466 @@
...
+
+## Overview ##
+
+The Vlan Stack feature includes two functions, vlan translation and QinQ.
+
+![qinq](figures/vlan_stack/vlan_stack_QinQ_usecase.png "Figure 1: QinQ usage scenario")
+__Figure 1: QinQ usage scenario__
+
+![vlan_translate](figures/vlan_stack/vlan_stack_vlan_xlate_usecase.png "Figure 2: Vlan translation usage scenario")
+__Figure 2: Vlan Translation usage scenario__
```
prvattem, line 24: If L3 is enabled (ex: ip addr configured) on the translated VLAN and if the incoming packets are copied to CPU, what would be the VLAN on the packet? is it the CVLAN or the Translated VLAN?

kuanyu99: In some cases, the packets are copied to CPU before handling by the switch chip. For those packets, the VLAN will be the one before translated.
But this may be different for other chip vendors.

inc/saitypes.h
```diff
@@ -1308,6 +1323,9 @@ typedef union _sai_attribute_value_t
 
     /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_PORT_ERR_STATUS_LIST */
     sai_port_err_status_list_t porterror;
+
+    /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_VLAN_STACKING_VID */
+    sai_vlan_stacking_vid_t vlanstacking_vid;
```
kcudnik, line 1328: don't use underscore, all other attr values don't use them

meta/parse.pl
```diff
@@ -359,6 +359,8 @@ sub ProcessTagDefault
 
     return $val if $val eq "\"\"";
 
+    return $val if $val =~ /^0\,0$/;
```
kcudnik, line 362: not need for \\, since comma is not regex special

meta/parse.pl
```diff
@@ -1569,6 +1571,8 @@ sub ProcessDefaultValueType
 
     return "SAI_DEFAULT_VALUE_TYPE_CONST" if $default eq "\"\"";
 
+    return "SAI_DEFAULT_VALUE_TYPE_CONST" if $default =~ /^0\,0$/;
```
kcudnik, line 1574: not need for \\, since comma is not regex special

meta/parse.pl
```diff
@@ -1640,6 +1644,10 @@ sub ProcessDefaultValue
     {
         WriteSource "$val = { .mac = { 0, 0, 0, 0, 0, 0 } };";
     }
+    elsif ($default =~ /^0\,0$/ and $type =~ /^(sai_vlan_stacking_vid_t)/)
```
kcudnik, line 1647: not need for \\, since comma is not regex special

meta/saimetadatatypes.h
```diff
@@ -451,6 +451,11 @@ typedef enum _sai_attr_value_type_t
      */
     SAI_ATTR_VALUE_TYPE_AUTH_KEY,
 
+    /**
+     * @brief Attribute value is vlan id.
```
kcudnik, line 455: description not matching actual field

kcudnik (changes requested):
please address comments

kcudnik (approved):

doc/SAI-Proposal-VlanStack-QinQ.md
ashutosh-agrawal: Should we define the expected behavior as part of this proposal? I understand that some of the older switching chips may not be compliant with the SAI requirements.

kuanyu99: I think this PR already defined the expected behavior of the vlan header when the 802.1ad rule was set.
All packets transferred in Layer 2 should follow the rule.
It is another thing how the switch chip handle copy-to-CPU packets technically.

kuanyu99:
Hi @rlhui @ashutosh-agrawal  @itaibaz @kcudnik ,
Can we merge this PR? Or what should I do to proceed it?

kcudnik:
please wait for others to apprive

inc/saivlan.h
```diff
@@ -487,6 +487,170 @@ typedef enum _sai_vlan_stat_t
...
+    /**
+     * @brief The packet priority (3 bits) in the vlan tag
+     * Range from 0 to 7, priority from low to high.
+     * Default 0xFF will inherit the original vlan tag priority
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0xFF
+     */
+    SAI_VLAN_STACK_ATTR_VLAN_APPLIED_PRI,
```
prvattem, line 587: The HLD definition of Priority attribute is not consistent with what has been described here.
"Rule priority from low to high can assign this from 0 to 7. If a packet can match multiple vlan stacking rule, only the highest priority rule will take effect."

could you please give an example, how this attribute is used?

kuanyu99: I think maybe my statement make mistake easily. In both places, I try to describe the same thing.
0 is the loweset priority and 7 is the highest priority.

For example:
I set a (10-to-20) vlan translation with priority 0 on the port1. After that, I set a (30-to-40) vlan translation with priority 7 on the port1.
Only the (30-to-40) vlan translation will take effect.

If my explanation reasonable, I will then append it into the HLD to make the attribute function more clear.

kuanyu99: The statement in the HLD is wrong. I will correct it now.
The attribute there is used to fill the "priority" column of the VLAN tag.

If set this value to 0xFF, the priority of the transformed VLAN tag will be equal to the one before the transformation.

inc/saitypes.h
```diff
@@ -1335,6 +1350,9 @@ typedef union _sai_attribute_value_t
 
     /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_LATCH_STATUS */
     sai_latch_status_t latchstatus;
+
+    /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_VLAN_STACKING_VID */
+    sai_vlan_stacking_vid_t vlanstackingvid;
```
srj102, line 1355: Do we need a new type ? 
Can we not add more attributes instead for the inner and outer vlan. ?
Not adding a new type to the union helps avoid all the meta changes in SAI and also simplifies meta changes in sonic.

inc/saivlan.h
```diff
@@ -487,6 +487,170 @@ typedef enum _sai_vlan_stat_t
...
+    SAI_VLAN_STACK_ATTR_ACTION,
+
+    /**
+     * @brief Original Vlan ID of the inner or outer
+     * The action will perform on the packets whose vlan tag match the corresponding vlan ID
+     *
+     * @type sai_vlan_stacking_vid_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_VLAN_STACK_ATTR_ORIGINAL_VLAN_ID,
```
srj102, line 556: To avoid meta changes can we use the foll:

SAI_VLAN_STACK_ATTR_ORIGINAL_VLAN_ID_OUTER
SAI_VLAN_STACK_ATTR_ORIGINAL_VLAN_ID_INNER

inc/saivlan.h
```diff
@@ -657,7 +821,10 @@ typedef struct _sai_vlan_api_t
     sai_get_vlan_stats_fn               get_vlan_stats;
     sai_get_vlan_stats_ext_fn           get_vlan_stats_ext;
     sai_clear_vlan_stats_fn             clear_vlan_stats;
-
+    sai_create_vlan_stack_fn            create_vlan_stack;
```
srj102, line 824: vlan stack/translation is a different stage in the pipeline.
It involves tag manipulations only. 

Should we be modifying the sai_vlan_api_t ?

Instead Is it better to add a new API sai_vlan_stack_api_t and add 
the vlan_stack related functions there ?

srj102 (changes requested):

inc/saitypes.h
kuanyu99: If considering this situation, I think change to use individual attributes can achieve the same thing.

inc/saivlan.h
kuanyu99: It is also an option. But if I do so, the sai_vlan_stack_api_t should be put in another header file, right?

srj102: yes that would be another header file. However since we this is another stage in the pipeline it should be fine.

kuanyu99: Hi @srj102 ,
I have split the vlan stack into another header file. Please check if it helps.

inc/saivlan.h
kuanyu99: Done. Please check.

meta/saimetadatatypes.h
kuanyu99: Done

meta/parse.pl
kuanyu99: Corrected

meta/parse.pl
kuanyu99: Corrected

inc/saitypes.h
kuanyu99: Corrected

meta/parse.pl
kuanyu99: Corrected

inc/saivlan.h
kuanyu99: Sorry, I don't fully understand the operation you mentioned.
Can you check if the current design is enough for your desire? Or can you give me some examples about your mentioned operation.

kuanyu99:
Hi @ashutosh-agrawal @srj102 @prvattem 
I have made some updates based on the comments recently.
Please check if anything needs to be corrected or can we let the PR be merged.

kcudnik (approved):

inc/saivlanstack.h
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * @brief The packet priority (3 bits) in the vlan tag
+     * Range from 0 to 7, priority from low to high.
+     * Default 0xFF will inherit the original vlan tag priority
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0xFF
+     */
+    SAI_VLAN_STACK_ATTR_VLAN_APPLIED_PRI,
```
AnandhiDhanabalan, line 157: Similar to _PRI, the CFI bit in vlan tag should be handled.

PR #1402, opened 2022-01-27
[Port] 8-lanes interface types

MaksymPry:
This commit adds 8-lanes port interface types: CR8, KR8, SR8, LR8.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -316,6 +316,18 @@ typedef enum _sai_port_interface_type_t
     /** Interface type XGMII */
     SAI_PORT_INTERFACE_TYPE_XGMII,
 
+    /** Interface type CR8 */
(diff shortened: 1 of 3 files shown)


itaibaz:
@kcudnik 
There seems to be meta check issue with SAI_PORT_INTERFACE_TYPE_MAX changing it's enum value. We think <enum>_MAX should be in exception list, similar to <enum>_END
Can you please look

kcudnik:
> @kcudnik There seems to be meta check issue with SAI_PORT_INTERFACE_TYPE_MAX changing it's enum value. We think _MAX should be in exception list, similar to _END Can you please look

yes, it should be excluded

kcudnik (approved):

rlhui (approved):

rlhui:
This commit adds 8-lanes port interface types: CR8, KR8, SR8, LR8.

PR #1408, opened 2022-02-07
Support for overlay for ACL match fields

JaiOCP:
This flag indicates if hw can support overlaying of ACL match fields.
IPv6 DIP and IPV4 DIP keys can use either 128+32 bits in HW or only 128 bits.
If is_field_overlay_capability return TRUE, this means that in such cases HW will only use 128 bit wide field


inc/saitypes.h
```diff
@@ -764,6 +764,16 @@ typedef struct _sai_acl_capability_t
...
     sai_s32_list_t action_list;
+
+    /**
+     * @brief Output from get function
+     *
+     * Flag indicating whether overlay of match field is supported or not
+     * For example, IPv6 DIP and IPv4 DIP can be overlay on a 128 bit wide
+     * field
+     */
+    bool is_field_overlay_supported;
```
kcudnik, line 775: older version of SAI may not populate this field, and access to this field can cause to read invalid memory or some garbage

JaiOCP: > older version of SAI may not populate this field, and access to this field can cause to read invalid memory or some garbage

You mean older version of SAI adapter using new headers, right?
If this is the case, older version of SAI adapter doesn't know about this field and should not be accessing it.

kcudnik: i had in mind scenario, that if sonic is compiled with this  feature, but sai lib would not have this feature (older sai) but this scenario should not happen, since headers should always match on SAI and sonic

PR #1414, opened 2022-02-17
Add support for building under Doxygen 1.9.1 / Debian 11

alexrallen:
As of Doxygen 1.9 there is a new standard for defining newlines in Doxyfiles (`^^`) that is not backwards compatible with the old method (`\n`)

As such I modified the Doxyfile in meta to meet this new standard and also made a copy of the old one. I then added logic to meta/Makefile to detect the version of Doxygen running on the build environment and choose the appropriate Doxyfile thusly.

At some point we will want to drop Doxygen 1.8 support but that can be done at a later time now allowing any environments which build SAI to upgrade to Debian 11(or backport Doxygen 1.9) before that point.

### DIFF
--- a/meta/Makefile
+++ b/meta/Makefile
@@ -95,8 +95,15 @@ toolsversions:
 
 CONSTHEADERS = saimetadatatypes.h saimetadatalogger.h saimetadatautils.h saiserialize.h
 
-xml: $(DEPS) Doxyfile $(CONSTHEADERS)
-	doxygen Doxyfile 2>&1 | perl -npe '$$e=1 if /warning/i; END{exit $$e}'
+DOXYGEN_VERSION_CHECK = $(shell printf "$$(doxygen -v)\n1.8.16" | sort -V | head -n1)
+ifeq (${DOXYGEN_VERSION_CHECK},1.8.16)
+	DOXYFILE = Doxyfile
+else
+	DOXYFILE = Doxyfile.compat
+endif
+
+xml: $(DEPS) Doxyfile Doxyfile.compat $(CONSTHEADERS)
+	doxygen ${DOXYFILE} 2>&1 | perl -npe '$$e=1 if /warning/i; END{exit $$e}'
 	touch xml
 
 EXTRA = acronyms.txt aspell.en.pws *.pm *.cap
--- a/meta/Doxyfile.compat
+++ b/meta/Doxyfile.compat
@@ -218,23 +218,23 @@ TAB_SIZE               = 4
 # "Side Effects:". You can put \n's in the value part of an alias to insert
 # newlines.
 
-ALIASES                += "type           =@par Value Type:^^             @xmlonly @@type           @endxmlonly"
-ALIASES                += "flags          =@par Flags:^^                  @xmlonly @@flags          @endxmlonly"
-ALIASES                += "objects        =@par Allowed object types^^    @xmlonly @@objects        @endxmlonly"
-ALIASES                += "allownull      =@par Allows NULL object ID^^   @xmlonly @@allownull      @endxmlonly"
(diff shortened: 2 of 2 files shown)


dgsudharsan:
@kcudnik Can you please review?

meta/Makefile
```diff
@@ -95,8 +95,15 @@ toolsversions:
 
 CONSTHEADERS = saimetadatatypes.h saimetadatalogger.h saimetadatautils.h saiserialize.h
 
+DOXYGEN_VERSION_CHECK = $(shell echo -e "$$(doxygen -v)\n1.9.0" | sort -V | head -n1)
+ifeq (${DOXYGEN_VERSION_CHECK},1.9.0)
```
kcudnik, line 99: this is only checking one version, and will fall back if doxygen changes to let say 1.9.1 and error will still be shown, and actually new lines "^^" are introduced from doxygen 1.8.16

meta/Makefile
```diff
@@ -95,8 +95,15 @@ toolsversions:
...
+	DOXYFILE = Doxyfile
+else
+	DOXYFILE = Doxyfile.compat
+endif
+
 xml: $(DEPS) Doxyfile $(CONSTHEADERS)
```
kcudnik, line 105: ```xml: $(DEPS) Doxyfile Doxyfile.compat $(CONSTHEADERS)```

meta/Makefile
kcudnik: i yesterday thought about this how make this more generic compare, but i didn't found easy and compact way to compare version strings, the closes way i found was perl version:
```
#!/usr/bin/perl
use version;
print version->parse(1.8.15) < version->parse(1.9.0);
```

kcudnik: this could be actually moved to separate perl file, and that file could be used to output result for Makefile ifeq check

alexrallen: Hi Kamil `sort -V` compares versions this should be extendable.

kcudnik: > Hi Kamil `sort -V` compares versions this should be extendable.

yes, it compares, but you still compare EQ to 1.9.0, so this will fail for 1.9.1, so operator "<" should be used, thats why i proposed perl version compare which does exactly that

alexrallen: It is using `sort` against version `1.9.0` and the version of Doxygen. This command will print the lower version first and then the higher version. The code checks that the lower version is `1.9.0` which means that the version of Doxygen is `1.9.0` or higher which is what we are intending to test. 

I have tested this change already in 1.9.1 and it was successful.

kcudnik: this still does not solve issue for version v1.8.16 which should use Doxygen file, and v1.8.15 should use Doxygen.compat, it will only be working from version 1.9.0 and above, you would need to use 1.8.16 string instead 1.9.0

kcudnik:
ping, can we make this change faster?

alexrallen:
@kcudnik I have been working on other projects.

I have resolved your concerns please take another look.

kcudnik:
build is still failing, seems like it's choosing wrong file:
```
/__w/1/s/inc/saimymac.h:47: warning: Illegal command par as part of a title section
/__w/1/s/inc/saimymac.h:48: warning: Illegal command par as part of a title section
/__w/1/s/inc/saimymac.h:56: warning: Illegal command par as part of a title section
```

alexrallen:
@kcudnik Build is passing now, had a bit of behavior difference between `echo` on my system and the AZP. 

I think all your checkers / feedback should be satisfied now.

kcudnik (approved):

kcudnik:
can you port this change to branch v1.9 ?

kcudnik:
ping

PR #1418, opened 2022-02-24
[Fix]Correct enum check on branch 1.9

richardyu-ms:
When checking the enum we should base on the current branch(1.9)
Test done:
local build

### DIFF
--- a/meta/checkenumlock.sh
+++ b/meta/checkenumlock.sh
@@ -29,9 +29,9 @@ rm -rf temp
 
 mkdir temp
 
-git --work-tree=temp/ checkout origin/master inc
+git --work-tree=temp/ checkout origin/v1.9 inc
 
-echo "Checking for possible enum values shift (current branch vs origin/master) ..."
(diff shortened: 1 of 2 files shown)


meta/checkenumlock.sh
```diff
@@ -29,9 +29,9 @@ rm -rf temp
 
 mkdir temp
 
-git --work-tree=temp/ checkout origin/master inc
+git --work-tree=temp/ checkout origin/v1.9 inc
 
-echo "Checking for possible enum values shift (current branch vs origin/master) ..."
+echo "Checking for possible enum values shift (current branch vs origin/v1.9) ..."
```
kcudnik, line 34: this should still be master here, since enums on branch v1.9 (current) still must have binary values exact the same as on master branch, this is correct, if you want this PR, then you can just add explanation comment to that check enum lock

kcudnik (changes requested):

kcudnik (approved):

PR #1407, opened 2022-02-04
Add label attribute for counter

msomasundaran:
Counter does not have any mandatory attribute to identify object uniquely. Adding a label attribute that can be used to uniquely identify object during warmboot scenarios. This is similar to existing label attribute used by LAG and virtual router.
The attribute is considered as user data attached to the object.

Related to:
#1158 

Signed-off-by: Midhun Somasundaran <msomasundaran@fb.com>

PR #1430, opened 2022-03-06
Add label attribute for ACL counter

shri-khare:
Summary:

This is similar to #1158 and #1407.

Those PRs added label attribute for LAG/virtual router and counter. This
PR adds similar label attribute for ACL counter.

ACL counter does not have any mandatory attribute to identify object
uniquely. Adding a label attribute that can be used to uniquely identify
ACL counter object during warmboot. The attribute is considered as user
data attached to the object.

Signed-off-by: Shrikrishna Khare <skhare@fb.com>

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -3017,6 +3017,15 @@ typedef enum _sai_acl_counter_attr_t
     SAI_ACL_COUNTER_ATTR_BYTES,
 
     /**
+     * @brief Attribute used to uniquely identify ACL counter.
+     *
+     * @type char
+     * @flags CREATE_AND_SET
+     * @default ""
+     */
+    SAI_ACL_COUNTER_ATTR_LABEL,
+
+    /**
      * @brief End of attributes
      */
     SAI_ACL_COUNTER_ATTR_END,


kcudnik (approved):

shri-khare:
Thanks @kcudnik for the quick review.

I notice that 1.10 was tagged only yesterday morning #1429.
Is there a possibility that this PR could be included in the 1.10 revision? 

Otherwise,  we would have to wait till the next SAI Spec revision (what is the timeline for that?) before we could configure multiple ACL counters in the same table and warmboot. cc @rlhui 

Thank you!
Shri

shri-khare:
@kcudnik , @rlhui,  following up my question.

This is a relatively straightforward PR that will help during warmboot with ACL counters. It just missed 1.10 revision. Would it be possible to include it in 1.10.2 (say), so we could configure multiple ACL counters in the same table without needing to wait for the next SAI spec revision (which might be some time away?)

Thanks,
Shri

> Thanks @kcudnik for the quick review.
> 
> I notice that 1.10 was tagged only yesterday morning #1429. Is there a possibility that this PR could be included in the 1.10 revision?
> 
> Otherwise, we would have to wait till the next SAI Spec revision (what is the timeline for that?) before we could configure multiple ACL counters in the same table and warmboot. cc @rlhui
> 
> Thank you! Shri

rlhui:
@JaiOCP , would you like to review/approve this? thanks.

JaiOCP (approved):
Looks Good

JaiOCP:
> @JaiOCP , would you like to review/approve this? thanks.

Reviewed. Thanks

PR #1442, opened 2022-03-17
Ecmp member 

JaiOCP:
Two new Switch attributes are introduced.
Read Only:
SAI_SWITCH_ATTR_MAX_ECMP_MEMBER_COUNT
This attribute is queried during switch init to find out device specific max number of ecmp members supported.

CREATE_AND_SET (read/write):
SAI_SWITCH_ATTR_ECMP_MEMBER_COUNT
This attribute is set based on the query for MAX_ECMP_MEMBER_COUNT and can be changed dynamically.
If the SAI adapter doesn't support dynamic change of this attribute based on certain conditions like if ECMP groups are already configured then MUST return error.

Typical Workflow:

Switch object create
Switch get SAI_SWITCH_ATTR_MAX_ECMP_MEMBER_COUNT: say SAI adapter returns 2k
Switch set SAI_SWITCH_ATTR_ECMP_MEMBER_COUNT
If step 3 is invoked after the system is fully configured and forwarding traffic
SAI adapter MAY return error if HW is not capable of dynamically adjusting the ECMP group size

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -2696,6 +2696,23 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_IPSEC_SA_STATUS_CHANGE_NOTIFY,
 
     /**
+     * @brief Number of ECMP members supported by My MAC entries installed on the switch
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_MAX_ECMP_MEMBER_COUNT,
+
+    /**
+     * @brief Number of  ECMP Members configured on the switch
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 64
+     */
+    SAI_SWITCH_ATTR_ECMP_MEMBER_COUNT,
+
+    /**
      * @brief End of attributes
      */
     SAI_SWITCH_ATTR_END,
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -764,6 +764,13 @@ typedef struct _sai_acl_capability_t
      * @passparam &sai_metadata_enum_sai_acl_action_type_t
      */
     sai_s32_list_t action_list;
+
+    /**
+     * @brief Output from get function
+     * Flag indicating whether overlay of match field is supported or not
+     * For example, IPv6 DIP and IPv4 DIP can be overlayed on a 128 bit wide field
+     */
+     bool is_field_overlay_supported;
 } sai_acl_capability_t;
 
 /**
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -765,12 +765,6 @@ typedef struct _sai_acl_capability_t
      */
     sai_s32_list_t action_list;
 
-    /**
-     * @brief Output from get function
-     * Flag indicating whether overlay of match field is supported or not
-     * For example, IPv6 DIP and IPv4 DIP can be overlayed on a 128 bit wide field
-     */
-     bool is_field_overlay_supported;
 } sai_acl_capability_t;
 
 /**


inc/saiswitch.h
```diff
@@ -2695,6 +2695,23 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_IPSEC_SA_STATUS_CHANGE_NOTIFY,
 
+    /**
+     * @brief Number of ECMP members supported by My MAC entries installed on the switch
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_MAX_ECMP_MEMBER_COUNT,
```
itaibaz, line 2704: what's the different vs existing attribute 

    /**
     * @brief ECMP number of members per group
     *
     * Default is 64
     *
     * @type sai_uint32_t
     * @flags READ_ONLY
     */
    SAI_SWITCH_ATTR_ECMP_MEMBERS,

inc/saiswitch.h
```diff
@@ -2695,6 +2695,23 @@ typedef enum _sai_switch_attr_t
...
+
+    /**
+     * @brief Number of  ECMP Members configured on the switch
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 64
+     */
+    SAI_SWITCH_ATTR_ECMP_MEMBER_COUNT,
+
```
rck-innovium, line 2714: The meaning of this is not clear.  

1) Is this applicable for all types of sai_next_hop_group_type_t?

2) Can this attribute be made a per Next Hop Group attribute instead of a switch attribute?

3) Can you also see if we can relax SAI_NEXT_HOP_GROUP_ATTR_CONFIGURED_SIZE to be applicable for other next_hop_group types?

rck-innovium (changes requested):

inc/saiswitch.h
```diff
@@ -2695,6 +2695,23 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_IPSEC_SA_STATUS_CHANGE_NOTIFY,
 
+    /**
+     * @brief Number of ECMP members supported by My MAC entries installed on the switch
```
lguohan, line 2699: why is this only My MAC entries?

inc/saiswitch.h
```diff
@@ -2695,6 +2695,23 @@ typedef enum _sai_switch_attr_t
...
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_MAX_ECMP_MEMBER_COUNT,
+
+    /**
+     * @brief Number of  ECMP Members configured on the switch
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 64
```
lguohan, line 2711: suggest value to be internal and then have comment to say if KV MAX_ECMP_MEMBER exists, then use that value, otherwise 64?

rck-innovium:
@JaiOCP 

1) Can you please clarify what is the expected hardware behavior when SAI_SWITCH_ATTR_ECMP_MEMBER_COUNT is set to a value < SAI_SWITCH_ATTR_MAX_ECMP_MEMBER_COUNT

We should define the expected behavior for the various nexthop group types: Unordered ECMP, Ordered ECMP, Fine Grained ECMP etc..

2) How does this work for weighted ECMP? Is this the max number of nexthops or is it the max of the sum of the weights of the member nexthops (SAI_NEXT_HOP_GROUP_MEMBER_ATTR_WEIGHT)?

inc/saiswitch.h
JaiOCP: Existing attribute SAI_SWITCH_ATTR_ECMP_MEMBERS is used in conjunction with the boot time SAI_KV_NUM_ECMP_GROUPS and is read only. We can not configure dynamically the ECMP_MEMBER size based on HW capability

inc/saiswitch.h
JaiOCP: typo will fix it.

inc/saiswitch.h
JaiOCP: Changed to  @default internal

JaiOCP:
> @JaiOCP
> 
> 1. Can you please clarify what is the expected hardware behavior when SAI_SWITCH_ATTR_ECMP_MEMBER_COUNT is set to a value < SAI_SWITCH_ATTR_MAX_ECMP_MEMBER_COUNT
> 
> We should define the expected behavior for the various nexthop group types: Unordered ECMP, Ordered ECMP, Fine Grained ECMP etc..
> 
> 2. How does this work for weighted ECMP? Is this the max number of nexthops or is it the max of the sum of the weights of the member nexthops (SAI_NEXT_HOP_GROUP_MEMBER_ATTR_WEIGHT)?

I will add these details in PR Description so that it is captured in the release notes.

inc/saiswitch.h
JaiOCP: 1. This is a switch level attribute that helps carve out memory for max ecmp members and ecmp groups. Nothing to do with next group configuration.
2. No, there is different attribute to control nexthop count in nexthop group : SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_COUNT
3. I believe we can but that is beyond the scope of this PR

rck-innovium:
Thanks, Jai. It would help if you can add these as comments in the header files rather than the PR description.

inc/saiswitch.h
lguohan: how is the ECMP_MEMBER_COUNT related to https://github.com/opencomputeproject/SAI/blob/master/inc/sainexthopgroup.h#L160 ?

PR #1445, opened 2022-03-20
[SAI-PTF]Add platform helper for saiswitch test and refactor saiswitch

richardyu-ms:
Make the sai test on different platforms can use their own helpers for
making some platform related settings.

- Register the helper in the common_sai_helper
- Override the helper in the platfrom helper as needed
- Use the method from the helper for platform related settings
- Platform choice the helper and its helper method from the
corresponding platform
- refactor the test from method to class

### DIFF
--- a/ptf/platform_helper/brcm_helper/brcm_saiswitch_helper.py
+++ b/ptf/platform_helper/brcm_helper/brcm_saiswitch_helper.py
@@ -1,19 +1,38 @@
-THIS_DIR = os.path.dirname(os.path.abspath(__file__))
-sys.path.append(os.path.join(THIS_DIR, '..'))
-from common_helper import saiswitch_helper
+# Copyright (c) 2021 Microsoft Open Technologies, Inc.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+#    THIS CODE IS PROVIDED ON AN *AS IS* BASIS, WITHOUT WARRANTIES OR
+#    CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT
+#    LIMITATION ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS
+#    FOR A PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
+#
+#    See the Apache Version 2.0 License for specific language governing
+#    permissions and limitations under the License.
+#
+#    Microsoft would like to thank the following companies for their review and
+#    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
+#    Dell Products, L.P., Facebook, Inc., Marvell International Ltd.
+
+
+from platform_helper.common_helper.saiswitch_helper import *
 from sai_base_test import * # pylint: disable=wildcard-import; lgtm[py/polluting-import]
-from saiswitch import * # pylint: disable=wildcard-import; lgtm[py/polluting-import]
 
-class BrcmSaiSwitchHelper(saiswitch_helper):
-    def __init__(self):
-        pass
+"""
+Brcm sai switch helper for sai switch tests.
+"""
 
+class BrcmSaiSwitchHelper(SaiSwitchHelper):
     
-    def create_route_entry_from_default_vrf(self):
+    def create_route_entry_from_default_vrf(self, client):
+        print("BrcmSaiSwitchHelper::create_route_entry_from_default_vrf.")
         route_entry = sai_thrift_route_entry_t(
-            switch_id=self.switch_id,
-            vr_id = self.default_vrf,
+            switch_id=client.switch_id,
+            vr_id = client.default_vrf,
             destination=sai_ipprefix('0.0.0.0/0'))
         status = sai_thrift_create_route_entry(
-            self.client, route_entry, next_hop_id=nhop)
-        self.assertEqual(status, SAI_STATUS_SUCCESS)
+            client.client, route_entry, next_hop_id=client.nhop)
+        client.assertEqual(status, SAI_STATUS_SUCCESS)
+        client.route_number += 1
--- a/ptf/platform_helper/brcm_helper/brcm_saiswitch_helper.py
+++ b/ptf/platform_helper/brcm_helper/brcm_saiswitch_helper.py
@@ -1,0 +1,19 @@
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.join(THIS_DIR, '..'))
+from common_helper import saiswitch_helper
+from sai_base_test import * # pylint: disable=wildcard-import; lgtm[py/polluting-import]
+from saiswitch import * # pylint: disable=wildcard-import; lgtm[py/polluting-import]
+
+class BrcmSaiSwitchHelper(saiswitch_helper):
+    def __init__(self):
+        pass
+
+    
+    def create_route_entry_from_default_vrf(self):
+        route_entry = sai_thrift_route_entry_t(
+            switch_id=self.switch_id,
+            vr_id = self.default_vrf,
+            destination=sai_ipprefix('0.0.0.0/0'))
+        status = sai_thrift_create_route_entry(
+            self.client, route_entry, next_hop_id=nhop)
+        self.assertEqual(status, SAI_STATUS_SUCCESS)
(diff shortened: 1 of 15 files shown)


ptf/saiswitch.py
```diff
@@ -224,29 +254,24 @@ def availableIPv6RouteEntryTest(self):
...
                 else:
                     self.fail("Route creation failed after creating %d "
                               "entries, status %u" % (route_number, status))
 
                 attr = sai_thrift_get_switch_attribute(
-                    self.client, available_ipv6_route_entry=True)
-                self.assertEqual(attr["available_ipv6_route_entry"],
+                    self.client, available_ipv4_route_entry=True)
+                self.assertEqual(attr["available_ipv4_route_entry"],
                                  max_route_entry - route_number)
```
RyoYang, line 271: This test case may lack a logic that if the host table is not full until the loop ends, we may assign the value to`self.available_v4_host_routes` which is used in test case `self.availableIPv4NeighborEntryTest()`
```python
self.available_v4_host_routes = route_number
```

richardyu-ms: test cases method will be split into smaller classes, and those variables will be assigned in the new test class.

ptf/saiswitch.py
```diff
@@ -100,7 +287,6 @@ def setUp(self):
 
     def runTest(self):
         self.availableIPv4RouteEntryTest()
```
ravi861, line 289: Can we remove this too since you have duplicated it above?

ptf/warm_boot/warm_saiswitch.py
```diff
@@ -0,0 +1,73 @@
...
+#    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
+#    Dell Products, L.P., Facebook, Inc., Marvell International Ltd.
+
+from warm_test_utils import *  # pylint: disable=wildcard-import; lgtm[py/polluting-import]
+
+"""
+Thrift SAI interface Switch tests
+"""
+
+class WarmAvailableIPv4RouteEntryTest(AvailableIPv4RouteEntryTest):
```
ravi861, line 25: So this is the template to be used for reusing any tests for warm reboot scenarios?

ptf/warm_boot/warm_saiswitch.py
```diff
@@ -0,0 +1,73 @@
...
+        AvailableIPv4RouteEntryTest.setUp(self)
+
+
+    def setUp_starting(self):
+        print("Skip setUp_starting WarmAvailableIPv4RouteEntryTest")
+
+
+    def setUp_post_start(self):
+        print("setUp_post_start WarmAvailableIPv4RouteEntryTest")
+        SaiHelperBase.setUp(self)
```
ravi861, line 41: why is this setUp call required here?

ptf/warm_boot/warm_saiswitch.py
richardyu-ms: Thanks for your review.
Yes, i plan to reuse existing test cases and make all of them cover the warmboot 
There is a design doc for the structure for warmboot testing
https://github.com/opencomputeproject/SAI/pull/1449

ptf/warm_boot/warm_saiswitch.py
richardyu-ms: We have three mode, pre-start, starting, post-start, 
in starting, we will not use any SAI_PTF apis to handle the swtich, as the sai sdk doesn't start.
in post start, we expect we can call some sai apis.

Then, after the reboot, we also need to mapping the data with the test framework, including remap the bridge port id.
then take bridge for example;
- for  the devices that need to recreated we will make more operation in this step, as those data should be consistent with the restored data from warmboot, (will make it when start landing those kinds of device for testing)
- for other platform don't need to recreate the bridge, we can use those setup to remap them again.

then in the port_start test, we can use those data to handle and check switch status with SAI-PTF apis

ptf/saiswitch.py
richardyu-ms: removed.

then, what's your opinion about smaller those cases or use the PTF structure to call them in methods but not class level? or, we can use both of them in different scenarios?

there is the code for refactor in PTF the https://github.com/p4lang/ptf/pull/168

ravi861: I think both in different scenarios is a good way forward. There are some tests which have multiple methods but do very little in each method. Splitting them into multiple classes is not worth our time.

ptf/warm_boot/warm_saiswitch.py
ravi861: Got it. Thanks for the explanation. Is it possible to do this automatically in the base class instead of having every warm boot class do this?

richardyu-ms: Yes, this could be and it is a good idea.

But, I have some concerns about the refactor from the fundamental layer, so choose to wrap some methods in the test class and introduce few modifications to the original test class. Then, just re-orgainze the method in a new place.

Apart from that, this is also a proposal for reusing the existing test cases, I will try to make some experiments with this new structure and see if there any other improvements could be.

thanks.

ptf/platform_helper/brcm_helper/brcm_saiswitch_helper.py
```diff
@@ -0,0 +1,38 @@
...
+    def create_route_entry_from_default_vrf(self, client):
+        print("BrcmSaiSwitchHelper::create_route_entry_from_default_vrf.")
+        route_entry = sai_thrift_route_entry_t(
+            switch_id=client.switch_id,
+            vr_id = client.default_vrf,
+            destination=sai_ipprefix('0.0.0.0/0'))
+        status = sai_thrift_create_route_entry(
+            client.client, route_entry, next_hop_id=client.nhop)
+        client.assertEqual(status, SAI_STATUS_SUCCESS)
+        client.route_number += 1
```
aljer, line 38: Where is this variable defined?

ptf/saiswitch.py
```diff
@@ -125,64 +104,87 @@ def availableIPv4RouteEntryTest(self):
...
 
-        finally:
-            for ip_p_m in routes:
-                sai_thrift_remove_route_entry(self.client, routes.get(ip_p_m))
-            sai_thrift_remove_next_hop(self.client, nhop)
+    def check_amount(self):
+        attr = sai_thrift_get_switch_attribute(
+            self.client, available_ipv4_route_entry=True)
+        self.assertEqual(attr["available_ipv4_route_entry"], 0)
+
```
aljer, line 170: If we want to be compliant also with pycodestyle standards there should be only one blank line between methods.

ptf/saiswitch.py
```diff
@@ -125,64 +104,87 @@ def availableIPv4RouteEntryTest(self):
...
+
+    def runTest(self):
```
aljer, line 172: minor: In all our test files we tried to keep test classes methods in the same order: setUp, runTest, tearDown, and then the rest of test methods.

ptf/saiswitch.py
```diff
@@ -17,12 +17,12 @@
 """
 from ipaddress import ip_address
 
-from sai_thrift.sai_headers import *
+from sai_thrift.sai_headers import *  # pylint: disable=wildcard-import; lgtm[py/polluting-import]
```
aljer, line 20: If you use pylint, I guess there are some continued indentation issues and some lines are too long

ptf/saiswitch.py
```diff
@@ -125,64 +104,87 @@ def availableIPv4RouteEntryTest(self):
...
-                self.assertEqual(attr["available_ipv4_route_entry"],
-                                 max_route_entry - route_number)
 
-            print("%s LPM routes have been created"
-                  % (max_route_entry - max_host_route))
-            self.assertEqual(attr["available_ipv4_route_entry"], 0)
+    def tearDown(self):
+        for ip_p_m in self.routes:
+            sai_thrift_remove_route_entry(self.client, self.routes.get(ip_p_m))
+        sai_thrift_remove_next_hop(self.client, self.nhop)
```
aljer, line 163: In tearDown() method we should just clean up what was configured in setUp(). I would suggest do to either: use try-finally clause in check_data_creation() or create self.routes dictionary and self.nhop object in setUp().

ptf/saiswitch.py
```diff
@@ -83,38 +83,17 @@ def generate_mac_list(no_of_addr):
...
-        self.availableNexthopGroupMemberEntryTest()
-        self.availableFdbEntryTest()
-        self.availableAclTableTest()
-        self.readOnlyAttributesTest()
-        self.refreshIntervalTest()
-        self.availableSnatEntryTest()
-        self.availableDnatEntryTest()
 
-    def availableIPv4RouteEntryTest(self):
+    def check_data_creation(self):
```
aljer, line 96: If this is a subtests method the name should be in camelCase style. 
same for check_amount

ptf/saiswitch.py
```diff
@@ -125,64 +104,87 @@ def availableIPv4RouteEntryTest(self):
...
+            self.client, available_ipv4_route_entry=True)
+        self.assertEqual(attr["available_ipv4_route_entry"], 0)
+
+
+    def runTest(self):
+        self.check_data_creation()
+        self.check_amount()
+
+
+class AvailableIPv6RouteEntryTest(PlatformSaiHelper):
```
aljer, line 177: All above comments apply also to this class.

RyoYang (approved):

PR #1458, opened 2022-03-24
[saimacsec]: Add comments for the byte order of SCI and SSCI

Pterosaur:
Signed-off-by: Ze Gan <ganze718@gmail.com>

### DIFF
--- a/inc/saimacsec.h
+++ b/inc/saimacsec.h
@@ -612,6 +612,8 @@ typedef enum _sai_macsec_sc_attr_t
 
     /**
      * @brief SCI value for this Secure Channel, carried in MACsec packet SecTAG.
+     * The MAC address of SCI is stored in the lowest memory address,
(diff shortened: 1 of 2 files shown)


kcudnik (approved):

inc/saimacsec.h
```diff
@@ -825,6 +827,7 @@ typedef enum _sai_macsec_sa_attr_t
 
     /**
      * @brief SSCI value for this Secure Association
+     * Host Byte Order.
```
SarathBug, line 830: Hi  @Pterosaur 
Should this be in Network byte order?

Pterosaur: The current implementation we used is host byte order. Meanwhile,  some vendors had implemented their SAI according to Host Byte Order.
In my opinion, SSCI will be encoded into frame, So, I think to use Host Byte Order looks reasonable too.

SarathBug (approved):

PR #1463, opened 2022-04-01
[SAI-PTF]Align the sai rpc front name with swss and syncd

richardyu-ms:
In order to make the comparison with SONiC default config environment, we need to init the saiserver within a complete sonic environment.
In order to do that, we need to
- Combine new saiserver with syncd
- saiserver can be initalized by syncd
- SAI-PTF can operate on the saiserver instance from syncd

In order to make the new saiserver integrate with syncd use the existing variable gSwitchId.

Test done:
Test the syncd-rpc container within DUT.

### DIFF
--- a/meta/gensairpc.pl
+++ b/meta/gensairpc.pl
@@ -281,7 +281,7 @@ sub generate_server_template_from_skeleton {
 
                 # Define global variable before "class"
                 print {$server_template}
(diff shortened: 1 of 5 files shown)


rlhui:
@ravi861 , would you review this? Thanks.

PR #1452, opened 2022-03-23
[v1.7] Tunnel Scoped TC MAP and Remarking

bingwang-ms:
Signed-off-by: bingwang <bingwang@microsoft.com>

This PR is a cherry-pick of PR https://github.com/opencomputeproject/SAI/pull/1413, and resolve cherry-pick conflicts as well.

This PR add following behavior to tunnel encap and decap

**Tunnel Encap:**

 - `SAI_TUNNEL_ATTR_ENCAP_QOS_TC_AND_COLOR_TO_DSCP_MAP` for remarking the packet based on normalized TC as per port TC MAP binding
 - `SAI_TUNNEL_ATTR_ENCAP_QOS_TC_TO_QUEUE_MAP `for queue selection based on normalized TC as per port TC MAP binding

**Tunnel Decap:**

 - `SAI_TUNNEL_ATTR_DECAP_QOS_DSCP_TO_TC_MAP` new TC MAP can be bind'ed to the tunnel for decap. This MAP overrides port MAP if configured.
 - `SAI_TUNNEL_ATTR_DECAP_QOS_TC_TO_PRIORITY_GROUP_MAP` for mapping the normalized TC to priority group mainly for PFC generation.
DSCP mode takes effect as configured for both cases.

PR #1344, opened 2021-11-22
Align headers from v1.7.1 to v1.9.1

kcudnik:
Align enum values to match those from v1.9.1 so later on
branch v1.7.1 could be used to warm boot to v1.9.1

PR #1468, opened 2022-04-08
[SAI]Fix build in v1.7 and remove unnecessary change

richardyu-ms:
In order to enable the function for Tunnel Scoped TC MAP in #1452, we
need to enable the tunnel related interfaces, attributes and method
tables.
But other changes like changes related to ECN, IPSec, and VxLan UDP source port in PR #1452, should not be needed.

Besides, some other interfaces are needed for compiling the meta in SAI.

Cause a lack of the reference to the switch tunnel object and some unnecessary interface added, there are some errors as below
```
ERROR: invalid enum specified 'sai_tunnel_vxlan_udp_sport_mode_t' on SAI_SWITCH_TUNNEL_ATTR_TUNNEL_VXLAN_UDP_SPORT_MODE
ERROR: invalid enum specified 'sai_tunnel_vxlan_udp_sport_mode_t' on SAI_TUNNEL_ATTR_VXLAN_UDP_SPORT_MODE
ERROR: condition value 'SAI_TUNNEL_TYPE_IPINIP_ESP' in 'SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_ESP' on SAI_TUNNEL_ATTR_SA_INDEX is not present in sai_tunnel_type_t
ERROR: condition value 'SAI_TUNNEL_TYPE_IPINIP_UDP_ESP' in 'SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_IPINIP_UDP_ESP' on SAI_TUNNEL_ATTR_SA_INDEX is not present in sai_tunnel_type_t
ERROR: condition value 'SAI_TUNNEL_TYPE_VXLAN_UDP_ESP' in 'SAI_TUNNEL_ATTR_TYPE == SAI_TUNNEL_TYPE_VXLAN_UDP_ESP' on SAI_TUNNEL_ATTR_SA_INDEX is not present in sai_tunnel_type_t
```
Some of the objects need other related objects and some of them need to be removed.

what I have done:
- Add objects, methods, and attributes in sai headers files to satisfy build dependencies.
- Remove unnecessary changes related to ECN, IPSec, and VxLan UDP source port

Test done:
make -C ./meta/

**TODO: Add the enum check when compare with master meta**


inc/saitunnel.h
```diff
@@ -1149,8 +1171,6 @@ typedef struct _sai_tunnel_api_t
     sai_remove_tunnel_map_entry_fn               remove_tunnel_map_entry;
     sai_set_tunnel_map_entry_attribute_fn        set_tunnel_map_entry_attribute;
     sai_get_tunnel_map_entry_attribute_fn        get_tunnel_map_entry_attribute;
-    sai_bulk_object_get_attribute_fn             get_tunnels_attribute;
```
kcudnik, line 1152: why this was removed ?

richardyu-ms: Corrected it.
I removed it cause i did not get the method definition in saitunnel.h, and thought those method should not be needed.

I got those two types def in the saitype.h.

thanks for pointing it out.

richardyu-ms:
@bingwang-ms please help confirm the interfaces in the modification are necessary and sufficient.

kcudnik (approved):

inc/saitunnel.h
```diff
@@ -323,8 +323,30 @@ typedef enum _sai_tunnel_type_t
 
     SAI_TUNNEL_TYPE_MPLS,
 
+    SAI_TUNNEL_TYPE_IPINIP_ESP,
```
itaibaz, line 326: this tunnel types are relevant to ipsec
what's the use of introducing to 1.7 if we don't bring the full ipsec feature

itaibaz: perhaps it's better to clean up
    SAI_TUNNEL_ATTR_SA_INDEX,
    SAI_TUNNEL_ATTR_IPSEC_SA_PORT_LIST,
that come from https://github.com/opencomputeproject/SAI/commit/231e67e4fb627265f0d583ad3be88271395e247a
as we don't have ipsec in 1.7
then we don't need the new _ESP types

meta/aspell.en.pws
```diff
@@ -148,6 +148,7 @@ pre
 serdes
 idriver
 ipredriver
+IPsec
```
itaibaz, line 151: also not needed if we don't bring ipsec

inc/saitunnel.h
```diff
@@ -323,8 +323,30 @@ typedef enum _sai_tunnel_type_t
...
+    SAI_TUNNEL_TYPE_IPINIP_UDP_ESP,
+
+    SAI_TUNNEL_TYPE_VXLAN_UDP_ESP,
+
 } sai_tunnel_type_t;
 
+/**
+ * @brief Defines VXLAN tunnel UDP source port mode
+ */
+typedef enum _sai_tunnel_vxlan_udp_sport_mode_t
```
itaibaz, line 337: have we confirmed if udp sport is needed at 1.7 ?
as far as i understand, we are addressing the dual tor scenario
for which we need the loopback action, and the qos maps

we don't need, as i understand, the vxlan sport attributes at switch tunnel and tunnel

we don't need, as i understand, 
SAI_SWITCH_TUNNEL_ATTR_TUNNEL_ENCAP_ECN_MODE
SAI_SWITCH_TUNNEL_ATTR_ENCAP_MAPPERS
SAI_SWITCH_TUNNEL_ATTR_TUNNEL_DECAP_ECN_MODE
SAI_SWITCH_TUNNEL_ATTR_DECAP_MAPPERS

these were already committed in previous commit, and should be cleaned imo
backporting should be minimal to what is required, and not to full objects

inc/saitunnel.h
richardyu-ms: thank you!
Clean up the ipsec related interfaces and attributes.

Test with `make -C ./meta/`

meta/aspell.en.pws
richardyu-ms: yes, that's right!
thank you!

inc/saitunnel.h
```diff
@@ -323,8 +323,28 @@ typedef enum _sai_tunnel_type_t
 
     SAI_TUNNEL_TYPE_MPLS,
 
+    SAI_TUNNEL_TYPE_IPINIP_UDP_ESP,
```
itaibaz, line 326: this is also ipsec, not needed

inc/saitunnel.h
```diff
@@ -323,8 +323,28 @@ typedef enum _sai_tunnel_type_t
...
+
+    SAI_TUNNEL_TYPE_VXLAN_UDP_ESP,
```
itaibaz, line 328: this is also ipsec

inc/saitunnel.h
richardyu-ms: check the change in related PR #1205 #1173 and #1206. 
remove the unnecessary attributes.
thanks!

inc/saitunnel.h
richardyu-ms: done

inc/saiswitch.h
```diff
@@ -2860,20 +2794,78 @@ typedef sai_status_t (*sai_clear_switch_stats_fn)(
...
+    sai_create_switch_fn                   create_switch;
+    sai_remove_switch_fn                   remove_switch;
+    sai_set_switch_attribute_fn            set_switch_attribute;
+    sai_get_switch_attribute_fn            get_switch_attribute;
+    sai_get_switch_stats_fn                get_switch_stats;
+    sai_get_switch_stats_ext_fn            get_switch_stats_ext;
+    sai_clear_switch_stats_fn              clear_switch_stats;
+    sai_switch_mdio_read_fn                switch_mdio_read;
+    sai_switch_mdio_write_fn               switch_mdio_write;
+    sai_create_switch_tunnel_fn            create_switch_tunnel;
```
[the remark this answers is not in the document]
richardyu-ms: those APIs needed by switch_tunnel for CRUDs

bingwang-ms (approved):

bingwang-ms:
Thanks for the fix. @richardyu-ms

bingwang-ms (approved):

itaibaz (approved):

inc/saiswitch.h
```diff
@@ -331,6 +331,8 @@ typedef enum _sai_switch_failover_config_mode_t
  * per tunnel type global configuration.
  * SAI_OBJECT_TYPE_TUNNEL object configuration
  * overrides the switch scoped global configuration.
+ *
+ * @flags Contains flags
```
kcudnik, line 335: this should not contain flags, please remove this, and make all attributes continuous

inc/saiswitch.h
```diff
@@ -499,6 +425,8 @@ typedef enum _sai_switch_tunnel_attr_t
 /**
  * @brief Attribute Id in sai_set_switch_attribute() and
  * sai_get_switch_attribute() calls
+ *
+ * @flags Contains flags
```
kcudnik, line 429: please remove as above

inc/saitunnel.h
```diff
@@ -448,6 +448,8 @@ typedef enum _sai_tunnel_peer_mode_t
 
 /**
  * @brief Defines tunnel attributes
+ *
+ * @flags Contains flags
```
kcudnik, line 452: please remove

kcudnik (changes requested):

kcudnik (approved):

JaiOCP:
Looks good to me

JaiOCP:
> @bingwang-ms @JaiOCP please help confirm the interfaces in the modification are necessary and sufficient.

I do not see the review button on this PR. Don't know why.
But changes looks good to me.

inc/saiswitch.h
```diff
@@ -2427,6 +2351,16 @@ typedef enum _sai_switch_attr_t
...
 
+    /**
+     * @brief Switch scoped Tunnel objects
+     *
+     * @type sai_object_list_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_SWITCH_TUNNEL
+     * @default empty
+     */
+    SAI_SWITCH_ATTR_TUNNEL_OBJECTS_LIST,
```
lguohan, line 2362: why do we need this at all?

inc/saiswitch.h
```diff
@@ -2429,12 +2431,24 @@ typedef enum _sai_switch_attr_t
...
     /**
      * @brief Packet action when a packet ingress and gets routed back to same tunnel
+     * This enum is only available in v1.7.
      *
      * @type sai_packet_action_t
      * @flags CREATE_AND_SET
      * @default SAI_PACKET_ACTION_FORWARD
      */
-    SAI_SWITCH_ATTR_TUNNEL_LOOPBACK_PACKET_ACTION,
+    SAI_SWITCH_ATTR_TUNNEL_LOOPBACK_PACKET_ACTION = 186,
```
richardyu-ms, line 2440: this is the attribute that is only available in v1.7 but not in master, then if there is any other attribute with the same value that might be added in master, how can we handle that?

inc/saitunnel.h
```diff
@@ -325,6 +325,22 @@ typedef enum _sai_tunnel_type_t
...
+{
+    /**
+     * @brief User define value
+     */
+    SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_USER_DEFINED,
+
+    /**
+     * @brief RFC6335 Computed hash value in range 49152-65535
+     */
+    SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_EPHEMERAL,
```
richardyu-ms, line 341: this is another way to keep the enum values compatible with the latter version than v1.7, i left them there

inc/saitunnel.h
```diff
@@ -686,44 +704,17 @@ typedef enum _sai_tunnel_attr_t
...
+     * Set to the same value as master branch
      *
      * @type sai_object_id_t
      * @flags CREATE_AND_SET
      * @objects SAI_OBJECT_TYPE_QOS_MAP
      * @allownull true
      * @default SAI_NULL_OBJECT_ID
      */
-    SAI_TUNNEL_ATTR_ENCAP_QOS_TC_AND_COLOR_TO_DSCP_MAP,
+    SAI_TUNNEL_ATTR_ENCAP_QOS_TC_AND_COLOR_TO_DSCP_MAP = 25,
```
richardyu-ms, line 717: cause we removed the attribute with ipsec, then, there is a hole in the enum value, we need to set the value.

inc/saiswitch.h
```diff
@@ -499,6 +499,8 @@ typedef enum _sai_switch_tunnel_attr_t
 /**
  * @brief Attribute Id in sai_set_switch_attribute() and
  * sai_get_switch_attribute() calls
+ *
+ * @flags skip enum value check
```
richardyu-ms, line 503: this is the tag that is used to skip the enum check as the enum value must be increased >=0 and <=1

inc/saiswitch.h
```diff
@@ -2429,12 +2431,24 @@ typedef enum _sai_switch_attr_t
...
+    /**
+     * @brief Switch scoped Tunnel objects
+     * Set to the same value as master branch.
+     *
+     * @type sai_object_list_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_SWITCH_TUNNEL
+     * @default empty
+     */
+    SAI_SWITCH_ATTR_TUNNEL_OBJECTS_LIST = 190,
```
richardyu-ms, line 2451: align with master

inc/saiswitch.h
kcudnik: why this is added? on attributes each attribute should increase +1

inc/saitunnel.h
```diff
@@ -448,6 +464,8 @@ typedef enum _sai_tunnel_peer_mode_t
 
 /**
  * @brief Defines tunnel attributes
+ *
+ * @flags skip enum value check
```
kcudnik, line 468: why this is added ? please remove

meta/test.pm
```diff
@@ -530,6 +530,30 @@ sub CreateStatEnumTest
...
+    WriteTest "    val = SAI_SWITCH_ATTR_TUNNEL_OBJECTS_LIST;";
+    WriteTest "    TEST_ASSERT_TRUE(val == 190, \"enum SAI_SWITCH_ATTR_TUNNEL_OBJECTS_LIST is not with value 190.\");";
+    WriteTest "    val = SAI_TUNNEL_ATTR_ENCAP_QOS_TC_AND_COLOR_TO_DSCP_MAP;";
+    WriteTest "    TEST_ASSERT_TRUE(val == 25, \"enum SAI_TUNNEL_ATTR_ENCAP_QOS_TC_AND_COLOR_TO_DSCP_MAP is not with value 25.\");";
+    WriteTest "    val = SAI_OBJECT_TYPE_SWITCH_TUNNEL;";
+    WriteTest "    TEST_ASSERT_TRUE(val == 95, \"enum SAI_OBJECT_TYPE_SWITCH_TUNNEL is not with value 95.\");";
+
+    WriteTest "}";
+}
+
```
kcudnik, line 556: i would move this to saisanitycheck.c
also, why those enums must have specific values ? why this is sudenny needed here?

kcudnik (changes requested):

meta/test.pm
richardyu-ms: briefly, we try to match the enum value between master and v1.7 (for warm reboot).

examples:
0. SAI_SWITCH_ATTR_SUPPORTED_FAILOVER_MODE
The is the precondition, this value already been there before current change.
in order to specify the situation happening between 1.7 and master, let's check the SAI_SWITCH_ATTR_SUPPORTED_FAILOVER_MODE, between v1.7 and master.
**in v1.7 SAI_SWITCH_ATTR_SUPPORTED_FAILOVER_MODE ==185**
**in master SAI_SWITCH_ATTR_SUPPORTED_FAILOVER_MODE == 189**

then there is already a difference, but in my current, i just want to add more attributes from master to v1.7, but not want to touch the historical items. 

With those difference, i want to add more attributes, but keep them with the same value with master
1. SAI_SWITCH_ATTR_TUNNEL_LOOPBACK_PACKET_ACTION
 this is a v1.7 only attribute, it only available in v1.7, then it must have a value
**in v1.7 WAS SAI_SWITCH_ATTR_TUNNEL_LOOPBACK_PACKET_ACTION = 186**
**in master None of this enum**

2. SAI_SWITCH_ATTR_TUNNEL_OBJECTS_LIST
 this is the attribute i need to add into v1.7, cause it is from master, then i can set the value
**in v1.7 set SAI_SWITCH_ATTR_TUNNEL_OBJECTS_LIST= 190**
**in master SAI_SWITCH_ATTR_TUNNEL_OBJECTS_LIST= 190**

3. SAI_TUNNEL_ATTR_ENCAP_QOS_TC_AND_COLOR_TO_DSCP_MAP
Cause we don't need IPSec related attributes, i need to set it
**in v1.7 set SAI_TUNNEL_ATTR_ENCAP_QOS_TC_AND_COLOR_TO_DSCP_MAP= 25**
or it will be
**in v1.7 set SAI_TUNNEL_ATTR_ENCAP_QOS_TC_AND_COLOR_TO_DSCP_MAP= 23**

cause I set this value for SAI_TUNNEL_ATTR_ENCAP_QOS_TC_AND_COLOR_TO_DSCP_MAP
then i need the flag
**@flags skip enum value check**

then those are values that need to be set and keep them same with master.

Any suggestion are welcomed.

Thanks.

inc/saitunnel.h
richardyu-ms: reason as below

meta/test.pm
kcudnik: i made syncd between v1.7 and v1.9 which was master at a time here: https://github.com/opencomputeproject/SAI/pull/1344, but it was hanging too long and got out of sync before merging, this was also discussed on SAI community meeting, that if this syncd will be made then it will fail to warm boot for some vendors from v1.7 to v1.7 synced since binary values on SAI will differ :(

there is no good way to do this sync, and you adding forced enum values will still hit some issues in the near future since some enums will start overlap and attributes will have different meaning

my suggestion here is to keep v1.7 branch as a temporary, and phase out this branch and move to v1.9 or later and forget about v1.7

richardyu-ms: For the values mismatch between versions, how about if we use the master branch as anchor and check previous version with the master? can we use the saisanitycheck, or saimetadatatest for go through all the enums?

richardyu-ms: we mentioned 1.7 to 1.9, did we make sure 1.5 to 1.7 can make the upgrade warm reboot?

kcudnik: was that v1.5 to v1.7 or v1.4 to v1.7 ? i lost track of warm boot on lower versions, and as for tracking via master it's not possible via sanity check since that is compiled against 1 set of headers, and it can't have multiple headers compiled, i used checkheaders.pl to find miss matched values and used that to manually create PR #1344 and it's a paint to do this manually, but there is no other way, since between those versions there are some new enums introduced, some were removed and changed type, and entire whole api was remove SR6 and was replaced by segmented route, and this is causing issues when comparing those branches

richardyu-ms: > it's a paint to do this manually

Cannot be more agree with that and thanks for the explicit details.

richardyu-ms: removed this validation in EnumValueTest

lguohan (approved):

kcudnik (approved):

PR #1471, opened 2022-04-20
[v1.9] [meta] Disable doxygen TCL_SUBST and COLS_IN_ALPHA_INDEX (#1457)

alexrallen:
Those features are no longer present on new doxygen versions and are not used by metadata parser

Backports #1457


kcudnik (approved):
thanks for the fix

PR #1482, opened 2022-05-24
Location Specific Policer/Trap Binding

JaiOCP:
Existing SAI spec supports binding of a trap to a policer without the need to know if the trap is in ingress or egress state.
This PR enhances binding of policer by specifying if trap is an ingress or egress policer and then binding the policer from ingress policer pool or egress policer pool.
Workflow:
1. SAI_HOSTIF_TRAP_ATTR_LOCATION Read only attribute specifies if trap belongs to ingress or egress stage of the pipeline
2. SAI_POLICER_ATTR_POLICER_POOL specifies to use ingress or egress policer pool. Default is common for backward compatibility.


kcudnik (approved):

rck-innovium:
1. This notion of attaching a direction/location could be used for other object types like counters, mirrors etc.
2. So, we should create a new SAI type to convey the location

-  For bind points like hostif-trap-group there would be a new read-only attribute of this type
-  For action objects like like policer, this would be a create-only object (with default value as BOTH to ensure backward compatibility)
 
4. We should create a new SAI error status to denote location mismatch

rck-innovium:
1. This notion of attaching a direction/location could be used for other object types like counters, mirrors etc.
2. So, we should create a new SAI type to convey the location

-  For bind points like hostif-trap-group there would be a new read-only attribute of this type
-  For action objects like like policer, this would be a create-only object (with default value as BOTH to ensure backward compatibility)
 
4. We should create a new SAI error status to denote location mismatch

inc/saipolicer.h
```diff
@@ -200,6 +216,15 @@ typedef enum _sai_policer_attr_t
      */
     SAI_POLICER_ATTR_ENABLE_COUNTER_PACKET_ACTION_LIST = 0x0000000a,
 
+    /**
+     * @brief Policer Pool
+     *
+     * @type sai_policer_pool_t
+     * @flags CREATE_AND_SET
```
rck-innovium, line 223: this should be create-only and not create-and-set.

rck-innovium, line 223: this should be create-only and not create-and-set.

inc/saipolicer.h
```diff
@@ -84,6 +84,22 @@ typedef enum _sai_policer_color_source_t
...
+    /** Common Pool */
+    SAI_POLICER_POOL_COMMON,
+
+    /** Ingress Pool */
+    SAI_POLICER_POOL_INGRESS,
+
+    /** Egress Pool */
+    SAI_POLICER_POOL_EGRESS
+
+} sai_policer_pool_t;
```
rck-innovium, line 101: Move to saitypes.h as

```
typedef enum _sai_object_direction_t
{
    /** Common Pool */
    SAI_OBJECT_DIRECTION_BOTH,

    /** Ingress Pool */
    SAI_OBJECT_DIRECTION_INGRESS,


    /** Egress Pool */
    SAI_OBJECT_DIRECTION_EGRESS,

} sai_object_location_t;


```

rck-innovium, line 101: Move to saitypes.h as

```
typedef enum _sai_object_direction_t
{
    /** Common Pool */
    SAI_OBJECT_DIRECTION_BOTH,

    /** Ingress Pool */
    SAI_OBJECT_DIRECTION_INGRESS,


    /** Egress Pool */
    SAI_OBJECT_DIRECTION_EGRESS,

} sai_object_location_t;


```

inc/saipolicer.h
```diff
@@ -200,6 +216,15 @@ typedef enum _sai_policer_attr_t
      */
     SAI_POLICER_ATTR_ENABLE_COUNTER_PACKET_ACTION_LIST = 0x0000000a,
 
+    /**
+     * @brief Policer Pool
+     *
+     * @type sai_policer_pool_t
```
rck-innovium, line 222: * @type sai_object_location_t

rck-innovium, line 222: * @type sai_object_location_t

inc/saihostif.h
```diff
@@ -581,6 +597,14 @@ typedef enum _sai_hostif_trap_attr_t
      */
     SAI_HOSTIF_TRAP_ATTR_COUNTER_ID,
 
+    /**
+     * @brief Hostif Trap Location
+     *
+     * @type sai_hostif_trap_location_t
```
rck-innovium, line 603: @type sai_object_location_t

rck-innovium, line 603: @type sai_object_location_t

lguohan:
can you give example sai calling flow? also, how do we know if a trap belongs to ingress or egress?

JaiOCP:
> can you give example sai calling flow? also, how do we know if a trap belongs to ingress or egress?

Added a document with workflow

kcudnik:
```
WARNING: Both enums have the same value SAI_HOSTIF_TRAP_ATTR_OBJECT_STAGE and SAI_HOSTIF_TRAP_ATTR_LOCATION = 0x00000007
WARNING: Both enums have the same value SAI_POLICER_ATTR_OBJECT_STAGE and SAI_POLICER_ATTR_POLICER_POOL = 0x0000000b
```

inc/saistatus.h
```diff
@@ -173,6 +173,11 @@
  */
 #define SAI_STATUS_NOT_EXECUTED                     SAI_STATUS_CODE(0x00000017L)
 
+/**
+ * @brief Invalid object location
+ */
+#define SAI_STATUS_STAGE_MISMATCH                   SAI_STATUS_CODE(0x00000018L)
```
kcudnik, line 179: description don't match define name, maybe explain it more here why invalid object location is same as stage mismatch

rck-innovium (approved):

PR #1486, opened 2022-06-02
Add virtual IP entry

marian-pritsak:
Add entry to let overlay pipeline know which
destination IP to accept.

### DIFF
--- a/experimental/saitypesextensions.h
+++ b/experimental/saitypesextensions.h
@@ -48,6 +48,8 @@ typedef enum _sai_object_type_extensions_t
 
     SAI_OBJECT_TYPE_ENI,
 
+    SAI_OBJECT_TYPE_VIP_ENTRY,
+
     SAI_OBJECT_TYPE_DASH_ACL,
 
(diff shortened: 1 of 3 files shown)


experimental/saitypesextensions.h
```diff
@@ -48,6 +48,8 @@ typedef enum _sai_object_type_extensions_t
 
     SAI_OBJECT_TYPE_ENI,
 
+    SAI_OBJECT_TYPE_VIP_ENTRY,
```
kcudnik, line 51: why adding in the middle and not at the end?

marian-pritsak: It was autogenerated
Is it required to keep these enums in order?

prsunny:
@kcudnik , I just noticed one outstanding comment . Are you ok that this got merged?

kcudnik:
> @kcudnik , I just noticed one outstanding comment . Are you ok that this got merged?

should be good enough, but i would prefer this to be in order, since currently extensions are not checked for that, but if we start using them in warm boot then there will be an issue with this

PR #1493, opened 2022-06-13
Add mdio access type (clause 45/clause 22) attribute

jiahua-wang:
Signed-off-by: Jiahua Wang <jiahua.wang@broadcom.com>

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -2777,6 +2777,20 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_ECMP_MEMBER_COUNT,
 
     /**
+     * @brief Enable/disable Clause 22 MDIO access
+     *
+     * set mdio access type (clause 45 or clause 22) before calling sai_switch_mdio_read_fn and sai_switch_mdio_write_fn
+     *
+     * FALSE - sai_switch_mdio_read_fn and sai_switch_mdio_write_fn use clasue 45 mdio access.
+     * TRUE - sai_switch_mdio_read_fn and sai_switch_mdio_write_fn use clasue 22 mdio access.
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_SWITCH_ATTR_MDIO_CLAUSE_22,
+
+    /**
      * @brief End of attributes
      */
     SAI_SWITCH_ATTR_END,


inc/saiswitch.h
```diff
@@ -2776,6 +2776,20 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_ECMP_MEMBER_COUNT,
 
+    /**
+     * @brief Enable/disable Clause 22 MDIO access
+     *
+     * set mdio access type (clause 45 or clause 22) before calling sai_switch_mdio_read_fn and sai_switch_mdio_write_fn
+     *
+     * FALSE - sai_switch_mdio_read_fn and sai_switch_mdio_write_fn use clasue 45 mdio access.
+     * TRUE - sai_switch_mdio_read_fn and sai_switch_mdio_write_fn use clasue 22 mdio access.
```
kcudnik, line 2785: could there be other clauses more in the future ? maybe this should be enum?, i think enum would be better here, since bool value should be used for something that is enabled or not, and here you have explicit types 22 and 45 to set

jiahua-wang: MDIO was originally defined in Clause 22 of IEEE RFC802.3. Clause 45 of the 802.3ae specification provided the ability to access 65,536 registers in 32 different devices on 32 different ports. 802.3ae is 20 years old. There is no new MDIO access mode in the last 20 years. MDIO frame format starts with 32 bits of logical 1 preamble and 2 bits of start of frame. The start of frame is 01 for clause 22 and 00 for clause 45. If the first bit of the start frame has to be logical 0, then no new MDIO frame format will be possible.

kcudnik: i see, but it still would make sense to make it as enum, like for example you would have ipv4 and ipv6, it's better specify it as enum rather then bool, maybe lets ask community on this

jiahua-wang: Changed to attribute SAI_SWITCH_ATTR_MDIO_ACCESS_TYPE and enum sai_switch_mdio_clause_t

lguohan:
@jiahua-wang , can you check why build is failing?

prgeor:
> @jiahua-wang , can you check why build is failing?

@jiahua-wang failure is due to your change, check the error log in build:-

```
2022-06-14T22:10:40.9552117Z ERROR: enum starts by object type SWITCH but not ending on MDIO_CLAUSE in SWITCH_MDIO_CLAUSE
2022-06-14T22:10:40.9552651Z ERROR: sai_switch_mdio_clause_t == SAI_SWITCH_ATTR_MDIO_ACCESS_TYPE not ending on enum name SWITCH_MDIO_CLAUSE
2022-06-14T22:10:40.9553117Z no need for END enum for SAI_IN_DROP_REASON_CUSTOM_RANGE_BASE
2022-06-14T22:10:40.9553526Z no need for END enum for SAI_OUT_DROP_REASON_CUSTOM_RANGE_BASE
2022-06-14T22:10:40.9553858Z non range base SAI_UDF_ATTR_BASE
2022-06-14T22:10:40.9554193Z make: *** [Makefile:114: saimetadata.h] Error 1
2022-06-14T22:10:40.9598190Z ##[error]Bash exited with code '2'.
2022-06-14T22:10:40.9681525Z ##[section]Finishing: Metadata check
```

inc/saiswitch.h
```diff
@@ -2776,6 +2789,17 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_ECMP_MEMBER_COUNT,
 
+    /**
+     * @brief Switch MDIO access type
+     *
+     * set mdio access type (clause 22 or clause 45) before calling sai_switch_mdio_read_fn and sai_switch_mdio_write_fn
```
prgeor, line 2795: why can't the platform decide to choose between cl22 or cl45 for the mdio r/w? why should a higher SAI layer tell which one to choose? i am not able to understand the usecase. Can you add more to the PR description?

jiahua-wang:
With PR #1507 replacing this PR, I will close this PR.

PR #1498, opened 2022-06-17
Add fabric port isolation attribute

jfeng-arista:
Add fabric port isolation attribute

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -2137,8 +2137,8 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_FEC_ALIGNMENT_LOCK,
 
     /**
-     * @brief Fabric port isolation setting
-     * 
+     * @brief Fabric port isolation setting.
+     *
      * true: The link may be enabled in serdes level and the
      * MAC level, but the link partner will not use
      * it for traffic distribution.
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -2138,6 +2138,12 @@ typedef enum _sai_port_attr_t
 
     /**
      * @brief Fabric port isolation setting
+     * 
+     * true: The link may be enabled in serdes level and the
+     * MAC level, but the link partner will not use
+     * it for traffic distribution.
+     * false: Undo the isolation operation.
+     * This attribute is for fabric links only.
      *
      * @type bool
      * @flags CREATE_AND_SET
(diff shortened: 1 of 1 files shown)


kcudnik:
WARNING: line ends in whitespace saiport.h 1888:      * @brief Fabric port isolation setting

kcudnik:
still some issues with build, please fix

kcudnik (approved):

rlhui (approved):

SarathBug (approved):

inc/saiport.h
```diff
@@ -2104,6 +2104,21 @@ typedef enum _sai_port_attr_t
...
     SAI_PORT_ATTR_SUPPORTED_LINK_TRAINING_MODE,
 
+    /**
+     * @brief Fabric port isolation setting.
+     *
+     * true: The link may be enabled in serdes level and the
+     * MAC level, but the link partner will not use
+     * it for traffic distribution.
+     * false: Undo the isolation operation.
+     * This attribute is for fabric links only.
```
abdosi, line 2114: can we add something like this in the comment section. 
Refer sairouterinterface.h

@condition  SAI_PORT_ATTR_TYPE == SAI_PORT_TYPE_FABRIC

abdosi (changes requested):
please check above comment.

skeesara-nokia (approved):

inc/saiport.h
jfeng-arista: Hi, 

SAI_PORT_ATTR_TYPE is a read only attribute , and as far as I know that @condition not works for read only attribute.  So checking with you again about if we need to make this change ? 

thank you 

Jie

jfeng-arista: I tried add the @condition, and the metadata check failed, 

it looks like based on the following 3 checks, we can not use @condition check for read only attribute. thank you 

https://github.com/opencomputeproject/SAI/blob/f69394f590362b1b02852ec80f00f9a27a02ed97/meta/saisanitycheck.c#L1284
https://github.com/opencomputeproject/SAI/blob/f69394f590362b1b02852ec80f00f9a27a02ed97/meta/saisanitycheck.c#L1455
https://github.com/opencomputeproject/SAI/blob/f69394f590362b1b02852ec80f00f9a27a02ed97/meta/saisanitycheck.c#L1489

JaiOCP (approved):

abdosi (approved):

kcudnik (approved):

rlhui:
@jfeng-arista  - would you please resolve conflicts?

jfeng-arista:
Hi, 
Can any one help me re-trigger the build for my branch and I could not trigger it by my last comment. 
thank you 
Jie

jfeng-arista:
thank you !

PR #1500, opened 2022-06-20
Add DASH HA API

marian-pritsak:


kcudnik (approved):

marian-pritsak:
Yes, that's the purpose of the backup role - to not accept new connections unless explicitly said to. If the desired behavior is to accept connections immediately without waiting for a new peer, two peers may be configured to active role. Then in case of failure new connections will be accepted immediately without any direction from a controller.

PR #1508, opened 2022-06-24
add vlan test case

ms-junyi:
add vlan test case Group 1~5


test/sai_test/sai_vlan_test.py
```diff
@@ -79,7 +81,383 @@ def tearDown(self):
...
+
+    def tearDown(self):
+        """
+        Test the basic tearDown process
+        """
+        sai_thrift_flush_fdb_entries(self.client, entry_type=SAI_FDB_FLUSH_ENTRY_TYPE_ALL)
+
+        super().tearDown()
+
+    def create_fdb_entry(self,mac_address_):
```
richardyu-ms, line 268: reuse the method in fdb_configer?

PR #1509, opened 2022-06-24
add vlan test case

ms-junyi:
Test vlan cases from group1 to foup5 according to [vlan test plan](https://github.com/opencomputeproject/SAI/blob/master/doc/sai-ptf/vlan_test_plan.md)

-  verify the VLAN function around untag and access ports
-  verify the frame filtering
-  verify the VLAN flooding with untagged and tagged pkts
-  verify the VLAN broadcast with untagged and tagged pkts
- verify the VLAN mac learning with untagged and tagged pkts

ALL cases are tested on DUT

### DIFF
--- a/test/sai_test/sai_vlan_test.py
+++ b/test/sai_test/sai_vlan_test.py
@@ -17,6 +17,8 @@
 #    Dell Products, L.P., Facebook, Inc., Marvell International Ltd.
 #
 #
+from time import sleep
+
 
 from sai_test_base import T0TestBase
 from sai_thrift.sai_headers import *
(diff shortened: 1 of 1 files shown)


test/sai_test/sai_vlan_test.py
```diff
@@ -79,7 +81,364 @@ def tearDown(self):
...
+                send_packet(self, self.dev_port_list[1], pkt)
+                verify_no_other_packets(self, timeout=1)
+        finally:
+            pass
+
+    def tearDown(self):
+        """
+        Test the basic tearDown process
+        """
+        sai_thrift_flush_fdb_entries(self.client, entry_type=SAI_FDB_FLUSH_ENTRY_TYPE_ALL)
```
richardyu-ms, line 179: do we really need those methods?

PR #1517, opened 2022-06-30
add read config_db.json

ms-junyi:
read config_db.json to set port attributes

### DIFF
--- a/test/sai_test/config/port_configer.py
+++ b/test/sai_test/config/port_configer.py
@@ -94,6 +94,8 @@ class PortConfiger(object):
         """
         self.test_obj = test_obj
         self.client = test_obj.client
+        config_driver = ConfigDBOpertion()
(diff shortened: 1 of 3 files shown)


test/sai_test/config/port_configer.py
```diff
@@ -381,14 +384,27 @@ def turn_up_and_check_ports(self, port_list):
...
+                        admin_state=True,
+                        fec_mode=self.get_fec_mode())
             if all_ports_are_up:
                 print("Retry {} times turn up port.".format(num_of_tries))
                 break
         if not all_ports_are_up:
             print("Not all the ports are up after {} rounds of retries.".format(retries))
+    
+        
+    def get_fec_mode(self):
```
richardyu-ms, line 397: please add comment for each methods

richardyu-ms (changes requested):
please add description in the PR

PR #1541, opened 2022-07-22
Build saithrift with python3.9 in bullseye

richardyu-ms:
Bullseye upgrade the python to python3.9 and other image, like buster still using python 2.7.

To build with different python in different image, add a condition for copy differernt install file in different type of image.


debian/copy_installer.sh
```diff
@@ -3,6 +3,12 @@ then
    echo "Copy python-saithriftv2.install as python-saithrift.install"
    cp ./debian/installerFiles/python-saithriftv2.install ./debian/python-saithrift.install
 else
-   echo "Copy python-saithriftv1.install as python-saithrift.install"
-   cp ./debian/installerFiles/python-saithriftv1.install ./debian/python-saithrift.install
+   if test -n "lsb_release -a | grep bullseye"
```
xumia, line 6: How about the bookworm, debian 12?
Suggest to check if buster, then do something.

xumia (approved):

PR #1543, opened 2022-07-25
[SAI Test] Add lag test

ms-junyi:
Test LAG cases from group1 to goup4 according to [lag test plan](https://github.com/opencomputeproject/SAI/blob/master/doc/sai-ptf/lag_test_plan.md)
1. verify the  PortChannel Load balancing based on different attributes(source IP, destination IP, source port, destination port, and protocol). 
2. verify traffic drop on the disabled lag member after disabling ingress or egress on a lag member
3. verify traffic to drop/appear on the lag member after removing/adding the lag members
4. verify the ingress ports should not be as a Hash Factor in Lag load balance. When forwarding the packet from different ingress ports, if only the ingress port changed, then the load balance should not happen among lag members.

Test:
Test all lag cases at DUT
Skip the LoadbalanceOnProtocalTest and DisableIngressTest


test/sai_test/sai_lag_test.py
```diff
@@ -121,6 +122,197 @@ def runTest(self):
...
+    """
+    Test load balance of l3 by destinstion port.
+    """
+    def setUp(self):
+        T0TestBase.setUp(self)
+
+    def runTest(self):
+        try:
+            print("Lag l3 load balancing test based on des port")
+            sai_thrift_create_router_interface(self.client, virtual_router_id=self.default_vrf, type=SAI_ROUTER_INTERFACE_TYPE_PORT, port_id=self.port_list[21])
```
richardyu-ms, line 135: this line is too long, please reformat the code

test/sai_test/config/route_configer.py
```diff
@@ -38,10 +38,17 @@ def t0_route_config_helper(test_obj, is_create_route=True, is_create_route_for_l
 
         ip_addr2 = '10.1.2.100'
         mac_addr2 = '02:04:02:01:02:01'
-        route_configer.create_route_and_neighbor_entry_for_port(ip_addr=ip_addr1, 
+        route_configer.create_route_and_neighbor_entry_for_port(ip_addr=ip_addr2, 
             mac_addr=mac_addr2, 
             port_id=test_obj.lag2.lag_id, 
             virtual_router_id=test_obj.default_vrf)
+        
+        port21_addr = '192.168.0.1'
```
richardyu-ms, line 46: for those common config items, we should maintain them in a list that can be operated in the different test step, like testMethod and teardown

PR #1550, opened 2022-07-29
Add VNET direct, direct and drop actions to DASH routing

marian-pritsak:

### DIFF
--- a/experimental/saiexperimentaldashvnet.h
+++ b/experimental/saiexperimentaldashvnet.h
@@ -47,6 +47,21 @@ typedef enum _sai_inbound_routing_entry_action_t
 } sai_inbound_routing_entry_action_t;
 
 /**
+ * @brief Attribute data for #SAI_OUTBOUND_ROUTING_ENTRY_ATTR_ACTION
+ */
+typedef enum _sai_outbound_routing_entry_action_t
+{
+    SAI_OUTBOUND_ROUTING_ENTRY_ACTION_ROUTE_VNET,
+
+    SAI_OUTBOUND_ROUTING_ENTRY_ACTION_ROUTE_VNET_DIRECT,
+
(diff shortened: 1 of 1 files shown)


experimental/saiexperimentaldashvnet.h
```diff
@@ -285,13 +300,33 @@ typedef enum _sai_outbound_routing_entry_attr_t
...
+    /**
+     * @brief Action route_vnet_direct parameter OVERLAY_NH_IP
+     *
+     * @type sai_ip_address_t
+     * @flags CREATE_AND_SET
+     * @default 0.0.0.0
+     * @validonly SAI_OUTBOUND_ROUTING_ENTRY_ATTR_ACTION == SAI_OUTBOUND_ROUTING_ENTRY_ACTION_ROUTE_VNET_DIRECT
      */
-    SAI_OUTBOUND_ROUTING_ENTRY_ATTR_DEST_VNET_VNI = SAI_OUTBOUND_ROUTING_ENTRY_ATTR_START,
+    SAI_OUTBOUND_ROUTING_ENTRY_ATTR_OVERLAY_NH_IP,
```
prsunny, line 329: I think this should be OVERLAY_IP instead of OVERLAY_NH_IP.

kcudnik (approved):

experimental/saiexperimentaldashvnet.h
marian-pritsak: Isn't it the true though? It's a next hop for the overlay IP of the packet.

prsunny: mostly yes, but IMO, this is an IP used for a specific lookup and lets just have it without calling it NH.

prsunny (approved):

PR #1551, opened 2022-08-01
SAI Generic Programmable Extensions proposal

ravi861:
Signed-off-by: Ravi Vantipalli ravi.vantipalli@intel.com

This document defines the technical specification for the API to support SAI Generic Extensions. The accompanying code changes are added via PR #1585


kcudnik (approved):

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+
+# Scope #
+This document defines the technical specification for the API to support SAI Extensions. Extensions are an abstraction over niche/device-specific features. They facilitate the configuration of newly supported features of ASICs via the familiar SAI API model. The exact feature enabled via the extensions itself is outside the scope of this document but an example is provided below to better explain this feature. These extensions are not a replacement to the existing SAI objects but instead provide a complementary path to achieve the goals set out below.
+
+# Overview #
+SAI Generic extensions introduce a new SAI object SAI_OBJECT_TYPE_GENERIC_PROGRAMMABLE of object_type_oid. This new object uses a single attribute of type sai_json_t defined in this proposal. This object and its attribute provide an abstraction of target-specific configurable features otherwise not described in SAI. This new object enables the 
+- Support ability to add user features
+- Support device-specific capabilities
+- Rapid application prototyping
+
```
rck-innovium, line 44: Most  if not all of the above can be achieved using SAI vendor extensions. Please specify that the main motivation is to avoid changing the SAI header files.

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+                "brief": "Brief Attribute Description",
+                "sai_attr_flags": "<SAI_ATTR_FLAGS_T>",
+                "allowed_object_types": [ "<LIST OF ALLOWED OBJECT TYPES>" ],
+                "default_value": "<DEFAULT ATTR VALUE>"
+            },
+            "value": <VALUE of the attribute>
+        }
+    }
+]
+\endverbatim
```
rck-innovium, line 70: The creation of an object class includes specifying an object name, the attributes valid for this object name, and the metadata for each of this attribute. 

Once the object class is defined, the application can create one or more instances of this object class by specifying the attribute names and values.

In this proposal, both the metadata and values are combined together. This leads to ambiguity.
For eg, what is the behavior if I call create_generic_programmable() twice with the same value for SAI_GENERIC_PROGRAMMABLE_ATTR_OBJECT_NAME but each time I specify different metadata for the attributes.
Let us say, in the second call, I introduce a new mandatory attribute inside the sai_json_t. Now, the first object has been created w/o this mandatory attribute.

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+    {
+        <attribute_name>: {
+            "sai_metadata": {
+                "sai_attr_value_type": "<SAI_ATTR_VALUE_TYPE_T>",
+                "brief": "Brief Attribute Description",
+                "sai_attr_flags": "<SAI_ATTR_FLAGS_T>",
+                "allowed_object_types": [ "<LIST OF ALLOWED OBJECT TYPES>" ],
+                "default_value": "<DEFAULT ATTR VALUE>"
+            },
+            "value": <VALUE of the attribute>
```
rck-innovium, line 66: how do we do a get/set on a specific attribute name? Say virtual_ip in the below example? It is inefficient to get/set the entire SAI_GENERIC_PROGRAMMABLE_ATTR_ENTRY?

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+## sai_json_t ##
+sai_json_t defines a generic data type using JSON as the data format with a predefined grammar allowed to be used within. This grammar allows validators like SAI metadata to verify data and corresponding types accurately. The user of this data type copies a JSON string into this type. The exact structure or content of the API is opaque and is considered an agreed contract between the application and SAI implementation.
+
+```c
+/**
+ * @brief
+\verbatim
+"attributes": [
+    {
+        <attribute_name>: {
```
rck-innovium, line 58: Please specify how the SAI application knows the h/w table (vip_table in this example) and the attributes of the h/w table (virtual_ip and nh_list). 

We should avoid making this vendor specific. Doesn't this make the SAI application locked-in to a particular SAI vendor?

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+
+## Counters ##
+Counters in SAI use 2 patterns.
+- Counters are retrieved using the objects OID and call the associated get_stats() API. sai_get_vlan_stats_fn for SAI_OBJECT_TYPE_VLAN is an example.
+- Counters are retrieved by attaching an explicit counter object to an attribute of the object counters are associated with. SAI_ACL_ENTRY_ATTR_ACTION_COUNTER is used to attach a SAI_OBJECT_TYPE_ACL_COUNTER object. The stats are then obtained by calling sai_get_acl_counter_attribute_fn.
+
+TBD: Use one of the above patterns to retrieve counters based on feedback
+
+## CRM ##
+Table availability and other characteristics can be queried similarly to how the rest of the SAI objects work. sai_object_type_get_availability() allows passing sai_json_t type with the object_name encoded in the attribute list.
```
rck-innovium, line 215: We do not have a sai_object_type per generic object name. i.e. we do not have a unique object type for vip_table objects.  So attribute based filtering is not available in the CRM APIs for these Generic Programmable Entries.

This constraints the functioning of several existing SAI capability query and CRM APIs. 
For example, how will the below work?

- sai_query_attribute_capability()
- sai_query_attribute_enum_values_capability()

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+
+# API Specification #
+
+## New header file saigenericextensions.h ##
+The new header file defines interfaces for a single object of type SAI_OBJECT_TYPE_GENERIC_PROGRAMMABLE.
+
```
rck-innovium, line 221: Why can't this saigenericextensions.h  be under sai vendor extensions API? 
We add this once in SAI vendor extensions and then reuse it for different h/w blocks without recompiling the SAI headers.

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+The list of all generic programmable objects can be queried using the existing saiobject.h infrastructure in SAI.
+
+## CRM usage ##
+```c
+sai_attribute_t attr;
+vector<sai_attribute_t> attrs;
+ 
+// set object name
+std::string object_name = “vip_table”;
+attr.id = SAI_GENERIC_PROGRAMMABLE_ATTR_OBJECT_NAME;
```
rck-innovium, line 361: Attribute based filtering is lost; for eg. let us assume VIP_table has an attribute to say if a given entry is v4 or v6. How do I filter based on v6 entries?

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+The sai_json_t type can include references to other SAI extension blocks. It is allowed to pass the new object type as a reference in the JSON like any other SAI object.
+```json
+{
+    "object_id": {
+        "sai_metadata": {
+            "sai_attr_value_type": "SAI_ATTR_VALUE_TYPE_OBJECT_ID",
+            "allowed_object_types": ["SAI_OBJECT_TYPE_GENERIC_PROGRAMMABLE"]
+        },
+        "value": "0x150000000008ba"
+    }
```
rck-innovium, line 400: How do we handle warm-upgrade?

1)  Say if the metadata of an attribute is changed across releases?

2) I introduce a new mandator attribute in the future release.

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+]
+```
+
+## SAI ACL considerations ##
+The new VIP table is conditionally evaluated based on policies in the ACL tables. This requires a new action type in saiacl.h.
+SAI ACL action type is today exposed via 2 mechanisms. * sai_acl_action_type_t - This type is only used today to query the SAI ACL action capabilities * SAI_ACL_ENTRY_ATTR_ACTION_START/END - An attribute specifiying the ACL action type for an entry
+While there are multiple mechanisms to extend these, one recommendation is as below. 
+- Extend sai_acl_acion_type_t with a custom range just like all the attribute enums for SAI objects 
+  - SAI_ACL_ACTION_TYPE_CUSTOM_RANGE 
+  - SAI_ACL_ACTION_TYPE_DO_LB = SAI_ACL_ACTION_TYPE_CUSTOM_RANGE + 1
```
rck-innovium, line 190: This requires changing the SAI header, thereby defeating the state purpose of avoding header file changes/recompilation.

rck-innovium (changes requested):

doc/SAI_Proposal_Generic_Extensions.md
ravi861: We will recommend sticking with the SAI semantics of performing an update on a single attribute in one call.
Having said that, it is certainly possible for an implementation to allow update for multiple attributes within the JSON.

doc/SAI_Proposal_Generic_Extensions.md
ravi861: Did you mean the physical location of the header file?

doc/SAI_Proposal_Generic_Extensions.md
ravi861: I agree. The design of this particular application required changing the ACL object to get it to work. You will agree though that programming the vip_table here does not require any changes to SAI? That is ultimately the goal to allow programming of niche H/W blocks without modifying SAI.

doc/SAI_Proposal_Generic_Extensions.md
```diff
@@ -0,0 +1,423 @@
...
+  - [CRM API](#crm-usage)
+- [Usage Examples for sai_json_t](#usage-examples-for-saijsont)
+  - [Refer to an OID - Port](#refer-to-an-oid---port)
+  - [Refer to an OID - Generic Programmable Block](#refer-to-an-oid---generic-programmable-block)
+  - [Refer to a key based object - Route](#refer-to-a-key-based-object---route)
+
+# Scope #
+This document defines the technical specification for the API to support SAI Extensions. Extensions are an abstraction over niche/device-specific features. They facilitate the configuration of newly supported features of ASICs via the familiar SAI API model. The exact feature enabled via the extensions itself is outside the scope of this document but an example is provided below to better explain this feature. These extensions are not a replacement to the existing SAI objects but instead provide a complementary path to achieve the goals set out below.
+
+# Overview #
```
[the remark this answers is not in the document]
ravi861: Please find description below.

doc/SAI_Proposal_Generic_Extensions.md
ravi861: Added a description. Please review.

doc/SAI_Proposal_Generic_Extensions.md
ravi861: My thinking is this behavior would mirror the existing SAI objects. The 2 different SAI implementations before and after warmboot should be able handle the difference in attributes.

doc/SAI_Proposal_Generic_Extensions.md
rck-innovium: Yes, I had originally meant placing it. I agree it is ok to leave it here based on the new discussions we had.

rck-innovium (approved):

doc/SAI_Proposal_Generic_Extensions.md
ravi861: @rck-innovium As discussed, we will add this file in the main directory.

ravi861:
@rlhui @kcudnik I will create a 2nd PR with the header file and new data type described in this PR. It will contain some enhancements to SAI meta to incorporate the new type.

ravi861:
Pull request for actual header and code changes are at https://github.com/opencomputeproject/SAI/pull/1585

rlhui:
@JaiOCP , @marian-pritsak , would you please review/approve this?

rlhui:
@ravi861 - please add more PR description.

JaiOCP (approved):

marian-pritsak (approved):

Issue #1531, opened 2022-07-11
[sairpcgen] Extension headers missing from some include paths in Makefile


chrispsommers:
When building saithriftv2, extension headers are missing from some include paths, for example see the build run below. I will follow-up with a PR to fix this.
```
ar rcs ./obj/librpcserver.a obj/sai_rpc.o obj/sai_types.o obj/sai_rpc_server.o
g++ -I/usr/include/sai -I. -std=c++11 -O0 -ggdb -c src/saiserver.cpp  -o obj/saiserver.o -I/usr/include/sai -I. -std=c++11 -O0 -ggdb -DBRCMSAI -I./gen-cpp -I../../inc
In file included from /usr/include/sai/sai.h:48,
                 from src/saiserver.cpp:28:
/usr/include/sai/saiobject.h:43:10: fatal error: saiexperimentalbmtor.h: No such file or directory
   43 | #include <saiexperimentalbmtor.h>
```

kcudnik:
probably addressed in here: https://github.com/opencomputeproject/SAI/pull/1533, when experimental directory headers needs to be added to compilation path

richardyu-ms:
cause this issue block the saithrift v2 related docker's build (ptf-sai), try to resolve it here https://github.com/opencomputeproject/SAI/pull/1552/files

chrispsommers:
Hi kind reviewers @kcudnik @richardyu-ms I determined that this issue is bogus, and that copying SAI extension headers to `/usr/include/sai` as explained in [SAI_PTF_user-guide](https://github.com/opencomputeproject/SAI/blob/master/ptf/SAI_PTF_user-guide.md#generate--build) eliminates the problem. I will back out the corresponding changes in https://github.com/opencomputeproject/SAI/pull/1533. Thanks for your advice. - chris

PR #1552, opened 2022-08-02
[SAI-PTF]Support init swtich from syncd-rpc container

richardyu-ms:
Support init swtich from syncd-rpc container

In Syncd starting process, it will try to start the switch and assign the gSwtichId to the sai implementation.
In the syncd rpc container, in order to get the swtich id after syncd initalization, we need to get and check the switch id.

Meanwhile support build the saithrift with vs platform

Test done:
Test the syncd-rpc container can start and the switch id can be retrieve from thrift client

### DIFF
--- a/meta/gensairpc.pl
+++ b/meta/gensairpc.pl
@@ -281,7 +281,7 @@ sub generate_server_template_from_skeleton {
 
                 # Define global variable before "class"
                 print {$server_template}
-"\nextern sai_object_id_t switch_id;\nsai_object_id_t switch_id;\n\n\n";
+"\nextern sai_object_id_t switch_id;\nsai_object_id_t switch_id;\nextern sai_object_id_t gSwitchId;\n\n\n";
 
                 # Define helper functions
                 print {$server_template} "[% PROCESS helper_functions %]\n\n\n";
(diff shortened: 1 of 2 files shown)


richardyu-ms:
@chrispsommers please help make a review

test/saithriftv2/Makefile
```diff
@@ -88,7 +88,7 @@ $(ODIR)/sai_rpc_server.o: ../../meta/sai_rpc_frontend.cpp
 	$(CXX) $(CPPFLAGS) -c $^ -o $@ $(CPPFLAGS) -I./gen-cpp -I../../meta -I../../inc -I../../experimental
 
 $(ODIR)/saiserver.o: src/saiserver.cpp $(CPP_SOURCES) directories
-	$(CXX) $(CPPFLAGS) -c src/saiserver.cpp  -o $@ $(CPPFLAGS) $(CDEFS) -I./gen-cpp -I../../inc
+	$(CXX) $(CPPFLAGS) -c src/saiserver.cpp  -o $@ $(CPPFLAGS) $(CDEFS) -I./gen-cpp -I../../inc -I../../experimental
```
chrispsommers, line 91: This might not be needed anymore, see https://github.com/opencomputeproject/SAI/issues/1531#issuecomment-1203359282

chrispsommers (approved):

kcudnik (approved):

test/saithriftv2/Makefile
richardyu-ms: thanks. Is there any code copy those files to /usr/include/sai/?

PR #1562, opened 2022-08-10
Add VNET object to DASH

marian-pritsak:


experimental/saiexperimentaldashvnet.h
```diff
@@ -354,6 +305,38 @@ typedef enum _sai_outbound_routing_entry_attr_t
 
 } sai_outbound_routing_entry_attr_t;
 
+/**
+ * @brief Attribute ID for dash_vnet_vnet
```
prsunny, line 309: should this be just dash_vnet?

prsunny (commented):
lgtm, minor comment

kcudnik (approved):

experimental/saiexperimentaldashvnet.h
marian-pritsak: dash_vnet is the API name, and vnet is the object type. They should've been separated, but there must be mistake in the API generation. It appears in all enum headers. Let me fix it separately to not pollute this change for every enum.

PR #1571, opened 2022-08-16
Fix DASH inbound routing

marian-pritsak:
Add VNET classification

### DIFF
--- a/experimental/saiexperimentaldashvnet.h
+++ b/experimental/saiexperimentaldashvnet.h
@@ -85,9 +85,31 @@ typedef struct _sai_inbound_routing_entry_t
     sai_object_id_t switch_id;
 
     /**
+     * @brief Exact matched key eni_id
+     *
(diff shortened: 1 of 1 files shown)


experimental/saiexperimentaldashvnet.h
```diff
@@ -84,11 +84,33 @@ typedef struct _sai_inbound_routing_entry_t
...
+
     /**
      * @brief Exact matched key VNI
      */
     sai_uint32_t vni;
 
+    /**
+     * @brief Ternary matched key sip
+     */
+    sai_ip_address_t sip;
```
prsunny, line 102: should this be ip_prefix?

marian-pritsak: It is a ternary match (priority based) with explicit priority. Prefix is used for LPM only.

prsunny (approved):

PR #1573, opened 2022-08-18
[SAI-PTF]  Added support to run tests on devices with fewer ports

aputriax:
- Add support to run tests on devices that have fewer ports
    than required by SONiC common setup.
  - Adopted applicable sairif test cases to use new approach.
  - Updated PTF SAI proposal doc.

Details:
  - SaiHelperUtilsMixin class created with API for SAI objects creation.
  - SaiHelperSimplified class created for DUT with limited port resources
    without initial switch ports setup.
  - SaiHelper refactored (inherited from the new SaiHelperUtilsMixin class).
  - sairif.py contains examples of new classes usage.

Signed-off-by: Anton Putria <antonx.putria@intel.com>
Signed-off-by: Volodymyr Mytnyk <volodymyrx.mytnyk@intel.com>

### DIFF
--- a/doc/SAI-Proposal-SAI-PTF.md
+++ b/doc/SAI-Proposal-SAI-PTF.md
@@ -5,7 +5,8 @@
 | Authors     | Intel           |
 | Status      | In review       |
 | Type        | Standards track |
-| Created     | 04/14/2022      |
+| Created     | 10/13/2021      |
+| Updated     | 06/07/2022      |
 | SAI-Version | 1.9             |
 
 ## Overview
--- a/doc/SAI-Proposal-SAI-PTF.md
+++ b/doc/SAI-Proposal-SAI-PTF.md
@@ -5,7 +5,7 @@
 | Authors     | Intel           |
 | Status      | In review       |
 | Type        | Standards track |
-| Created     | 10/13/2021      |
+| Created     | 04/14/2022      |
 | SAI-Version | 1.9             |
 
 ## Overview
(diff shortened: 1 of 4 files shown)


richardyu-ms:
in this change, the dest port(egress port base one some route or forward rule) is moved to port1, could it be more dynamic, like binding the dest port when create the forwarding rule in fdb table, route? like a object for simulate the dest device with a certain ip, and that device have a dest port, and that port index defined in that device.
when create route we set

device.ip = 1.0.0.1
device.egress_port_index = 1

and when testing
verify_packet(self, device.egress_port_index)

aputriax:
> in this change, the dest port(egress port base one some route or forward rule) is moved to port1, could it be more dynamic, like binding the dest port when create the forwarding rule in fdb table, route? like a object for simulate the dest device with a certain ip, and that device have a dest port, and that port index defined in that device. when create route we set
> 
> device.ip = 1.0.0.1 device.egress_port_index = 1
> 
> and when testing verify_packet(self, device.egress_port_index)
> 
> by this way, we can define the dest port when create the config (still can use previous basic_test class), and can make the adjustment base on the actual device?

Do I understand correctly your suggestion?
Ixia has a feature to create virtual hosts with IPs to send traffic. PTF doesn't have this. But what you suggest is to create object that will store endpoint information. So when we do IP configuration (e.g. in sairif.py:L3InterfaceSimplifiedTest.setUp() we do not rely on portN and dev_portN numbers, but rather define self.device1.ingress_port_index & self.device1.egress_port_index at this stage to use them in test cases.

We can create those self.deviceN in sai_base_test.py:SaiHelperUtilsMixin:create_routing_interfaces() and fill in ingress_port_index/ip and egress_port_index/ip values in setUp() methods, when we add actual IP configuration.

So tests will rely on deviceN and not on portN/dev_portN configuration.

aputriax:
As this is an improvement and not an issue with the current implementation I would like to have this in a separate PR. Is this OK?

We already started this improvement implementation for the new tests. Once we have finalized look and feel of it, we can update existing tests. It doesn't look like a big change, but it will require a lot of renaming in all test suites from the PR #1592 as well. So I would like to separate this work.

richardyu-ms (approved):

reshmaintel (commented):
These are the changes from Intel to add Test Harness for DASH support by extending SAI-PTF

PR #1575, opened 2022-08-19
[SAI-PTF]add EcmpLagTest

zhiyuan0112:
Signed-off-by: zhiyuanliang-ms <liangzy512@gmail.com>

### DIFF
--- a/test/sai_test/config/lag_configer.py
+++ b/test/sai_test/config/lag_configer.py
@@ -42,6 +42,7 @@ def t0_lag_config_helper(test_obj: 'T0TestBase', is_create_lag=True):
     if is_create_lag:
         test_obj.dut.lag1 = lag_configer.create_lag([17, 18])
         test_obj.dut.lag2 = lag_configer.create_lag([19, 20])
+        test_obj.dut.lag3 = lag_configer.create_lag([21, 22])
 
     """
     lag_configer.set_lag_hash_algorithm()
--- a/test/sai_test/config/route_configer.py
+++ b/test/sai_test/config/route_configer.py
@@ -30,6 +30,7 @@ from data_module.lag import Lag
 from typing import Dict, List
 
 from data_module.nexthop import Nexthop
+from data_module.ecmp import Ecmp
 
 if TYPE_CHECKING:
     from sai_test_base import T0TestBase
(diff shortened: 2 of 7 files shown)


test/sai_test/config/route_configer.py
```diff
@@ -739,6 +749,31 @@ def create_nexthop_by_rif(self, rif, nexthop_device: Device):
...
+            self.test_obj.client,
+            next_hop_group_id=self.test_obj.nhop_group1,
+            next_hop_id=self.test_obj.dut.lag2.nexthopv4.nexthop_id)
+
+        self.test_obj.nh_group_member1 = sai_thrift_create_next_hop_group_member(
+            self.test_obj.client,
+            next_hop_group_id=self.test_obj.nhop_group1,
+            next_hop_id=self.test_obj.dut.lag3.nexthopv4.nexthop_id)
+
+        self.test_obj.ecmp_route0 = sai_thrift_route_entry_t(
```
richardyu-ms, line 769: you can create a list to save nexthop groups
after create the nexthop group, it should be returned as other method

test/sai_test/config/route_configer.py
```diff
@@ -739,6 +749,31 @@ def create_nexthop_by_rif(self, rif, nexthop_device: Device):
 
         return nhopv4, nhopv6
 
+    def create_nexthop_group_by_nexthops(self, dest_device: Device):
```
richardyu-ms, line 752: for a method, the parameters can always to illustrate what kinds of data that will be needed in this method, is it only need the device?

test/sai_test/config/route_configer.py
```diff
@@ -739,6 +749,31 @@ def create_nexthop_by_rif(self, rif, nexthop_device: Device):
...
+        """
+        Create nexthop group by nexthops.
+        """
+        self.test_obj.nhop_group1 = sai_thrift_create_next_hop_group(
+            self.test_obj.client, type=SAI_NEXT_HOP_GROUP_TYPE_ECMP)
+
+        self.test_obj.nh_group_member1 = sai_thrift_create_next_hop_group_member(
```
richardyu-ms, line 759: we define two kind of data objects, device and dut, for persist then in a file then we can load them.
Why add those attribute in test_obj?

test/sai_test/config/lag_configer.py
```diff
@@ -42,6 +42,7 @@ def t0_lag_config_helper(test_obj: 'T0TestBase', is_create_lag=True):
     if is_create_lag:
         test_obj.dut.lag1 = lag_configer.create_lag([17, 18])
         test_obj.dut.lag2 = lag_configer.create_lag([19, 20])
+        test_obj.dut.lag3 = lag_configer.create_lag([21, 22])
```
richardyu-ms, line 45: we have more than three lags, maybe add them to a list can make other operation on them easier

test/sai_test/sai_ecmp_test.py
```diff
@@ -0,0 +1,150 @@
...
+                    exp_pkt2 = simple_tcp_packet(eth_dst=self.t1_list[3][100].mac,
+                                        eth_src=ROUTER_MAC,
+                                        ip_dst=ip_dst,
+                                        ip_src=ip_src,
+                                        tcp_sport= src_port,
+                                        ip_id=105,
+                                        ip_ttl=63)   
+
+                    send_packet(self, 1, pkt)
+                    rcv_idx= verify_any_packet_any_port(self, [exp_pkt1, exp_pkt2], [19, 20, 21, 22])
```
richardyu-ms, line 61: in the nexthop_group object, when we create it, we already know how many ports belongs to it, then we can use the data base on how we create them but not a index group, like 19~22, without any meaning

richardyu-ms:
please follow this PR to add more description 
https://github.com/opencomputeproject/SAI/pull/1553

PR #1585, opened 2022-08-25
SAI Generic Programmable Extensions header

ravi861:
Signed-off-by: Ravi Vantipalli <ravi.vantipalli@intel.com>

This PR adds the header files and relevant meta code changes to SAI. The proposal is added and reviewed in PR #1551.

### DIFF
--- a/inc/saigenericprogrammable.h
+++ b/inc/saigenericprogrammable.h
@@ -1,0 +1,155 @@
+/**
+ * Copyright (c) 2014 Microsoft Open Technologies, Inc.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *    not use this file except in compliance with the License. You may obtain
+ *    a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    THIS CODE IS PROVIDED ON AN *AS IS* BASIS, WITHOUT WARRANTIES OR
+ *    CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT
+ *    LIMITATION ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS
+ *    FOR A PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
+ *
+ *    See the Apache Version 2.0 License for specific language governing
+ *    permissions and limitations under the License.
+ *
+ *    Microsoft would like to thank the following companies for their review and
+ *    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
+ *    Dell Products, L.P., Facebook, Inc., Marvell International Ltd.
+ *
+ * @file    saigenericprogrammable.h
+ *
+ * @brief   This module defines SAI Genetic Programmable Extensions (GPE)
+ */
+
+#if !defined (__SAIGENERICPROGRAMMABLE_H_)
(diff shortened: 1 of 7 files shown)


inc/saitypes.h
```diff
@@ -989,6 +990,35 @@ typedef struct _sai_segment_list_t
...
+ * sai_attr_value_type - Data type of the attribute
+ * brief - Optional description of the field
+ * sai_attr_flags - Optional Usage flags for the field
+ * allowed_object_types - If data type is OID, then this is the list of object types allowed as data
+ */
+typedef struct _sai_json_t
+{
+    /** String in JSON format */
+    sai_s8_list_t json;
+} sai_json_t;
```
kcudnik, line 1020: why we need special type if it's just wrapper ? and not use sai_s8_list_t directly?

kcudnik (changes requested):
check comments

inc/saitypes.h
ravi861: Having a new SAI_ATTR_VALUE_TYPE for json is very useful in generic code like sonic-sairedis. The JSON is allowed to carry other SAI datatypes like OID, IP address, etc and knowing that the attribute is a JSON helps with parsing it.
If I use straight, sai_s8_list_t, I don't think there is a way to distinguish between a simple string and a JSON string.
Would it help if we use a count/list within sai_json_t directly?

kcudnik: no, lets leave this as is then

kcudnik (approved):

reshmaintel:
@JaiOCP @marian-pritsak @rck-innovium Please do complete this review. Thanks.

rlhui:
@JaiOCP @marian-pritsak @rck-innovium - please review/approve?  Thanks.

rlhui:
@ravi861  - please add description of this PR and can we not use "GPE" in title but expand it?

inc/saigenericprogrammable.h
```diff
@@ -0,0 +1,155 @@
...
+    SAI_GENERIC_PROGRAMMABLE_ATTR_OBJECT_NAME = SAI_GENERIC_PROGRAMMABLE_ATTR_START,
+
+    /**
+     * @brief JSON string carrying HW block entry information
+     *
+     * @type sai_json_t
+     * @flags CREATE_AND_SET
+     * @default vendor
+     */
+    SAI_GENERIC_PROGRAMMABLE_ATTR_ENTRY,
```
rck-innovium, line 61: If not already done, please add API examples that cover:

1) creating multiple entries of the same object name. For example, creating multiple VIP entries.


2) Updating a single attribute of an entry. For example, changing the nexthop-id of a VIP entry.

rck-innovium (approved):

marian-pritsak (approved):

rlhui:
Can we please add more description to the PR. Currently it only has "sign-off by ..."

inc/saigenericprogrammable.h
rlhui: Hi, I don't see example of "creating multiple VIP entries" in https://github.com/opencomputeproject/SAI/blob/19ba5116264829d3a7b933212ab6bd0d977a3a55/doc/SAI_Proposal_Generic_Extensions.md, and also changing the entry?

ravi861: Adding multiple different entries is the same as the single create example. The same invocation as existing example would work with modified IP and nexthop information.
I added an code example for update scenario also.

rlhui (approved):

PR #1599, opened 2022-09-14
[SAI-PTF]Proposal for enhancement points in next hop cases

zhoudongxu:
## Description
For next hop cases, there are some issue need to be addressed

1. **init and setup the ports but not recreate port**
   - Issue 
   between different platform, some of them doesn't support recreate port  but need to init and setup the ports at once after start the switch.
   - Solution
   We use PlatformSaiHelper to help set basic common config base on different platform instead of  base class SaiHelper. When setting Platform = brcm,  PlatformSaihelper actually is BrcmSaihelper.
  - Details
   #1603
  - Current Result
     image

     image

2.  **default route for default VRF is mandartory.**  
    - Issue
   Creating route in  default VRF will failed if there aren't default routes.
    - Solution
   Create default route before create route in detaul VRF
    - Details 
       #1606

    - Current Result
       
        image

3. **Creating object in order**  
   - Issue 
   If object A have reference to  object B, Creating A before B is necessary in some platform. 
   For example, Creating object must be in order neighbor->nexthop->route, otherwise, it will failed.
   - Solution
  Ordering the calling sequence
   -  Details 
#1607
   - Current Result
     
     image

4. **Deleting object in order**
   - Issue
   If object A have reference to  object B, Deleting A after B is necessary
   For example, Deleting object order is route->nexthop->neighbor
   - Solution
  Ordering the calling sequence
   -  Details 
#1607

5. **Verification against unsupport interface/feature**
   - Issue
     Case failed when verfication agsinst some unsupported interface and feature
   - Solution
     Temp solution, removed check points.
   - Original Result
     image
   - Current Result
     we choose disable this check because of   other platform maybe don't support the counter

> Note: For Points 3 & 4, there might be some other error cause by the calling sequence, need work when hit them.

> Todo: For point 1, There should be some common way to initalize the port (like sonic behavior mode), there will be another PR to replace the platform specific method and make the test structure support current cases meanwhile support other platforms

> Todo: need better soltuion for the unsupport interface

### DIFF
--- a/ptf/sai_base_test.py
+++ b/ptf/sai_base_test.py
@@ -471,6 +471,7 @@ class SaiHelperBase(ThriftInterfaceDataPlane):
                     self.client, port, oper_status=True)
                 if attr['oper_status'] != SAI_SWITCH_OPER_STATUS_UP:
                     allup = False
+                    print("prot is down")
                     break
             if allup:
                 break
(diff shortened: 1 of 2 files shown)


ptf/sai_base_test.py
```diff
@@ -946,6 +946,35 @@ def setUp(self):
         self.assertEqual(self.status(), SAI_STATUS_SUCCESS)
         self.def_rif_list.append(self.port13_rif)
 
+        self.create_default_v4_v6_route_entry()
+
+    def create_default_v4_v6_route_entry(self):
```
zhoudongxu, line 951: Broadcom asic must create default route for  default VRF.

ptf/sainexthop.py
```diff
@@ -23,10 +23,10 @@
 from ptf.thriftutils import *
 
 from sai_base_test import *
-
+import pdb
 
 @group("draft")
-class L3NexthopTest(SaiHelper):
+class L3NexthopTest(PlatformSaiHelper):
```
zhoudongxu, line 29: replace saihelper with brcmsaihelper as basic setting up proccess

ptf/sainexthop.py
```diff
@@ -23,10 +23,10 @@
 from ptf.thriftutils import *
 
 from sai_base_test import *
-
+import pdb
```
richardyu-ms, line 26: please remove the unnecessary import

richardyu-ms (approved):

PR #1608, opened 2022-09-22
Fix for metadata serialization of enum list.

j-bos:
Fix a couple of bugs in the metadata enum-list serialization function, sai_serialize_enum_list(), which prevented it from producing useful output.


kcudnik (approved):

kcudnik:
good catch, thanks!

PR #1609, opened 2022-09-25
[SAI-PTF] Add decorator for skipping test on specified error

richardyu-ms:
Add decorator method in the perl template  for skipping test on specified error
try to resolve issue 
#1610 

details:
1. add decorator method in sai_adapter_utils.tt
2. use decorator when set parameters ``skip_error``
3. add parameters when building from template, like ``perl -Irpc gensairpc.pl  --skip_error=-2,-3``

Test done:
GEN_SAIRPC_OPTS="--skip_error=-2,-3" make rpc


richardyu-ms:
@ravi861 please help make a review, thanks.

ravi861:
While this decorator has its uses, it should not be used to solve the issue raised in #1610 . The correct approach is by verifying if the attribute is supported using sai attribute query capability.

ravi861 (approved):

richardyu-ms:
> While this decorator has its uses, it should not be used to solve the issue raised in #1610 . The correct approach is by verifying if the attribute is supported using sai attribute query capability.

Yes, this is a approach in sai latter than 1.6, but consider we need to use this framework in differernt versions, use this as a first edition, we can make other improvement for different versions of sai

PR #1615, opened 2022-10-03
Implement a mechanism that to enable a NOS running SAI and SDK, to detect SDK and FW failures

amandyuk:
Add new health event, for SAI to notify severity, timestamp and description of a SDK/FW health event

Signed-off-by: Andrii Mandiuk <amandiuk@nvidia.com>

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -1429,6 +1429,15 @@ typedef enum _sai_acl_table_attr_t
     SAI_ACL_TABLE_ATTR_AVAILABLE_ACL_COUNTER,
 
     /**
+     * @brief Match type for the table
+     *
+     * @type sai_acl_table_match_type_t
+     * @flags CREATE_ONLY
+     * @default SAI_ACL_TABLE_MATCH_TYPE_TERNARY
+     */
+    SAI_ACL_TABLE_ATTR_ACL_TABLE_MATCH_TYPE,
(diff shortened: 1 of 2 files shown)


kcudnik (approved):

mikeberesford:
@JaiOCP this looks to be partially overlapping with https://github.com/opencomputeproject/SAI/pull/1307

rlhui:
better if nos doesn't need to parse string but the info is in a defined format.  Please also checkout https://github.com/sonic-net/SONiC/blob/5677242770e3774a5eab0f951dc14642d81422ec/doc/event-alarm-framework/events-producer.md

reshmaintel (commented):
@itaibaz I like this proposal that this SDK failure detection and reporting is generic and gives the control to the vendor to push specific failure logs with related severity that help user to debug the given vendor SDK/HW. I think that if this feature is only severity level based it may resort to being an alternative to syslog. If there is additional categorization for the failures it may be more valuable to the user to monitor certain categories. Certain broad categories can be suggested. Given that using enums for categories seem restrictive and not having to add new enums for additional categories is one of your goals, please explore if we can make it extensible by adding the category as a field just like the description field.

itaibaz:
opened https://github.com/opencomputeproject/SAI/pull/1777/ instead

PR #1616, opened 2022-10-05
Allow for CREATE_ONLY attributes in non objects

marian-pritsak:
Primarily desired for action selectors to make
the dependent attributes that are object mandatory

### DIFF
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -3690,6 +3690,7 @@ void check_non_object_id_object_attrs()
             {
                 case SAI_ATTR_FLAGS_MANDATORY_ON_CREATE | SAI_ATTR_FLAGS_CREATE_AND_SET:
                 case SAI_ATTR_FLAGS_CREATE_AND_SET:
+                case SAI_ATTR_FLAGS_CREATE_ONLY:
                     break;
 
                 default:


marian-pritsak:
The example is as follows:

We have a route entry with an action selector:
```
    /**
     * @brief Packet action
     *
     * @type sai_packet_action_t
     * @flags CREATE_AND_SET
     */
    SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION = SAI_ROUTE_ENTRY_ATTR_START,
```

And another attribute that is an object:
```
    /**
     * @brief Next hop or next hop group id for the packet, or a router interface
     * in case of the directly reachable route or the CPU port in case of IP2ME route
     *
     * The next hop id is only effective when the packet action is one of the
     * following: FORWARD, COPY, LOG, TRANSIT.
     *
   ...
     *
     * When it is SAI_NULL_OBJECT_ID, then the packet will be dropped.
     *
     * @type sai_object_id_t
     * @flags CREATE_AND_SET
     * @objects SAI_OBJECT_TYPE_NEXT_HOP, SAI_OBJECT_TYPE_NEXT_HOP_GROUP, SAI_OBJECT_TYPE_ROUTER_INTERFACE, SAI_OBJECT_TYPE_PORT
     * @allownull true
     * @default SAI_NULL_OBJECT_ID
     */
    SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID,
```

The next hop ID only makes sense if the action is not DROP. If the action is FORWARD, and next hop ID is NULL, it is just another way of saying action == DROP.

There may be action selectors other than sai_packet_action_t that don't allow for drop and object attributes like next hop ID cannot be NULL, which the metadata checker does not allow for today.

kcudnik (approved):

PR #1655, opened 2022-11-16
Update saiport.h to add new interface type for USXGMII

SidharajU:
Adding a new USXGMII interface


rlhui:
Please fix the DCO error. Thanks.

PR #1660, opened 2022-11-20
[SAI-PTF]Include sai expermential for generate the rpc headers

richardyu-ms:
When generate the sai_adapter it depends on the files in  sai expermental https://github.com/opencomputeproject/SAI/blob/master/meta/Makefile#L72
```
CFLAGS += -I../inc -I../experimental $(WARNINGS)
```
In gensairpc
https://github.com/opencomputeproject/SAI/blob/master/meta/gensairpc.pl#L132
```
our $EXPERIMENTAL_DIR = catdir( $sai_dir,      'experimental' );
```
But when generate the py headers for RPC service, it doesn't include that. In order to make them matched, then add that dependences.

Test done:
Local compile and checked the headers

### DIFF
--- a/test/saithriftv2/Makefile
+++ b/test/saithriftv2/Makefile
@@ -82,7 +82,7 @@ $(PY_SOURCES):  $(METADIR)sai.thrift
 	$(THRIFT) -o ./ --gen py -r $^
 
 $(SAI_PY_HEADERS): $(SAI_HEADERS)
-	$(CTYPESGEN) --output-language=py32 -I/usr/include -I$(SAI_HEADER_DIR) --include /usr/include/linux/limits.h $^ -o $@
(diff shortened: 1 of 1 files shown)


kcudnik:
@richardyu-ms what errors this causing ? since i test this locally and with experimental include or not i still get the same gen-py/sai/sai_headers.py generated

richardyu-ms:
> @richardyu-ms what errors this causing ? since i test this locally and with experimental include or not i still get the same gen-py/sai/sai_headers.py generated
> 
> can you share exact steps how you reproduce that the experimental are missing?

@kcudnik I cannot recall the actual vendor OS version which caused this issue, but i remember the RC was the local SAI headers(from SAI repo) did not match the vendor SAI headers within their SDK

kcudnik:
yes, differences between vendor and actual SAI headers can lead to weird errors, and it could be even hard to track, since they can pop up while runtime at some point

PR #1670, opened 2022-12-01
[TAM] Granular counter subscription

j-bos:
Enhancements to TAM API to enable a high-speed polling feature to repeatedly poll one or a few counters at short time intervals.

- A new telemetry type for granular counter subscription to request specific counters at runtime.

- Introduce SAI_TAM_COUNTER_SUBSCRIPTION object to represent subscribed counters and allow the user to associate a label for the counter in telemetry reports.

- Add a time units configuration to specify units for the report interval. Default is backwards-compatible to the current units (microseconds).

### DIFF
--- a/doc/TAM/Granular-Counter-Subscription.md
+++ b/doc/TAM/Granular-Counter-Subscription.md
@@ -1,0 +1,208 @@
+#  Telemetry Granular Counter Subscription
+-------------------------------------------------------------------------------
+ Title       | Telemetry Granular Counter Subscription
+-------------|-----------------------------------------------------------------
+ Authors     | Jason Bos, Cisco
+ Status      | In review
+ Type        | Standards track
+ Created     | 2023-01-11 - Initial Draft
+ SAI-Version | 1.12
+-------------------------------------------------------------------------------
+
+
+## 1.0  Introduction
+
+This spec enhances the existing TAM (Telemetry and Monitoring) spec to add granular counter subscription.
+
+The TAM API  allows for predefined telemetry groups, like SAI_TAM_TELEMETRY_TYPE_SWITCH or SAI_TAM_TELEMETRY_TYPE_PORT. This provides extensibility to the API only to define new fixed collections. When collecting counters, all supported counters for the relevant objects are reported.
+
+However, the user may wish to enable telemetry for a more limited set of counters, selected at runtime. For example, a user may wish to receive only port byte counters, or queue watermarks. Limiting the scope of the telemetry collection permits a device to generate counters samples faster than possible when all counters are collected, and fit many more samples into each network packet, allowing a higher rate of delivery.
+
+In this mode, the collector may require a meaningful identifier for a counter, which identifies both the object as well as the specific counter on the object.
+Depending on the report type, this may be carried directly in the report itself. Or it may be delivered separately, for example as an enterprise-specific
+information element ID in an IPFIX template set.
+
+## 2.0 Configuration
+
+The subscription is represented by an object of SAI_TAM_COUNTER_SUBSCRIPTION. The creation of this object will indicate that a specific counter should be monitored.
+
+```c
+typedef enum _sai_tam_counter_subscription_attr_t
+{
+    /**
+     * @brief Start of Attributes
+     */
+    SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_START,
+
+    /**
+     * @brief TAM telemetry type object
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_TAM
+     */
+    SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_TEL_TYPE = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_START,
+
+    /**
+     * @brief Subscribed object
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_OBJECT_ID,
+
+    /**
+     * @brief Subscribed stat enum
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_STAT_ID,
+
+     /**
+     * @brief Telemetry label
+     *
+     * Label to identify the subscribed counter in telemetry reports.
+     *
+     * @type sai_uint64_t
+     * @flags CREATE_ONLY
+     * @default 0
+     */
+    SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_LABEL,
+```
+
+The attribute SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_LABEL may be used to provide the NOS-specific counter ID.
+
+
+## 3 Configuration example
+
+```c
+
+// Example: Create a bulk report object
+sai_attr_list[0].id = SAI_TAM_REPORT_ATTR_TYPE;
+sai_attr_list[0].value.s32 = SAI_TAM_REPORT_TYPE_PROTO;
+
+sai_attr_list[1].id = SAI_TAM_REPORT_ATTR_REPORT_MODE;
+sai_attr_list[1].value.s32 = SAI_TAM_REPORT_MODE_BULK;
+
+attr_count = 2;
+
+sai_create_tam_report_fn(
+   &sai_tam_report_obj,
+   switch_id,
+   attr_count,
+   sai_attr_list);
+
+// Example: Create telemetry type object to collect object stats
+sai_attr_list[0].id = SAI_TAM_TEL_TYPE_ATTR_TAM_TELEMETRY_TYPE;
+sai_attr_list[0].value.u32 = SAI_TAM_TELEMETRY_TYPE_OBJECT_STAT;
+
+sai_attr_list[1].id = SAI_TAM_TEL_TYPE_ATTR_REPORT_ID;
+sai_attr_list[1].value.oid = sai_tam_report_obj;
+
+attr_count = 2;
+sai_create_tam_tel_type_fn(
+    &sai_tam_tel_type_obj,
+    switch_id,
+    attr_count,
+    sai_attr_list);
+
+// Example: Create counter subscription(s) to collect queue length and watermark
+sai_attr_list[0].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_TEL_TYPE;
+sai_attr_list[0].value.u32 = sai_tam_tel_type_obj;
+
+sai_attr_list[1].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_OBJECT_ID;
+sai_attr_list[1].value.oid = sai_queue_obj;
+
+sai_attr_list[2].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_STAT_ID;
+sai_attr_list[2].value.u32 = SAI_QUEUE_STAT_CURR_OCCUPANCY_BYTES;
+
+sai_attr_list[3].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_LABEL;
+sai_attr_list[3].value.u64 = 1;
+
+attr_count = 4;
+sai_create_tam_counter_subscription_fn(
+    &sai_tam_subscription_obj1,
+    switch_id,
+    attr_count,
+    sai_attr_list);
+
+sai_attr_list[0].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_TEL_TYPE;
+sai_attr_list[0].value.u32 = sai_tam_tel_type_obj;
+
+sai_attr_list[1].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_OBJECT_ID;
+sai_attr_list[1].value.oid = sai_queue_obj;
+
+sai_attr_list[2].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_STAT_ID;
+sai_attr_list[2].value.u32 = SAI_QUEUE_STAT_WATERMARK_BYTES;
+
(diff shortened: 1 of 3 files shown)


j-bos:
Capturing here comments from review on SAI weekly call:

- PERIODIC may not be needed. Use existing attribute for the counter polling interval.
- Report should include a report / device id to identify the originating device. 
- Move document to .md format

inc/saitam.h
```diff
@@ -1858,6 +1921,7 @@ typedef enum _sai_tam_event_type_t
      * @brief Buffer service pool threshold event
      */
     SAI_TAM_EVENT_TYPE_BSP,
+
```
j-bos, line 1924: Remove whitespace change.

doc/TAM/Timeslice-Telemetry-and-Granular-Counter-Subscription.md
```diff
@@ -0,0 +1,138 @@
...
+
+## 2.0 Timeslice Telemetry
+
+The Timeslice telemetry format is an output report format for exporting one or more counters correlated with collection timestamps. The goal is to provide a compact format to observe multiple counters correlated closely correlated in time.
+
+In implementations, the reports may be generated by the datapath, so the exact format of the packet will be vendor-defined. One sample format:
+
+| Node ID | Time-0 timestamp | Time-1 timestamp | Time-2 timestamp | Time-3 timestamp | Time-4 timestamp | Time-5 timestamp |
+| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
+| Counter-ID 0 | Counter-0 at Time-0 | Counter-0 at Time-1 | Counter-0 at Time-2 | Counter-0 at Time-3 | Counter-0 at Time-4 | Counter-0 at Time-5  |
```
rck-innovium, line 28: What is the mapping between the counter-id in the report to the actual SAI Object id?

How is the node-id mapped to a particular SAI switch instance?

doc/TAM/Timeslice-Telemetry-and-Granular-Counter-Subscription.md
```diff
@@ -0,0 +1,138 @@
...
+| Counter-ID 1 | Counter-1 at Time-0 | Counter-1 at Time-1 | Counter-1 at Time-2 | Counter-1 at Time-3 | Counter-1 at Time-4 | Counter-1 at Time-5  |
+| Counter-ID 2 | Counter-2 at Time-0 | Counter-2 at Time-1 | Counter-2 at Time-2 | Counter-2 at Time-3 | Counter-2 at Time-4 | Counter-2 at Time-5  |
+
+To configure the format of the report, the TAM report API is extended with a new type and attribute to declare the number of samples per report, i.e. the width of the table.
+
```
rck-innovium, line 33: Can you please give some example for a reference implementation. Something like:


| Switch-id 0x1023456 | UTC 1673848590 | UTC 1673848595 | UTC 1673848600 |
| :--- | :--- | :--- | :--- |
| Eth0-Q2 SAI\_QUEUE\_STAT\_CURR\_OCCUPANCY\_BYTES | 0 | 3000 | 4000 |
| Eth1-Q3 SAI\_QUEUE\_STAT\_CURR\_OCCUPANCY\_BYTES | 4000 | 3000 | 0 |
| Eth2-IPG2 SAI\_INGRESS\_PRIORITY\_GROUP\_STAT\_CURR\_OCCUPANCY\_BYTES | 1000 | 1000 | 1000|

rck-innovium (changes requested):

rck-innovium (commented):
We need capability APIs to expose:

1.  the timeslice interval that can be supported by the device
2. the max number and types of counters that can be polled in one polling interval atomically

doc/TAM/Timeslice-Telemetry-and-Granular-Counter-Subscription.md
```diff
@@ -0,0 +1,138 @@
...
+    &sai_tam_flow_tel_type_obj,
+    switch_id,
+    attr_count,
+    sai_attr_list);
+
+// Example: Create TAM object and attach to monitored objects:
+sai_attr_list[0].value.objlist.count = 1;
+sai_attr_list[0].value.objlist.list[0] = sai_tam_telemetry_obj;
+
+sai_attr_list[1].id = SAI_TAM_ATTR_EVENT_OBJECTS_LIST;
```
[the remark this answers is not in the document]
j-bos: Yes, will do.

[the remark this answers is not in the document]
j-bos: Expanded example.

j-bos:
Significant Updates:

- Decouple the telemetry report format from this subscription proposal. Existing report types like proto or ipfix may be used.
- A counter label is required for the collector to identify a counter. Since this needs to be bound to a specific object (with object id), introduce a new object to represent the relationship.

JaiOCP (approved):
Looks Good

inc/saitam.h
```diff
@@ -2074,6 +2122,118 @@ typedef sai_status_t (*sai_set_tam_event_attribute_fn)(
...
+     * @brief Start of Attributes
+     */
+    SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_START,
+
+    /**
+     * @brief TAM telemetry type object
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_TAM
```
j-bos, line 2140: Offline comment received, object-type needs to be TEL_TYPE rather than TAM

inc/saitam.h
```diff
@@ -1239,7 +1277,7 @@ typedef enum _sai_tam_report_attr_t
     SAI_TAM_REPORT_ATTR_REPORT_MODE,
 
     /**
-     * @brief Report Interval in micro seconds
```
kcudnik, line 1242: why remove unit ?

j-bos: The new units attribute means the units will not necessarily be microseconds, though they are microseconds by default, for backward-compatibility. The user may change it, e.g. to nanoseconds (usually hundreds of nanoseconds) for a specific use-case.

kcudnik: woot ? how come parameter can chance type of units ? this seems like a wrong idea, please elaborate more

j-bos: Hi, this is modelled after: SAI_TAM_TELEMETRY_ATTR_TAM_REPORTING_UNIT and SAI_TAM_TELEMETRY_ATTR_REPORTING_INTERVAL. That one uses seconds/minutes/days.

The requirement is to support sub-microsecond interval for this attribute. By defaulting to microsecond units, but allowing the user to change the units to nanoseconds, it can allow configuring for reduced interval while maintaining backward compatibility.

j-bos: @kcudnik Is it okay? If not, any suggestion? Just simply changing the units to be in nanoseconds would be the other approach but seems like a landmine if a user or implementation misses the update.

rlhui:
@rck-innovium would you like to approve this or any concern/feedback?

kcudnik (approved):

doc/TAM/Granular-Counter-Subscription.md
```diff
@@ -0,0 +1,208 @@
...
+    attr_count,
+    sai_attr_list);
+
+// Example: Attach the TAM to a queue
+
+sai_attr.id = SAI_QUEUE_ATTR_TAM;
+sai_attr.value.oid = sai_tam_obj;
+
+sai_set_queue_attribute_fn(
+    sai_queue,
```
rck-innovium, line 188: Typo? Instead of sai\_queue, it should be sai\_queue\_obj used as SAI\_TAM\_COUNTER\_SUBSCRIPTION\_ATTR\_OBJECT\_ID earlier. 

If yes, this is a redundancy in this new API model.

rck-innovium:
> @rck-innovium would you like to approve this or any concern/feedback?

@rlhui  @j-bos 

The current spec has changed quiet a bit from what was last discussed in the community. The comment about mapping the counter id in the report to the SAI OID has been addressed using the labels. 

However, the other changes including the new API needs more discussion.  I believe that the example workflow may have some typos and/or missing attributes. I want to ensure that there is no redundancy in the proposed API model. 

Typos and/or missing attributes in the given workflow:

*   SAI\_TAM\_TEL\_TYPE\_ATTR\_COUNTER\_SUBSCRIPTION\_LIST is missing in the example workflow. Do we construct a TAM object that references the TAM counter subscription object?
*   SAI\_TAM\_COUNTER\_SUBSCRIPTION\_ATTR\_OBJECT\_ID   points to the target queue\_obj\_id. And later, we bind the TAM object to the queue\_obj\_id using SAI\_QUEUE\_ATTR\_TAM.  The example uses different variable names for the queue OID. Is this a typo? If yes, we should avoid this bidirectional reference?

doc/TAM/Granular-Counter-Subscription.md
```diff
@@ -0,0 +1,208 @@
...
+   attr_count,
+   sai_attr_list);
+
+// Example: Create telemetry type object to collect object stats
+sai_attr_list[0].id = SAI_TAM_TEL_TYPE_ATTR_TAM_TELEMETRY_TYPE;
+sai_attr_list[0].value.u32 = SAI_TAM_TELEMETRY_TYPE_OBJECT_STAT;
+
+sai_attr_list[1].id = SAI_TAM_TEL_TYPE_ATTR_REPORT_ID;
+sai_attr_list[1].value.oid = sai_tam_report_obj;
+
```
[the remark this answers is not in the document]
j-bos: The subscription list is a read-only attribute. It is enforced by the SAI metadata library, that such a list attribute must exist when defining objects as a binding.

The semantics are similar to SAI_NEXT_HOP_GROUP_ATTR_NEXT_HOP_MEMBER_LIST for ECMP.

doc/TAM/Granular-Counter-Subscription.md
j-bos: Yes, it is a typo (more like a missed edit when adding _obj to follow the style of other examples). 

Agree it is a redundancy. I also considered the alternative of just using the TAM attachment point at the switch as an on/off for a TAM defined for this kind of counter collection. What are your thoughts on that? 

In particular, it might increase extensibility to other objects which do not currently have a defined TAM bind-point (and be a bit easier to use since the telemetry can be enabled/disabled all-at-once if desired (rather than object-by-object)

j-bos: @JaiOCP , I would like to get your thoughts on this as well. Whether we can/should just use an attachment to switch for this kind of telemetry.

inc/saitam.h
```diff
@@ -891,7 +891,14 @@ typedef enum _sai_tam_telemetry_type_t
      * @brief INT TAM
      * All the data relevant on a per packet basis
      */
-    SAI_TAM_TELEMETRY_TYPE_INT
+    SAI_TAM_TELEMETRY_TYPE_INT,
+
+    /**
+     * @brief Data based on counter subscriptions
+     * Collect statistics for specific counters, using
```
kcudnik, line 898: please add empty line before  this

inc/saitam.h
kcudnik: thanks i get it now

kcudnik (approved):

kcudnik:
please fix build errors

j-bos:
> please fix build errors

Thanks, looks related to the git history scanning, will squash the commits.

inc/saitam.h
```diff
@@ -2079,6 +2128,118 @@ typedef sai_status_t (*sai_set_tam_event_attribute_fn)(
...
+
+/**
+ * @brief Set value for a specified counter subscription object attribute
+ *
+ * @param[in] tam_counter_subscription_id Counter Subscription object id
+ * @param[in] attr Attribute
+ *
+ * @return #SAI_STATUS_SUCCESS on success, failure status code on error
+ */
+typedef sai_status_t (*sai_set_tam_counter_subscription_attribute_fn)(
```
kcudnik, line 2239: Delare set function before gry funcion

j-bos: Done

kcudnik (approved):

inc/saitam.h
```diff
@@ -2079,6 +2128,118 @@ typedef sai_status_t (*sai_set_tam_event_attribute_fn)(
...
+     * @objects SAI_OBJECT_TYPE_BUFFER_POOL, SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP, SAI_OBJECT_TYPE_PORT, SAI_OBJECT_TYPE_QUEUE
+     */
+    SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_OBJECT_ID,
+
+    /**
+     * @brief Subscribed stat enum
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
```
srikrishnagopu, line 2164: These stat ID correspond to certain objects which has stat_t defined. Would it be also possible to report based on the COUNTER_ID object itself ? 

For eg, to report a route or a trap hit counters - https://github.com/opencomputeproject/SAI/blob/master/inc/sairoute.h#L129

inc/saitam.h
```diff
@@ -2079,6 +2128,118 @@ typedef sai_status_t (*sai_set_tam_event_attribute_fn)(
...
+    SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_STAT_ID,
```
srikrishnagopu, line 2165: Any reason behind why this is not a list of STAT IDs and its expected to create one TAM subscription object per object ?

j-bos: Yes, to allow for a user label to be assigned per counter. Very early we thought to have a structure (possibly in a list) rather than an object per monitored counter. However, it is disallowed to reference object-ids in a structure. That is enforced by the SAI metadata library. It makes sense from an object and dependency-management perspective.

This is like the next-hop group member object as far as the object model goes.

srikrishnagopu (approved):

j-bos:
> * SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_OBJECT_ID   points to the target queue_obj_id. And later, we bind the TAM object to the queue_obj_id using SAI_QUEUE_ATTR_TAM.  The example uses different variable names for the queue OID. Is this a typo? If yes, we should avoid this bidirectional reference?

Removed the bidirectional reference as discussed offline. Documentation now indicates binding to the switch is used to enable/disable the feature in a simpler way after the subscriptions are created.

rck-innovium (approved):

PR #1674, opened 2022-12-04
[SAI-PTF][1.10]Fix issue when transfer make parameter GEN_SAIRPC_OPTS

richardyu-ms:
Fix issue when transfer make param GEN_SAIRPC_OPTS with space in it

when transfer GEN_SAIRPC_OPTS with multi parameters in it and use the space to split it, there is a issue for translate from build system to actual make system.
```
--adapter_logger --skip_error=-2
```

add quota when transfer the parameter

Test Done:
Verified in local container
Verified in build system

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ test:
 	make -C test
 
 saithrift-build:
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

Issue #1604, opened 2022-09-20
[SAI-PTF] Need to divide the test case according to test objective


Gfrom2016:
# Summary
Currently, there are multiple test cases in a class. During the testing process, multiple small test cases in a class are likely to affect each other and cause the test to fail. 

## Details
Take FdbLearnTest class for example, there a two subtest: dynamicMacLearnTest() and macLearnErrorTest(). 
```python
class FdbLearnTest(SaiHelper):
    '''
    Verify different cases of FDB learning
    '''
...
    def runTest(self):
        self.dynamicMacLearnTest()
        self.macLearnErrorTest()
```
## Error message
When we run saifdb.FdbLearnTest, it will run two subtests in turn. We ran into the following error on some platforms.
![image](https://user-images.githubusercontent.com/19384917/191159029-21e20eb9-5ca8-42bb-b9e6-f5864594a2ab.png)

### Another example
Take sairoute.L3RouteTest() as an example. Currently, all 8 subtests are run in sequence at one time. But it failed at the 6th test *sviNeighborTest* which stopped the *cpuForwardTest* and *routeNbrColisionTest* from running.

https://github.com/opencomputeproject/SAI/blob/1b268a820ddb29341f71d1b2f4d6a2444f965bc3/ptf/sairoute.py#L28

```
class L3RouteTest(SaiHelper):
    '''
    Route test class
    '''
    def runTest(self):
        self.multipleRoutesTest()
        self.dropRouteTest()
        self.routeUpdateTest()
        self.routeIngressRifTest()
        self.emptyECMPGroupTest()
        self.sviNeighborTest() <-- failed
        self.cpuForwardTest() <-- not running with result unknown
        self.routeNbrColisionTest() <-- not running with result unknown
```

![image](https://user-images.githubusercontent.com/19384917/195055132-39a91b3e-5b83-4c7c-ba52-ffcaa37e71de.png)

But if we run these test cases separately, *cpuForwardTest* and *routeNbrColisionTest* will all pass like below.
![image](https://user-images.githubusercontent.com/19384917/195055349-71ecab97-7f4f-43e8-b64a-61080e3de95b.png)

![image](https://user-images.githubusercontent.com/19384917/195058275-c0044f99-19e9-4a22-9516-49fd34d79e23.png)

# Proposal
[...]

PR #1677, opened 2022-12-04
[master][SAI-PTF]Split L3 subtests into individual tests

ravi861:
This PR addresses issues #1604 and #1599 

This PR updates following objects to address above issues

sairoute
saineighbor
sainexthop
sainexthopgroup

### DIFF
--- a/ptf/sai_base_test.py
+++ b/ptf/sai_base_test.py
@@ -305,6 +305,14 @@ class SaiHelperBase(ThriftInterfaceDataPlane):
         self.assertEqual(self.status(), SAI_STATUS_SUCCESS)
 
 
+    def warm_start_switch(self):
+        """
+        Warm start switch.
+        """
+        self.switch_id = sai_thrift_create_switch(
+        self.client, init_switch=True, warm_recover=True)
+        self.assertEqual(self.status(), SAI_STATUS_SUCCESS)
+
     def setUp(self):
         super(SaiHelperBase, self).setUp()
         self.set_logger_name()
--- a/ptf/sai_base_test.py
+++ b/ptf/sai_base_test.py
@@ -52,21 +52,109 @@ platform_map = {'broadcom': 'brcm', 'barefoot': 'bfn',
 class ThriftInterface(BaseTest):
     """
     Get and format a port map, retrieve test params, and create an RPC client
+
+    Have the following class attributes:
+        port_map_loaded: If the Port map loaded when Test init
+        transport: Thrift socket object
+        test_reboot_mode: reboot mode, which will be read from system env
+        test_reboot_stage: reboot stage, which will be read from system env
+        test_params: All the values passed via test-params if present
+        interface_to_front_mapping: Config from port_map_file for the interface (local) to front(PTF) mapping
+        protocol: Thrift protocol object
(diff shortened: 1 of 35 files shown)


ptf/sairoute.py
```diff
@@ -1050,10 +1089,20 @@ def gleanEndForwardTest(self):
             sai_thrift_remove_neighbor_entry(self.client, nbr_entry2)
```
richardyu-ms, line 1089: local variable nbr_entry1 referenced before assignment

ptf/sainexthop.py
```diff
@@ -677,22 +721,19 @@ def setUp(self):
...
             encap_ttl_mode=self.ttl_mode,
             decap_ttl_mode=self.ttl_mode,
             underlay_interface=self.urif_lb)
 
-    def runTest(self):
-        self.tunnelVniTest()
-
     def tearDown(self):
         sai_thrift_remove_tunnel(self.client, self.tunnel)
         sai_thrift_remove_tunnel_map_entry(self.client,
```
richardyu-ms, line 739: self.decap_tunnel_map_entry is not an integer

ptf/sairoute.py
richardyu-ms: fixed in https://github.com/ravi861/SAI/pull/3 and https://github.com/ravi861/SAI/pull/2

ptf/sainexthop.py
richardyu-ms: fixed in https://github.com/ravi861/SAI/pull/3 and https://github.com/ravi861/SAI/pull/2

richardyu-ms (approved):

ptf/sai_base_test.py
```diff
@@ -203,9 +290,38 @@ def get_active_port_list(self):
...
+            port_up = True
+            if port_attr['oper_status'] != SAI_PORT_OPER_STATUS_UP:
+                port_up = False
+                for num_of_tries in range(retries):
+                    port_attr = sai_thrift_get_port_attribute(
+                        self.client, oid, oper_status=True)
+                    if port_attr['oper_status'] == SAI_PORT_OPER_STATUS_UP:
+                        port_up = True
+                        break
+                    time.sleep(5)
```
kcudnik, line 318: @ravi861 please explain this sleeps if they are needed, they slow down out tests, waiting and retrying all time

ravi861: Although the changes came in via this PR, this specific timeout was initiated by Richard during his testing with BRCM devices as the comment says above.
Typically, the test would not have to retry so many times if the links come up fast enough. An alternative to hard-coding 5 sec is to pass in a test param and make it configurable.

kcudnik: in the DASH repo scenario where those tests are played, they are played gains p4/bmv2 switch, and those ports are never get up, at least what i observed notifications are not supported yet, so it waits 5x10 - 50 seconds to retry, and this is not the only tests, there are couple of them

kcudnik: you can see the retrying here: https://github.com/sonic-net/DASH/actions/runs/6053205361/job/16428197918?pr=432#step:17:80

PR #1449, opened 2022-03-22
[SAI-PTF]Add SAI-PTF Warm reboot design doc

richardyu-ms:
Add design doc for the SAI-PTF warm reboot.

It contains the introduction about the warm reboot in SAI-PTF, and
provide the sample code, command line.

Signed-off-by: richardyu-ms <richard.yu@microsoft.com>

PR #1702, opened 2022-12-27
Add warmboot design doc

x-junyi:
Why
Design and use instructions for the newly added feature that tests in the warm reboot scenario.

How
skipped

Test
skipped

Signed-off-by: ms-junyi <t-junyixiao@microsoft.com>


richardyu-ms:
contained in https://github.com/opencomputeproject/SAI/pull/1449

PR #1718, opened 2023-01-11
Add ACL table attribute to specify Dst IPv6 valid bits

msomasundaran:
This diff adds an ACL table parameter to specify the valid bits of Src/Dst IPv6 to use in ACL table creation. There are two main uses cases.

1. For TCAM ACLs, this attribute allows creation of table with shorter keys as users can specify a lower number of bits instead of whole 128 bits
2. For exact match lookup (ref PR 1717), masks cannot be specified at ACL entry level since all entries in a table need to have the same mask. The new attribute allows users to specify the mask for exact match tables.    

Signed-off-by: Midhun Somasundaran <msomasundaran@meta.com>, Shrikrishna Khare <skhare@meta.com>

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -1429,6 +1429,44 @@ typedef enum _sai_acl_table_attr_t
     SAI_ACL_TABLE_ATTR_AVAILABLE_ACL_COUNTER,
 
     /**
+     * @brief Start of Table Match valid bits
+     *
+     * The valid bits specify the bits of match field that should be
+     * included in the lookup key. If a match field does not have
+     * valid bits specified, all bits in the field are valid.
+     *
+     * For tables implemented using Exact Match, there is no further
+     * key masking supported in table entries. The mask for such
+     * entries needs to be set to all 1s.
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_START = 0x00003000,
+
+    /**
+     * @brief Src IPv6 Valid bits
+     *
+     * @type sai_acl_field_data_mask_t sai_ip6_t
+     * @flags CREATE_ONLY
+     * @default ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff
+     * @validonly SAI_ACL_TABLE_ATTR_FIELD_SRC_IPV6 == true
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_SRC_IPV6 = SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_START,
+
+    /**
+     * @brief Dst IPv6 Valid bits
+     *
+     * @type sai_acl_field_data_mask_t sai_ip6_t
+     * @flags CREATE_ONLY
+     * @default ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff
+     * @validonly SAI_ACL_TABLE_ATTR_FIELD_DST_IPV6 == true
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_DST_IPV6 = SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_START + 1,
+
+    /**
+     * @brief End of Table Match Field Mask
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_END = SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_DST_IPV6,
+
+    /**
      * @brief End of ACL Table attributes
      */
     SAI_ACL_TABLE_ATTR_END,
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -1320,6 +1320,12 @@ typedef union _sai_attribute_value_t
      */
     sai_acl_action_data_t aclaction;
 
+    /**
+     * @passparam meta
+     * @validonly meta->isaclmask == true
+     */
+    sai_acl_field_data_mask_t aclmask;
+
     /** @validonly meta->attrvaluetype == SAI_ATTR_VALUE_TYPE_ACL_CAPABILITY */
     sai_acl_capability_t aclcapability;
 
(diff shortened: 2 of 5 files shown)


inc/saiacl.h
```diff
@@ -1419,6 +1419,26 @@ typedef enum _sai_acl_table_attr_t
...
+    SAI_ACL_TABLE_ATTR_FIELD_MASK_START = 0x00003000,
+
+    /**
+     * @brief Dst IPv6 Valid bits
+     *
+     * @type sai_acl_field_data_mask_t sai_ip6_t
+     * @flags CREATE_ONLY
+     * @default ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff
+     * @validonly SAI_ACL_TABLE_ATTR_FIELD_DST_IPV6 == true
+     */
```
rck-innovium, line 1434: We should rename this to mean they are valid bits.

rck-innovium (changes requested):
We should add comments in the header file to specify that in an ACL table of type Exact match, the ACL match fields used in an ACL entry should have all the  bits in the mask set to 1.

rck-innovium:
One way to expose device capability that only contiguous bits can be specified in table level valid bits (SAI\_ACL\_TABLE\_ATTR\_FIELD\_VALID\_BITS\_DST\_IPV6) is to have a new field in sai\_acl\_capability\_t like:

1.  bool is\_non\_contiguous\_exact\_match\_supported

rck-innovium:
One way to expose device capability that only contiguous bits can be specified in table level valid bits (SAI\_ACL\_TABLE\_ATTR\_FIELD\_VALID\_BITS\_DST\_IPV6) is to have a new field in sai\_acl\_capability\_t like:

1.  bool is\_non\_contiguous\_exact\_match\_supported

inc/saiacl.h
msomasundaran: renamed to valid bits

msomasundaran:
> One way to expose device capability that only contiguous bits can be specified in table level valid bits (SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_DST_IPV6) is to have a new field in sai_acl_capability_t like:
> 
> 1. bool is_non_contiguous_exact_match_supported

I am not sure whether there is a use case currently for supporting non contiguous valid bits. Let me know if you prefer to add it regardless

msomasundaran:
> > One way to expose device capability that only contiguous bits can be specified in table level valid bits (SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_DST_IPV6) is to have a new field in sai_acl_capability_t like:
> > 
> > 1. bool is_non_contiguous_exact_match_supported
> 
> I am not sure whether there is a use case currently for supporting non contiguous valid bits. Let me know if you prefer to add it regardless

Added the capability discovery for non contiguous bit mapping in PR1730 which is the one that added EM

shri-khare (approved):

inc/saiacl.h
```diff
@@ -1419,6 +1419,44 @@ typedef enum _sai_acl_table_attr_t
...
+    SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_START = 0x00003000,
+
+    /**
+     * @brief Src IPv6 Valid bits
+     *
+     * @type sai_acl_field_data_mask_t sai_ip6_t
+     * @flags CREATE_ONLY
+     * @default ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff
+     * @validonly SAI_ACL_TABLE_ATTR_FIELD_SRC_IPV6 == true
```
JaiOCP, line 1441: I believe this is only for all match type
SAI_ACL_MATCH_TYPE_TERNARY_AND_EXACT
and
SAI_ACL_MATCH_TYPE_EXACT.

Add a valid only condition

msomasundaran: > I believe this is only for all match type SAI_ACL_MATCH_TYPE_TERNARY_AND_EXACT and SAI_ACL_MATCH_TYPE_EXACT.
> 
> Add a valid only condition

This is valid for SAI_ACL_MATCH_TYPE_TERNARY also as it allows us to reduce the tcam key width

rck-innovium (approved):

JaiOCP (approved):

ashutosh-agrawal (approved):

kcudnik:
i just noticed taht you introduce new type but you are reusing existing field mask, for aclmask you should add newattr valuetype,since this type is not valid acl_field_data, since acl field data contains enable data and mask fields

are you planning do add other fields similar to field /reuse from acl_field data or you only needed mask ?

msomasundaran:
> i just noticed taht you introduce new type but you are reusing existing field mask, for aclmask you should add newattr valuetype,since this type is not valid acl_field_data, since acl field data contains enable data and mask fields
> 
> are you planning do add other fields similar to field /reuse from acl_field data or you only needed mask ?

For the table level mask, there will not be any value to specify. Hence it is of type sai_acl_field_data_mask_t. For exact match tables, the entries will have masks set to all 1s since the mask is associated with table and not entry. This is documented in https://github.com/opencomputeproject/SAI/blob/master/doc/ACL/SAI-Proposal-Exact-Match.md Section "Acl Enhancements", item 4.

kcudnik:
I'd the fields are always ones then this field is not needed at all, again asking whether this type will always be ipv6 or other types unit OID etc ? Since you need to define new atte value type for each one

msomasundaran:
> I'd the fields are always ones then this field is not needed at all, again asking whether this type will always be ipv6 or other types unit OID etc ? Since you need to define new atte value type for each one

The type is sai_acl_field_data_mask_t which is a union of all possible mask types.

https://github.com/opencomputeproject/SAI/blob/master/inc/saitypes.h#L481

So each field will use the union with corresponding member type depending on the field. I did not understand your concern

kcudnik:
Are you planning to use ech field ? Since you only used ip mask, and you added wrong type, since that attribute type is for sai ack field, which contains enable field data and mask, it's totally different than mask only, if you need specific attribute with mask only, you need separate attr type or use entire acl field type

PR #1730, opened 2023-01-31
Add ACL table attribute to identify match type

msomasundaran:
Please refer PR1717 for details

Signed-off-by: Midhun Somasundaran <msomasundaran@meta.com>, Shrikrishna Khare <skhare@meta.com>

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -1429,6 +1429,15 @@ typedef enum _sai_acl_table_attr_t
     SAI_ACL_TABLE_ATTR_AVAILABLE_ACL_COUNTER,
 
     /**
+     * @brief Match type for the table
+     *
+     * @type sai_acl_table_match_type_t
+     * @flags CREATE_ONLY
+     * @default SAI_ACL_TABLE_MATCH_TYPE_TERNARY
+     */
+    SAI_ACL_TABLE_ATTR_ACL_TABLE_MATCH_TYPE,
+
+    /**
      * @brief Start of Table Match valid bits
      *
      * The valid bits specify the bits of match field that should be
--- a/inc/saitypes.h
+++ b/inc/saitypes.h
@@ -784,6 +784,19 @@ typedef struct _sai_map_list_t
 } sai_map_list_t;
 
 /**
+ * @brief Attribute data for SAI_ACL_TABLE_ATTR_SUPPORTED_MATCH_TYPE
+ */
+typedef enum _sai_acl_table_supported_match_type_t
+{
+    /** Ternary */
+    SAI_ACL_TABLE_SUPPORTED_MATCH_TYPE_TERNARY,
+
+    /** Exact */
+    SAI_ACL_TABLE_SUPPORTED_MATCH_TYPE_EXACT,
+
+} sai_acl_table_supported_match_type_t;
+
+/**
  * @brief Structure for ACL attributes supported at each stage.
  * action_list alone is added now. Qualifier list can also be added
  * when needed.
@@ -807,6 +820,20 @@ typedef struct _sai_acl_capability_t
      * @passparam &sai_metadata_enum_sai_acl_action_type_t
      */
     sai_s32_list_t action_list;
+
+    /**
+     * @brief Supported match type
+     *
+     * Indicates the match type supported by ACL stage
+     */
+    sai_acl_table_supported_match_type_t supported_match_type;
+
+    /**
+     * @brief Non contiguous bit match supported
+     *
+     * Indicates whether EM can support non contiguous bits matching
+     */
+    bool is_non_contiguous_bits_exact_match_supported;
 
 } sai_acl_capability_t;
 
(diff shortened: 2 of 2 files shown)


inc/saiacl.h
```diff
@@ -354,6 +354,15 @@ typedef enum _sai_acl_table_group_attr_t
...
     SAI_ACL_TABLE_GROUP_ATTR_MEMBER_LIST,
 
+    /**
+     * @brief Match type for the table group
+     *
+     * @type sai_acl_match_type_t
+     * @flags CREATE_ONLY
+     * @default SAI_ACL_MATCH_TYPE_PRIORITY_BASED
+     */
+    SAI_ACL_TABLE_GROUP_ATTR_ACL_MATCH_TYPE,
```
rck-innovium, line 364: As discussed during the meeting, we should 

*   remove this attribute and 
*   add the switch acl table capability to specify whether the platform can support mixing Exact and Ternary match in the same ACL Table Group.

rck-innovium (changes requested):

inc/saiacl.h
msomasundaran: Suggested changes are incorporated in the latest diffs

inc/saitypes.h
```diff
@@ -855,6 +855,19 @@ typedef enum _sai_acl_bind_point_type_t
 
 } sai_acl_bind_point_type_t;
 
+/**
+ * @brief Attribute data for SAI_ACL_TABLE_ATTR_MATCH_TYPE
+ */
+typedef enum _sai_acl_match_type_t
+{
+    /** Match Type Priority Based */
+    SAI_ACL_MATCH_TYPE_PRIORITY_BASED,
```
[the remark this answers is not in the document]
msomasundaran: Suggested changes are incorporated in the latest diffs

shri-khare (approved):

inc/saitypes.h
```diff
@@ -855,6 +885,19 @@ typedef enum _sai_acl_bind_point_type_t
...
+ * @brief Attribute data for SAI_ACL_TABLE_ATTR_MATCH_TYPE
+ */
+typedef enum _sai_acl_match_type_t
+{
+    /** Match Type Ternary */
+    SAI_ACL_MATCH_TYPE_TERNARY,
+
+    /** Match Type Exact */
+    SAI_ACL_MATCH_TYPE_EXACT
+
```
JaiOCP, line 898: Missing following type:   
 /** Ternary and Exact */
     SAI_ACL_MATCH_TYPE_TERNARY_AND_EXACT,

msomasundaran: thanks for pointing out. Added

ashutosh-agrawal (approved):

inc/saitypes.h
rck-innovium: What is the difference between TERNARY\_AND\_EXACT and TERNARY? A Ternary type can behave as both exact and ternary. 

What kind of h/w resource would return their type as TERNARY\_AND\_EXACT instead of TERNARY? Also the comments are not adding any additional info.

msomasundaran: > What is the difference between TERNARY_AND_EXACT and TERNARY? A Ternary type can behave as both exact and ternary. 
> 
> What kind of h/w resource would return their type as TERNARY_AND_EXACT instead of TERNARY? Also the comments are not adding any additional info.

For match type TERNARY_AND_EXACT, the implementation will host some entries in TCAM and some in EM. For example matches that include /120 prefixes will be placed in TCAM while all /128s will be placed in EM. If there is a match in both TCAM and EM, then hardware resolves it by using the EM result. This allows supporting increased scale.

inc/saitypes.h
```diff
@@ -855,6 +885,22 @@ typedef enum _sai_acl_bind_point_type_t
 
 } sai_acl_bind_point_type_t;
 
+/**
+ * @brief Attribute data for SAI_ACL_TABLE_ATTR_MATCH_TYPE
+ */
+typedef enum _sai_acl_match_type_t
```
rck-innovium, line 891: Can you please rename the enums as:

*   SAI\_ACL\_TABLE\_MATCH\_TYPE\_XXX

And rename the enum type as:

*   sai\_acl\_table\_match\_type\_t

inc/saitypes.h
rck-innovium: As discussed, we need to define the semantics of operations like ACL entry priority, available ACL table entries etc.

inc/saitypes.h
msomasundaran: renamed

inc/saitypes.h
msomasundaran: removed type SAI_ACL_MATCH_TYPE_TERNARY_AND_EXACT based on our discussion. it can be added back when needed with more details on semantics

rck-innovium (approved):

JaiOCP (approved):

rlhui:
please resolve conflicts. Thanks.

rlhui:
Can we please not use "force-push" as history of commits/diff's will be lost.

msomasundaran:
> please resolve conflicts. Thanks.

Resolved conflicts. Sorry about the force push. Will keep it in mind for future.

PR #1741, opened 2023-02-08
[master]Enhance the check enum lock script

richardyu-ms:
Why
fix for
```
 git --work-tree=temp/ checkout 
```
after checkout from other branch, data will be left in git

then in when running command from sonic-buildimage/rules/sairedis.dep, it will report
```
[sonic-buildimage/rules/sairedis.dep](make: *** No rule to make target 'src/sonic-sairedis/SAI/inc/saigenericprogrammable.h', needed by 'target/debs/buster/libsairedis_1.0.0_amd64.deb.smdep'.)
```

How
clean up the git env

Verify
run script
```
checkenumlock.sh
```

### DIFF
--- a/meta/checkenumlock.sh
+++ b/meta/checkenumlock.sh
@@ -27,17 +27,12 @@ set -e
 
 rm -rf temp
 
-mkdir temp
+sairepo=`git remote get-url origin`
 
-git --work-tree=temp/ checkout origin/master inc
-git --work-tree=temp/ checkout origin/master experimental
+git clone $sairepo temp
 
 echo "Checking for possible enum values shift (current branch vs origin/master) ..."
 
 ./checkheaders.pl -s ../inc/ temp/inc/
 
 rm -rf temp
-# clean up the git changes as well
-# workaround fix for git --work-tree=temp/ checkout ...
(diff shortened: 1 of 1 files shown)


meta/checkenumlock.sh
```diff
@@ -37,3 +37,7 @@ echo "Checking for possible enum values shift (current branch vs origin/master)
 ./checkheaders.pl -s ../inc/ temp/inc/
 
 rm -rf temp
+# clean up the git changes as well
+# workaround fix for git --work-tree=temp/ checkout ...
+# after checkout from other branch, data will be left in git
+git stash
```
xumia, line 43: Looks like we should not use "git --work-tree=temp/ checkout origin/master inc", it has impact on the working directory. Maybe we cannot assumption the source folder is clean.

richardyu-ms: yes, so this workaround will try to stash all the changes, that includes the intended changes. for example, if someone changed some files in the inc and experimental folder but not committed, then by this method, there sill be a chance to recover them.

richardyu-ms: enhanced the approach

xumia (approved):

PR #1744, opened 2023-02-09
SRv6 VPN enhancements

rameshms-work:
List of SRv6 enhancements added through this PR.

Disaggregation of VPN SID from underlay SID List by use of tunnel map.
Selection of underlay nexthop to use for a given SID list
L3VPN only solution, without the need to use SID List.
Allow prefix based derivation of VPN SID.
Ability for multiple prefixes to same destination but different VPN SID to share same NH/NHG using Prefix aggregation id.
Statistics collection on SID List


doc/SAI-IPv6-Segment-Routing-VPN.md
```diff
@@ -0,0 +1,674 @@
...
+
+The diagram below shows the objects and relationship. The diagram only shows relevant fields in the different tables.
+
+NOTE: Below object types have been combined to keep the diagram simple.
+* Next hop group and member list
+* Tunnel map and Tunnel map entry table
+
+New objects are highlighed in RED.
+
+![SAI SRv6 Encap model](figures/sai_srv6_encap_model.png "Figure 1: SRv6 VPN Encap model ")
```
svshah-intel, line 173: SIDLIST OID field from SRv6 NH object should be pointing to an object. why is it pointing to content of an object?

rameshsanth: Thank you for your comments. I've updated the diagram. Hopefully its clearer now.

rameshms-work: updated the diagram

reshmaintel (approved):

j-bos (approved):

kcudnik (approved):

PR #1759, opened 2023-02-23
Add SRV6 encap support to BFD 

mthatty:
This PR enables adding SRV6 Encapsulation to BFD.

This is accomplished by extending sai_bfd_encapsulation_type_t - adding the enum SAI_BFD_ENCAPSULATION_TYPE_SRV6.

A new attribute SAI_BFD_SESSION_ATTR_SRV6_SIDLIST_ID is also introduced, which refers to a SAI object of type SAI_OBJECT_TYPE_SRV6_SIDLIST.

conditions for SAI_BFD_SESSION_ATTR_TUNNEL_TTL, SAI_BFD_SESSION_ATTR_TUNNEL_TOS and SAI_BFD_SESSION_ATTR_TUNNEL_SRC_IP_ADDRESS were updated to include SAI_BFD_ENCAPSULATION_TYPE_SRV6.

SAI_BFD_SESSION_ATTR_TUNNEL_DST_IP_ADDRESS was intentionally excluded as these details are defined in SAI_BFD_SESSION_ATTR_SRV6_SIDLIST_ID

### DIFF
--- a/inc/saibfd.h
+++ b/inc/saibfd.h
@@ -89,7 +89,7 @@ typedef enum _sai_bfd_encapsulation_type_t
     SAI_BFD_ENCAPSULATION_TYPE_NONE,
 
     /**
-     * @brief SRV6 Tunnel Encapsulation | L2 Ethernet header | SRV6 Header | IP header | Original BFD packet
+     * @brief SRV6 Tunnel Encapsulation | L2 Ethernet header | IPV6 Header | SRV6 Header | Inner IP header | Original BFD packet
      */
     SAI_BFD_ENCAPSULATION_TYPE_SRV6,
 
--- a/inc/saibfd.h
+++ b/inc/saibfd.h
@@ -88,6 +88,11 @@ typedef enum _sai_bfd_encapsulation_type_t
      */
     SAI_BFD_ENCAPSULATION_TYPE_NONE,
 
+    /**
+     * @brief SRV6 Tunnel Encapsulation | L2 Ethernet header | SRV6 Header | IP header | Original BFD packet
+     */
+    SAI_BFD_ENCAPSULATION_TYPE_SRV6,
+
 } sai_bfd_encapsulation_type_t;
 
 /**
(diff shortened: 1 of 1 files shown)


inc/saibfd.h
```diff
@@ -88,6 +88,11 @@ typedef enum _sai_bfd_encapsulation_type_t
      */
     SAI_BFD_ENCAPSULATION_TYPE_NONE,
 
+    /**
+     * @brief SRV6 Tunnel Encapsulation | L2 Ethernet header | SRV6 Header | IP header | Original BFD packet
```
ashutosh-agrawal, line 92: Shouldn't the IP (ipv6) header come before SRV6 header in the comment?

ashutosh-agrawal:
I didn't understand the comment (intentionally excluded) on SAI_BFD_SESSION_ATTR_TUNNEL_DST_IP_ADDRESS. The condition attached to this attributed needs to be updated.

inc/saibfd.h
mthatty: Updated

mthatty:
> I didn't understand the comment (intentionally excluded) on SAI_BFD_SESSION_ATTR_TUNNEL_DST_IP_ADDRESS. The condition attached to this attributed needs to be updated.

May have added more confusion with the original comment. To rephrase - SAI_BFD_SESSION_ATTR_TUNNEL_DST_IP_ADDRESS still only applies to IPinIP not this new attribute. For this reason I left the condition unchanged.

ashutosh-agrawal:
> > I didn't understand the comment (intentionally excluded) on SAI_BFD_SESSION_ATTR_TUNNEL_DST_IP_ADDRESS. The condition attached to this attributed needs to be updated.
> 
> May have added more confusion with the original comment. To rephrase - SAI_BFD_SESSION_ATTR_TUNNEL_DST_IP_ADDRESS still only applies to IPinIP not this new attribute. For this reason I left the condition unchanged.

Thanks. I still don't get why the original attribute is conditional for IP-in-IP and not IP-GRE tunnels.

rameshms-work:
> > > I didn't understand the comment (intentionally excluded) on SAI_BFD_SESSION_ATTR_TUNNEL_DST_IP_ADDRESS. The condition attached to this attributed needs to be updated.
> > 
> > 
> > May have added more confusion with the original comment. To rephrase - SAI_BFD_SESSION_ATTR_TUNNEL_DST_IP_ADDRESS still only applies to IPinIP not this new attribute. For this reason I left the condition unchanged.
> 
> Thanks. I still don't get why the original attribute is conditional for IP-in-IP and not IP-GRE tunnels.

I agree. This seems to be a miss from the original PR that introduced SAI BFD specification.
It would be better to track this miss in a different PR .

mthatty:
Hi @ashutosh-agrawal, is there anything else we need to address in this PR? or are you OK to approve? 

I would like to leave the condition for IP-GRE tunnels untouched as it does not directly related to this PR, and can be taken care of in a followup commit.

ashutosh-agrawal (approved):

j-bos (approved):

PR #1743, opened 2023-02-09
Port based hash algorithm and offset.

erohsik:
Provide the ability to set Hash algorithm, offset for a port that
takes precedence over the corresponding switch wide attribute.

Signed-off-by: Kishore Gummadidala <kishorg@google.com>

PR #1775, opened 2023-03-02
Provide the ability to set ECMP Hash algorithm, offset, seed for a port that

erohsik:
takes precedence over the corresponding switch wide attribute.


erohsik:
@JaiOCP @marian-pritsak @rlhui This is essentially #1743 plus a port attribute to set seed.

JaiOCP (approved):
Looks Good

PR #1790, opened 2023-04-26
Add DASH cluster tag API

marian-pritsak:


kcudnik (approved):

experimental/saiexperimentaldashacl.h
```diff
@@ -110,6 +110,38 @@ typedef enum _sai_dash_acl_rule_attr_t
...
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_DASH_ACL_RULE_ATTR_DST_TAG_MASK,
+
+    /**
+     * @brief Ternary matched key src_tag
+     *
+     * @type sai_uint32_t
```
prsunny, line 132: This should be a new type of bitvector of size, say 512 since we can have more than 32 tags.

PR #1793, opened 2023-05-02
VoQ Switch Isolation

JaiOCP:
Switch Isolate validonly condition addition for VoQ switch.

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -2852,6 +2852,7 @@ typedef enum _sai_switch_attr_t
      * true: Trigger the switch isolate process
      * false: Undo the isolation operation.
      * This attribute is for Fabric or VOQ Chassis only
+     * Setting this flag to true will stop data traffic from flowing on local fabric ports. Fabric control traffic like credit grants and control messages will still continue to flow. Setting this flag to false will (re)enable data traffic on fabric ports.
      * If this attribute is set to true, it overrides port level isolation setting.
      *
      * @type bool
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -2852,7 +2852,10 @@ typedef enum _sai_switch_attr_t
      * true: Trigger the switch isolate process
      * false: Undo the isolation operation.
      * This attribute is for Fabric or VOQ Chassis only
-     * Setting this flag to true will stop data traffic from flowing on local fabric ports. Fabric control traffic like credit grants and control messages will still continue to flow. Setting this flag to false will (re)enable data traffic on fabric ports.
+     * Setting this flag to true will stop data traffic from flowing
+     * on local fabric ports. Fabric control traffic like credit grants
+     * and control messages will still continue to flow. Setting this
+     * flag to false will (re)enable data traffic on fabric ports.
      * If this attribute is set to true, it overrides port level isolation setting.
      *
      * @type bool
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -2497,7 +2497,7 @@ typedef enum _sai_switch_attr_t
      *
      * @type sai_uint32_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
-     * @condition SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_VOQ
+     * @condition SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_VOQ or SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_FABRIC
      */
     SAI_SWITCH_ATTR_SWITCH_ID,
 
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -2851,13 +2851,13 @@ typedef enum _sai_switch_attr_t
      *
      * true: Trigger the switch isolate process
      * false: Undo the isolation operation.
-     * This attribute is for Fabric Chassis only
+     * This attribute is for Fabric or VOQ Chassis only
      * If this attribute is set to true, it overrides port level isolation setting.
      *
      * @type bool
      * @flags CREATE_AND_SET
      * @default false
-     * @validonly SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_FABRIC
+     * @validonly SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_FABRIC or SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_VOQ
      */
     SAI_SWITCH_ATTR_SWITCH_ISOLATE,
 


inc/saiswitch.h
```diff
@@ -2873,6 +2879,32 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_HOSTIF_OPER_STATUS_UPDATE_MODE,
 
+    /**
+     * @brief VOQ chassis isolation setting.
+     *
+     * This attribute is for VOQ Chassis only
+     * This attribute is for configuring threshold value of active fabric ports below which VOQ Chassis isolation mechanism will trigger. Value of 0 means this attribute is not considered by switch.
```
shri-khare, line 2886: We need to explicitly call out that SAI adapter triggers callback when the number of UP fabric ports fall below the configured threshold and SAI adapter host callback processing should isolate by setting SAI_SWITCH_ATTR_SWITCH_ISOLATE to true.

This will use sai_switch_state_change_notification_fn as notification function? In that case that needs to be captured here as well?

inc/saiswitch.h
```diff
@@ -2873,6 +2879,32 @@ typedef enum _sai_switch_attr_t
...
+     * @default 0
+     * @validonly SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_VOQ
+     */
+    SAI_SWITCH_ATTR_FABRIC_PORT_THRESHOLD_LOW,
+
+    /**
+     * @brief VOQ chassis isolation setting.
+     *
+     * This attribute is for VOQ Chassis only
+     * This attribute is for configuring threshold value of active fabric ports above which VOQ Chassis isolation mechanism will be aborted or VOQ chassis will be eligible to become part of dis-aggregated chassis. Value of 0 means this attribute is not considered by switch.
```
shri-khare, line 2899: Similar to above, here, we should call out that SAI adapter triggers callback when then number of UP fabric ports exceeds this threshold and SAI adapter host callback processing should unisolate by setting to false.

This will use sai_switch_state_change_notification_fn as notification function? In that case that needs to be captured here as well?

inc/saiswitch.h
```diff
@@ -2873,6 +2879,32 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     * @validonly SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_VOQ
+     */
+    SAI_SWITCH_ATTR_FABRIC_PORT_THRESHOLD_HIGH,
```
shri-khare, line 2906: Would it be easier to follow if below are renamed as:

SAI_SWITCH_ATTR_FABRIC_PORT_THRESHOLD_LOW => SAI_SWITCH_ATTR_FABRIC_PORT_ISOLATE_THRESH
SAI_SWITCH_ATTR_FABRIC_PORT_THRESHOLD_HIGH => SAI_SWITCH_ATTR_FABRIC_PORT_UNISOLATE_THRESH

If we don't want to tie the action (isolate/unisolate) to the threshold configuration, then we could consider naming it as SAI_SWITCH_ATTR_FABRIC_PORT_UP_THRESHOLD_LOW, SAI_SWITCH_ATTR_FABRIC_PORT_UP_THRESHOLD_HIGH i.e. add 'UP' since this is counting how many FABRIC ports are UP?

inc/saiswitch.h
```diff
@@ -60,6 +60,12 @@ typedef enum _sai_switch_oper_status_t
     /** Switch encountered a fatal error */
     SAI_SWITCH_OPER_STATUS_FAILED,
 
+    /** VOQ Switch in critical state with available fabric ports below configured configured threshold */
+    SAI_SWITCH_OPER_STATUS_VOQ_FABRIC_PORT_THRESHOLD_LOW,
```
shri-khare, line 64: If we choose to rename SAI_SWITCH_ATTR_FABRIC_PORT_THRESHOLD_LOW, _HIGH (see comments below), then we could similarly rename this as well.

inc/saiswitch.h
```diff
@@ -60,6 +60,12 @@ typedef enum _sai_switch_oper_status_t
...
+
+    /** VOQ Switch in recovery state with available fabric ports above configured configured threshold. Once the recovery is complete switch transitions to UP state */
+    SAI_SWITCH_OPER_STATUS_VOQ_FABRIC_PORT_THRESHOLD_HIGH,
+
 } sai_switch_oper_status_t;
```
shri-khare, line 69: In the current form, the callback is defined to be triggered  for numUpFabricPorts < THRESH_LOW or numUpFabricPorts > THRESH_HIGH, and not when numUpFabricPorts <= THRESH_LOW or numUpFabricPorts >= THRESH_HIGH.

Thus, there should not be a risk that both the  _LOW and _HIGH conditions are met simultaneously and SAI adapter needs to set both THRESHOLD_LOW and THRESHOLD_HIGH in the same callback: that won't be possible as sai_switch_oper_status_t is an enum not a bitmask. 

Thus, might be good idea for SAI_SWITCH_ATTR_FABRIC_PORT_THRESHOLD_LOW description to be more explicit about semantics being "less then" instead of "less than equal to".

shri-khare: Thinking more about it, given the PR defines the new attribute to trigger callback when "above" or "below" threshold, we are good here, pls treat this comment as resolved.

inc/saiswitch.h
```diff
@@ -60,6 +60,12 @@ typedef enum _sai_switch_oper_status_t
     /** Switch encountered a fatal error */
     SAI_SWITCH_OPER_STATUS_FAILED,
 
+    /** VOQ Switch in critical state with available fabric ports below configured configured threshold */
+    SAI_SWITCH_OPER_STATUS_VOQ_FABRIC_PORT_THRESHOLD_LOW,
+
+    /** VOQ Switch in recovery state with available fabric ports above configured configured threshold. Once the recovery is complete switch transitions to UP state */
+    SAI_SWITCH_OPER_STATUS_VOQ_FABRIC_PORT_THRESHOLD_HIGH,
```
rlhui, line 67: "fabric ports too low" is one of reasons for a switch in an error situations.  There can be other reasons, e.g. buffer or memory low, other hardware/memory errors etc.  Let's not add the error conditions here so that we can keep the switch oper status simple/clean.  thanks.

reshmaintel:
Since the usage of this attribute SAI_SWITCH_OPER_STATUS_VOQ_FABRIC_PORT_THRESHOLD_HIGH seems to indicate the fabric port stability, please consider a name like SAI_SWITCH_OPER_STATUS_VOQ_FABRIC_PORT_STABLE

inc/saiswitch.h
JaiOCP: https://github.com/opencomputeproject/SAI/pull/1793/commits/64d77e084762eeb7ee35ac1ab1641b8584104c6d

inc/saiswitch.h
JaiOCP: https://github.com/opencomputeproject/SAI/pull/1793/commits/64d77e084762eeb7ee35ac1ab1641b8584104c6d

inc/saiswitch.h
JaiOCP: Please look at the latest diffs and let me know what you think

inc/saiswitch.h
JaiOCP: https://github.com/opencomputeproject/SAI/pull/1793/commits/64d77e084762eeb7ee35ac1ab1641b8584104c6d

JaiOCP:
> Since the usage of this attribute SAI_SWITCH_OPER_STATUS_VOQ_FABRIC_PORT_THRESHOLD_HIGH seems to indicate the fabric port stability, please consider a name like SAI_SWITCH_OPER_STATUS_VOQ_FABRIC_PORT_STABLE

https://github.com/opencomputeproject/SAI/pull/1793/commits/64d77e084762eeb7ee35ac1ab1641b8584104c6d

Please look at the latest diffs.

JaiOCP:
@jasmeetbagga @shri-khare @rck-innovium @rlhui 
Please look at the latest diffs.

inc/saiswitch.h
JaiOCP: Created a new voq switch isolate state for isolation and corresponding callback. Removed the switch over state changes.

https://github.com/opencomputeproject/SAI/pull/1793/commits/0d07fb18dc37f419ec400f7383685f1cf104fe46

inc/saiswitch.h
```diff
@@ -2851,13 +2864,13 @@ typedef enum _sai_switch_attr_t
      *
      * true: Trigger the switch isolate process
      * false: Undo the isolation operation.
-     * This attribute is for Fabric Chassis only
+     * This attribute is for Fabric or VOQ Chassis only. This attribute is set as an action to the isolation status change notification callback. Setting this flag to true means that there is no data traffic flowing through the fabric ports. Fabric control traffic like credit grants and control messages will still continue to flow.
```
jasmeetbagga, line 2867: > This attribute is set as an action to the isolation status change notification callback.

That's not the only use case. Switch can be isolated for operational reasons as well. The ask from meeting on 05/04 was to define what happens in response to set/unset of this attribute. That part is covered by

"Setting this flag to true means that there is no data traffic flowing through the fabric ports. Fabric control traffic like credit grants and control messages will still continue to flow."

Consider wording it as

"Setting this flag to true will stop data traffic from flowing on local fabric ports. Fabric control traffic like credit grants and control messages will still continue to flow. Setting this flag to false will (re)enable data traffic on fabric ports."

kcudnik (approved):

inc/saiswitch.h
JaiOCP: Updated the comments

inc/saiswitch.h
```diff
@@ -2851,13 +2851,14 @@ typedef enum _sai_switch_attr_t
      *
      * true: Trigger the switch isolate process
      * false: Undo the isolation operation.
-     * This attribute is for Fabric Chassis only
+     * This attribute is for Fabric or VOQ Chassis only
+     * Setting this flag to true will stop data traffic from flowing on local fabric ports. Fabric control traffic like credit grants and control messages will still continue to flow. Setting this flag to false will (re)enable data traffic on fabric ports.
```
kcudnik, line 2855: break the line

shri-khare (approved):

kcudnik (approved):

JaiOCP:
Threshold attributes are removed as NOS can maintain the fabric port OPER_UP state and can trigger isolation sequence based on up or down fabric port thresholds. There is no need to push down this information to SAI layer and expect a callback.

@rlhui 
Same isolation sequence is also applicable to VOQ De-Agg Chassis.

jasmeetbagga (approved):

abdosi (approved):

PR #1799, opened 2023-05-19
Tunnel Term Attributes validation fixed for all MP2P and MP2MP attributes

mholankar:
For Tunnel Termination Tables -  for MP2P type, we want to be able to provide the dst IP+ mask & src IP, so made sure the
dst IP attribute relects that

Also for MP2MP type, made sure dst IP and src IP attributes are allowed for the type

    /** Tunnel termination table multi point to point entry match on dst IP+mask & src IP & tunnel type */
    SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_MP2P,

    /** Tunnel termination table multi point to multi point entry match on dst IP+mask & src IP+mask & tunnel type */
    SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_MP2MP,


mholankar:
@kcudnik @JaiOCP could you review the changes please?

kcudnik (approved):

inc/saitunnel.h
```diff
@@ -957,7 +957,6 @@ typedef enum _sai_tunnel_term_table_entry_attr_t
      *
      * @type sai_ip_address_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
-     * @condition SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TYPE == SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_P2P or SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TYPE == SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_P2MP
      */
```
JaiOCP, line 960: Not clear why the condition check is removed

kcudnik:
removing those conditions is a breaking backward compatibility change, since studently those 2 attributes will be required mandatory when creating tunnel term table entry

inc/saitunnel.h
mholankar: When condition is met, then the attribute is mandatory during create, if condition is not met, then attribute is not valid on create and meta checker will not allow it to be passed. However we do want the attribute SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_DST_IP for new tunnel types MP2P and MP2MP as well.

abdosi:
@marian-pritsak can you please review/approve this.

abdosi:
@rck-innovium please help sign-off this.

abdosi:
@mholankar can you see the comment on test failure. It seems PR checkers are passing here.

mikeberesford:
My apologies, the test that was failing is in SONiC sairedis not SAI. @kcudnik this appears to be BCM56971B0/acl_limit.rec, which is making use of the fact that the default SRC_IP mask is 0.0.0.0, so SRC_IP does not need to be specified for SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_P2MP.

Example entry:
```
2022-03-22.05:32:44.773768|c|SAI_OBJECT_TYPE_TUNNEL_TERM_TABLE_ENTRY:oid:0x2b0000000014ed|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_VR_ID=oid:0x300000000007c|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TYPE=SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_P2MP|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TUNNEL_TYPE=SAI_TUNNEL_TYPE_IPINIP|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_ACTION_TUNNEL_ID=oid:0x2a0000000014ec|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_DST_IP=192.168.0.1
```

kcudnik:
> My apologies, the test that was failing is in SONiC sairedis not SAI. @kcudnik this appears to be BCM56971B0/acl_limit.rec, which is making use of the fact that the default SRC_IP mask is 0.0.0.0, so SRC_IP does not need to be specified for SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_P2MP.
> 
> Example entry:
> 
> ```
> 2022-03-22.05:32:44.773768|c|SAI_OBJECT_TYPE_TUNNEL_TERM_TABLE_ENTRY:oid:0x2b0000000014ed|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_VR_ID=oid:0x300000000007c|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TYPE=SAI_TUNNEL_TERM_TABLE_ENTRY_TYPE_P2MP|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TUNNEL_TYPE=SAI_TUNNEL_TYPE_IPINIP|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_ACTION_TUNNEL_ID=oid:0x2a0000000014ec|SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_DST_IP=192.168.0.1
> ```

its not the only test that fail, its about 237 lines of tests that will fail because now this change will require to DST_IP and SRC_IP to be passed every time, which is not required now, please reffer to https://github.com/sonic-net/sonic-swss/issues/2829

kcudnik:
great, but we will need to wait on sairedis for pulling those new headers probably

inc/saitunnel.h
JaiOCP: Instead of removing the condition, would it make sense to add new conditions (even if its the complete set of enums)

JaiOCP:
> @JaiOCP Please help sign off this.

Still under review. Waiting for reply on comments.

inc/saitunnel.h
mikeberesford: I don't think so. It's semantically the same, and less confusing to understand "this is required for all tunnel types" than "this is required for P2P AND P2MP AND MP2P AND MP2MP"

JaiOCP: I agree that semantically it's the same but to keep backward compatibility more obvious adding new condition checks seems more reasonable then completely removing the condition check. It's also quite possible that there may be some other attribute in future that MAY need the condition check. So what looks like a full set right now may not be so in future.

mholankar: @JaiOCP Are other existing attributes designed for a possibility of new tunnel types? The way I see it, right now this is broken. We cannot use MP2P and MP2MP tunnels. We should fix the case for existing tunnel types. Can we please come to a conclusion on this? We could meet if required.

JaiOCP: My preference is to add the new checks instead of removing the condition flag completely.
Rational is that adding new checks makes the intent clear vs removing the condition check can be interpreted as if no condition is check is needed or is needed for all the options i.e. intent is ambiguous.

mholankar: @kcudnik Do you agree with this? If so, we'll go ahead and make the change

kcudnik: @JaiOCP do you mean add new check like adding condition for all existing enums then ?

JaiOCP: > @JaiOCP do you mean add new check like adding condition for all existing enums then ?

yes. If we add new enums in future then we need to again add the condition flag.
Also removing condition flag does tooling treat as backward incompatible change ?

kcudnik: i think that was the point when i talked with google, that we want to remove condition at all, so even if we add new enums in the future this attribute will be required for them too

JaiOCP: Then as I mentioned my preference is to keep the condition flag.

kcudnik: fine with me

mholankar: @kcudnik @JaiOCP could you please take a look now?

JaiOCP (approved):

mholankar:
@kcudnik could you please help merge this PR?

kcudnik (approved):

mholankar:
@kcudnik @JaiOCP  I don't have permissions to merge this PR. Could one of you please help?

kcudnik:
@mholankar we will be reverting this change back, and discuss this on SAI OCP meeting today, since it's a breaking change and apparently this was not deeply discussed back in the day to allow this to merge

kcudnik:
@mholankar hey, there is a problem still, even if i relax the condition, so you could set it even if condition is not met, this is still create_only attribute, so you will be only able to set it on create() function, and not later on, just fyi

mholankar:
@kcudnik - yes for these attributes it is completely fine.

mikeberesford:
@kcudnik  We can discuss making the src/dst ip and mask able to be modified separately (unless there's a reason to combine these discussions). I do think we'd like to be able to modify these as well at some point though.

PR #1787, opened 2023-04-10
Chained ACL

JaiOCP:
This PR introduces a concept of Chained Table Group.
For details please look at at the ./doc/ACL/Chained-ACL.pptx

PR #1803, opened 2023-06-06
Chained ACL Capability Attr

JaiOCP:
This PR moves the chain acl list information out of sai_acl_capability_t as requested by Kamil for toolchain requirements

### DIFF
--- a/experimental/saiextensions.h
+++ b/experimental/saiextensions.h
@@ -25,6 +25,7 @@
 #ifndef __SAIEXTENSIONS_H_
 #define __SAIEXTENSIONS_H_
 
+#include <sai.h>
 #include <saitypes.h>
 
 /* existing enum extensions */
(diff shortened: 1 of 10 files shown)


kcudnik (approved):

kcudnik:
hey, this spouse to fix https://github.com/opencomputeproject/SAI/pull/1787/files which extended attribute value size, and break structures size, but seems like not doing that

abdosi:
@JaiOCP can you please check on Kamil comment.

kcudnik:
the PR i mentioned is changing structure size of sai_attribute_value_t

JaiOCP:
> the PR i mentioned is changing structure size of sai_attribute_value_t

It's not possible to add a new struct without changing sai_attribute_value_t. Isn't this struct used mainly for toolchains? Can we live the change in this struct size.

kcudnik:
yes its possible since thats an union, but by adding bigger struct than entire value, will change size of entire union, and we want to keep union size constant

JaiOCP:
> @JaiOCP can you please check on Kamil comment.

There was a left over field. Deleted it and all should be good now.

kcudnik:
please fix build errors

JaiOCP:
closing this PR as new enums are causing previous commits to conflict.

kcudnik:
you could force push

PR #1807, opened 2023-06-14
Fix serialize/deseralize functions in SAI meta

juntaog:
This PR is fixing several issues we saw in serialize/deserialize:

1 Fix several not implemented deserialize functions.
    sai_deserialize_enum_list()
    sai_deserialize_attr_id()
    sai_deserialize_attribute()
By doing this, the serialize/deserialize functions are complete.

2 Mark several attributes as aclfiled for SAI_OBJECT_TYPE_UDF_MATCH. The corresponding attribute values are in type of sai_acl_field_data_t.
    SAI_UDF_MATCH_ATTR_L2_TYPE ~ SAI_UDF_MATCH_ATTR_GRE_TYPE
    SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE
By doing this, serialize/deserialize for the above attributes works as expected. 

3 Add sai_metadata_get_attr_metadata_by_attr_id_name_ext which can support the case of attribute id name is in deserialized buffer and terminated by characters listed in function sai_serialize_is_char_allowed().
By doing this, sai_deserialize_attr_id() and sai_deserialize_attribute() can get the attribute metadata by the deserialized buffer.

4 Return SAI_SERIALIZE_ERROR in case of the data type is not find for the attribute value.
By doing this, the user can know that serialize/deserialize is failed and do proper response.

Thanks!

### DIFF
--- a/meta/parse.pl
+++ b/meta/parse.pl
@@ -2056,6 +2056,8 @@ sub ProcessIsAclField
 
     return "true" if $attr =~ /^SAI_ACL_ENTRY_ATTR_(USER_DEFINED_)?FIELD_\w+$/;
 
+    return "true" if $attr =~ /^SAI_UDF_MATCH_ATTR_\w+_TYPE$/;
+
     return "false";
 }
 
--- a/meta/saimetadatautils.c
+++ b/meta/saimetadatautils.c
@@ -150,6 +150,75 @@ const sai_attr_metadata_t* sai_metadata_get_attr_metadata_by_attr_id_name(
     return NULL;
 }
 
+static int sai_metadata_attr_id_name_cmp(
+        _In_ const char * str1,
+        _In_ const char * str2)
+{
+    char c1 = 0;
+    char c2 = 0;
+
+    while (true)
+    {
+        c1 = *str1++;
+        c2 = *str2++;
+
+        if (sai_serialize_is_char_allowed(c1) || sai_serialize_is_char_allowed(c2) || c1 != c2)
+        {
+            if (sai_serialize_is_char_allowed(c1))
+            {
+                c1 = 0;
+            }
+
+            if (sai_serialize_is_char_allowed(c2))
+            {
+                c2 = 0;
+            }
+
+            return c1 - c2;
+        }
+    }
+}
+
+const sai_attr_metadata_t* sai_metadata_get_attr_metadata_by_attr_id_name_ext(
+        _In_ const char *attr_id_name)
+{
+    if (attr_id_name == NULL)
+    {
+        return NULL;
+    }
+
+    /* use binary search */
+
+    ssize_t first = 0;
+    ssize_t last = (ssize_t)(sai_metadata_attr_sorted_by_id_name_count - 1);
+
+    while (first <= last)
+    {
+        ssize_t middle = (first + last) / 2;
+
+        int res = sai_metadata_attr_id_name_cmp(attr_id_name, sai_metadata_attr_sorted_by_id_name[middle]->attridname);
+
+        if (res > 0)
+        {
+            first = middle + 1;
+        }
+        else if (res < 0)
+        {
+            last = middle - 1;
+        }
+        else
+        {
+            /* found */
+
+            return sai_metadata_attr_sorted_by_id_name[middle];
+        }
+    }
+
+    /* not found */
+
+    return NULL;
+}
+
 const sai_attr_metadata_t* sai_metadata_get_ignored_attr_metadata_by_attr_id_name(
         _In_ const char *attr_id_name)
 {
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -449,6 +449,19 @@ bool sai_metadata_is_acl_field_or_action(
         }
     }
 
+    if (metadata->objecttype == SAI_OBJECT_TYPE_UDF_MATCH)
+    {
+        if (metadata->attrid <= SAI_UDF_MATCH_ATTR_GRE_TYPE)
+        {
+            return true;
+        }
+
+        if (metadata->attrid == SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE)
+        {
+            return true;
+        }
+    }
+
     return false;
 }
 
@@ -2324,7 +2337,7 @@ void check_attr_acl_field_or_action(
         META_ASSERT_FALSE(md->defaultvalue->aclaction.enable, "enable should be false");
     }
 
-    if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY)
+    if (md->objecttype != SAI_OBJECT_TYPE_ACL_ENTRY && md->objecttype != SAI_OBJECT_TYPE_UDF_MATCH)
     {
         META_ASSERT_FALSE(md->isaclfield, "field should be not marked as acl field");
         META_ASSERT_FALSE(md->isaclaction, "field should be not marked as acl action");
@@ -2332,18 +2345,36 @@ void check_attr_acl_field_or_action(
         return;
     }
 
-    if (md->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_START &&
-            md->attrid <= SAI_ACL_ENTRY_ATTR_FIELD_END)
+    if (md->objecttype == SAI_OBJECT_TYPE_ACL_ENTRY)
     {
-        META_ASSERT_TRUE(md->isaclfield, "field should be marked as acl field");
-        META_ASSERT_FALSE(md->isaclaction, "field should be not marked as acl action");
+        if (md->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_START &&
+                md->attrid <= SAI_ACL_ENTRY_ATTR_FIELD_END)
+        {
+            META_ASSERT_TRUE(md->isaclfield, "field should be marked as acl field");
+            META_ASSERT_FALSE(md->isaclaction, "field should be not marked as acl action");
+        }
+
+        if (md->attrid >= SAI_ACL_ENTRY_ATTR_ACTION_START &&
+                md->attrid <= SAI_ACL_ENTRY_ATTR_ACTION_END)
+        {
+            META_ASSERT_FALSE(md->isaclfield, "field should not be marked as acl field");
+            META_ASSERT_TRUE(md->isaclaction, "field should be marked as acl action");
+        }
     }
 
-    if (md->attrid >= SAI_ACL_ENTRY_ATTR_ACTION_START &&
-            md->attrid <= SAI_ACL_ENTRY_ATTR_ACTION_END)
+    if (md->objecttype == SAI_OBJECT_TYPE_UDF_MATCH)
     {
-        META_ASSERT_FALSE(md->isaclfield, "field should not be marked as acl field");
-        META_ASSERT_TRUE(md->isaclaction, "field should be marked as acl action");
+        if (md->attrid <= SAI_UDF_MATCH_ATTR_GRE_TYPE)
+        {
+            META_ASSERT_TRUE(md->isaclfield, "field should be marked as acl field");
+            META_ASSERT_FALSE(md->isaclaction, "field should be not marked as acl action");
+        }
+
+        if (md->attrid == SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE)
+        {
+            META_ASSERT_TRUE(md->isaclfield, "field should be marked as acl field");
+            META_ASSERT_FALSE(md->isaclaction, "field should be not marked as acl action");
+        }
     }
 }
 
@@ -3755,6 +3786,13 @@ void check_attr_sorted_by_id_name()
         META_ASSERT_NOT_NULL(found);
 
         META_ASSERT_TRUE(strcmp(found->attridname, am->attridname) == 0, "search attr by id name failed to find");
+
+        const sai_attr_metadata_t *found_ext = sai_metadata_get_attr_metadata_by_attr_id_name_ext(am->attridname);
+
+        META_ASSERT_NOT_NULL(found_ext);
+
+        META_ASSERT_TRUE(strcmp(found_ext->attridname, am->attridname) == 0, "search attr by id name ext failed to find");
+
     }
 
     META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name(NULL));     /* null pointer */
@@ -3763,6 +3801,13 @@ void check_attr_sorted_by_id_name()
     META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name("SAI_P"));  /* in the middle of attr names */
     META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name("SAI_W"));  /* in the middle of attr names */
     META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name("ZZZ"));    /* after all attr names */
+
+    META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name_ext(NULL));     /* null pointer */
+    META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name_ext("AAA"));    /* before all attr names */
+    META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name_ext("SAI_B"));  /* in the middle of attr names */
+    META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name_ext("SAI_P"));  /* in the middle of attr names */
+    META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name_ext("SAI_W"));  /* in the middle of attr names */
+    META_ASSERT_NULL(sai_metadata_get_attr_metadata_by_attr_id_name_ext("ZZZ"));    /* after all attr names */
 }
 
 void list_loop(
--- a/meta/saiserialize.c
+++ b/meta/saiserialize.c
@@ -30,6 +30,7 @@
 #include <limits.h>
 #include <stdio.h>
 #include <string.h>
+#include <stdlib.h>
 #include <sai.h>
 #include "saimetadatautils.h"
 #include "saimetadata.h"
@@ -37,6 +38,24 @@
 
 #define PRIMITIVE_BUFFER_SIZE 128
 #define MAX_CHARS_PRINT 25
+
+/* Expect macros */
+
+#define EXPECT(x) { \
+    if (strncmp(buf, x, sizeof(x) - 1) == 0) { buf += sizeof(x) - 1; } \
+    else { \
+        SAI_META_LOG_WARN("expected '%s' but got '%.*s...'", x, (int)sizeof(x), buf); \
+        return SAI_SERIALIZE_ERROR; } }
+#define EXPECT_KEY(k)    EXPECT("\"" k "\":")
+#define EXPECT_NEXT_KEY(k) { EXPECT(","); EXPECT_KEY(k); }
+#define EXPECT_CHECK(expr, suffix) {                                 \
+    ret = (expr);                                                  \
+    if (ret < 0) {                                                 \
+        SAI_META_LOG_WARN("failed to deserialize " #suffix "");      \
+        return SAI_SERIALIZE_ERROR; }                              \
+    buf += ret; }
+#define EXPECT_QUOTE_CHECK(expr, suffix) {\
+    EXPECT("\""); EXPECT_CHECK(expr, suffix); EXPECT("\""); }
 
 bool sai_serialize_is_char_allowed(
         _In_ char c)
@@ -1154,8 +1173,8 @@ int sai_serialize_enum_list(
                 SAI_META_LOG_WARN("failed to serialize enum_list");
                 return SAI_SERIALIZE_ERROR;
             }
-            buf += ret;
 
+            buf += ret;
             buf += sprintf(buf, "\"");
         }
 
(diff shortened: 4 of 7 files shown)


itaibaz:
@kcudnik 
Can you please review / merge ?
This PR isn't changing any APIs, instead fixing various issues we saw in serialize / desrialize
Thanks

meta/saimetadatautils.c
```diff
@@ -127,7 +146,7 @@ const sai_attr_metadata_t* sai_metadata_get_attr_metadata_by_attr_id_name(
     {
         ssize_t middle = (first + last) / 2;
 
-        int res = strcmp(attr_id_name, sai_metadata_attr_sorted_by_id_name[middle]->attridname);
+        int res = sai_metadata_attr_id_name_cmp(attr_id_name, sai_metadata_attr_sorted_by_id_name[middle]->attridname, strlen(sai_metadata_attr_sorted_by_id_name[middle]->attridname) + 1);
```
kcudnik, line 149: why this needs to be updated ?

meta/saisanitycheck.c
```diff
@@ -2324,26 +2337,43 @@ void check_attr_acl_field_or_action(
...
+    if (md->objecttype == SAI_OBJECT_TYPE_UDF_MATCH)
     {
-        META_ASSERT_FALSE(md->isaclfield, "field should not be marked as acl field");
-        META_ASSERT_TRUE(md->isaclaction, "field should be marked as acl action");
+        if (md->attrid <= SAI_UDF_MATCH_ATTR_GRE_TYPE)
+        {
+            META_ASSERT_TRUE(md->isaclfield, "field should be marked as acl field");
+            META_ASSERT_FALSE(md->isaclaction, "field should be not marked as acl action");
+        }
+        if (md->attrid >= SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE &&
```
kcudnik, line 2371: add empty line before IF

meta/saiserialize.c
```diff
@@ -803,8 +808,20 @@ int sai_serialize_ip_address(
 
         default:
 
-            SAI_META_LOG_WARN("invalid ip address family: %d", ip_address->addr_family);
-            return SAI_SERIALIZE_ERROR;
+            for (idx = 0; idx < sizeof(*ip_address); idx++) {
```
kcudnik, line 811: put { in new line, add empty lines before each if and after each }

meta/serialize.pm
```diff
@@ -847,18 +848,18 @@ sub CreateSerializeEmitMacros
...
+    WriteHeader "#define EMIT_KEY(k)    EMIT(\"\\\"\" k \"\\\":\")";
+    WriteHeader "#define EMIT_NEXT_KEY(k) { EMIT(\",\"); EMIT_KEY(k); }";
+    WriteHeader "#define EMIT_CHECK(expr, suffix) {                                 \\";
+    WriteHeader "    ret = (expr);                                                  \\";
+    WriteHeader "    if (ret < 0) {                                                 \\";
+    WriteHeader "        SAI_META_LOG_WARN(\"failed to serialize \" #suffix \"\");      \\";
+    WriteHeader "        return SAI_SERIALIZE_ERROR; }                              \\";
+    WriteHeader "    buf += ret; }";
+    WriteHeader "#define EMIT_QUOTE_CHECK(expr, suffix) {\\";
+    WriteHeader "    EMIT_QUOTE; EMIT_CHECK(expr, suffix); EMIT_QUOTE; }";
```
kcudnik, line 862: move this back to source, i specially put that in source file to not collide with other defines, this cant be in header

meta/serialize.pm
```diff
@@ -488,6 +488,7 @@ sub EmitSerializeFooter
         WriteSource "{";
         WriteSource "SAI_META_LOG_WARN(\"nothing was serialized for '$name', bad condition?\");";
         # WriteSource "return SAI_SERIALIZE_ERROR;";
+        WriteSource "return SAI_SERIALIZE_ERROR;" if $name eq "sai_attribute_value_t";
```
kcudnik, line 491: why this is addded?

meta/serialize.pm
```diff
@@ -874,22 +875,22 @@ sub CreateDeserializeEmitMacros
...
+    WriteHeader "#define EXPECT_KEY(k)    EXPECT(\"\\\"\" k \"\\\":\")";
+    WriteHeader "#define EXPECT_NEXT_KEY(k) { EXPECT(\",\"); EXPECT_KEY(k); }";
+    WriteHeader "#define EXPECT_CHECK(expr, suffix) {                                 \\";
+    WriteHeader "    ret = (expr);                                                  \\";
+    WriteHeader "    if (ret < 0) {                                                 \\";
+    WriteHeader "        SAI_META_LOG_WARN(\"failed to deserialize \" #suffix \"\");      \\";
+    WriteHeader "        return SAI_SERIALIZE_ERROR; }                              \\";
+    WriteHeader "    buf += ret; }";
+    WriteHeader "#define EXPECT_QUOTE_CHECK(expr, suffix) {\\";
+    WriteHeader "    EXPECT_QUOTE; EXPECT_CHECK(expr, suffix); EXPECT_QUOTE; }";
```
kcudnik, line 893: move back to source

meta/serialize.pm
```diff
@@ -1093,6 +1094,7 @@ sub EmitDeserializeFooter
         WriteSource "{";
         WriteSource "SAI_META_LOG_WARN(\"nothing was deserialized for '$name', bad condition?\");";
         # WriteSource "return SAI_SERIALIZE_ERROR;";
+        WriteSource "return SAI_SERIALIZE_ERROR;" if $name eq "sai_attribute_value_t";
```
kcudnik, line 1097: why this is added?

meta/saiserialize.c
```diff
@@ -1224,15 +1336,42 @@ int sai_serialize_attribute(
...
+            }
+            ret = sai_serialize_uint8(buf, ((const uint8_t*)&attribute->value)[idx]);
+            if (ret < 0)
+            {
+                SAI_META_LOG_WARN("failed to serialize attribute value index %u", idx);
+                return SAI_SERIALIZE_ERROR;
+            }
+            buf += ret;
+        }
+        buf += sprintf(buf, "}");
```
kcudnik, line 1373: please improve your code quality, take a look at the file how lines are written, add empty line before each IF and after each }

meta/saiserialize.c
```diff
@@ -803,8 +808,20 @@ int sai_serialize_ip_address(
 
         default:
 
-            SAI_META_LOG_WARN("invalid ip address family: %d", ip_address->addr_family);
```
kcudnik, line 806: why this is removed? make no sense to serialize IP address if it's not ip4 or ip6, since you dont know what you want to serialize

meta/saiserialize.c
```diff
@@ -832,10 +851,22 @@ int sai_deserialize_ip_address(
         return res;
     }
 
-    SAI_META_LOG_WARN("failed to deserialize '%.*s' as ip address",
-            INET6_ADDRSTRLEN, buffer);
+    for (idx = 0; idx < sizeof(*ip_address); idx++)
```
kcudnik, line 854: make no sense to deserialize numbers, since you have 2 types of ip address 4 and 6, other types are not supported

meta/saimetadatautils.c
kcudnik: please create separate function for this like sai_metadata_get_attr_metadata_by_attr_id_name_ext, since sai_metadata_get_attr_metadata_by_attr_id_name is performance critical, and you are doing a lot of stuff here for each character in the attribute name and also strlen

meta/saiserialize.c
```diff
@@ -1243,6 +1382,77 @@ int sai_deserialize_attribute(
...
+    {
+        EXPECT_CHECK(sai_deserialize_uint32(buf, &attribute->id), uint32);
+        SAI_META_LOG_WARN("id: %u, name not found", attribute->id);
+    }
+
+    EXPECT("\"");
+
+    EXPECT_NEXT_KEY("value");
+
+    if (NULL != meta)
```
kcudnik, line 1416: please move NULL to the right of compare, keep the same code quality as rest of the file

meta/saiserializetest.c
```diff
@@ -1532,6 +1551,195 @@ void test_deserialize_pointer()
...
+    sai_attribute_t attribute = {0};
+    sai_attribute_value_t value;
+
+    buf = "{\"id\":\"SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS\",\"value\":{\"u32\":3}}";
+    res = sai_deserialize_attribute(buf, &attribute);
+    ASSERT_TRUE(res == (int)strlen(buf), "expected true");
+    ASSERT_TRUE(attribute.id == SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS, "expected true");
+    ASSERT_TRUE(attribute.value.u32 == 3, "expected true");
+
+    buf = "{\"id\":\"99999\",\"value\":{85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85,85}}";
```
kcudnik, line 1735: serialize/deserialize attribute IDs which are not valid attributes that are not defined in sai headers make no sense

meta/saiserialize.c
```diff
@@ -813,8 +830,10 @@ int sai_deserialize_ip_address(
         _Out_ sai_ip_address_t *ip_address)
 {
     int res;
+    uint32_t idx;
+    const char *buf = buffer;
 
-    /* try first deserialize ip4 then ip6 */
+    /* try first deserialize ip4 then ip6 then uint8 array*/
```
kcudnik, line 836: make no sense to try this as uint8, since you already know ip4 and ip6, no other types are supported

meta/saisanitycheck.c
```diff
@@ -2324,26 +2337,43 @@ void check_attr_acl_field_or_action(
...
 
         return;
     }
 
-    if (md->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_START &&
-            md->attrid <= SAI_ACL_ENTRY_ATTR_FIELD_END)
-    {
-        META_ASSERT_TRUE(md->isaclfield, "field should be marked as acl field");
-        META_ASSERT_FALSE(md->isaclaction, "field should be not marked as acl action");
+    if (md->objecttype == SAI_OBJECT_TYPE_ACL_ENTRY) {
```
kcudnik, line 2348: put { in new line, please keep code quality high

meta/saisanitycheck.c
```diff
@@ -449,6 +449,19 @@ bool sai_metadata_is_acl_field_or_action(
         }
     }
 
+    if (metadata->objecttype == SAI_OBJECT_TYPE_UDF_MATCH)
+    {
+        if (metadata->attrid <= SAI_UDF_MATCH_ATTR_GRE_TYPE)
+        {
+            return true;
+        }
+        if (metadata->attrid >= SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE &&
```
kcudnik, line 458: add empty line before IF

meta/saimetadatautils.c
```diff
@@ -110,6 +110,25 @@ const sai_attr_metadata_t* sai_metadata_get_attr_metadata(
     return NULL;
 }
 
+static int sai_metadata_attr_id_name_cmp(_In_ const char * str1, _In_ const char * str2, _In_ size_t num)
+{
+    char c1 = '\0';
+    char c2 = '\0';
```
kcudnik, line 116: you can just type = 0;

meta/saimetadatautils.c
```diff
@@ -110,6 +110,25 @@ const sai_attr_metadata_t* sai_metadata_get_attr_metadata(
...
+    while (num > 0)
+    {
+        c1 = *str1++;
+        c2 = *str2++;
+        if (sai_serialize_is_char_allowed(c1) || c1 != c2) {
```
kcudnik, line 121: add empty line before IF, put { in next line, add empty line after each }

kcudnik (changes requested):

meta/saimetadatautils.c
```diff
@@ -110,6 +110,25 @@ const sai_attr_metadata_t* sai_metadata_get_attr_metadata(
     return NULL;
 }
 
+static int sai_metadata_attr_id_name_cmp(_In_ const char * str1, _In_ const char * str2, _In_ size_t num)
```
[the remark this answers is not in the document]
juntaog: ACK.

meta/saimetadatautils.c
juntaog: First of all, one attribute id name may be substring of another attribute id name:
    .attridname                    = "SAI_PORT_ATTR_ADVERTISED_FEC_MODE",
    .attridname                    = "SAI_PORT_ATTR_ADVERTISED_FEC_MODE_EXTENDED",
Secondly, when deserializing an attribute, the input buffer looks like this:
"{\"id\":\"SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS\",\"value\":{\"u32\":3}}"
The deserialize function then pass the pointer of the first character of the id name to the get meta by name function. If it is using strcmp, the attribute id will not be found because the buffer is much longer than just attribute id name string. If it is using strncmp, there is a chance that the length of the substring attribute id is used, so that it will match its longer one.
Will create a separate function for this and do refactor to make it more efficient by avoiding the strlen calling.

meta/saisanitycheck.c
juntaog: ACK

meta/saiserialize.c
juntaog: ACK

meta/serialize.pm
juntaog: The macros are also used by other c files such as saiserialize.c.  If they are moved back, then the same code have to be defined in other c files. Does this acceptable? Thanks!

meta/serialize.pm
juntaog: This is for giving the caller a chance to serialize the attribute value as array of uint8 if nothing is deserialized for it.

meta/saiserialize.c
juntaog: See this corner case.
SAI implementation code may want to serialize an attribute passed by NOS. The attribute is _Inout_ parameter, there is a chance that only attribute.id is set by NOS, but the attribute.value(which contains ipaddress) is random data. In this case, the whole serialize of attribute will be failed if the random addr_family is causing error.
So that I suggest do the serialize in best effort way: do serialize according with the data structure, if failed, serialize the data as array of uint8.

meta/saiserialize.c
juntaog: See the comments in the serialize counterpart.

meta/saiserialize.c
juntaog: ACK

meta/saiserializetest.c
juntaog: Invalid attribute IDs maybe passed to SAI under some error condition. SAI code of course will do sanity check and return failure. In this case, it will be nice if SAI code can serialize and record the invalid IDs.

meta/saisanitycheck.c
juntaog: ACK

meta/saisanitycheck.c
juntaog: ACK

meta/saimetadatautils.c
juntaog: ACK

meta/saimetadatautils.c
juntaog: ACK

meta/saiserialize.c
juntaog: Please see the comments for line 806

meta/saimetadatautils.c
kcudnik: substring dont matter, since strcmp will compare to zero at the end of string so it will always choose correct attribute

meta/serialize.pm
kcudnik: sure, im fine with duplicate

meta/saiserialize.c
kcudnik: random data is invalid, should be set to zero

meta/saiserializetest.c
kcudnik: what conditions? if attribute id is invalid it will be logged by warning message to syslog, thats where you get actual id which is invalid, no need to actually serialize invalid id

meta/saimetadatautils.c
juntaog: Yes, usually strcmp works well if input attribute id name is "SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS".
However, when deserializing an attribute, the input attribute id name is "SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS","value":{"u32":3}}", so that we need a special version of strcmp which takes all the allowed char as '\0'.
Another way is that first extract and copy the attribute id into a separate buffer, then pass it to the get meta by name function. This effects the performance since an additional memory allocation and copy is needed.

meta/serialize.pm
juntaog: ACK

meta/saiserialize.c
juntaog: Even if addr_family is 0, it's still an invalid value when serializing the ip address, that's because only 4 and 6 are valid values.

meta/saiserializetest.c
juntaog: Two possible conditions: 1. Code error in NOS 2. Test cases which are checking if SAI code can deal with the invalid id in right way.
Yes, the invalid id can also be logged by warning messages. However, there are still benefits if the invalid id can be serialized. Firstly, this makes sure the serializing code does not crush under this condition, the NOS may have a chance to know the id is invalid by checking SAI return values. Secondly, this makes the serialized data complete (SAI can serialize and record all the API calling data), which is good for debugging SAI related issues.

meta/saimetadatautils.c
kcudnik: yes, then i proposed to add _ext version of this api especially for that, since i want existing version to be fast

juntaog: ACK,

juntaog:
Updated the PR description.

juntaog:
Updated the patch, removed the process for invalid address family and attribute id.

itaibaz:
@kcudnik 
Hi, we did the asked fixes, can you please help review ?
Thanks

meta/saiserialize.c
```diff
@@ -1172,25 +1186,80 @@ int sai_deserialize_enum_list(
...
+    const sai_attr_metadata_t *meta;
+    size_t len;
+
+    meta = sai_metadata_get_attr_metadata_by_attr_id_name_ext(buffer);
+
+    if (meta != NULL)
+    {
+        len = strlen(meta->attridname);
+        *attr_id = meta->attrid;
+        return (int)len;
```
kcudnik, line 1259: remove len, parameter and just return strlen(meta->attridname);

meta/saiserialize.c
```diff
@@ -1113,11 +1132,6 @@ int sai_serialize_enum_list(
         _In_ const sai_enum_metadata_t *meta,
         _In_ const sai_s32_list_t *list)
 {
-    if (meta == NULL)
-    {
-        return sai_serialize_s32_list(buf, list);
-    }
-
```
kcudnik, line 1120: please bring back this check

meta/saisanitycheck.c
```diff
@@ -449,6 +449,20 @@ bool sai_metadata_is_acl_field_or_action(
...
 
+    if (metadata->objecttype == SAI_OBJECT_TYPE_UDF_MATCH)
+    {
+        if (metadata->attrid <= SAI_UDF_MATCH_ATTR_GRE_TYPE)
+        {
+            return true;
+        }
+
+        if (metadata->attrid >= SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE &&
+                metadata->attrid <= SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE)
```
kcudnik, line 460: why this is range ? you are using the same attribute here SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE , use metadata->attrid == SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE

meta/saisanitycheck.c
```diff
@@ -2324,26 +2338,45 @@ void check_attr_acl_field_or_action(
...
-        META_ASSERT_FALSE(md->isaclfield, "field should not be marked as acl field");
-        META_ASSERT_TRUE(md->isaclaction, "field should be marked as acl action");
+        if (md->attrid <= SAI_UDF_MATCH_ATTR_GRE_TYPE)
+        {
+            META_ASSERT_TRUE(md->isaclfield, "field should be marked as acl field");
+            META_ASSERT_FALSE(md->isaclaction, "field should be not marked as acl action");
+        }
+
+        if (md->attrid >= SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE &&
+                md->attrid <= SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE)
```
kcudnik, line 2375: use == sign, since you use the same attribute SAI_UDF_MATCH_ATTR_L4_DST_PORT_TYPE

meta/parse.pl
```diff
@@ -2056,6 +2056,8 @@ sub ProcessIsAclField
 
     return "true" if $attr =~ /^SAI_ACL_ENTRY_ATTR_(USER_DEFINED_)?FIELD_\w+$/;
 
+    return "true" if $attr =~ /^SAI_UDF_MATCH_ATTR_\w+_TYPE$/;
```
kcudnik, line 2059: this is optimistic, since currently all TYPE ending attr are alc_field_data, but later on this may not be the case, but maybe this would be catched by current saisanitycheck, but if not then attributes would need to be typed here explicitly

meta/saimetadatautils.c
```diff
@@ -150,6 +150,75 @@ const sai_attr_metadata_t* sai_metadata_get_attr_metadata_by_attr_id_name(
...
+            {
+                c2 = 0;
+            }
+
+            return c1 - c2;
+        }
+    }
+}
+
+const sai_attr_metadata_t* sai_metadata_get_attr_metadata_by_attr_id_name_ext(
```
kcudnik, line 182: please add test in saisanitycheck.c that will compare results from sai_metadata_get_attr_metadata_by_attr_id_name_ext and sai_metadata_get_attr_metadata_by_attr_id_name, iterating via all attributes to see if the results will be the same

meta/saiserialize.c
juntaog: ACK.

meta/saiserialize.c
juntaog: ACK.

meta/saisanitycheck.c
juntaog: ACK.

meta/saisanitycheck.c
juntaog: ACK.

meta/parse.pl
juntaog: Yes, in check_attr_acl_field_or_action(), the TYPE ending attr are checked. If new TYPE ending attr is added as not aclfield data, the check will be failed. Then we can add 
return "false" if $attr == SAI_UDF_MATCH_ATTR_NEW_TYPE;
in front of line 2059.

meta/saimetadatautils.c
juntaog: ACK.

juntaog:
Thanks for the comments! Updated the commit accordingly, please help to review, thanks!

kcudnik (approved):

juntaog:
@kcudnik 
Hi, could you please help merge the PR?
Thanks!

PR #1814, opened 2023-06-22
fix the issue where port sai oid not correct when speed change is done from 400G to 100G

abdosi:
What I did:
Fixes:  https://github.com/sonic-net/sonic-buildimage/issues/14706

Why I did:
Without this change any HWSKU with native speed of 400G (8lanes) running as 100G port speed (4 lanes) the sai port oid are always return as NULL cauing all QOS test to fail.

How I did:
Change is generic where instead of doing exact match of lane value set return by SAI_PORT_ATTR_HW_LANE_LIST with the port_config.ini define value we are checking if it's subset of port_config.ini set.

How I verify:
Manual Verification.

### DIFF
--- a/test/saithrift/src/switch_sai_rpc_server.cpp
+++ b/test/saithrift/src/switch_sai_rpc_server.cpp
@@ -1685,13 +1685,29 @@ public:
           port_lane_list_attribute.value.u32list.count = 8;
           port_api->get_port_attribute(port_list_object_attribute.value.objlist.list[i], 1, &port_lane_list_attribute);
 
-          std::set<int> port_lanes;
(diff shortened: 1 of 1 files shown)


abdosi:
@rlhui we need to this for SAI 1.10 and need to backport for 202205 sonic-sairedis

abdosi:
cc @anamehra for viz

abdosi:
@arlakshm can you please help review this.

arlakshm (approved):
lgtm

rraghav-cisco:
@abdosi : Is this ported to on 1.12 as well, pls?

PR #1815, opened 2023-06-26
ACL Chain Capability Attribute

JaiOCP:
Move the ACL chain struct out of sai_acl_capability as it was exceeding the struct size.

### DIFF
--- a/meta/parse.pl
+++ b/meta/parse.pl
@@ -2549,6 +2549,7 @@ sub ProcessStructValueType
     return "SAI_ATTR_VALUE_TYPE_MAC"            if $type eq "sai_mac_t";
     return "SAI_ATTR_VALUE_TYPE_IP_ADDRESS"     if $type eq "sai_ip_address_t";
     return "SAI_ATTR_VALUE_TYPE_IP_PREFIX"      if $type eq "sai_ip_prefix_t";
+    return "SAI_ATTR_VALUE_TYPE_ACL_CHAIN_LIST" if $type eq "sai_acl_chain_list_t";
     return "SAI_ATTR_VALUE_TYPE_PRBS_RX_STATE"  if $type eq "sai_prbs_rx_state_t";
     return "SAI_ATTR_VALUE_TYPE_UINT16"         if $type eq "sai_vlan_id_t";
     return "SAI_ATTR_VALUE_TYPE_UINT32"         if $type eq "sai_label_id_t";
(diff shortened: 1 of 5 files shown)


JaiOCP:
@kcudnik Please review

meta/saisanitycheck.c
```diff
@@ -2286,6 +2290,21 @@ void check_attr_acl_capability(
     }
 }
 
+void check_attr_acl_chain_list(
```
kcudnik, line 2293: you are making sure that acl chain list here is always read only, i think this check is not necessary

meta/saisanitycheck.c
```diff
@@ -2286,6 +2290,21 @@ void check_attr_acl_capability(
...
+    if (md->attrvaluetype != SAI_ATTR_VALUE_TYPE_ACL_CHAIN_LIST)
+    {
+        return;
+    }
+
+    if (md->flags != SAI_ATTR_FLAGS_READ_ONLY)
+    {
+        META_MD_ASSERT_FAIL(md, "attribute marked as acl chain list should be READ_ONLY");
+    }
+}
```
kcudnik, line 2307: add empty line after }

meta/parse.pl
```diff
@@ -2549,6 +2549,7 @@ sub ProcessStructValueType
     return "SAI_ATTR_VALUE_TYPE_MAC"            if $type eq "sai_mac_t";
     return "SAI_ATTR_VALUE_TYPE_IP_ADDRESS"     if $type eq "sai_ip_address_t";
     return "SAI_ATTR_VALUE_TYPE_IP_PREFIX"      if $type eq "sai_ip_prefix_t";
+    return "SAI_ATTR_VALUE_TYPE_ACL_CHAIN_LIST" if $type eq "sai_acl_chain_list_t";
```
kcudnik, line 2552: please remove, why this was added here? this struct value should not be LIST, this is probably leftover since you removed acl_chain_list from acl_capability

meta/saisanitycheck.c
JaiOCP: https://github.com/opencomputeproject/SAI/pull/1815/commits/f56ef7b31276c8228817804428421cbf2870c1db

meta/saisanitycheck.c
JaiOCP: https://github.com/opencomputeproject/SAI/pull/1815/commits/f56ef7b31276c8228817804428421cbf2870c1db

meta/parse.pl
JaiOCP: https://github.com/opencomputeproject/SAI/pull/1815/commits/f56ef7b31276c8228817804428421cbf2870c1db

lguohan:
@kcudnik , can you check?

kcudnik (approved):

PR #1820, opened 2023-07-12
Add hostif trap for MSTP

hamnarauf:
Signed-off-by: Hamna Rauf [hamna.rauf@xflowresearch.com](hamna.rauf@xflowresearch.com)


divyachandralekha:
Since Destination MAC (DMAC) is same for STP and MSTP , we should be able to use same Hostif trap for MSTP as well.

PR #1826, opened 2023-07-18
Add Priority for Tunnel Term Table Entries

mholankar:
Since Tunnel Term Table supports ternary matching for Dest IP+Mask and/or SRC IP+Mask, added priority


kcudnik:
please fix build errors

rlhui:
@mholankar , would you like to present this in 8/10 SAI meeting?  please join if so. Thanks.

mholankar:
@mikeberesford please take a look

PR #1827, opened 2023-07-24
Add inner src and dst mac fields to ACL

kperumalbfn:
Added INNER_SRC_MAC and INNER_DST_MAC fields to ACL table and ACL entry.


kcudnik:
please fix build errors. ps. you will need to do force push here

PR #1829, opened 2023-07-28
[dash]: Add TAG for ACL

Pterosaur:

PR #1834, opened 2023-08-01
[DASH] Update dash headers to most recent

kcudnik:
Updates DASH experimental headers to headers generated from https://github.com/sonic-net/DASH repo on origin/main

Copied generated headers from DASH project after building code with commit b322bc6 (https://github.com/sonic-net/DASH/commit/b322bc6e3e4625de9418fdc1615ee7144ca76ecb)

### DIFF
--- a/experimental/saiexperimentaldashacl.h
+++ b/experimental/saiexperimentaldashacl.h
@@ -111,42 +111,76 @@ typedef enum _sai_dash_acl_rule_attr_t
     SAI_DASH_ACL_RULE_ATTR_DASH_ACL_GROUP_ID,
 
     /**
-     * @brief List matched key dip
+     * @brief Ternary matched key dst_tag
      *
-     * @type sai_ip_prefix_list_t
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_DASH_ACL_RULE_ATTR_DST_TAG,
+
+    /**
+     * @brief Ternary matched mask dst_tag
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_DASH_ACL_RULE_ATTR_DST_TAG_MASK,
+
+    /**
+     * @brief Ternary matched key src_tag
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_DASH_ACL_RULE_ATTR_SRC_TAG,
+
+    /**
+     * @brief Ternary matched mask src_tag
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_DASH_ACL_RULE_ATTR_SRC_TAG_MASK,
+
+    /**
+     * @brief Optional matched key dip
+     *
+     * @type sai_ip_address_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      */
     SAI_DASH_ACL_RULE_ATTR_DIP,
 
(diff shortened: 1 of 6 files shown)


kcudnik:
@marian-pritsak, @Pterosaur, @chrispsommers please review

chrispsommers:
@reshmaintel Could you or your team review? Thanks.

chrispsommers (approved):
Changes seem OK, but it would be helpful to include in the PR description from which commit of [DASH](https://github.com/sonic-net/DASH), these correspond to. If my understanding is correct, these are hand-copied from generated .h files in DASH repo. There is no way to automatically link back to the code commit. Thanks!

kcudnik:
> Changes seem OK, but it would be helpful to include in the PR description from which commit of [DASH](https://github.com/sonic-net/DASH), these correspond to. If my understanding is correct, these are hand-copied from generated .h files in DASH repo. There is no way to automatically link back to the code commit. Thanks!

yes, those are hand copied from generated files, i discussed this with Guohan, that we should treat DASH repo like a ground truth for DASH api

what description do you have in mind?

chrispsommers:
It could be as simple as "Copied generated headers from DASH project after building code with commit XYZ."

experimental/saiexperimentaldashacl.h
```diff
@@ -111,42 +111,76 @@ typedef enum _sai_dash_acl_rule_attr_t
...
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_DASH_ACL_RULE_ATTR_SRC_TAG_MASK,
+
+    /**
+     * @brief Optional matched key dip
+     *
+     * @type sai_ip_address_t
```
Pterosaur, line 148: I have submitted a PR (https://github.com/opencomputeproject/SAI/pull/1829) for the TAG feature, this design is reviewed by @marian-pritsak

Pterosaur (changes requested):

experimental/saiexperimentaldashacl.h
```diff
@@ -111,42 +111,76 @@ typedef enum _sai_dash_acl_rule_attr_t
...
     SAI_DASH_ACL_RULE_ATTR_SRC_PORT,
 
     /**
-     * @brief Range_list matched key dst_port
+     * @brief Optional matched key dst_port
      *
-     * @type sai_u16_range_list_t
+     * @type sai_uint16_t
      * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @isvlan false
```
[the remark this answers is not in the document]
kcudnik: why this needs to be unmarked ?

experimental/saiexperimentaldashacl.h
kcudnik: is that different from auto generated ?

experimental/saiexperimentaldashacl.h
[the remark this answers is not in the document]
Pterosaur: In our design for DASH project, here should be a list, like u16_range_list, u8_list, and ip_prefix_list. But the current implementation of BMv2 doesn't support the list type. Most of the APIs are developed on BMv2 as the prototype. So, if we want to run P4 code on BMv2, this comment should be marked, if we want to generate SAI headers, this comment should be unmarked.

experimental/saiexperimentaldashacl.h
Pterosaur: The TAG feature is still being discussed, so we don't have a final conclusion for it.

kcudnik: does it matter at this stage? since it's still experimental, does anyone use it at all at that point? it could be updated to keep headers in sync, and changed then at any point

kcudnik:
obsolete

PR #1839, opened 2023-08-08
SAI attributes for midpoint for path tracing.

aparnapattekar-gh:
New SAI attributes for configuring path tracing on midpoint node.


kcudnik:
you can fix errors locally by building locally using command: cd meta; make

PR #1845, opened 2023-08-10
Add Priority for Tunnel Term Table Entries

mholankar:
Since Tunnel Term Table supports ternary matching for Dest IP+Mask and/or SRC IP+Mask, added priority


mholankar:
@mikeberesford please take a look

mikeberesford:
@JaiOCP any thoughts?

kcudnik:
to fix errors you will need to squas those commits and force push them on this PR

mholankar:
opening another PR as the commit history in this one is off.

PR #1852, opened 2023-08-18
DCO Error fixing PR

ridahanif96:
Currently, there are two tagging modes for a bridge port (untagged, tagged). A Hybrid Switchport Mode allows both untagged or tagged traffic which needs this hybrid bridge port tagging attribute.

This is in support of  [Hybrid Switchport Mode](https://github.com/sonic-net/SONiC/issues/1136) SONiC

Signed-off-by: Rida Hanif <ridahanif958@gmail.com>

PR #1853, opened 2023-08-18
MSTP Traps for Control Packet

ridahanif96:
This PR brings MSTP  Trap for control packets.  Modified saihostif.h for adding MSTP Trap

This is based on [MSTP Support in SONIC](https://github.com/sonic-net/SONiC/blob/b789c19a6b965b1633200ab710468f049c510a6f/doc/stp/MSTP_HLD.md)
 
Signed-off-by: Rida Hanif <ridahanif958@gmail.com>


ridahanif96:
@rck-innovium Hi, this PR is created to discuss in reference to MSTP Traps. During the community review it was discussed that we don't need this trap as switch internal STP trap will be enough to handle control packet traps. Please let me know about your concerns on this.

rlhui:
> @rck-innovium Hi, this PR is created in reference to MSTP Traps. During the community review it was discussed that we don't need this trap as switch internal STP trap will be enough to handle control packet traps. Can you please help with this. Thanks.

Hi @ridahanif96 , what exact help is needed from @rck-innovium ?  thanks.

ridahanif96:
> > @rck-innovium Hi, this PR is created in reference to MSTP Traps. During the community review it was discussed that we don't need this trap as switch internal STP trap will be enough to handle control packet traps. Can you please help with this. Thanks.
> 
> Hi @ridahanif96 , what exact help is needed from @rck-innovium ? thanks.

Hi @rlhui , thanks for help. 
Earlier, we purpose a separte trap for MSTP Control packet. Later, during HLD review @rck-innovium suggested that there is no need for this new trap.We can use SAI_HOSTIF_TRAP_TYPE_STP, as DMAC and LLC fields are identical for STP and MSTP. If we create a separate trap types for STP and MSTP , it would require the hardware to parse/match the BPDU protocol identifier.
We also agree with this, just created this PR for getting more suggestions from @rck-innovium so we can proceed further with our work as per suggestions/guidance.

kcudnik (approved):

rck-innovium (commented):
@ridahanif96 

Yes. We should be able to use the existing STP trap for STP, RSTP and MSTP.

Can you please explain about the other change in this PR about SAI\_BRIDGE\_PORT\_TAGGING\_MODE\_HYBRID.

ridahanif96:
> @ridahanif96 
> 
> Yes. We should be able to use the existing STP trap for STP, RSTP and MSTP.
> 
> Can you please explain about the other change in this PR about SAI_BRIDGE_PORT_TAGGING_MODE_HYBRID.

Thanks @rck-innovium , I will close this PR. The other change  in this PR about SAI_BRIDGE_PORT_TAGGING_MODE_HYBRID is added by mistake with this PR. Actually we have another [PR](https://github.com/opencomputeproject/SAI/pull/1852) related to feature Switchport mode hybrid. In that PR we proposed a change in saibridge.h for bridge tagging mode hybrid to support both tagged and untagged traffic.

PR #1861, opened 2023-08-27
[meta] Use history file for checking enum ancestry

kcudnik:
since checking ancestry history is taking longer time each commit, we will use history file to load all the history from previous processed commits, in this way we just load entire history to perl directly and save time on processing all those previous commits, this process can be repeated later on if the processing time will increase too much


kcudnik:
This reduced time from metadata make check, from 52 seconds to 14 seconds, on az pipeline this would be hardly noticed since other things take several minutes to build, but this improvement is mainly for local build

PR #1867, opened 2023-08-28
[meta] [rpc] Use location in local inc headers for gensairpc.pl

kcudnik:
it may happen that metadata is already generated, and SAI direcoty is mounted in docker, then the location will be invalid inside docker, so let's use only filename and inc directory

### DIFF
--- a/meta/gensairpc.pl
+++ b/meta/gensairpc.pl
@@ -302,35 +302,46 @@ sub generate_server_template_from_skeleton {
 }
 
 # Find the api name within the header related to the currenct XML file.
-sub get_api_name {
+sub get_api_name
+{
+    # it may happen that metadata is already generated, and SAI direcoty is
+    # mounted in docker, then the location will be invalid inside docker, so
+    # let's use only filename and inc directory
+
     my $location = shift;
-    my $header;
-    my $api;
+
+    $location =~ s!.+/!!; # leave only filename
 
     # The hash should be static - don't open and traverse the header
     # if we already found the API name inside.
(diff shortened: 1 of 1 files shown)


meta/gensairpc.pl
```diff
@@ -302,35 +302,46 @@ sub generate_server_template_from_skeleton {
 }
 
 # Find the api name within the header related to the currenct XML file.
-sub get_api_name {
+sub get_api_name
+{
+    # it may happen that metadata is already generated, and SAI direcoty is
```
chrispsommers, line 307: s/direcoty/directory/

meta/gensairpc.pl
```diff
@@ -302,35 +302,46 @@ sub generate_server_template_from_skeleton {
...
-        close $header or croak;
     }
 
-    $api //= $1 if $location =~ /inc\/sai(\w+)[.]h$/;
-    $api //= 'common';
-    $api = 'common' if $api eq 'types';
-
-    $api_names{$location} = $api;
-    return $api;
+    die "File $file don't contain api struct!";
```
chrispsommers, line 344: s/don't/doesn't/

kcudnik: fixed https://github.com/opencomputeproject/SAI/pull/1870

meta/gensairpc.pl
kcudnik: fixed https://github.com/opencomputeproject/SAI/pull/1870

PR #1886, opened 2023-09-06
Tunnel Term Table support for P2MP type and backward compatible change

mholankar:
For Tunnel Termination Tables - P2MP type, we want to be able to provide the dst IP+ mask & src IP. This change is backward compatible, while relaxing the validations.

### DIFF
--- a/inc/saitunnel.h
+++ b/inc/saitunnel.h
@@ -911,7 +911,7 @@ typedef sai_status_t (*sai_clear_tunnel_stats_fn)(
  */
 typedef enum _sai_tunnel_term_table_entry_type_t
 {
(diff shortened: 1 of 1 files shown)


mholankar:
This is an example change proposed- for discussion for the SAI OCP call on Thu Sep 7, 2023. This change is proposed instead of the already merged PR:
https://github.com/opencomputeproject/SAI/pull/1799

kcudnik:
this is backward compatibility breaking change

fyi @lguohan

kcudnik (changes requested):
this is backward compatibility breaking change, will need to be discussed on OCP SAI meeting

PR #1893, opened 2023-09-08
[meta] Add relaxed condition type

kcudnik:
If attribute is MANDATORY_ON_CREATE and relaxed flag is set to true then given attribute can be passed to create function even if the condition is not met. If relaxed flag is set to false, then attribute is forbidden to be passed to create function is condition is not met.

Mark SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_DST_IP and
SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_SRC_IP as relaxed


mikeberesford:
Thank you!

PR #1899, opened 2023-09-14
Add Flow Abstraction of SAI [Draft, Don't Review]

zhixiongniu:


kcudnik:
why this is route entry copy ?

PR #1901, opened 2023-09-19
[saithriftv2] Use libsaimetadata.so instead of linked meta objects

kcudnik:
This will reuse existing libsaimetadata.so in vendor libsai.so and saithrift server


kcudnik:
FYL @chrispsommers, later on this will be included in libsai.so DASH repo, so both saithrift and libsai.so could reuse libsaimetadata.so

chrispsommers:
Awesome @kcudnik, thanks!

PR #1904, opened 2023-09-20
Update version to 1.13.1

kcudnik:
Update version to 1.13.1


kcudnik:
@rlhui FYI, i forgot to update version when added tag v1.13.1, just added another commit on top and updated tag on github

PR #1914, opened 2023-10-17
Allow null on SAI_PORT_ATTR_PORT_SERDES_ID

kcudnik:
This attribute is readonly, and if port serdes was removed then this attribute can return null object id

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -1730,6 +1730,7 @@ typedef enum _sai_port_attr_t
      * @type sai_object_id_t
      * @flags READ_ONLY
      * @objects SAI_OBJECT_TYPE_PORT_SERDES
+     * @allownull true
      * @default internal
      */
     SAI_PORT_ATTR_PORT_SERDES_ID,


Junchao-Mellanox:
This PR is to fix a crash issue while removing port from orchagent portsorch.cpp. The crash is at line https://github.com/sonic-net/sonic-sairedis/blob/65323ca51db8947e8a39db5c9c0ef1fe4b27a59f/syncd/SaiSwitch.cpp#L1105 . Error flow like this:

orchagent calls portsorch.cpp::removePortBulk:

    a. it removes port serdes first by calling removePortSerdesAttribute, this will remove serdes from VIDTORID and RIDTOVID

    b. it continues and calls remove port which will call Syncd.cpp::processOidRemove

    c. In Syncd.cpp::processOidRemove, it calls SaiSwitch.cpp::collectPortRelatedObjects

    d. In SaiSwitch.cpp::collectPortRelatedObjects, it gets serdes object as a related object to the port. As serdes object does not allow null, it gets a real object ID.

    e. It calls SaiSwitch.cpp::postPortRemove, and it calls getVidForRid for each related object

    f. As serdes VID RID mapping has been removed by step#a, it crashes at  https://github.com/sonic-net/sonic-sairedis/blob/65323ca51db8947e8a39db5c9c0ef1fe4b27a59f/syncd/SaiSwitch.cpp#L1109

srikrishnagopu (commented):
Couple of questions for my understanding

- In step d, what OID is being returned here ? The serdes OID that has been deleted ?
- With this fix, portRelatedObjects will not contain the serdes object anymore ?

Junchao-Mellanox:
> Couple of questions for my understanding
> 
> * In step d, what OID is being returned here ? The serdes OID that has been deleted ?
> * With this fix, portRelatedObjects will not contain the serdes object anymore ?

1. serdes OID. The RIDTOVID mapping has been deleted
2. yes. With the fix, Vendor SAI is allowed to return NULL object ID after removing serdes, and NULL object ID would not be added to portRelatedObjects

Junchao-Mellanox:
Hi @kcudnik , many thanks to have this PR merged. Are we going to take this new commit to sonic?

kcudnik:
Yes, we can update sai pointer

kcudnik:
here is PR https://github.com/sonic-net/sonic-sairedis/pull/1311 including that change

PR #1918, opened 2023-11-01
Flush all FDB entries if no attributes specified

andriy-kokhan:
As per enum `sai_fdb_flush_attr_t` description:
```
 * For example:
 *
 * 1) Flush all entries in FDB table - Do not specify any attribute
   . . . . .
```
Also, it's logically to flush all types of FDB entries in case either no attributes specified or just BV_ID or BRIDGE_PORT_ID.

### DIFF
--- a/inc/saifdb.h
+++ b/inc/saifdb.h
@@ -272,7 +272,7 @@ typedef enum _sai_fdb_flush_attr_t
      *
      * @type sai_fdb_flush_entry_type_t
      * @flags CREATE_ONLY
-     * @default SAI_FDB_FLUSH_ENTRY_TYPE_DYNAMIC
+     * @default SAI_FDB_FLUSH_ENTRY_TYPE_ALL
      */
     SAI_FDB_FLUSH_ATTR_ENTRY_TYPE,
 
--- a/inc/saifdb.h
+++ b/inc/saifdb.h
@@ -229,11 +229,11 @@ typedef enum _sai_fdb_flush_entry_type_t
  *
  * For example:
  *
- * 1) Flush all entries in FDB table - Do not specify any attribute
- * 2) Flush all entries by bridge port - Set #SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID
- * 3) Flush all entries by VLAN - Set #SAI_FDB_FLUSH_ATTR_BV_ID with object id as vlan object
- * 3) Flush all entries by bridge - Set #SAI_FDB_FLUSH_ATTR_BV_ID with object id as bridge object
- * 4) Flush all entries by bridge port and VLAN - Set #SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID
+ * 1) Flush all dynamic entries in FDB table - Do not specify any attribute
+ * 2) Flush dynamic entries by bridge port - Set #SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID
+ * 3) Flush dynamic entries by VLAN - Set #SAI_FDB_FLUSH_ATTR_BV_ID with object id as vlan object
+ * 3) Flush dynamic entries by bridge - Set #SAI_FDB_FLUSH_ATTR_BV_ID with object id as bridge object
+ * 4) Flush dynamic entries by bridge port and VLAN - Set #SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID
  *    and #SAI_FDB_FLUSH_ATTR_BV_ID
  * 5) Flush all static entries by bridge port and VLAN - Set #SAI_FDB_FLUSH_ATTR_ENTRY_TYPE,
  *    #SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID, and #SAI_FDB_FLUSH_ATTR_BV_ID
@@ -272,7 +272,7 @@ typedef enum _sai_fdb_flush_attr_t
      *
      * @type sai_fdb_flush_entry_type_t
      * @flags CREATE_ONLY
-     * @default SAI_FDB_FLUSH_ENTRY_TYPE_ALL
+     * @default SAI_FDB_FLUSH_ENTRY_TYPE_DYNAMIC
      */
     SAI_FDB_FLUSH_ATTR_ENTRY_TYPE,
 


kcudnik:
This is breaking change and I this was done on purpose, since probably you don't want to flush entries you created statically, let say your network hosts, but flush dynamic by default of all learned entries + @lguohan, let's discuss this on SAI community meeting

andriy-kokhan:
> This is breaking change and I this was done on purpose, since probably you don't want to flush entries you created statically, let say your network hosts, but flush dynamic by default of all learned entries + @lguohan, let's discuss this on SAI community meeting

Yeh, you are right. But anyway we should either fix a default value or enum description. At this point, they are contradictory. I'm fine with both ways.

kcudnik:
> > This is breaking change and I this was done on purpose, since probably you don't want to flush entries you created statically, let say your network hosts, but flush dynamic by default of all learned entries + @lguohan, let's discuss this on SAI community meeting
> 
> Yeh, you are right. But anyway we should either fix a default value or enum description. At this point, they are contradictory. I'm fine with both ways.

ahh i see your point now, ok, in the syncd we just forward call to vendor: https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/Syncd.cpp#L598 and then we process it https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/Syncd.cpp#L639 based on attributes https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/RedisClient.cpp#L877 and default type is obtained here:
https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/Syncd.cpp#L609 which is encoded, and set as SAI says, but this should be obtained from sai metadata to be more correct, based on that , the description should be changed, and in flush default value should be obtained if not attribute is specified, which in this case should be DYNAMIC
and description here should be fixed

andriy-kokhan:
> > > This is breaking change and I this was done on purpose, since probably you don't want to flush entries you created statically, let say your network hosts, but flush dynamic by default of all learned entries + @lguohan, let's discuss this on SAI community meeting
> > 
> > 
> > Yeh, you are right. But anyway we should either fix a default value or enum description. At this point, they are contradictory. I'm fine with both ways.
> 
> ahh i see your point now, ok, in the syncd we just forward call to vendor: https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/Syncd.cpp#L598 and then we process it https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/Syncd.cpp#L639 based on attributes https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/RedisClient.cpp#L877 and default type is obtained here: https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/Syncd.cpp#L609 which is encoded, and set as SAI says, but this should be obtained from sai metadata to be more correct, based on that , the description should be changed, and in flush default value should be obtained if not attribute is specified, which in this case should be DYNAMIC and description here should be fixed

I'd expect that in case no attributes specified then SAI library implementation should take the default values based on SAI headers (or meta). I mean it's not mandatory for the application (syncd in our case) to obtain the default values and provide them explicitly to SAI library.

Also, I've just check Linux `bridge` command description (skipping not relevant parameters):
```
bridge fdb flush dev DEV [ brport DEV ] [ vlan VID ] [ [no]permanent | [no]static | [no]dynamic ]
```
Looks like `bridge fdb flush dev DEV` should flush all FDB entries.
So, I still tending to change the default to `_ALL` for SAI as well.

@lguohan , your input will be greatly appreciated. Thanks

kcudnik:
added fix in syncd https://github.com/sonic-net/sonic-sairedis/pull/1317 to get that value form SAI metadata

kcudnik:
> bridge fdb flush dev DEV [ brport DEV ] [ vlan VID ] [ [no]permanent | [no]static | [no]dynamic ]
> ```
> 
> Looks like `bridge fdb flush dev DEV` should flush all FDB entries. So, I still tending to change the default to `_ALL` for SAI as well.

thats a point to discuss whether we should behave like linux bridge command

kcudnik:
Instead of chainging default value can you update description?

andriy-kokhan:
> Instead of chainging default value can you update description?

done

inc/saifdb.h
```diff
@@ -229,11 +229,11 @@ typedef enum _sai_fdb_flush_entry_type_t
...
- * 3) Flush all entries by bridge - Set #SAI_FDB_FLUSH_ATTR_BV_ID with object id as bridge object
- * 4) Flush all entries by bridge port and VLAN - Set #SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID
+ * 1) Flush all dynamic entries in FDB table - Do not specify any attribute
+ * 2) Flush dynamic entries by bridge port - Set #SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID
+ * 3) Flush dynamic entries by VLAN - Set #SAI_FDB_FLUSH_ATTR_BV_ID with object id as vlan object
+ * 3) Flush dynamic entries by bridge - Set #SAI_FDB_FLUSH_ATTR_BV_ID with object id as bridge object
+ * 4) Flush dynamic entries by bridge port and VLAN - Set #SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID
  *    and #SAI_FDB_FLUSH_ATTR_BV_ID
  * 5) Flush all static entries by bridge port and VLAN - Set #SAI_FDB_FLUSH_ATTR_ENTRY_TYPE,
  *    #SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID, and #SAI_FDB_FLUSH_ATTR_BV_ID
```
kcudnik, line 239: i think this also don't specifies all scenarios since if only SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID is set, then the dynamic entries will be flushed, butwhen we set SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID and SAI_FDB_FLUSH_ATTR_ENTRY_TYPE set to static then static entries will be flused or all entries if type will be all so those descriptions it is in assumption that type is not dpecified

so meybe it should be something similar to this:

By default (when SAI_FDB_FLUSH_ATTR_ENTRY_TYPE is not specified) only dynamic entries are affected when doinf flush. By setting SAI_FDB_FLUSH_ATTR_ENTRY_TYPE only dynamic entries can be flushed, static or both, depending on attribute value. Futher more, setting other attributes, can restrict flush only to speciffic enrtires, filetered by bridge port, vlan or bridge. Combinded with type atribute, this can affect static,dynamic or both entries. If bride_port_id attribute is not set, it's not participating into filter. Same for bv_id attribute.

kcudnik (approved):

rck-innovium (approved):
Looks good.

PR #1920, opened 2023-11-02
Queue delay watermark

j-bos:
Introduce a watermark to report maximum delay experienced for packets traversing a queue.

This value should increase as queue congestion increases due to more packets in the queue or lower scheduling rate. Lower scheduling rate may occur, for example, due to packets in higher priority queues or competing WFQ queues.


inc/saiqueue.h
```diff
@@ -393,6 +393,9 @@ typedef enum _sai_queue_stat_t
     /** Get packets deleted when the credit watch dog expires for VOQ System [uint64_t] */
     SAI_QUEUE_STAT_CREDIT_WD_DELETED_PACKETS = 0x00000026,
 
+    /** Queue delay watermark in nanoseconds [uint64_t] */
+    SAI_QUEUE_STAT_DELAY_WATERMARK = 0x00000027,
```
srikrishnagopu, line 397: The name gives an intuition that watermark value has been reported with a delay. Should we consider naming it differently ? For eg: SAI_QUEUE_STAT_WATERMARK_DELAY_TIME or SAI_QUEUE_STAT_DELAYED_WATERMARK_TIME.

kcudnik (approved):
approve only for code quality, please wait for others to review

inc/saiqueue.h
j-bos: How about SAI_QUEUE_STAT_DELAY_WATERMARK_NS  or SAI_QUEUE_STAT_WATERMARK_NS ?

srikrishnagopu: That works.

srikrishnagopu: @j-bos  do you plan to change this ?

srikrishnagopu (approved):

inc/saiqueue.h
j-bos: updated

j-bos:
> You need to squash commits since you changed the name of the enum

Done.

kcudnik (approved):

rlhui (approved):

PR #1921, opened 2023-11-07
Add sai_double_t as double in C. so that SAI can support attributes f…

jjin62:
…or high precison scalar values, such as optical power in dB or dBm


kcudnik (changes requested):
We agreed with community long time ago to not use floating point in SAI only specific const values like microseconds or nanoseconds etc, you should fallow this, +@lguohan

kcudnik:
Please use uint64_t or int64_t with fixed precision milli micro nano pico etc

meta/saimetadatautils.c
```diff
@@ -376,6 +377,9 @@ static bool sai_metadata_is_condition_value_eq(
         case SAI_ATTR_VALUE_TYPE_UINT64:
             return cvalue->u64 == value->u64;
 
+        case SAI_ATTR_VALUE_TYPE_DOUBLE:
+            return fabs(cvalue->d64 - value->d64) < 1E-6;
```
kcudnik, line 381: This can be NaN compare is not valid for all values

meta/saiserialize.c
```diff
@@ -1355,3 +1355,17 @@ int sai_deserialize_attribute(
 
     return (int)(buf - buffer);
 }
+
+int sai_serialize_double(
+        _Out_ char *buffer,
+        _In_ sai_double_t d64)
+{
+    return sprintf(buffer, "%.2lf", d64);
```
kcudnik, line 1363: Throwing off most of significant places make no sense using double use uint64 or int64

jjin62:
> We agreed with community long time ago to not use floating point in SAI only specific const values like microseconds or nanoseconds etc, you should fallow this, +@lguohan

The motivation of this PR is to support yang decimal64 data type, such as:
```
_type decimal64 {
        fraction-digits 2;
 }_
```
Decimal is used as in openconfig as a basic data type: https://github.com/openconfig/public/blob/master/release/models/types/openconfig-types.yang

Now I realize that it is not easy to define a general double type in SAI, because the precision needed is different for each applications. So can we add sai_decimal_precision1_t and sai_decimal_precision2_t? This would be enough to support openconfig yang without conversions.

> We agreed with community long time ago to not use floating point in SAI only specific const values like microseconds or nanoseconds etc, you should fallow this, +@lguohan

Hi, Kamil,
thanks for the response.
The motivation of this PR is to support decimal64 type in yang model, such as:
      type decimal64 {
        fraction-digits 2;
      }
Decimal yang type is used in openconfig as a basic type:
https://github.com/openconfig/public/blob/master/release/models/types/openconfig-types.yang.

Now I realize that it is not easy to support a general double type in SAI, as the precision requirement is case by case. So instead of sai_double_t, can we add sai_decimal64_precision1 and sai_decimal64_precision2? This would be enough to support all openconfig attributes without conversion code.   If you agree, I can amend the commit to address your other comments too.

regards,
Jimmy

kcudnik:
> > We agreed with community long time ago to not use floating point in SAI only specific const values like microseconds or nanoseconds etc, you should fallow this, +@lguohan
> 
> Hi, Kamil, thanks for the response. The motivation of this PR is to support decimal64 type in yang model, such as: type decimal64 { fraction-digits 2; } Decimal yang type is used in openconfig as a basic type: https://github.com/openconfig/public/blob/master/release/models/types/openconfig-types.yang.
> 
> Now I realize that it is not easy to support a general double type in SAI, as the precision requirement is case by case. So instead of sai_double_t, can we add sai_decimal64_precision1 and sai_decimal64_precision2? This would be enough to support all openconfig attributes without conversion code. If you agree, I can amend the commit to address your other comments too.
> 
> regards, Jimmy

i would avoid this at all, and use uint64_t and int64_t if necessary, and specify per attribute basis what precision and what unit is needed, for example if you need mdB unit and you need 2 extra digicts after coma, use micro decibel and you will get 3 most significant digits

jjin62:
> > > We agreed with community long time ago to not use floating point in SAI only specific const values like microseconds or nanoseconds etc, you should fallow this, +@lguohan
> > 
> > 
> > Hi, Kamil, thanks for the response. The motivation of this PR is to support decimal64 type in yang model, such as: type decimal64 { fraction-digits 2; } Decimal yang type is used in openconfig as a basic type: https://github.com/openconfig/public/blob/master/release/models/types/openconfig-types.yang.
> > Now I realize that it is not easy to support a general double type in SAI, as the precision requirement is case by case. So instead of sai_double_t, can we add sai_decimal64_precision1 and sai_decimal64_precision2? This would be enough to support all openconfig attributes without conversion code. If you agree, I can amend the commit to address your other comments too.
> > regards, Jimmy
> 
> i would avoid this at all, and use uint64_t and int64_t if necessary, and specify per attribute basis what precision and what unit is needed, for example if you need mdB unit and you need 2 extra digicts after coma, use micro decibel and you will get 3 most significant digits

Kamil,

I don't know SAI good enough to understand you exactly how to use int64. What we needed is to define the following attributes:

    /**
     * @brief Output power total
     *
     * @type sai_double_t
     * @unit dBm
     * @precision precision2
     * @iscounter false
     */
    SAI_OPTICAL_OUTPUT_POWER_TOTAL,

Can you write a equivalent using sai_int64_t?

kcudnik:
> I don't know SAI good enough to understand you exactly how to use int64. What we needed is to define the following attributes:
> 
> ```
> /**
>  * @brief Output power total
>  *
>  * @type sai_double_t
>  * @unit dBm
>  * @precision precision2
>  * @iscounter false
>  */
> SAI_OPTICAL_OUTPUT_POWER_TOTAL,
> ```
> 
> Can you write a equivalent using sai_int64_t?

use 
```
@type sai_uint64_t or @type sai_int64_t (if you need negative values as well)
```
and type in description what unit is that, please also build metadata locally to fix all errors, there is no @unit, @precision tag, for example take a look at this attribute: https://github.com/opencomputeproject/SAI/blob/b4331a048ec7f80cdf20304b80894fb011dc7b6e/inc/saiars.h#L78

jjin62:
int32 or int64 won't work, that is why we submit this PR . Many openconfig yang model (North bound interface) attributes are decimal. If we use int in SAI and redis DB for these attributes, we need to write code in mgmt transformer or orchagent to convert back and forth between int and decimal. And these are hard coded per attribute.  With sai support double (or double with 2 precision), we can avoid all these messy conversions.

for the build fail at 
```
2023-11-08T03:50:28.1406159Z generated/gen-cpp/sai_types.h:5339:22: error: comparing floating point with == or != is unsafe [-Werror=float-equal]
2023-11-08T03:50:28.1406717Z      if (!(d64 == rhs.d64))
2023-11-08T03:50:28.1406887Z                       ^~~
2023-11-08T03:50:31.8987154Z cc1plus: all warnings being treated as errors
```
 It is because:
1) thrift code generation use == for all data type, which is an issue that thrift need to fix
2) SAI pipeline use flag-Wfloat-equal. 

if I remove -Wfloat-equal, the build will pass. Before thrift fix, SAI pipeline should remove -Wfloat-equal. 
```
jimmy@jimmy-VirtualBox:~/SAI/meta$ g++ -I../inc -I../experimental -fPIC -ansi -Wall -Wcast-align -Wcast-qual -Wconversion -Wdisabled-optimization -Werror -Wextra -Wextra -Wformat=2 -Wformat-nonliteral -Wformat-security -Wformat-y2k -Wimport -Winit-self -Winline -Winvalid-pch -Wmissing-field-initializers -Wmissing-format-attribute -Wmissing-include-dirs -Wmissing-noreturn -Wno-aggregate-return -Wno-padded -Wno-switch-enum -Wno-unused-parameter -Wpacked -Wpointer-arith -Wredundant-decls -Wshadow -Wstack-protector -Wstrict-aliasing=3 -Wswitch -Wswitch-default -Wunreachable-code -Wunused -Wvariadic-macros -Wwrite-strings -std=c++11 -Wno-return-type -c generated/gen-cpp/sai_rpc_server.skeleton.cpp -o generated/gen-cpp/sai_rpc_server.skeleton.o
jimmy@jimmy-VirtualBox:~/SAI/meta
```

kcudnik:
Im not familiar with your model, in uint64 you can fit any number, no conversion is needed, what mess you have in mind? We want to avoid double to not deal with NaN values and normalized and truncated precision and comparison issues that's why we want to use int uint everywhere, if you need double, use uint64 and cast that internally to double,it's 1 line of code

jjin62:
> Im not familiar with your model, in uint64 you can fit any number, no conversion is needed, what mess you have in mind? We want to avoid double to not deal with NaN values and normalized and truncated precision and comparison issues that's why we want to use int uint everywhere, if you need double, use uint64 and cast that internally to double,it's 1 line of code

**Here is the use case of this PR:** 
we can not change the openconfig yang model, as it is tandard yang model for network device management. It uses decimal with 2 precision for optical devices. Config Flow in sonic:
1.  REST API has decimal parameters (A, B, C, D....) sent to sonic management container and set/get to CONFIG-DB
2. SWSS process the config change and modify the APP-DB/ASIC-DB and 
3. syncd will process the ASIC-DB change on HW

To use int in SAI, we need to convert/cast the decimal to int and save it in the REDIS, back and forth,  in either step 1 or 2:
`if (attr is A or B or C or D) then convert ()`. Currently both management and swss containers have infrastructure that is attribute generic, so we don't need to look at each attributes. 

**Deal with NaN**
Instead of have a general sai_doule_t, I propose to have a more specific type sai_decimal_p2. Please check my chnage update.

kcudnik:
we are not operating here on OA/swss level, but on SAI api level, im agains using double in sai specification, if if you need 2 decilal places in precisiosn, use 1 magniture more and have 3 decimal places and use uing64, not sure where your problem arises, if you need double inside SAI library do this:
```c
double val = (double)attr.value.i64;
```

meta/Makefile
```diff
@@ -35,7 +35,6 @@ WARNINGS = \
 	-Werror \
 	-Wextra \
 	-Wextra \
-	-Wfloat-equal \
```
kcudnik, line 38: don't remove warnings, you are making trouble with that, this warning is for a reason here, and reason is specifically what you are removing it

meta/parse.pl
```diff
@@ -72,6 +72,8 @@
 my $FLAGS = "MANDATORY_ON_CREATE|CREATE_ONLY|CREATE_AND_SET|READ_ONLY|KEY";
 my $ENUM_FLAGS_TYPES = "(none|strict|mixed|ranges|free)";
 
+my $DECIMAL_P2_REGEX = '([+-]?([0-9]*[.])?[0-9]{0,2})';
```
kcudnik, line 75: double can be written in expotential form, as you are introducing so many changes instead of using int/uint

jjin62: this is how yang type defined. not me.

kcudnik: that's not an argument to force this specific defintion here if your generator don't generate that

jjin62:
I was assuming supporting double, at least a double with a specific precision, is good thing for SAI in long term. Looks like this is not the case? If so, I can close this PR. Please let me know. Thanks for the feedbacks.

kcudnik:
i mentioned in first comment that we agreed in SAI community to not support floating point numbers because of problems related to them

jjin62:
Ok. We can revisit this if there is other use cases in the future.

Issue #1925, opened 2023-11-10
outside the range of type ‘sai_object_type_t’


oplklum:
I add 5 new objects in experimental, so the SAI_OBJECT_TYPE_EXTENSIONS_MAX is increased from 124 to 129. But when I make meta, an error reported like this:

In file included from saidepgraphgen.cpp:33:
saidepgraphgen.cpp: In function ‘void process_colors()’:
saimetadata.h:489:60: error: the result of the conversion is unspecified because ‘129’ is outside the range of type ‘sai_object_type_t’ {aka ‘_sai_object_type_t’} [-Werror=conversion]
  489 | #define SAI_OBJECT_TYPE_EXTENSIONS_MAX ((sai_object_type_t)129)

I think it's the reason of enum underlying type is not fixed, so it'll be all right when SAI_OBJECT_TYPE_EXTENSIONS_MAX <= 127, otherwise conversion outside the range. I change the value to 127 manually and it works, but report error [-Werror=conversion] when change to 128. So is this a bug?

kcudnik:
yes, probably a bug, can you give link to commit? i will pull that locally and fix that

oplklum:
> yes, probably a bug, can you give link to commit? i will pull that locally and fix that

I've commit my code to https://github.com/oplklum/SAI/commit/54c18db9490b5f1c5f9200c9a31e5c05581296e2
Please have a test, thanks.

kcudnik:
thanks, i cherry picked that, seems like gcc compiler internally chooses some 7 bit size to store enum values for sai_object_type_t, even that sizeof(sai_object_type_t) is sizeof(int32_t), and error pops up on cpp, not C, i'm not sure how to elegantly fix this, i can add pragam to ignore this conversion, but im not sure if this could have some implications later, if at some point compiler will decide to actual 7 bits, at some operations

kcudnik:
i tried to add
```
#pragma GCC diagnostic ignored "-Wconversion"
```

but seems like gcc is ignoring this pragma, since warning is still present if that pragma is applied, not sure if this is gcc bug or a feature

kcudnik:
@oplklum i added temporary workaround: https://github.com/opencomputeproject/SAI/pull/1926, you can cherrypick that and try to build and test locally, but we need to discuss this change on Thursday SAI community meeting

oplklum:
@kcudnik I'm sorry for the late response. #1925 is all right now, but the same problem happened for sai_api_t.

[...]

PR #1926, opened 2023-11-13
[meta] Add custom range start/end to sai_object_type_t and sai_api_t as workaround for internal enum size

kcudnik:
Workadound for https://github.com/opencomputeproject/SAI/issues/1925

Since SAI_OBJECT_TYPE_EXTENSIONS_MAX is declared as define, compiler warns about:

error: the result of the conversion is unspecified because .129. is outside the range of type .sai_object_type_t. {aka
    ._sai_object_type_t.} [-Werror=conversion]

when next value exceeds 7 bits, since that's enough for encode all values for sai_object_type_t. Unfortunetly #pragma GCC diagnostic ignored "-Wconversion" is not working (probably gcc bug) then workaround needs to be addressed in other way.

Those custom ranges can be removed after actual SAI_OBJECT_TYPE_MAX will reach 128 value

### DIFF
--- a/inc/sai.h
+++ b/inc/sai.h
@@ -147,6 +147,12 @@ typedef enum _sai_api_t
     SAI_API_ARS_PROFILE      = 49, /**<sai_ars_api_profile_t */
     SAI_API_TWAMP            = 50, /**< sai_twamp_api_t */
     SAI_API_MAX,                   /**< total number of APIs */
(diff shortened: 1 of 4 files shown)


kcudnik:
issue demonstration here: https://godbolt.org/z/8Mdse7jva

PR #1940, opened 2024-01-05
SAI proposal for icmp echo offload

chikkaiah-work:
This request is to proposes the following:
1. Proposes the concept of an ICMP echo object or session that can be offloaded to HW/ASIC or other modules for expedited link detection.
2. Introduces SAI APIs to define ICMP echo session properties
3. Introduces a switch-level notification API for tracking changes in ICMP echo session states.


kcudnik:
can you add like empty commit, since seems like SAI pipeline is hanging, and azp run can't restart this

kcudnik:
it's still stuck, something is wrong with the pipeline :/

chikkaiah-work:
> it's still stuck, something is wrong with the pipeline :/

it seems it took wrong git user name for commits. I closed this PR and opened new one - https://github.com/opencomputeproject/SAI/pull/1942

chikkaiah-work:
> it's still stuck, something is wrong with the pipeline :/

Hi Kamil, I have created new PR - https://github.com/opencomputeproject/SAI/pull/1943 for this. Looks like the build has not started. I'm not sure what is missing. Can you please let me know what is holding this up from running SAI build and help.

kcudnik:
yea, i wrote email internally to figure out why this is not staring, no need for this new PR, after we fix pipeline then we should be able to restart it for each pr

kcudnik:
ok i enabled the pipeline, issue should be fixed

PR #1944, opened 2024-01-10
Add port extensions for DASH project

r12f:
This port extension is described in the SAI extension design doc (PR #835): https://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Extensions.md, but currently missing in the experiment folder. This change is trying to add it in.

With port extensions, we will be able to add port level counters or other attributes for DASH project for tracking global stats.

### DIFF
--- a/experimental/saiextensions.h
+++ b/experimental/saiextensions.h
@@ -31,6 +31,7 @@
 /* existing enum extensions */
 #include "saitypesextensions.h"
 #include "saiswitchextensions.h"
+#include "saiportextensions.h"
 
 /* new experimental object type includes */
 #include "saiexperimentaldashvip.h"
--- a/experimental/saiportextensions.h
+++ b/experimental/saiportextensions.h
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) 2018 Microsoft Open Technologies, Inc.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *    not use this file except in compliance with the License. You may obtain
+ *    a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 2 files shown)


inc/saiport.h
```diff
@@ -3133,6 +3133,9 @@ typedef enum _sai_port_stat_t
     /** Port stat out drop reasons range end */
     SAI_PORT_STAT_OUT_DROP_REASON_RANGE_END = 0x00002fff,
 
+    /** Port stat range end */
+    SAI_PORT_STAT_END = SAI_PORT_STAT_OUT_DROP_REASON_RANGE_END
```
kcudnik, line 3137: this kind of enum is not supported, can you use sai metadata generate to get all enums ?

r12f: i see. then I can use SAI_PORT_STAT_OUT_DROP_REASON_RANGE_END directly as the extension start. is it ok?

kcudnik:
seems like build pipeline is broken

r12f:
> seems like build pipeline is broken

you mean CodeQL is not running correctly and caused by this change?

inc/saiport.h
r12f: updated now.

r12f:
looks like we will need a meeting on how to move on with stats extension support. i will get it scheduled.

r12f:
I have removed the port stats extension part. And after that, sai metadata generation works again.

image

r12f:
I have removed the port stats part from the PR description as well, so it matches with the code.

kcudnik:
> > seems like build pipeline is broken
> 
> you mean CodeQL is not running correctly and caused by this change?

not related to this change, other PRs are also not running pipelines

kcudnik:
please squash your commits to 1 commit, to trigger build, seems like you changed tam api previously and it fails to build becasue of tthat

r12f:
closing in favor of #1947

PR #1946, opened 2024-01-17
Update the Doxyfile for doxygen in Debian Bookworm

saiarcot895:
In Bookworm, the LATEX_SOURCE_CODE and CLASS_DIAGRAMS options are deprecated/removed. Comment these options out in the Doxyfile.


kcudnik (approved):

kcudnik:
can you also update Doxygen.compat ?

saiarcot895:
I think `Doxyfile.compat` is meant for Doxygen that is older than 1.8.16, and this change is needed for at least 1.9.4 and newer, so I didn't update that file.

PR #1948, opened 2024-01-19
[Suspend] Implement sai_bulk_query for Object Retrieval with Query Support 

zhixiongniu:
In this PR, we introduce the sai_bulk_query_attribute function, designed to retrieve objects (such as flow_entry) using various query conditions. This is particularly useful since each flow table may contain more than 10 million flow entries, making it impractical to fetch and filter all entries. By supporting queries with specific conditions, we enable direct retrieval of the necessary entries from the hardware, enhancing efficiency and performance.

### DIFF
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -249,7 +249,7 @@ sai_status_t sai_bulk_get_attribute(
  * @param[in] switch_id Switch object ID to perform the query.
  * @param[in] object_type Type of SAI objects to query.
  * @param[in] query_attr_count Count for query attributes.
- * @param[in] query_attr_list List of quary attributes.
+ * @param[in] query_attr_list List of query attributes.
  * @param[inout] object_count Number of objects retrieved.
  * @param[inout] object_key Array of object keys retrieved.
  * @param[inout] attr_count List of attr_count. Caller passes the number
@@ -276,7 +276,7 @@ sai_status_t sai_bulk_query_attribute(
         _In_ sai_object_id_t switch_id,
         _In_ sai_object_type_t object_type,
         _In_ uint32_t query_attr_count,
-        _In_ sai_attribute_t *query_attr_list,
+        _In_ const sai_attribute_t *query_attr_list,
         _Inout_ uint32_t *object_count,
         _Inout_ sai_object_key_t *object_key,
         _Inout_ uint32_t *attr_count,
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -250,8 +250,8 @@ sai_status_t sai_bulk_get_attribute(
  * @param[in] object_type Type of SAI objects to query.
  * @param[in] query_attr_count Count for query attributes.
  * @param[in] query_attr_list List of quary attributes. 
- * @param[out] object_count Number of objects retrieved.
- * @param[out] object_key Array of object keys retrieved.
+ * @param[inout] object_count Number of objects retrieved.
+ * @param[inout] object_key Array of object keys retrieved.
  * @param[inout] attr_count List of attr_count. Caller passes the number
  *    of attribute allocated in. Callee returns with the actual
  *    number of attributes filled in. If the count is less than
@@ -273,15 +273,15 @@ sai_status_t sai_bulk_get_attribute(
  * @return sai_status_t Status of the operation.
  */
 sai_status_t sai_bulk_query_attribute(
-    _In_ sai_object_id_t switch_id,
-    _In_ sai_object_type_t object_type,
-    _In_ uint32_t query_attr_count,
-    _In_ sai_attribute_t *query_attr_list,
-    _Out_ uint32_t *object_count,
-    _Out_ sai_object_key_t *object_key,
-    _Inout_ uint32_t *attr_count,
-    _Inout_ sai_attribute_t **attr_list,
-    _Inout_ sai_status_t *object_statuses);
+        _In_ sai_object_id_t switch_id,
+        _In_ sai_object_type_t object_type,
+        _In_ uint32_t query_attr_count,
+        _In_ sai_attribute_t *query_attr_list,
+        _Inout_ uint32_t *object_count,
+        _Inout_ sai_object_key_t *object_key,
+        _Inout_ uint32_t *attr_count,
+        _Inout_ sai_attribute_t **attr_list,
+        _Inout_ sai_status_t *object_statuses);
 
 /**
  * @brief Query attribute capability
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -280,7 +280,7 @@ sai_status_t sai_bulk_query_attribute(
     _Out_ uint32_t *object_count,
     _Out_ sai_object_key_t *object_key,
     _Inout_ uint32_t *attr_count,
-    _Inout_ sai_attribute_t **_attr_list,
+    _Inout_ sai_attribute_t **attr_list,
     _Inout_ sai_status_t *object_statuses);
 
 /**
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -37,8 +37,6 @@
 #include <saisrv6.h>
 
 /* new experimental object type includes */
-#include <saiexperimentaldashflow.h>
-#include <saiexperimentaldashmeter.h>
 #include <saiexperimentaldashvip.h>
 #include <saiexperimentaldashpavalidation.h>
 #include <saiexperimentaldashvnet.h>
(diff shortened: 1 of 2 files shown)


kcudnik:
what is the gain of having this api compared to existing object get ?

r12f:
although the current version passes the check, but i think the first version of API makes more sense. let's revert the back to version 1 and have an offline design to see which way might be better.

kcudnik:
please squash your commits to 1, seems like you have modified some flow entries and brought this back, which is causing validator to fail

inc/saiobject.h
```diff
@@ -239,6 +239,50 @@ sai_status_t sai_bulk_get_attribute(
...
+ *    If the allocated attribute count is not large enough,
+ *    set the status to #SAI_STATUS_BUFFER_OVERFLOW.
+ *
+ * @return #SAI_STATUS_SUCCESS on success, #SAI_STATUS_BUFFER_OVERFLOW if lists size insufficient, failure status code on error
+ */
+sai_status_t sai_bulk_query_attribute(
+        _In_ sai_object_id_t switch_id,
+        _In_ sai_object_type_t object_type,
+        _In_ uint32_t query_attr_count,
+        _In_ const sai_attribute_t *query_attr_list,
```
kcudnik, line 279: query_ prefix is not needed, just attr_count and attr_list

inc/saiobject.h
```diff
@@ -239,6 +239,50 @@ sai_status_t sai_bulk_get_attribute(
...
+ * objects that match the query condition, their attributes, and statuses.
+ *
+ * @param[in] switch_id Switch object ID to perform the query.
+ * @param[in] object_type Type of SAI objects to query.
+ * @param[in] query_attr_count Count for query attributes.
+ * @param[in] query_attr_list List of query attributes.
+ * @param[inout] object_count Number of objects retrieved.
+ * @param[inout] object_key Array of object keys retrieved.
+ * @param[inout] attr_count List of attr_count. Caller passes the number
+ *    of attribute allocated in. Callee returns with the actual
```
kcudnik, line 256: description is confused, this is not list of attributes allocated in

inc/saiobject.h
```diff
@@ -239,6 +239,50 @@ sai_status_t sai_bulk_get_attribute(
...
+ */
+sai_status_t sai_bulk_query_attribute(
+        _In_ sai_object_id_t switch_id,
+        _In_ sai_object_type_t object_type,
+        _In_ uint32_t query_attr_count,
+        _In_ const sai_attribute_t *query_attr_list,
+        _Inout_ uint32_t *object_count,
+        _Inout_ sai_object_key_t *object_key,
+        _Inout_ uint32_t *attr_count,
+        _Inout_ sai_attribute_t **attr_list,
```
kcudnik, line 283: whats the difference between
```
 _In_ uint32_t query_attr_count,
 _In_ const sai_attribute_t *query_attr_list,

and
 _Inout_ uint32_t *attr_count,
_Inout_ sai_attribute_t **attr_list,
```

why naming is confusing here, should you have something like this:
```
         _In_ sai_object_id_t switch_id,  // switch on which do operation
        _In_ sai_object_type_t object_type, // object type determines types of all objects
        _Inout_ uint32_t object_count, // number of objects of type t
        _Inout_ sai_object_key_t *object_key, // object keys x object_count
        _Inout_ uint32_t *attr_count, // attr count x object_count
        _Inout_ sai_attribute_t **attr_list, // attr list x attr_count[i]

File: inc/saiobject.h [hunk 4]
```diff
@@ -239,6 +239,50 @@ sai_status_t sai_bulk_get_attribute(
         _Inout_ sai_attribute_t **attr_list,
         _Inout_ sai_status_t *object_statuses);
 
+/**
+ * @brief Queries attributes of objects in bulk based on a query condition.
+ *
+ * This function performs a bulk query on SAI objects of a specified type,
+ * based on the provided query condition. It retrieves the keys of the
+ * objects that match the query condition, their attributes, and statuses.
+ *
```
kcudnik, line 248: description is too vague, it does not explain why you need 2 sets of attr_cunt passed

can you give example how this api should work ?, because it seems like you want to do a generic bulk_get api for N number of objects which each one can have different attributes, we have bulk api for each specific objects already (no all all populated) but i think this could be added to each specific object type instead of making 1 generic api

zhixiongniu:
> please squash your commits to 1, seems like you have modified some flow entries and brought this back, which is causing validator to fail

Thank you for your comments. The primary reason for the failure is that the 'query_attr_count' is not permitted in this context. It should match the regex as specified in the validation. The relevant figure is shown below.

image

I'd like to explain why 'query_attr_count' is necessary and seek your feedback or help.

We're introducing flow_table and flow_entry abstractions in DASH SAI, referred to as flow_api. This allows us to add flow entries to a flow table based on the ENI for the flow HA project. 

Our design allows each table to handle up to 10 million flow entries for operations like create, remove, modify, query, and get. These entries are key-value pairs based on a five_tuple.

For instance, if we need all entries related to the IP address 1:1:1:1, it's not a simple get operation but a query. The SAI must iterate through the table based on the query condition and return the required results.

The query conditions would include attributes like:

In uint32_t query_attr_count (e.g., 2)
In const sai_attribute_t *query_attr_list (e.g., 1.1.1.1 and QUERY_BASED_ON_IP)


Additionally, we need *object_key, *attr_count, and **attr_list to retrieve the query results:

Inout sai_object_key_t *object_key
Inout uint32_t *attr_count (e.g., 200 attrs)
Inout sai_attribute_t **attr_list

The reasons we prefer not to combine these are as follows:

1. The query attributes consist of a one-dimensional array, while the results are in a two-dimensional array.
2. The space for the result attributes should be allocated by the caller. It is clearer to allocate the query attributes and result collection attributes separately.

inc/saiobject.h
zhixiongniu: Thanks for your comments.
I relied here https://github.com/opencomputeproject/SAI/pull/1948#issuecomment-1907369064.

zhixiongniu: > description is too vague, it does not explain why you need 2 sets of attr_cunt passed
> 
> can you give example how this api should work ?, because it seems like you want to do a generic bulk_get api for N number of objects which each one can have different attributes, we have bulk api for each specific objects already (no all all populated) but i think this could be added to each specific object type instead of making 1 generic api

Thanks for your comments.
I relied here https://github.com/opencomputeproject/SAI/pull/1948#issuecomment-1907369064.

inc/saiobject.h
zhixiongniu: > query_ prefix is not needed, just attr_count and attr_list

Thanks for your comments.
I relied here https://github.com/opencomputeproject/SAI/pull/1948#issuecomment-1907369064.

inc/saiobject.h
zhixiongniu: > whats the difference between
> 
> ```
>  _In_ uint32_t query_attr_count,
>  _In_ const sai_attribute_t *query_attr_list,
> 
> and
>  _Inout_ uint32_t *attr_count,
> _Inout_ sai_attribute_t **attr_list,
> ```
> 
> why naming is confusing here, should you have something like this:
> 
> ```
>          _In_ sai_object_id_t switch_id,  // switch on which do operation
>         _In_ sai_object_type_t object_type, // object type determines types of all objects
>         _Inout_ uint32_t object_count, // number of objects of type t
>         _Inout_ sai_object_key_t *object_key, // object keys x object_count
>         _Inout_ uint32_t *attr_count, // attr count x object_count
>         _Inout_ sai_attribute_t **attr_list, // attr list x attr_count[i]
> ```

Thanks for your comments.
I relied here https://github.com/opencomputeproject/SAI/pull/1948#issuecomment-1907369064.

kcudnik:
this sounds like this specific api is only designed for some dash DASH, currently we don't have any API in SAI that need to specify sope specific condition to do actual GET operation, this seems like a weird approach, not aligned with current SAI api, and this needs to be addressed on SAI thursday meeting

kcudnik (changes requested):
Need SAI OCP meeting

also if this is DAS H only api, it should be defined in dash headers not here, since you mentioned flow_entry, then this should be flow entry specific api, added in flow_api_t struct

zhixiongniu:
> Need SAI OCP meeting
> 
> also if this is DAS H only api, it should be defined in dash headers not here, since you mentioned flow_entry, then this should be flow entry specific api, added in flow_api_t struct

Thanks, @kcudnik. I will discuss with @r12f  and come back soon.

zhixiongniu:
Thanks, @kcudnik. After discussion with @r12f , we decide to  suspend this PR.

PR #1949, opened 2024-01-22
Enable sai_uint16_t in ProcessStructValueType Struct Member

zhixiongniu:
We discovered that in ProcessStructValueType, as a member of a struct, the types sai_uint8_t, uint32_t, and uint64_t are permitted, but uint16_t is not. Since we plan to incorporate flow API into DASH and require src_port and dst_port to be uint16_t, we are including uint16_t in ProcessStructValueType to address this.


r12f:
Hi @kcudnik , could you please help see if this is something intentional?

kcudnik:
this looks ok, does this is still a Draft ? it seems like it os for description,

zhixiongniu:
> this looks ok, does this is still a Draft ? it seems like it os for description,

Thanks, I  removed the [Draft] words.

kcudnik (approved):

PR #1962, opened 2024-02-12
Add word of ENI into aspell.

r12f:
The word ENI is a term which means elastic network interface. It is used in DASH project and its API design, hence adding this word into the aspell checker. 

This issue didn't show up before very likely because aspell ignores the code, however, once we use it in comments, it starts to fail.

### DIFF
--- a/meta/aspell.en.pws
+++ b/meta/aspell.en.pws
@@ -44,6 +44,9 @@ eg
 egressing
 encap
 Encaps
+eni
+Eni
+ENI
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

r12f:
thanks a lot Kamil!

and ouch! I forgot the DCO again.... sorry...

r12f:
woot! thanks Kamil!

PR #1963, opened 2024-02-12
Update DASH headers

marian-pritsak:
Update the inbound routing header to use a new
tunnel decapsulation action name.

### DIFF
--- a/experimental/saiexperimentaldashinboundrouting.h
+++ b/experimental/saiexperimentaldashinboundrouting.h
@@ -38,9 +38,9 @@
  */
 typedef enum _sai_inbound_routing_entry_action_t
 {
-    SAI_INBOUND_ROUTING_ENTRY_ACTION_VXLAN_DECAP,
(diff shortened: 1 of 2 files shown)


kcudnik (approved):

r12f (approved):

experimental/saiexperimentaldashinboundrouting.h
```diff
@@ -38,9 +38,9 @@
  */
 typedef enum _sai_inbound_routing_entry_action_t
 {
-    SAI_INBOUND_ROUTING_ENTRY_ACTION_VXLAN_DECAP,
+    SAI_INBOUND_ROUTING_ENTRY_ACTION_TUNNEL_DECAP,
```
r12f, line 41: the old attributes are already used by swss, we cannot simply change them.

r12f (changes requested):

r12f:
@theasianpianist as heads up on naming change.

PR #1966, opened 2024-02-16
Adding DASH SAI APIs for load balancer fast path scenario.

r12f:
This change adds the SAI API changes for enabling the load balancer fast path scenario.

- DASH load balancer fast path HLD: https://github.com/sonic-net/DASH/blob/main/documentation/load-bal-service/fast-path-icmp-flow-redirection.md
- Change that used to generate the headers: https://github.com/sonic-net/DASH/pull/517

### DIFF
--- a/experimental/saiportextensions.h
+++ b/experimental/saiportextensions.h
@@ -52,6 +52,18 @@ typedef enum _sai_port_stat_extensions_t
 {
     SAI_PORT_STAT_EXTENSIONS_RANGE_START = SAI_PORT_STAT_END,
 
+    /** DASH port LB_FAST_PATH_ICMP_IN_BYTES stat count */
+    SAI_PORT_STAT_LB_FAST_PATH_ICMP_IN_BYTES = SAI_PORT_STAT_EXTENSIONS_RANGE_START,
+
+    /** DASH port LB_FAST_PATH_ICMP_IN_PACKETS stat count */
+    SAI_PORT_STAT_LB_FAST_PATH_ICMP_IN_PACKETS,
+
+    /** DASH port LB_FAST_PATH_ENI_MISS_BYTES stat count */
+    SAI_PORT_STAT_LB_FAST_PATH_ENI_MISS_BYTES,
+
+    /** DASH port LB_FAST_PATH_ENI_MISS_PACKETS stat count */
+    SAI_PORT_STAT_LB_FAST_PATH_ENI_MISS_PACKETS,
+
     /* Add new experimental port stats above this line */
 
     SAI_PORT_STAT_EXTENSIONS_RANGE_END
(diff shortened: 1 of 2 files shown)


kcudnik (approved):

experimental/saiportextensions.h
```diff
@@ -52,6 +52,18 @@ typedef enum _sai_port_stat_extensions_t
 {
     SAI_PORT_STAT_EXTENSIONS_RANGE_START = SAI_PORT_STAT_END,
 
+    /** DASH port LB_FAST_PATH_ICMP_IN_BYTES stat count */
+    SAI_PORT_STAT_LB_FAST_PATH_ICMP_IN_BYTES = SAI_PORT_STAT_EXTENSIONS_RANGE_START,
+
+    /** DASH port LB_FAST_PATH_ICMP_IN_PACKETS stat count */
+    SAI_PORT_STAT_LB_FAST_PATH_ICMP_IN_PACKETS,
```
sharmasushant, line 59: is this to count number of icmp_redirect packets, or just icmp packets??

experimental/saiportextensions.h
```diff
@@ -52,6 +52,18 @@ typedef enum _sai_port_stat_extensions_t
...
+
+    /** DASH port LB_FAST_PATH_ENI_MISS_BYTES stat count */
+    SAI_PORT_STAT_LB_FAST_PATH_ENI_MISS_BYTES,
```
sharmasushant, line 62: What do ENI miss bytes indicate?

sharmasushant: got it from here: https://github.com/sonic-net/DASH/blob/main/documentation/load-bal-service/fast-path-icmp-flow-redirection.md#port-stats-attributes

sharmasushant: Please confirm if these are DPU level counters. e.g., if a packet if given to a DPU due to PA match, but no ENI match happened once it hits DPU and packet is dropped?

experimental/saiportextensions.h
r12f: these counters are specific to LB fast path ICMP packets, not regular ICMP packets or regular ICMP redirect packets that we might not care, e.g. IPv6 ICMP packets with non-zero payload size.

experimental/saiportextensions.h
r12f: yes, these are DPU level counters. and yes, it is designed to capture the exact case you are referring to. In this case, nothing will show up on ENI counters. With this counter, we will know what there are valid fast path packets gets dropped, because ENI cannot be found.

marian-pritsak (approved):

r12f:
thanks a lot for the review and approval, Kamil, Sushant and Marian! since we have all parties signed off, I think this change should be ready to merge now. :D

Issue #1972, opened 2024-03-01
Get vendor specific information from SAI


bluecmd:
Hello,

I would like to be able to drop constructs like `#ifdef BRCMSAI` and be able to deduce from the SAI library what vendor workarounds to apply. I did not find any `sai_query_api_vendor` or similar.

Are there any current ways of doing this, and if not, are you open to the idea of adding one?

kcudnik:
i think SAI was designed to be generic interface and not vendor specific, currently there is no specific interface to query underlying vendoer, and we should not specifically design any macros for that

kcudnik:
what is your motivation to have workaroudns for specific vendor ? if some apis are not implemented you will get not implemented or not supported error code and this should be your information

bluecmd:
Happy to hear! I agree that not relying on the vendor is a good stance to take and try to hold on to.

In my case it is only for debugging, so having something like `sai_query_api_vendor_version` that outputs an opaque string would suffice - like `bsv` (e.g. `BRCM SAI ver: [8.4.0.2], OCP SAI ver: [1.11.0], SDK ver: [sdk-6.5.27] CANCUN ver: [06.04.01]`).

Right now I have to resort to vendor-specific ways to get that information for my test logs, which ironically requires me to care about what module I load so I can query its non-SAI API to get the data.

That said, I based some debugging code on `saithrift` from this very repo and it used `ifdef` macros per vendor.

E.g.:
https://github.com/opencomputeproject/SAI/blob/e2a3d4487c93f1e0f45cd4c0ab85f07e72c75a18/test/saithrift/src/saiserver.cpp#L457-L460

In my case I just removed that particular `ifdef` and always spawned a debug shell, but if it is indeed the case that SAI tries to be vendor agnostic maybe we should remove those from the code people otherwise use as references.

kcudnik:
SAI headers are generic, and there will be no vendor specific macro here, that will let you decide which SAI vendor is internally, and at SAI headers this is even impossible, since there are no vendor specific headers here. investigate SAI_SWITCH_ATTR_SWITCH_SHELL_ENABLE attribute.

We could discuss to introduce vendor string api or something like this that will return something like that, or have an enum in which will return actual vendor as unique integer, but this is broader discussion at SAI community meeting  @rlhui

bluecmd:
For reference, this is how I currently get the data I want in my logs if anyone wants something similar:

```c++
typedef struct {
  const char* sai_api_version;
[...]

PR #1975, opened 2024-03-01
Remove use of vendor specific macros

bluecmd:
SAI should not care about which vendor is loaded, so remove the compile time requirement to specify which vendor will be.

Proposal from discussions in #1972


bm/sai_adapter/test/sai_thrift_src/saiserver.cpp
```diff
@@ -136,7 +136,6 @@ const service_method_table_t test_services = {
     test_profile_get_next_value
 };
 
-#ifdef BRCMSAI
```
kcudnik, line 139: this entire directory is being removed

kcudnik: https://github.com/opencomputeproject/SAI/pull/1954

kcudnik:
pleasse add authors of code that is being removed to disscuss whether this is safe

bluecmd:
> pleasse add authors of code that is being removed to disscuss whether this is safe

Sure, looking at older PRs I guess this is @chrispsommers and @richardyu-ms.

I hope this ping is sufficient to ask for a review :-)

richardyu-ms:
did you make test with your changes in different platforms, or did you take those change in build-image repo?
Indeed, we expect the SAI module does not depend on vendors, but when SAI module build on different vendor platform it related to different SAI implementation/driver, then there we might need different build parameters to build SONiC image with different SAI implementations from different vendors

bluecmd:
I have tested this successfully on Broadcom and Innovium SAI.

richardyu-ms:
> I have tested this successfully on Broadcom and Innovium SAI.

Cause in your changes, there are some changes to remove the build parameters across many platforms
```
ifeq ($(platform),MLNX)
CDEFS = -DMLNXSAI
else
ifeq ($(platform),BFT)
CDEFS = -DBFTSAI
else
ifeq ($(platform),CAVIUM)
CDEFS = -DCAVIUMSAI
else
CDEFS = -DBRCMSAI
endif
endif
endif
```
I think it is better to build the image with your code on those platforms at least, it can make sure the sai component will not impact the SONiC image build.

bluecmd:
@richardyu-ms I do not have any possibility of testing all those platforms.

It seems to me that this code is dead and unmaintained, and I would be surprised if anyone with Barefoot or Cavium runs this code.

If you are certain you require testing on these platforms to remove this code, I will be forced to give up on this PR.

PR #1977, opened 2024-03-06
SAI Proposal for PoE

DanielaMurin:
The following proposal is a draft proposal to add PoE support that will
allow  vendors to implement their own PoE solution.


inc/saipoe.h
```diff
@@ -0,0 +1,524 @@
...
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_POE_ATTR_DEVICE_ID = SAI_POE_PORT_ATTR_START,
+
+    /**
+     * @brief PoE port front panel ID
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE ONLY
```
rck-innovium, line 244: CREATE\_ONLY. (underscore is missing)

inc/saipoe.h
```diff
@@ -0,0 +1,524 @@
+/**
+ * Copyright 2024 Marvell.
```
rck-innovium, line 2: Please fix the copyright.

inc/saipoe.h
```diff
@@ -0,0 +1,524 @@
...
+
+#include <saitypes.h>
+
+/** poe device power limit mode
+ *  port limit - max power limit per port is configured by the user 
+ *  class limit - max power is set automatically according the class of the connected device
+ */
+typedef enum _sai_poe_device_type_limit_mode_t
+{
+    SAI_POE_DEVICE_LIMIT_MODE_TYPE_PORT_LIMIT = 0,
```
rck-innovium, line 26: SAI\_POE\_DEVICE\_TYPE\_LIMIT\_MODE\_PORT\_LIMIT instead of SAI\_POE\_DEVICE\_LIMIT\_MODE\_TYPE\_PORT\_LIMIT

inc/saipoe.h
```diff
@@ -0,0 +1,524 @@
...
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_POE_PSE_ATTR_DEVICE_ID = SAI_POE_PSE_ATTR_START,
+
+    /**
+     * @brief PoE PSE ID (index)
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE ONLY
```
rck-innovium, line 172: CREATE\_ONLY

inc/saipoe.h
```diff
@@ -0,0 +1,524 @@
...
+     *
+     * @type char
+     * @flags READ_ONLY
+     */
+    SAI_POE_PSE_ATTR_PSE_HARDWARE_VERSION,
+
+        /**
+     * @brief temperature (in celsius) of the PSE
+     *
+     * @type int16_t
```
rck-innovium, line 195: sai\_int16\_t

inc/saipoe.h
```diff
@@ -0,0 +1,524 @@
...
+     * @default false
+     */
+    SAI_POE_PORT_ATTR_ADMIN_ENABLED_STATE,
+
+    /**
+     * @brief PoE port power limit mode
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default maximum value that the port can provide
```
rck-innovium, line 271: default can be left as a special value, say 0, to indicate that no max limit is enforced.

inc/saitypes.h
```diff
@@ -1196,6 +1199,54 @@ typedef struct _sai_port_snr_list_t
...
 
+/**
+ * @brief poe port active channel (when delivering power)
+ */
+typedef enum _sai_poe_port_active_channel_id_t
+{
+    SAI_POE_PORT_ACTIVE_CHANNEL_TYPE_A = 0,
+    SAI_POE_PORT_ACTIVE_CHANNEL_TYPE_B = 1,
+    SAI_POE_PORT_ACTIVE_CHANNEL_TYPE_A_AND_B = 2
+} sai_poe_port_active_channel_id_t;
```
rck-innovium, line 1210: either rename the typedef to: sai\_poe\_port\_active\_channel\_type\_t  or the enums to SAI\_POE\_PORT\_ACTIVE\_CHANNEL\_ID\_A

inc/saitypes.h
```diff
@@ -1196,6 +1199,54 @@ typedef struct _sai_port_snr_list_t
...
+
+/**
+ * @brief poe port signature type (when delivering power) 
+ */
+typedef enum _sai_poe_port_signature_type_id_t
+{
+    SAI_POE_PORT_SIGNATURE_TYPE_SINGLE = 0,
+    SAI_POE_PORT_SIGNATURE_TYPE_DUAL = 1
+} sai_poe_port_signature_type_id_t;
```
rck-innovium, line 1219: please rename the typedef to sai\_poe\_port\_signature\_type\_t

inc/saitypes.h
```diff
@@ -1196,6 +1199,54 @@ typedef struct _sai_port_snr_list_t
...
+
+/**
+ * @brief poe port classification method (when delivering power)
+ */
+typedef enum _sai_poe_port_class_method_id_t
+{
+    SAI_POE_PORT_CLASS_METHOD_TYPE_REGULAR = 0,
+    SAI_POE_PORT_CLASS_METHOD_TYPE_AUTO_CLASS = 1
+} sai_poe_port_class_method_id_t;
```
rck-innovium, line 1228: sai\_poe\_port\_class\_method\_id\_t => sai\_poe\_port\_class\_method\_type\_t

inc/saipoe.h
```diff
@@ -0,0 +1,524 @@
...
+     */
+    SAI_POE_PORT_ATTR_START,
+
+    /**
+     * @brief PoE device ID
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_POE_ATTR_DEVICE_ID = SAI_POE_PORT_ATTR_START,
```
rck-innovium, line 238: SAI\_POE\_**PORT**\_ATTR\_DEVICE\_ID

rck-innovium (changes requested):

inc/saitypes.h
```diff
@@ -1196,6 +1199,54 @@ typedef struct _sai_port_snr_list_t
...
+/**
+ * @brief Defines a port consumption structure
+ * 
+ * Data that is needed and available when a port is delivering power
+ */
+typedef struct _sai_poe_port_power_consumption_t {
+
+    /** Port power consumption value */
+    sai_poe_port_active_channel_id_t active_channel; /* active channel a/b/ab */
+    uint32_t voltage_in_volts; /* volts */
```
kcudnik, line 1239: Unit should not be contained in field but in description, also volts resolution Is probably too big you probably need mV

kcudnik:
Please fix build errors

inc/saitypes.h
DanielaMurin: Yes, they should all be milli. Thanks!

inc/saipoe.h
DanielaMurin: Done, thanks.

inc/saipoe.h
DanielaMurin: Done, thanks.

inc/saipoe.h
DanielaMurin: Done, thanks.

inc/saipoe.h
DanielaMurin: Done, thanks.

inc/saitypes.h
DanielaMurin: Done, thanks.

inc/saitypes.h
DanielaMurin: Done, thanks.

inc/saipoe.h
DanielaMurin: Done, thanks.

inc/saipoe.h
DanielaMurin: Done, thanks.

inc/saitypes.h
```diff
@@ -1196,6 +1199,94 @@ typedef struct _sai_port_snr_list_t
...
+    /**
+     * @brief current in milliAmpere
+     */
+    uint32_t current_in_ma;
+
+    /**
+     * @brief consumption in milliWatts
+     */
+    uint32_t consumption_in_mw;
+
```
kcudnik, line 1257: please remove units from vairable names

kcudnik: also is power consumption not needed information ? you can calculate this multiplying A and V to het Watts, but maybe there is some other meaning of this ?
also not sure what are the expected ranges between voltage and current are 40-50 volts ?

kcudnik:
please fix build errors, you have a lot of warnings https://dev.azure.com/mssonic/build/_build/results?buildId=493535&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=1f6d5e21-f0ce-508d-13df-d1c50f7ea7b9&l=286 you will also need to squash commits to 1 commit, since each commit is processed separatly, please try to fix all errors locally, and then submit, you just need to type "make" in meta directory to see if build is correct

inc/saitypes.h
DanielaMurin: I'll remove them. :)

Theoretically yes, you can calculate the Watts. However, we need to also take into consideration the power loss that happens because of the cable length and other hardware components/designs. So, when we multiple A and V we get a number that is in the ballpark but not completely accurate. This is why we prefer to add a dedicated consumption field.

As for the second question, the expected ranges are based on the PoE class. Each class supports up to a certain number of watts. Here is a chart that has the ranges:
image

DanielaMurin:
@kcudnik 
Thanks for the info! Will do. :)

inc/saiswitch.h
```diff
@@ -3000,6 +3003,17 @@ typedef enum _sai_switch_attr_t
...
 
+    /**
+     * @brief Default SAI POE ID
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_POE_DEVICE
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
```
andriy-kokhan, line 3014: Shouldn't we have `@condition SAI_SWITCH_ATTR_TYPE == SAI_SWITCH_TYPE_POE` here?
Is it the only switch attribute that is valid for SAI_SWITCH_TYPE_POE ?

inc/saiswitch.h
```diff
@@ -3000,6 +3003,17 @@ typedef enum _sai_switch_attr_t
...
+    SAI_SWITCH_ATTR_POE_DEVICE_ID,
```
andriy-kokhan, line 3015: Looks like it's proposed to have one-to-one mapping of SAI_OBJECT_TYPE_POE_DEVICE to SAI_OBJECT_TYPE_SWITCH, correct?

inc/saipoe.h
```diff
@@ -0,0 +1,522 @@
...
+     */
+    SAI_POE_PORT_ATTR_START,
+
+    /**
+     * @brief POE port front panel ID
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_POE_PORT_ATTR_FRONT_PANEL_ID = SAI_POE_PORT_ATTR_START,
```
andriy-kokhan, line 260: Is it the same as `SAI_PORT_ATTR_HW_LANE_LIST` ? Is so, should we add `| KEY` as well?
```
     * @type sai_u32_list_t
     */
    SAI_PORT_ATTR_HW_LANE_LIST,
```
Can we have a single POE port that uses multiple HW lanes?

inc/saiswitch.h
DanielaMurin: Looking at it again, perhaps SAI_OBJECT_TYPE_POE_DEVICE should not be mapped to SAI_OBJECT_TYPE_SWITCH at all. Since PoE has its own software and hardware that is separated.

andriy-kokhan: In general, other use-cases (like PHY or Fabric/VoQ) use a switch object for slightly different reasons. E.g., PHY use case considers PHY as L1 Switch object which makes sense. From what I understand, that's not the case with PoE which uses switch object just as a container or as an entry point for PoE specific syncd application instance to create PoE abstraction.

kcudnik:
ASSERT FAILED (on line 5131): object SAI_OBJECT_TYPE_POE_DEVICE is disconnected from graph

that error means that there is no other dependency on this object on other objects, that means that dependency graph is disconnected, you would need to add a link from existing objects, for example use any existing object in that specific object type poe_device to have a link established in dependency or we could discuss exception on this case

inc/saiswitch.h
DanielaMurin: From the metadata check, I see to add the conditional statement does not allow the type to be an sai_object_id_t. (SAI_OBJECT_TYPE_POE_DEVICE)

(Conditional wants the flags to be: @flags MANDATORY_ON_CREATE | CREATE_ONLY, but this is not allowed with sai_object_id_t. 
SAI_SWITCH_ATTR_DEFAULT_POE_DEVICE_ID: marked as conditional, but invalid creation flags: 0x4
SAI_SWITCH_ATTR_DEFAULT_POE_DEVICE_ID: attribute is create_only and it's an object id, this is not allowed)

inc/saiswitch.h
DanielaMurin: Thank you!

DanielaMurin:
I've added the object into sai_switch_attr_t, as an entry point. Adding exceptions may cause an unwanted ripple effect.

inc/saiswitch.h
kcudnik: it cannot be allowed on the switch object, because it make no sense, if you have not switch yet, how you would get other objects ?

kcudnik:
> I've added the object into sai_switch_attr_t, as an entry point. Adding exceptions may cause an unwanted ripple effect.

what ripple effect, what will it cause ?

DanielaMurin:
In general, if we don't need to use an exception it's best not to, so I think it would probably be better to link to an existing object.
From my point of view, I'm open to both approaches.

Are there any objects that were given exceptions?

kcudnik:
Yes 1, debug counter

inc/saiswitch.h
kcudnik: can you explain what would be the purpose here if you set POE instance here ?

doc/SAI-Proposal-PoE.md
```diff
@@ -0,0 +1,591 @@
...
+    SAI_SWITCH_TYPE_PHY,
+
+    /** Switch type is VOQ based NPU */
+    SAI_SWITCH_TYPE_VOQ,
+
+    /** Switch type is Fabric switch device */
+    SAI_SWITCH_TYPE_FABRIC,
+
+    /** Switch type is PoE (Power over Ethernet) */
+    SAI_SWITCH_TYPE_POE,
```
lguohan, line 79: why is this a new type of switch?

inc/saiswitch.h
```diff
@@ -310,6 +310,9 @@ typedef enum _sai_switch_type_t
     /** Switch type is Fabric switch device */
     SAI_SWITCH_TYPE_FABRIC,
 
+    /** Switch type is POE (Power over Ethernet) */
+    SAI_SWITCH_TYPE_POE,
```
kcudnik, line 314: i think this we would need to discuss, this on SAI community meeting, i think this is fine, but i think this is more like ability, since for example maybe NPU switch could have some ethernet ports that supports PoE, not sure if any exists, but i think we could discuss this, i think PoE is more like a "feature" of the switch, rather then entire switch on its own, but maybe i could be wrong

DanielaMurin:
@lguohan @kcudnik 
If the same vendor provides a NPU and PoE switch, then it can be placed under NPU switch with PoE abilities like Kamil mentioned. 
However, if the switch and PoE vendors differ than we want to provide flexibility to create two instances for each functionality. (Two different hardwares, two different vendors, two different implementations)
For example, if vendor A already has a SAI implementation and they want external PoE support, they can keep using their solution and add another PoE implementation from vendor B.

inc/saiswitch.h
DanielaMurin: The POE instance here will be an entry point for initialization. (Graph dependency)
When creating a POE device object, you are initializing the software and hardware components that are needed for POE to run.

rck-innovium (approved):
Looks good based on the review in the SAI community meeting on 21/Mar/2024.

Please consider Taras and Jai's suggestion of having the below:

```
    /**
     * @brief On NPUs that support POE, read the associated POE Port Id.
     *
     * @type sai_object_id_t
     * @flags READ_ONLY
     * @objects SAI_OBJECT_TYPE_POE_PORT
     * @allownull true
     * @default SAI_NULL_OBJECT_ID
     */
    SAI_PORT_ATTR_POE_PORT_ID
```

DanielaMurin:
Thanks Ravi, I've added it. :)

inc/saiswitch.h
```diff
@@ -3000,6 +3003,19 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_AVAILABLE_SYSTEM_VOQS,
 
+    /**
+     * @brief POE DEVICE LIST
+     *
```
kcudnik, line 3008: No need for all capital letters here

inc/saiswitch.h
```diff
@@ -3000,6 +3003,19 @@ typedef enum _sai_switch_attr_t
...
+     * POE Device list.
+     * SAI_NULL_OBJECT_ID in the attribute value.
```
kcudnik, line 3010: Just repeqted brief description not needed

kcudnik: Null object id has no meaning here what is that description? It does not explain anything

inc/saiswitch.h
DanielaMurin: Updated

inc/saiswitch.h
DanielaMurin: I based the comments on SAI_SWITCH_ATTR_TAM_OBJECT_ID. I've removed the unneeded comments.

inc/saitypes.h
rk946043: Yes. That's correct. We need to account the power conversion factors.
Power = current * voltage * power_factor
Here the power factor could be 80% to 85%.
The PSE reports the power at it's side.  It means the power that the PSE is delivering. This doesn't include the power loss due to cable medium and length. The loss is accounted at PD side.
So please keep the power value fetched from the PoE controller that is accurate.

inc/saiswitch.h
rk946043: > Looking at it again, perhaps SAI_OBJECT_TYPE_POE_DEVICE should not be mapped to SAI_OBJECT_TYPE_SWITCH at all. Since PoE has its own software and hardware that is separated.

Yes. The PoE controller HW is entirely different from switch silicon. The switch (MAC device) and the PoE controllers are entirely different HW elements. The device Id for PoE controllers is different from the MAC/PHY device Id/type.

inc/saipoe.h
rk946043: The above diagram has a few glitches. The PSE ports are p1 ... pN. Here the N could be 4 or 8. This is subject to the PoE chip and the supported PoE mode of operations like AT, BT-Type3, BT-full (Both Type-3, Type-4).
Similarly the Host CPU to PoE MCU interface could be Shared Memory/UART/I2C/SPI etc. It is subject to the platform and the implementation.

inc/saitypes.h
DanielaMurin: I am keeping the power value/consumption field. :)

inc/saipoe.h
DanielaMurin: In general, the image wasn't generic enough. It was updated in PoE_Physical_Design.png to reflect something less constrictive.

doc/SAI-Proposal-PoE.md
rk946043: what is the purpose of this "SAI_SWITCH_TYPE_POE" ? Does this mean the switch is PoE capable and can deliver power via it's port?

doc/SAI-Proposal-PoE.md
```diff
@@ -0,0 +1,591 @@
...
+
+# Proposal
+
+The current SAI does not support PoE. We will implement PoE in a way that any vendor can provide their own PoE hardware and software solution.
+
+## Abbreviations and Information
+
+| **Term** | **Definition**                                                                              |
+| -------------- | ------------------------------------------------------------------------------------------------- |
+| PoE            | Power over Ether                                                                                  |
```
rk946043, line 23: Power Over Ethernet

doc/SAI-Proposal-PoE.md
DanielaMurin: Yes.

If the switch and PoE vendors differ than we want to provide flexibility to create two instances for each functionality. (Different hardware, two different vendors, two different implementations)
For example, if vendor A already has a SAI implementation and they want external PoE support, they can keep using their solution and add another PoE implementation from vendor B.

(If the same vendor provides a NPU and PoE switch, then it can be placed under NPU switch with PoE abilities.
Per the community's request, I added a new PoE attribute to sai port: SAI_PORT_ATTR_POE_PORT_ID)

DanielaMurin:
@kcudnik 
Hi Kamil, is there anything from my side that I can do to help move things along?

kcudnik:
if everything is set, we can merge this

DanielaMurin:
From my side, everything is set. 
(All questions were answered. and all requests have been added.)

doc/SAI-Proposal-PoE.md
```diff
@@ -0,0 +1,591 @@
...
+| ----------- | ---------------------- |
+| Authors     | Daniela Murin, Marvell |
+| Status      | In review              |
+| Type        | Standards track        |
+| Created     | 02/25/2024             |
+| SAI-Version | 1.13                   |
+
+# Overview
+
+Power over Ethernet, or PoE, describes any of several standards that pass electric power along with data on twisted-pair Ethernet cabling. This allows a single cable to provide both a data connection and enough electricity to power networked devices such as wireless access points (WAPs), IP cameras and VoIP phones.
```
rk946043, line 13: Reframe the sentence as shown below:

Power over Ethernet, or PoE, describes any of the 802.3(AF, AT, BT) standards that deliver electrical power along with data on twisted-pair Ethernet cables. This allows a single cable to provide both data connection and electricity to power the networked devices such as wireless access points (WAPs), IP cameras and VoIP phones etc.

doc/SAI-Proposal-PoE.md
```diff
@@ -0,0 +1,591 @@
...
+| -------------- | ------------------------------------------------------------------------------------------------- |
+| PoE            | Power over Ether                                                                                  |
+| PSE            | Power Sourcing Equipment                                                                          |
+| PD             | Powered Device, can power the device through an PD-capable Ethernet port                          |
+| PoE Port       | Ethernet port that can deliver power and info to devices, located on the front panel of the board |
+
+# Key Ideas behind PoE
+
+1. PoE provides a hardware solution for providing a data connection and electricity to power connected devices
+2. The PoE software delivers information about the board and its operational statuses
```
rk946043, line 31: The PoE software provides an interface to configure(program) and query the PoE controller and its internal peripherals such as ports etc.

DanielaMurin: Updated

doc/SAI-Proposal-PoE.md
DanielaMurin: Updated

DanielaMurin:
As per @rk946043's request, I've updated some sentence structures in the proposal MD file.

doc/SAI-Proposal-PoE.md
rk946043: Ok.

doc/SAI-Proposal-PoE.md
```diff
@@ -0,0 +1,591 @@
...
+     * Hardware information format is based on POE SAI implementations by vendors.
+     * Format is vendor specific.
+     * Example: Like PCI location, I2C address, UART etc.
+     * In case of NULL, use the default PoE associated switch id.
+     *
+     * @type char
+     * @flags CREATE_ONLY
+     * @default ""
+     */
+    SAI_POE_DEVICE_ATTR_HARDWARE_INFO = SAI_POE_DEVICE_ATTR_START,
```
rk946043, line 105: Does this attribute SAI_POE_DEVICE_ATTR_HARDWARE_INFO refer to initializing the PoE controller ? If so, it is very misleading.

doc/SAI-Proposal-PoE.md
```diff
@@ -0,0 +1,591 @@
...
+/** 
+ * Create a PoE device object
+ */
+  
+/* Allows the PoE software to initialize itself based on the PoE hardware */
+attr_list[0].id = SAI_POE_DEVICE_ATTR_HARDWARE_INFO;
+strncpy(attr.value.chardata, "integrated-mcu", sizeof("integrated-mcu");  /* string is only relevant to the driver, can be anything*/ 
+
+attr_count = 1;
+sai_create_poe_device_fn(&poe_device_id, switch_id, attr_count, attr_list);
```
rk946043, line 118: A Switch(MAC) device can have multiple ports and these ports can span across multiple PoE controllers like as shown below:
              ports 1 ... K   |---------- PoE Controller1 
MAC --- ports K ... M  |---------- PoE Controller2
              ports  M ... N |---------- PoE Controller3
                                and so on
Like some 'J' PoE controllers are serving 1 ... N ports of MAC(Switch) device.
We need the switch Id and the PoE controller (instance) Id here for initializing all the PoE devices under a switch.

doc/SAI-Proposal-PoE.md
```diff
@@ -0,0 +1,591 @@
...
+
+# SAI POE Components
+
+## New PoE SAI objects
+
+Three new SAI objects are introduced.
+
+### PoE Device ID
+
+This object is used to hold the global configuration and PoE hardware status. In many cases, one PoE device instance will have access to all the PoE hardware and components.
```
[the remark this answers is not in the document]
DanielaMurin: The PoE device represents the PoE subsystem(s). 
It can be a singleton, or you can make as many SAI_OBJECT_TYPE_POE_DEVICE objects as required by your system.

doc/SAI-Proposal-PoE.md
```diff
@@ -0,0 +1,591 @@
...
+
+### PoE PSE ID
+
+This object is used to access data from the Power Sourcing Equipment devices in the board. Such as PSE temperatures, versions, and statuses. There can be multiple PSEs in one board.
```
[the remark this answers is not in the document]
DanielaMurin: You create a SAI_OBJECT_TYPE_POE_PSE object and through its existing attributes you access the information for that specific PSE.
If you want to manage the PSE devices, you should create them as a PoE device object.

doc/SAI-Proposal-PoE.md
DanielaMurin: The SAI_POE_DEVICE_ATTR_HARDWARE_INFO refers to initializing the PoE subsystem according to the PoE hardware.
For example, if the board has a MCU you can set the hw_info to "integrated_mcu". If the device uses direct i2c access then you can set the hw_info to "i2c" or an address. You can also send an empty string if this is not relevant for the sytem. 
This allows the driver to understand how to communicate with the PoE hardware.

doc/SAI-Proposal-PoE.md
DanielaMurin: In our example, we have a microcontroller chip that runs PoE firmware (manages all PSE controllers), and the system only communicates with the MCU and not directly with all the PoE components.
If you wish to control each PoE controller individually, please create a PoE device for each relevant one.

doc/SAI-Proposal-PoE.md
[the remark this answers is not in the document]
rk946043: In most of the times, the PSE chip is underneath the PoE MCU. The MCU manages all the PSE chips. The PSE chips are abstracted under MCU.  Here the PoE controller (MCU + PSE chips) is considered as a single unit that manages all the ports underneath it.

MCU based Design:
----------------------
                                   | --> PSE1
CPU <---> MCU <-->| --> PSE2
                                   .........
                                   | --> PSEn 

MCU less Design:
-------------------
                  | --> PSE1
CPU <---> | --> PSE2
                     .........
                  | --> PSEn 

Ultimately the POE controller and the Ports are the primary objects that deal with PoE functionality.
Which design you are targeting here?
If non MCU design is considered, The CPU core has more job like triggering init sequence (reset, push firmware image), configurations, state monitoring etc. In such cases the design MUST cater to these scenarios.

doc/SAI-Proposal-PoE.md
rk946043: Ok.

doc/figures/PoE_Physical_Design.png
rk946043, line 1: Keep an MCU between Host CPU and the PSE chips.
You may keep a box with dotted lines indicating an optional MCU presence.

doc/SAI-Proposal-PoE.md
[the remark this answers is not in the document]
DanielaMurin: The PoE device object and the PSE object are two different objects and are not considered a single unit.

What you wrote is exactly the design. The PoE controller which is the PoE device object, and the ports which is the PoE port object are indeed the main functionality.

The PSE object has read-only attributes and is not used for management. For management, the PoE device object is used, “PoE controller” as you called it.
The PSE object is not mandatory. So, if it is not relevant to the PoE solution then it does not need to be created.

Creating the PoE device object(s), is what triggers initialization sequence. The sequence is completely internal. It should not affect SAI.

The objects allow both MCU based and MCU “less” designs.
For non-MCU, the firmware is loaded by the switch object. (Assuming all the PSEs use the same firmware)
For internal state monitoring, this should be handled inside the SAI SDK and does not require any SAI APIs.
The implementation is up to the vendor, they can initialize everything however they see fit. And then SAI objects are used to interact with the higher levels. (NOS)

doc/figures/PoE_Physical_Design.png
DanielaMurin: I prefer not to add an MCU since not all designs have one and it may cause confusion. 
The simplistic example above shows the flexibility, while highlighting the new SAI components.

JaiOCP (approved):
Looks Good

doc/figures/PoE_Physical_Design.png
rk946043: Since this document doesn't comment anything about MCU between the host CPU and PSEs, I advise you to keep an MCU with dotted lines indicating that MCU presence is optional and also mention the MCU based and MCU less model in the document.

kcudnik:
please resolve conflicts

rlhui:
@DanielaMurin - are you able to resolve conflicts by today (4/12 PST)?

DanielaMurin:
@rlhui @kcudnik
I've resolved the merge conflicts, thank you.

rlhui:
PR build failed.

DanielaMurin:
Hi Rita,

I rebased it, all checks have passed.

Thank you,
Daniela

From: Rita Hui ***@***.***>
Sent: Wednesday, April 17, 2024 5:23 AM
To: opencomputeproject/SAI ***@***.***>
Cc: Daniela Murin ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] Re: [opencomputeproject/SAI] SAI Proposal for PoE (PR #1977)

Prioritize security for external emails: Confirm sender and content safety before clicking links or opening attachments

PR build failed.

inc/saipoe.h
```diff
@@ -0,0 +1,522 @@
...
+    SAI_POE_DEVICE_ATTR_START,
+
+    /**
+     * @brief Device Information for POE initialization.
+     *
+     * Hardware information format is based on POE SAI implementations by vendors.
+     * Format is vendor specific.
+     * Example: Like PCI location, I2C address, UART etc.
+     *
+     * @type char
```
kcudnik, line 102: this will point to sai_attribute_value_t.chardata[32] type, is that what you want ?

DanielaMurin: Yes, a string allows each vendor the flexibility to pass whatever information they need on how to communicate with the PoE hardware.
For example, if the board has a MCU you can set the hw_info to "integrated_mcu". If the board uses direct i2c access then you can set the hw_info to "i2c" or an address. You can also send an empty string if this is not relevant for the system.

rk946043: A simple string will not be sufficient. Better define a model for possible interfaces.
1) UART :
    Dev name/node, baud rate, parity settings etc
2) I2C:
    Bus number, i2C address and the mode of I2C/SMBUS transaction etc
3) SPI:
    SPI address, CS, Cpol, Cphase etc
4) PCI or shared memory:
    Memory start, Memory size, R/W attributes, etc
and so on.

PR #1777, opened 2023-03-09
Notification NOS running SAI and SDK, about HW/FW failures

amandyuk:
Health event is a way to inform NOS about some HW issue in the switch. When some error on switch occur, SAI should in some way inform a NOS about the status. In order to provide flexibility depending on different types issues SAI generates health event with several parameters that describes that event. Hence, to inform NOS - sai_switch_asic_sdk_health_event_notification_fn callback is invoked.

PR #1979, opened 2024-03-06
SER Support

JaiOCP:
SER Support is added on top of base infra introduced by
https://github.com/opencomputeproject/SAI/pull/1777


inc/saitypes.h
```diff
@@ -1722,14 +1722,141 @@ typedef enum _sai_object_stage_t
...
+    uint32_t ser_log_type;
+} sai_ser_health_data_t;
+
+/**
+ * @extraparam sai_health_data_type_t data_type
+ */
+typedef union _sai_health_data_t
+{
+    /** @validonly data_type == SAI_HEALTH_DATA_TYPE_SER */
+    sai_ser_health_data_t ser;
```
kcudnik, line 1850: Why there is only one field where? Dont create stricte with one wntry

kcudnik:
Fix build errors

inc/saitypes.h
itaibaz: This is the original design discussed in 1777 so seems good
Per health event we create a struct
Each health event is only of one type at a time, so we use a union

So let's say we have 2 types of event
struct health_a_data
struct health_b_data

we will put in the union
union health_data {a_data, b_data}

JaiOCP:
Opened a new PR
https://github.com/opencomputeproject/SAI/pull/1982

PR #1980, opened 2024-03-07
protobuf file generation for SAI headers

raghunandan403:
The changes does the following:
- The above changes, when given "make rpc" from meta folder, will generate a sai.proto file.
- have made changes related to protobuf in:
    - rpc/SAI/Struct.pm
    - rpc/SAI/Function.pm
    - rpc/SAI/Typedef.pm
    - rpc/SAI/Variable.pm
    - special handling for protobufs added into: rpc/SAI/RPC/ProtoBufName/Type.pm & meta/rpc/SAI/RPC/ProtoBufName/Variable.pm.

- A new Template toolkit added for generating proto that consist of 
    - messages, rpc_service, Enums/attributes for all features supported in saiheaders.
- The Enums are generated. But, not all enums have a value initialised. So, rt now, I commented Enums generation. I will fix this in coming weeks

- A new gensaiprotorpc.pl is added in meta directory. this will have script to generate the .proto.
- A new target is added in meta/Makefile "protorpc". When the make is given, it runs the above pearl file. This will build a .proto and also generate .c and .h file.

As a part of the make of the SAI, both thrift and proto will be built. To check if the .proto is generated properly, the script will try to generate the .c and .h file based on the .proto file. If the .proto file has some issue, the make will fail and bail out.


kcudnik:
please fix build errors and sign your commit to pass DCO

kcudnik:
hmm, not sure why this errors shows on pipeline, maybe perl in pipeline needs some extra include directory for search in path

raghunandan403:
> hmm, not sure why this errors shows on pipeline, maybe perl in pipeline needs some extra include directory for search in path

How do i take it forward? 
Can i get access to the pipeline code and how they have written code there? I will check and try to fix. Also, I need to fix "install proto compiler".

kcudnik:
That's the problem there is no access to pipeline everything is in yml file in SAI repo you can see. Configuration there

kcudnik:
https://github.com/opencomputeproject/SAI/blob/master/azure-pipelines.yml

meta/gensaiprotorpc.pl
```diff
@@ -0,0 +1,529 @@
...
+# @file    gensaiprotorpc.pl
+#
+# @brief   This module generates RPC interface of SAI for PTF
+#
+
+use strict;
+use warnings;
+use diagnostics;
+
+use 5.020;
```
kcudnik, line 30: why explicit perl version ?

raghunandan403: Agree Kamil.  Hardcoding of pearl isnt mandatory.
But, i picked up from the gensairpc.pl file.

------

# @file    gensairpc.pl
#
# @brief   This module generates RPC interface of SAI for PTF
#

use strict;
use warnings;
use diagnostics;

use 5.020;

bhagatgit:
@raghunandan403 
Thank you for implementing protobuf file generation. We have been working on something similar. Is there a HLD associated with this code PR ? Are there any applications planned with this ?

raghunandan403:
> @raghunandan403 Thank you for implementing protobuf file generation. We have been working on something similar. Is there a HLD associated with this code PR ? Are there any applications planned with this ?

The code is at :
https://github.com/raghunandan403/proto-sai

I had some hurdles in committing the code. The sanity tests were failing for unknown reason.. I havent spent the time to get it committed. But, we are using this internally for DPU based implementation.

PR #1985, opened 2024-03-26
[meta] Fix mac default value

kcudnik:
Make mac default value in correct format


rlhui:
@kcudnik , can we add reason why this change is needed? thanks.

kcudnik:
usually mac address each byte has 2 hexadecimal values, just corrected this to have a valid form

rlhui (approved):

PR #1986, opened 2024-03-28
TAM Mirror-on-Drop to Local host via Generic Netlink

rck-innovium:
**Mirror-On-Drop (MOD)** is one of the monitoring events supported by the existing SAI TAM 2.0 API. This proposal introduces additional attributes and enums to configure the hardware to send MOD packets to localhost via GENETLINK.

### DIFF
--- a/doc/TAM/SAI-Proposal-TAM-MOD-Localhost.md
+++ b/doc/TAM/SAI-Proposal-TAM-MOD-Localhost.md
@@ -1,0 +1,438 @@
+# SAI TAM Enhancements to support MOD to Local host via Generic Netlink
+-------------------------------------------------------------------------------
+ Title       | Send Mirror-on-Drop packets to Local host via Generic Netlink
+-------------|-----------------------------------------------------------------
+ Authors     | Ravindranath C K, Pooja Panara (Marvell)
+ Status      | In review
+ Type        | Standards track
+ Created     | 2024-03-28 - Initial Draft
+ SAI-Version | 1.14
+-------------------------------------------------------------------------------
+
+## 1.0  Introduction
+
+**Mirror-On-Drop (MOD)** is one of the monitoring events supported by the existing SAI TAM 2.0 API. This proposal introduces additional attributes and enums to configure the hardware to send MOD packets to localhost via GENETLINK.
+
+## 2.0  Use case
+
+Using GENETLINK is a popular mechanism to hand off packets/events to user agents in the local host. The GENETLINK mechanism is supported in SAI for sFlow samples. The below example describes the usage of GENETLINK for sFlow in SONiC/SAI today.
+
+```mermaid
+sequenceDiagram
+ASIC ->> Driver : sFlow(sample)
+Driver ->> GENETLINK : encapsulated sample: IIF-INDEX, SAMPLE-RATE etc.
+GENETLINK ->> hsflowd : genetlink multicast message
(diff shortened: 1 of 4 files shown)


inc/saitam.h
```diff
@@ -1180,8 +1180,19 @@ typedef enum _sai_tam_report_type_t
 
     /**
      * @brief Report using vendor extensions
+     * deprecate and use CUSTOM_RANGE_START
```
kcudnik, line 1183: if one enum should be depreacated, then another should replace it maybe, but giving custom range make no sense here?

doc/TAM/SAI-Proposal-TAM-MOD-Localhost.md
```diff
@@ -0,0 +1,440 @@
...
+     /**
+      * @brief Report using Histogram
+      */
+     SAI_TAM_REPORT_TYPE_HISTOGRAM,
+
+     /**
+      * @brief Report using vendor extensions
++     * deprecate and use CUSTOM_RANGE_START
+      */
+     SAI_TAM_REPORT_TYPE_VENDOR_EXTN,
```
JaiOCP, line 187: vendor extension is kept as a valid report type instead of using the extension range. This was a specific request during the early TAM review

doc/TAM/SAI-Proposal-TAM-MOD-Localhost.md
```diff
@@ -0,0 +1,440 @@
...
++
++    /**
++     * @brief Report using GENETLINK
++     */
++    SAI_TAM_REPORT_TYPE_GENETLINK,
```
JaiOCP, line 192: Can you elaborate what exactly will be this report type format?

JaiOCP (commented):
PR mostly looks good. Have few comments, mostly for clarifications.

doc/TAM/SAI-Proposal-TAM-MOD-Localhost.md
rck-innovium: The report type format is defined by the GENETLINK family.

The GENETLINK message has a netlink header, genetlink header followed by the payload. The payload is a list of attributes that are defined by the GENETLINK family. Example of the attributes:

[https://github.com/torvalds/linux/blob/c85af715cac0a951eea97393378e84bb49384734/include/uapi/linux/net_dropmon.h#L71](https://github.com/torvalds/linux/blob/c85af715cac0a951eea97393378e84bb49384734/include/uapi/linux/net_dropmon.h#L71)

JaiOCP (approved):

kcudnik (approved):

kcudnik:
I'm not familiar with that feature, i review this pr from consistency and validation purpose, but currently for example swss-common have feature that can listen on netlink message https://github.com/sonic-net/sonic-swss-common/blob/master/common/netdispatcher.cpp

rck-innovium:
> @kcudnik @JaiOCP could you please explain the gnetlink message format for that DROPS like we have for psample: https://man7.org/linux/man-pages/man8/tc-sample.8.html. How it expected to be listen by user space deamon?

Please see:
1) Definition of the Generic Netlink Attributes: https://github.com/torvalds/linux/blob/master/include/uapi/linux/net_dropmon.h
2) Kernel implementation of registering the family along with the "events" MC group, and  also filling the above genl attributes: https://github.com/torvalds/linux/blob/master/net/core/drop_monitor.c

PR #1989, opened 2024-04-12
[meta] Add exception for sai_switch_health_data_t

kcudnik:
This will add exception for breaking change SER Data #1982. Since new field is added to existing structure and it changes it size, but it is only used in notification, we allow this breaking change.

It still can cause issues, since that struct will have different size on the stack, and potentially it can read garbage memory or cause crash when using old headers and new SAI library

### DIFF
--- a/meta/structs.pl
+++ b/meta/structs.pl
@@ -162,7 +162,8 @@ sub BuildCommitHistory
         # of union may not increase by adding members, and actual union size
         # check is performed by sai sanity check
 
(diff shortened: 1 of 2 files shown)


rlhui:
@kcudnik seems more change is needed around https://github.com/kcudnik/SAI/blob/62ee9c0edfe2581236dc4d6d374521df528a752b/meta/structs.pl#L174 for the exception.

PR #1999, opened 2024-04-18
Update SAI API for DASH metering.

r12f:
The DASH metering design is updated, hence updating the SAI API definition.

The latest metering HLD is published here: <https://github.com/sonic-net/DASH/pull/544/files>.


inc/saiobject.h
```diff
@@ -37,6 +37,7 @@
 #include <saisrv6.h>
 
 /* new experimental object type includes */
+#include <saiexperimentaldashmeter.h>
```
kcudnik, line 40: this change is not needed

r12f: thanks for checking, Kamil! It is updated now.

prsunny (approved):

kcudnik (approved):

PR #2002, opened 2024-04-22
Add sai_u32_range_t support for SAI struct entries.

r12f:
Currently, SAI checks will fail if we add port ranges using sai_u32_range_t type in the table entry. And range will be great to have as it can be easily mapped to range match in ASIC.

```c
typedef struct _sai_some_table_entry_t
{
    /**
     * @brief Range matched key dst_port_range
     */
    sai_u32_range_t dst_port_range;

    ....
```


kcudnik (approved):
looks good

PR #2004, opened 2024-04-24
Add DASH flow resimulation SAI APIs.

r12f:
This commit is for adding SAI APIs for DASH flow resimulation feature.

The design is following the HLD defined here: <https://github.com/sonic-net/DASH/pull/543>.


kcudnik:
Please fix build errors

r12f:
hi Kamil, thanks for checking and the build is fixed now :D.

experimental/saitypesextensions.h
```diff
@@ -104,5 +104,24 @@ typedef enum _sai_dash_tunnel_dscp_mode_t
...
+
+    SAI_DASH_ROUTING_ACTIONS_NAT = 2,
+
+    SAI_DASH_ROUTING_ACTIONS_NAT46 = 4,
+
+    SAI_DASH_ROUTING_ACTIONS_NAT64 = 8,
+
+    SAI_DASH_ROUTING_ACTIONS_NAT_PORT = 16,
+
+} sai_dash_routing_actions_t;
```
kcudnik, line 124: can you use 1 << x as in other strict flags ?

kcudnik (approved):

experimental/saitypesextensions.h
r12f: thanks Kamil and will do! i will fix the SAI generator in DASH, so we won't have this problem starting the very next type!

PR #2005, opened 2024-04-24
Add action type to ENI ether address map entry to be consistent with all other DASH tables.

r12f:
Currently all DASH SAI table entries (not objects) will have an action type defined for it, even for a single action. This is for future proof in case we need to add any new action type in, so we will not break the ABI. 

However, due to an issue in the template, this action type is missing for eni ether address map entry. And we have this recently fixed in DASH, hence updating the API accordingly.

### DIFF
--- a/experimental/saiexperimentaldasheni.h
+++ b/experimental/saiexperimentaldasheni.h
@@ -36,6 +36,15 @@
  */
 
 /**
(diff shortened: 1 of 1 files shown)


mukeshmv (approved):
lgtm

kcudnik (approved):

PR #2006, opened 2024-04-24
Update the inbound routing header to use a new tunnel decapsulation action name.

r12f:
The name of inbound routing stage actions is changed from VXLAN_DECAP to TUNNEL_DECAP for covering other type of tunnels, hence updating the SAI API accordingly to match the latest DASH design.


marian-pritsak (approved):

kcudnik (approved):
just not that default value change is not backward compatible but since this is experimental header it does not matter

PR #2007, opened 2024-04-24
Add basic DASH HA session APIs for concepts, state management and inline sync.

r12f:
This change add the HA session API for:

- Basic building blocks, such as HA set, HA scope and HA role.
- State management: How the role is attached and controlled.
- Inline sync: Data path definitions of HA set.

All APIs are following the HLD that is merged in DASH here: <https://github.com/sonic-net/DASH/pull/532>.


inc/saiobject.h
```diff
@@ -37,6 +37,7 @@
 #include <saisrv6.h>
 
 /* new experimental object type includes */
+#include <saiexperimentaldashha.h>
```
kcudnik, line 40: this change is not needed

r12f: thanks for checking Kamil! I have it removed now.

kcudnik (approved):

PR #2012, opened 2024-05-10
Next hop group with members.

erohsik:
Introduce a new type of Next hop group that allows the application to provide the list of next hops that are members of the Next hop group along with the list of weights.


kcudnik:
Please fix build errors

PR #2020, opened 2024-05-28
SAI proposal for icmp echo offload

chikkaiah-work:
This specification suggests:

    The idea of an ICMP echo object or session that can be transferred to HW/ASIC or other modules for faster link detection.
    The introduction of SAI APIs to outline ICMP echo session properties.
    The introduction of a switch-level notification API to monitor changes in ICMP echo session states.
    The switchover flow on link down detection. 
    
Creating this PR to recommit #1943.


zjswhhh (commented):
Please update with: 
1. Jai's questions in the offline emails
2. LAG solution

chikkaiah-work:
> Please update with:
> 
>     1. Jai's questions in the offline emails
> 
>     2. LAG solution

chikkaiah-work:
closed by mistake

chikkaiah-work:
> > Please update with:
> > ```
> > 1. Jai's questions in the offline emails
> > 
> > 2. LAG solution
> > ```

Added SAI_ICMP_ECHO_SESSION_ATTR_RX_PORT to specify RX port. 
Added doc section "“ICMP echo session with LAG”

inc/saiicmpecho.h
```diff
@@ -0,0 +1,387 @@
...
+    SAI_ICMP_ECHO_SESSION_ATTR_START,
+
+    /**
+     * @brief Hardware lookup valid
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default true
+     */
+    SAI_ICMP_ECHO_SESSION_ATTR_HW_LOOKUP_VALID = SAI_ICMP_ECHO_SESSION_ATTR_START,
```
JaiOCP, line 83: I believe hw lookup means that ICMP packet is forwarded using L3 lookup instead of forwarding to a port. Please rename the attribute and add comments to that intent.

inc/saiicmpecho.h
```diff
@@ -0,0 +1,387 @@
...
+
+    /**
+     * @brief Destination Port
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_PORT
+     * @condition SAI_ICMP_ECHO_SESSION_ATTR_HW_LOOKUP_VALID == false
+     */
+    SAI_ICMP_ECHO_SESSION_ATTR_PORT,
```
JaiOCP, line 103: This means that session is created between peering links. How do we specify more sessions i.e. monitoring between multiple peering links

inc/saiicmpecho.h
```diff
@@ -0,0 +1,387 @@
...
+    SAI_ICMP_ECHO_SESSION_ATTR_RX_INTERVAL,
+
+    /**
+     * @brief To enable protection group switch over on session state change
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_ICMP_ECHO_SESSION_ATTR_SET_NEXT_HOP_GROUP_SWITCHOVER,
```
JaiOCP, line 215: This is very fuzzy. What happens if there are multiple such protection NHGs. How does this attribute indicates which one to of the protection NHG takes effect. We just can't trigger all the protection NHG to take over.

inc/saiicmpecho.h
chikkaiah-work: You are correct. It's intended for L3 lookup for forwarding. I am willing to rename it, but the same naming convention is also used in saitwamp.h and saibfd.h, and I prefer to maintain consistency.  Please let me know if still want me to rename it.

JaiOCP: In that case please add details in comments clearly stating what happens when flag is set to true or false

inc/saiicmpecho.h
JaiOCP: This means that SONiC has to create multiple echo sessions; one per peering ports.
If thats the requirement then you may want to add the CRM attribute in switch. Something like
SAI_SWITCH_ATTR_AVAILABLE_ICMP_ECHO_SESSIONS read only attribute.
Initial read will give the max and as more sessions as added subsequent reads will give remaining.

inc/saiicmpecho.h
JaiOCP: @chikkaiah-work Please look at this comment as well

chikkaiah-work: You brought up a valid point. If it's possible for various NHGs to share the same monitor object, it seems reasonable to consider modifying this attribute to the NHG object ID. Should this attribute be left unset (which is null as default), automatic switchover will not occur. Instead, the switchover should be initiated based on the specified NHG. Please let me know if this sounds fine to you

doc/SAI-Proposal-ICMP-ECHO.md
```diff
@@ -0,0 +1,424 @@
...
+This defines state change notification fields
+
+icmp_echo_session_id:  Identifies the ICMP echo session related to the state change notification.
+
+session_state: Reflects the operational state of the ICMP echo session, which can be either SAI_ICMP_ECHO_SESSION_UP or SAI_ICMP_ECHO_SESSION_DOWN.
+
+#### sai_icmp_echo_session_attr_t ####
+This defines icmp echo session attributes. These attributes collectively define the configuration and behavior of ICMP echo sessions in the SAI framework.
+
+- SAI_ICMP_ECHO_SESSION_ATTR_VIRTUAL_ROUTER: Specifies whether hardware lookup is valid, allowing for virtual router configuration.
```
lolyu, line 138: should be `SAI_ICMP_ECHO_SESSION_ATTR_HW_LOOKUP_VALID`

doc/SAI-Proposal-ICMP-ECHO.md
```diff
@@ -0,0 +1,424 @@
...
+    8B, processing of this filed is implementation specific.
+- TLV Type
+    1B, SENTINAL: 0xFF
+- TLV Length
+    2B, set to 0
+
+```
+***GUID:*** Global unique identifier for the ICMP echo session to correlate probe and reply packets with a session, especially when the session is offloaded to hardware, a unique identifier is required.
+
+***Session Cookie:*** The offloading engine interprets this value to decide whether the packet should undergo processing by the ICMP echo session or not. Packets not matching the cookie are punted to NOS. This can also be used to differentiate regular icmp response and icmp link monitoring response.
```
lolyu, line 99: should we use the same cookie for sessions on the same device or we can use multiple cookies?

doc/SAI-Proposal-ICMP-ECHO.md
```diff
@@ -0,0 +1,424 @@
...
+
+- Sessions across a LAG bundle: Individual sessions are created on each of the member links. The device creates ICMP echo requests for each configured probing session, which are sent through the port specified by the Session attribute SAI_ICMP_ECHO_SESSION_ATT_PORT.
+
+- Session on a logical bundle (LAG interface): A single session is set up on the LAG interface. Given that a LAG interface may have multiple next hops or that a next hop can vary dynamically, the destination for the outbound ICMP echo packets is determined by the router's lookup procedure. The session should be configured with the attributes SAI_ICMP_ECHO_SESSION_ATTR_VIRTUAL_ROUTER and SAI_ICMP_ECHO_SESSION_ATTR_HW_LOOKUP_VALID both set to true.
+
+- Incoming ICMP echo response packets can arrive on any of the member links. If necessary, the Session attribute SAI_ICMP_ECHO_SESSION_ATTR_RX_PORT is provided to verify the incoming packets on a specific member link.
+
+### 4.0 Examples ###
+
+#### 4.0.1 Create session using egress interface: ####
```
lolyu, line 252: Please provide examples with LAG

inc/saiicmpecho.h
lolyu: if there are new NHGs added, will the new NHGs be modified based on the monitor session?

doc/SAI-Proposal-ICMP-ECHO.md
lolyu, line 252: also, please provide examples of sessions with protection group.

chikkaiah-work: Added example with section tiltled " ICMP echo session with protection group"

chikkaiah-work: Added a section "ICMP echo session with LAG" and explained session creation process.

inc/saiicmpecho.h
chikkaiah-work: Added SAI_SWITCH_ATTR_AVAILABLE_ICMP_ECHO_SESSION and SAI_SWITCH_ATTR_MAX_ICMP_ECHO_SESSION.

inc/saiicmpecho.h
chikkaiah-work: Added a brief description for this attribute.  Also, this attribute is mentioned explained in the examples.

inc/saiicmpecho.h
chikkaiah-work: @lolyu, you are right. Also I have updated the FRR document by mentioning echo session could be included as monitor object.  I have added an example under the section "ICMP echo session with protection group"

doc/SAI-Proposal-ICMP-ECHO.md
chikkaiah-work: fixed.

doc/SAI-Proposal-ICMP-ECHO.md
chikkaiah-work: Each session can share the same cookie, but if a device needs to communicate with multiple devices, each session can have its own unique cookie. The decision to assign cookies to sessions is implementation-specific, and the device can implement a check to enforce the desired behavior.

inc/saiicmpecho.h
JaiOCP: Thank You

inc/saiicmpecho.h
JaiOCP: Thank You

zjswhhh (approved):
LGTM. 

Please wait for Jai's approval before merging.

inc/saiicmpecho.h
JaiOCP: Thanks for updating the spec with an example workflow. Looks good to me.

JaiOCP (approved):

chikkaiah-work:
@kcudnik - can you please help merge this.

lolyu (approved):

kcudnik (approved):

chikkaiah-work:
> @kcudnik - can you please help merge this.

@rlhui - can we get this merged please ?

PR #2023, opened 2024-05-31
Egress SAI DSCP action behavior 

siqbal1986:
This document des not define any new behavior for the SAI DSCP change capabilty. It only solidifies the existing the behavour of the packet pipelines observed for various vendor devices.
This document lists different places in the packet pipeline where the Packet DSCP value can be changed. It further clearifies how changing of the DSCP value  at one stage may be overridden by the next stage and shows how changing this value at different stages may have impact the final packet egressing the pipeline.


siqbal1986:
@kperumalbfn @abdosi  can you please review.

PR #2024, opened 2024-06-02
[Marvell]  SAI Proposal for color aware classification and remarking with qosmaps

helloanandhi:
Proposal to to enable for color aware classification and remarking with qosmaps.


inc/saitypes.h
```diff
@@ -755,6 +758,9 @@ typedef struct _sai_qos_map_params_t
     /** Forwarding class */
     sai_uint8_t fc;
 
+    /** CFI value */
+    sai_uint8_t cfi;
```
kcudnik, line 762: this change is not binary backward compatible

kcudnik:
also please fix all warnings: https://dev.azure.com/mssonic/build/_build/results?buildId=562393&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=1f6d5e21-f0ce-508d-13df-d1c50f7ea7b9&l=169

inc/saitypes.h
kcudnik: this will not pass metadata checks, unless we decide manually to make this change go + @lguohan

inc/saiport.h
```diff
@@ -3447,6 +3447,59 @@ typedef enum _sai_port_pool_attr_t
...
+     * @brief Enable DOT1P and CFI -> TC and COLOR MAP on port.
+     *
+     * Map id = #SAI_NULL_OBJECT_ID to disable map on port.
+     * Default no map.
+     *
+     * @type sai_object_id_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_QOS_MAP
+     * @allownull true
+     * @validonly SAI_PORT_ATTR_QOS_DOT1P_TO_TC_MAP == SAI_NULL_OBJECT_ID and SAI_PORT_ATTR_QOS_DOT1P_TO_COLOR_MAP == SAI_NULL_OBJECT_ID
```
kcudnik, line 3460: those conditions are not valid, you cat use oid values, in validonly ccondition you can use attribute which is enum, bool or int/uint, but not oid

helloanandhi:
Will present HLD first and will update header changes post community discussion in a separate PR.

PR #2026, opened 2024-06-05
[DASH] Add Routing Group API

marian-pritsak:
Add intermediate container object for outbound
routing entries to attach them in all together
to ENI.

### DIFF
--- a/experimental/saiexperimentaldashoutboundrouting.h
+++ b/experimental/saiexperimentaldashoutboundrouting.h
@@ -65,16 +65,16 @@ typedef struct _sai_outbound_routing_entry_t
     sai_object_id_t switch_id;
 
     /**
-     * @brief Exact matched key eni_id
-     *
-     * @objects SAI_OBJECT_TYPE_ENI
-     */
-    sai_object_id_t eni_id;
-
-    /**
      * @brief LPM matched key destination
      */
     sai_ip_prefix_t destination;
+
+    /**
+     * @brief Exact matched key outbound_routing_group_id
+     *
+     * @objects SAI_OBJECT_TYPE_OUTBOUND_ROUTING_GROUP
+     */
+    sai_object_id_t outbound_routing_group_id;
 
 } sai_outbound_routing_entry_t;
 
@@ -182,7 +182,7 @@ typedef enum _sai_outbound_routing_entry_attr_t
     SAI_OUTBOUND_ROUTING_ENTRY_ATTR_OVERLAY_DIP_MASK,
 
     /**
-     * @brief Action route_service_tunnel parameter OVERLAY_SIP
+     * @brief Action parameter overlay sip
      *
      * @type sai_ip_address_t
      * @flags CREATE_AND_SET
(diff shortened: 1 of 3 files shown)


experimental/saiextensions.h
```diff
@@ -81,6 +82,8 @@ typedef enum _sai_api_extensions_t
 
     SAI_API_DASH_HA,
 
+    SAI_API_DASH_ROUTING_GROUP,
```
kcudnik, line 85: why this is not added at the end ? such changes will not be backward compatible later

experimental/saitypesextensions.h
```diff
@@ -74,6 +74,8 @@ typedef enum _sai_object_type_extensions_t
 
     SAI_OBJECT_TYPE_HA_SCOPE,
 
+    SAI_OBJECT_TYPE_ROUTING_GROUP,
```
kcudnik, line 77: why this is not added at the end ? such changes will not be backward compatible later

experimental/saiextensions.h
marian-pritsak: @r12f is there a way to control ordering in this enum?

prsunny (approved):

experimental/saiextensions.h
r12f: The current logic of generation is that we will scan all the existing attributes and ignore if the attribute is already added.

Since you are hitting this issue, it should be caused by you have generated the SAI headers in multiple branches without resetting. If you revert all the SAI changes by `git reset --hard` and regenerate it, it will be added in the end.

experimental/saitypesextensions.h
r12f: this is the same problem as the API type. we need to reset the generated SAI header before generating the new one.

experimental/saiexperimentaldashroutinggroup.h
```diff
@@ -0,0 +1,136 @@
...
+    SAI_ROUTING_GROUP_ATTR_START,
+
+    /**
+     * @brief Action set_routing_group_attr parameter ADMIN_STATE
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_ROUTING_GROUP_ATTR_ADMIN_STATE = SAI_ROUTING_GROUP_ATTR_START,
```
r12f, line 55: I somehow remember this should be changed to disabled instead of admin state. But somehow it is missing?

r12f (changes requested):
Hi Marian, please check my comment in the PR. We will need some update in the code.

r12f (changes requested):
Hi Marian, please check my comment in the PR. We will need some update in the code.

experimental/saiexperimentaldashroutinggroup.h
```diff
@@ -0,0 +1,136 @@
...
+ * @param[inout] attr_list Array of attributes
+ *
+ * @return #SAI_STATUS_SUCCESS on success Failure status code on error
+ */
+typedef sai_status_t (*sai_get_routing_group_attribute_fn)(
+        _In_ sai_object_id_t routing_group_id,
+        _In_ uint32_t attr_count,
+        _Inout_ sai_attribute_t *attr_list);
+
+typedef struct _sai_dash_routing_group_api_t
```
r12f, line 122: since the routing entry in DASH is outbound_routing_entry, do you mind to change this to outbound_routing_group to make them align?

also, I believe it is better to merged into outbound routing header, we can use the annotation dash_api in the p4 file to specify the API group.

experimental/saiexperimentaldashoutboundrouting.h
```diff
@@ -65,16 +65,16 @@ typedef struct _sai_outbound_routing_entry_t
     sai_object_id_t switch_id;
 
     /**
-     * @brief Exact matched key eni_id
-     *
-     * @objects SAI_OBJECT_TYPE_ENI
+     * @brief LPM matched key destination
      */
-    sai_object_id_t eni_id;
+    sai_ip_prefix_t destination;
```
r12f, line 70: eni seem to be removed...

experimental/saiextensions.h
marian-pritsak: There's no longer a separate API, this change is removed

experimental/saitypesextensions.h
marian-pritsak: Moved to the end thanks to new api gen infra.

experimental/saiexperimentaldashroutinggroup.h
marian-pritsak: Changed to disabled

experimental/saiexperimentaldashroutinggroup.h
marian-pritsak: Changed and merged

r12f (approved):

kcudnik (approved):

PR #2027, opened 2024-06-06
[DASH] Update headers changes due to autogen

marian-pritsak:


experimental/saiswitchextensions.h
```diff
@@ -80,26 +80,11 @@ typedef void (*sai_ha_set_event_notification_fn)(
...
-    SAI_HA_SCOPE_EVENT_FLOW_RECONCILE_NEEDED,
-} sai_ha_scope_event_t;
-
 /**
  * @brief Notification data format received from SAI HA scope callback
  */
 typedef struct _sai_ha_scope_event_data_t
 {
-    /** Event type */
-    sai_ha_scope_event_t event_type;
```
kcudnik, line 101: such changes will not be backward compatible

experimental/saitypesextensions.h
```diff
@@ -74,6 +74,8 @@ typedef enum _sai_object_type_extensions_t
 
     SAI_OBJECT_TYPE_HA_SCOPE,
 
+    SAI_OBJECT_TYPE_ROUTING_GROUP,
```
kcudnik, line 77: similar here adding at the middle

experimental/saiexperimentaldasheni.h
```diff
@@ -699,14 +711,17 @@ typedef enum _sai_eni_stat_t
...
     SAI_ENI_STAT_INLINE_FLOW_DELETE_ACK_RECV,
 
-    /** DASH ENI INLINE_FLOW_DELETE_ACK_FAILED_RECV stat count */
-    SAI_ENI_STAT_INLINE_FLOW_DELETE_ACK_FAILED_RECV,
+    /** DASH ENI INLINE_FLOW_DELETE_ACK_FAILED stat count */
+    SAI_ENI_STAT_INLINE_FLOW_DELETE_ACK_FAILED,
 
-    /** DASH ENI INLINE_FLOW_DELETE_ACK_IGNORED_RECV stat count */
-    SAI_ENI_STAT_INLINE_FLOW_DELETE_ACK_IGNORED_RECV,
+    /** DASH ENI INLINE_FLOW_DELETE_ACK_IGNORED stat count */
```
kcudnik, line 723: changing enums names also wil not be backward compatible

experimental/saiswitchextensions.h
r12f: if we put this event data at the end, will this be backward compatible?

kcudnik: no xD chanigng struct size will not be backward compatible at all, but if you need to add something, add it at the end

r12f:
Yea, we need to at least get this change in before the generator produces non-compatible code: https://github.com/sonic-net/DASH/pull/578.

It is pending on https://github.com/sonic-net/DASH/pull/576, which is pending on https://github.com/sonic-net/DASH/pull/577.

PR #2031, opened 2024-06-12
Support to get list of all reachable switch-id's from a fabric port

harjotsinghpawra:
Adding support to get a list of all reachable switch-id's from a fabric port, currently we can only check from switch-id if its reachable or not .

SAI_PORT_ATTR_FABRIC_REACHABILITY_VECTOR


inc/saiport.h
```diff
@@ -2529,7 +2529,15 @@ typedef enum _sai_port_attr_t
...
-    SAI_PORT_ATTR_CUSTOM_RANGE_END
+    SAI_PORT_ATTR_CUSTOM_RANGE_END,
+
+    /**
+    * @brief Fabric port reachability for all remote reachable switches
+    *
+    * @type sai_u16_list_t
+    * @flags READ_ONLY
+    */
+    SAI_PORT_ATTR_FABRIC_REACHABILITY_VECTOR
```
kcudnik, line 2540: dont put attributes after SAI_PORT_ATTR_CUSTOM_RANGE_END,

harjotsinghpawra: HI kcidnik,
I tried putting it in middle , saimetadata checks were failing by saying enum shift , do i need to makes some changes to handle that ?

kcudnik:
whiuch enum shift ? current metadata complains about ident, you will need to squash your commits since validator is only checking each commit separetly

kcudnik:
try first pass check on your dev machine

PR #2032, opened 2024-06-12
Use newer action name for IRE meter attributes

marian-pritsak:
The DASH inbound routing entry has switched from old action names that used `VXLAN_DECAP` to new - `TUNNEL_DECAP`.
However, when adding those actions, some of the attributes' conditions were not updated, like meter attributes.
This PR brings all the remaining attributes to using the new names.


kcudnik:
please provide description what is it for

prsunny (approved):

r12f (approved):

marian-pritsak:
> please provide description what is it for

Done

PR #2046, opened 2024-07-10
Add label attribute for Policer, RIF, Scheduler and UDF

erohsik:
Add label attribute to a few SAI objects that will help during warmboot object matching.

### DIFF
--- a/inc/saipolicer.h
+++ b/inc/saipolicer.h
@@ -232,6 +232,15 @@ typedef enum _sai_policer_attr_t
     SAI_POLICER_ATTR_SELECTIVE_COUNTER_LIST,
 
     /**
+     * @brief Label attribute used to uniquely identify identical policers.
+     *
+     * @type char
+     * @flags CREATE_AND_SET
+     * @default ""
+     */
+    SAI_POLICER_ATTR_LABEL,
+
+    /**
      * @brief End of attributes
      */
     SAI_POLICER_ATTR_END,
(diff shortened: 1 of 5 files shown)


kcudnik:
Does this needs to be a string ? Or it would be better to make this uint32 or 64? Maybe it's discussion for sai community meeting

erohsik:
> Does this needs to be a string ? Or it would be better to make this uint32 or 64? Maybe it's discussion for sai community meeting

As discussed in the community meeting, the string provides the option to use the key from the application/config db entry..

rlhui:
@JaiOCP , @rck-innovium , please help review? thanks.

rck-innovium (approved):

JaiOCP (approved):

tjchadaga:
@erohsik - could you please resolve branch conflicts?

tjchadaga:
> > > Does this needs to be a string ? Or it would be better to make this uint32 or 64? Maybe it's discussion for sai community meeting
> > 
> > 
> > As discussed in the community meeting, the string provides the option to use the key from the application/config db entry..
> 
> what config db entry ? SAI don't have any idea about any config db entry

@kcudnik - Since the label is populated from the application/NOS, the general consensus was that a string would be easier to manage. Is there a concern with using a string label?

kcudnik:
ok, i got it now it's just a label, please resolve conflicts

kcudnik:
You have spelling issue

kcudnik (approved):

PR #2049, opened 2024-07-10
SAI Egress DSCP action behavior

siqbal1986:
This document des not define any new behavior for the SAI DSCP change capabilty. It only solidifies the existing the behavior of the packet pipelines observed.
This document lists different places in the packet pipeline where the Packet DSCP value can be changed. It further clearifies how changing of the DSCP value at one stage may be overridden by the next stage and shows how changing this value at different stages may have impact on the final packet egressing the pipeline.


doc/QOS/SAI-Proposal-Egress-DSCP.md
```diff
@@ -0,0 +1,58 @@
...
+The DSCP value set by various components on the egress side of the ASIC pipeline shall have the following precedence in descending order. This precedence shall apply to packets in the egress pipeline and shall override any other DSCP value set by the Ingress stage.
+
+1. Egress ACL.
+
+2. Tunnel DSCP mapping 
+    - Tunnel TC-COLOR -> DSCP mapping
+    - Tunnel DSCP mode
+    - Switch Tunnel TC-COLOR -> DSCP map
+
+3. Port Qos TC-COLOR -> DSCP map
```
helloanandhi, line 49: Between Egress ACL and Port, Precedence could be based on the location of the component in the ASIC pipeline. So fixing a precedence could be biased for certain ASICs. Suggest to have some query mechanism to know the ASIC capability  for this.

doc/QOS/SAI-Proposal-Egress-DSCP.md
```diff
@@ -0,0 +1,58 @@
...
+
+4. Switch Qos TC-COLOR -> DSCP map
+
+## Examples of packet behavior. ##
+
+These following is a logical diagram which shows how the DSCP value of a packet may be changed by different components. This diagram does not represent the actual layout of the pipeline.
+
+<img src="../figures/egress_dscp_change_precedence.png" width="1200">
```
helloanandhi, line 57: Covering both Ingress and Egress can provide a complete view.

doc/QOS/SAI-Proposal-Egress-DSCP.md
siqbal1986: I have updated the proposal with the suggestions. I have also opened new issues to track the progress of the suggestions.

doc/QOS/SAI-Proposal-Egress-DSCP.md
siqbal1986: the doc has been updated and new issues created.

PR #2050, opened 2024-07-12
Update DASH SAI API comments

r12f:
This PR is intended to update the current comments of DASH SAI APIs.

Previously, all DASH SAI API comments are generated from P4 code using very simple rule, such as all words to upper case to pass the SAI checks. Now, we switched to python for better comment generation, which makes the comment more meaningful and readable.

### DIFF
--- a/inc/saiobject.h
+++ b/inc/saiobject.h
@@ -37,9 +37,6 @@
 #include <saisrv6.h>
 
 /* new experimental object type includes */
-#include <saiexperimentaldashha.h>
-#include <saiexperimentaldashmeter.h>
-#include <saiexperimentaldashtunnel.h>
 #include <saiexperimentaldashvip.h>
 #include <saiexperimentaldashpavalidation.h>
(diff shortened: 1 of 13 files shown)


Pterosaur (approved):

Pterosaur:
Hi @prsunny , could you please help review and merge this PR?

mukeshmv:
lgtm. This brings these SAI APIs in sync with the DASH generated APIs

marian-pritsak (approved):

inc/saiobject.h
```diff
@@ -37,6 +37,9 @@
 #include <saisrv6.h>
 
 /* new experimental object type includes */
+#include <saiexperimentaldashha.h>
```
kcudnik, line 40: This is not needed here, no changes are made to this file so no object includes are required

r12f: Thanks Kamil. I have removed these lines now.

PR #2051, opened 2024-07-12
Update DASH pipeline and HA related counters.

r12f:
These counters are recently added for checking drops in DASH pipeline.

It follows the DASH HLD here: https://github.com/sonic-net/DASH/blob/main/documentation/high-avail/ha-api-hld.md.


mukeshmv:
lgtm

marian-pritsak (approved):

PR #2052, opened 2024-07-12
[DASH]: Reorder outboundcatopa and outboundrouting attributes for ABI

Pterosaur:
There are two PRs: https://github.com/opencomputeproject/SAI/pull/2035 and https://github.com/opencomputeproject/SAI/pull/2025 that inserted attributes in the middle of the attributes enum which does break the ABI compatibility.
So, I reorder them to the tail of the attributes enum to keep the ABI compatibility.


Pterosaur:
@r12f Please take a look.

Pterosaur:
@prsunny @kcudnik , Could you please help to review this PR?

r12f (approved):
lgtm.

r12f (commented):
actually, this is not DASH repo. so we need to change DASH SAI spec to align with merged API.

PR #2056, opened 2024-07-24
Convert DASH meter bucket object to table entry

mukeshmv:
Bring in SAI API changes from this DASH PR
https://github.com/sonic-net/DASH/pull/605


KrisNey-MSFT:
Could someone add @prsunny and @marian-pritsak to the Reviewers here?

marian-pritsak (approved):

meta/style.pm
```diff
@@ -215,15 +215,21 @@ sub CheckStatsFunction
 
     if (not $fnparams =~ /^\w+_id number_of_counters counter_ids( (mode )?counters)?$/)
     {
-        LogWarning "invalid stat function $fname params names: $fnparams";
+        if (not $fnparams =~ /^\w+_entry number_of_counters counter_ids( (mode )?counters)?$/)
```
r12f, line 218: @mukeshmv  as I mentioned in the DASH PR, the right way to do this is to check in this file only and bring it to DASH. Now, the DASH SAI is diverged from OCP SAI, because it is not superset anymore. Any future DASH update is going to have a chance to break this feature, as the generated API is still going to use the old SAI object format for metering bucket. It is not good to leave DASH in this state. Can you help bring this change back to DASH?

Hi @prsunny , do you mind to help track this? I am OOF and My internet access is horrible, so I won't be able to do this myself.

mukeshmv: This change is not sufficient to get the Meter bucket change in since this is not a backward compatible change. Hence we had to resort to pushing this directly to SAI. I have a new PR open in Dash to pull this in along with your original commit that should resolve this.

kcudnik: This is not backward compatible ! Remove !!

kcudnik:
This style change is not backward compatible

kcudnik:
Please. Remove style change

kcudnik:
@prsunny don't merge changes like this !!

meta/style.pm
mukeshmv: @kcudnik, could you elaborate why you think this style change is not backward compatible ? The change actually relaxes the check by allowing another variation in addition to the existing stats API. I don't see how this breaks backward compatibility.

mukeshmv: Change is equivalent to adding this to existing check - allowing an extra pattern. 

```
-    if (not $fnparams =~ /^\w+_id number_of_counters counter_ids( (mode )?counters)?$/)
+    if (not $fnparams =~ /^\w+_id number_of_counters counter_ids( (mode )?counters)?$/ and
+        not $fnparams =~ /^\w+_entry number_of_counters counter_ids( (mode )?counters)?$/)
     {
         LogWarning "invalid stat function $fname params names: $fnparams";
     }
```

If you want I can change it to this.

prsunny:
@mukeshmv , lets remove the style change and only have the dash changes in these PRs. all other changes other than dash should go as seperate PR/approval

prsunny:
> This style change is not backward compatible

@kcudnik , can you check Mukesh's comment? If required, we can revert this PR.

kcudnik:
Why added API is not following proposed standard ?

mukeshmv:
> Why added API is not following proposed standard ?

@kcudnik, the reason is explained in @r12f 's Dash PR https://github.com/sonic-net/DASH/pull/605

jimmyzhai:
@mukeshmv, @r12f, @prsunny, @kcudnik the style change relaxes stats API check. While compiling sonic-sairedis with this PR, it will fail. SaiInterface ([meta/SaiInterface.h](https://github.com/sonic-net/sonic-sairedis/blob/master/meta/SaiInterface.h)) still does not support stats API for sai_xxx_entry.

PR #2010, opened 2024-05-08
Add sai-thrift support to enable/disable CREDIT WATCHDOG  for Voq switch

saksarav-nokia:
sonic-mgmt Qos tests disables the TX for the destination ports to induce the shared buffer full, nominal headroom full a, ingress drop and egress drop.

Voq systems like BCM DNX has credit watchdog timer running periodically to check if voq's and flush the data in the voq to prevent the data being stuck forever in the voq. So to avoid the credit watchdog timer flushing the voq during sonic-mgmt Qos tests, we need to disable the credit watchdog timer and enable it back after the test.

The SAI attribute SAI_SWITCH_ATTR_CREDIT_WD was not present in the older SAI versions and got added in SAI 10.X. So this PR adds the sai-thrift support to enable or disable the SAI_SWITCH_ATTR_CREDIT_WD for Voq systems.

PR #2072, opened 2024-08-28
Add sai-thrift support to enable/disable CREDIT WATCHDOG  for Voq switch (cherry-pick PR #2010)

saksarav-nokia:
sonic-mgmt Qos tests disables the TX for the destination ports to induce the shared buffer full, nominal headroom full a, ingress drop and egress drop.

Voq systems like BCM DNX has credit watchdog timer running periodically to check if voq's and flush the data in the voq to prevent the data being stuck forever in the voq. So to avoid the credit watchdog timer flushing the voq during sonic-mgmt Qos tests, we need to disable the credit watchdog timer and enable it back after the test.

The SAI attribute SAI_SWITCH_ATTR_CREDIT_WD was not present in the older SAI versions and got added in SAI 10.X. So this PR adds the sai-thrift support to enable or disable the SAI_SWITCH_ATTR_CREDIT_WD for Voq systems.

### DIFF
--- a/test/saithrift/src/switch_sai_rpc_server.cpp
+++ b/test/saithrift/src/switch_sai_rpc_server.cpp
@@ -1596,6 +1596,29 @@ public:
       }
       attr_list.push_back(thrift_port_list_attribute);
       free(port_list_object_attribute.value.objlist.list);
+
+      sai_attribute_t switch_attr;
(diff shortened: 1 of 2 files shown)


saksarav-nokia:
@abdosi @kcudnik , pls review and approve.
This is cherry-pick from SAI master to v1.14 for 202405

abdosi (approved):

PR #2074, opened 2024-09-07
Allow sparse mode fdb mac program on vlan member

luhting:
There is a single bridge port on the port, and this bridge port, along with a VLAN, is used to create a VLAN member. When a MAC address is learned with a specific VLAN on a bridge port that has multiple VLAN members, it is necessary to specify the VLAN member.


luhting:
There is no longer needed.

PR #2078, opened 2024-09-19
Round-robin packet spraying and packet spraying selection per ECMP/LAG

dkadia-marvell:
This proposal introduces the below two types of light weight *static* member selection modes for ECMP/LAG.

 * Random 
 * Round Robin


kcudnik (approved):

tjchadaga:
@dkadia-marvell, @rck-innovium - any conclusion on this? does this need another discussion?

kcudnik (approved):

rck-innovium:
Thanks @JaiOCP  for the discussion. Please find the summary below:

### Summary of the follow-up discussion on 02/27/2025

- **Offline Discussion with SAI Vendors:** The consensus was to use SAI_HASH_ALGORITHM_RANDOM for per-packet random spraying. We have expanded the existing comments to ensure this is unambiguous.

- **Hash Algorithm Selection:** We agreed on the need to select the SAI hash algorithm at either the LAG level or the Nexthop Group (NHG) level. The use case presented involved a TOR switch connected to multiple appliances, each via a LAG. Some appliances can handle packet reordering, while others cannot. This approach can also be extended to NHG if the appliance is connected via ECMP.

- **Concerns Raised by Jai:** Jai raised a concern about exposing the hash algorithm at the NHG/LAG level. In some devices, this could lead to running out of hash generation blocks (e.g., some NHGs use CRC, some use XOR, some use CRC_XOR, etc.). We agreed to address this concern in a separate proposal to discover capacity/CRM.


> @JaiOCP @tjchadaga
> 
> ### Summary of the discussion on 11/14/2024
> 1. **[Jai]** The option to enable packet spraying is in practice needed at the Switch level and not at a per Next Hop Group level.
> 2. **[Jai]** A possible interpretation of  SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_ALGORITHM:SAI_HASH_ALGORITHM_RANDOM is to enable packet spraying at a switch level.
>    **[Ravi]**  The only description available in the community is: _/** Random-based hash algorithm */._ This suggests it is still a hash algorithm i.e., it hashes on some fields of the packet. A possible example of such an algorithm could be: https://patents.google.com/patent/US8503456
> 
> ### Next steps agreed:
> **[Ravi]** and **[Jai]** to see if there are use-cases for enabling packet-spraying at a per NHG granularity instead of switch level. We should also see what happens to telemetry packets, control packets etc. We will discuss offline over the PR and if needed use a follow up meeting.

inc/saiacl.h
```diff
@@ -3327,10 +3331,19 @@ typedef enum _sai_acl_entry_attr_t
...
+     * @flags CREATE_AND_SET
+     * @default disabled
+     */
+    SAI_ACL_ENTRY_ATTR_ACTION_SET_ECMP_HASH_ALGORITHM = SAI_ACL_ENTRY_ATTR_ACTION_START + 0x3c,
+
     /**
      * @brief End of Rule Actions
      */
-    SAI_ACL_ENTRY_ATTR_ACTION_END = SAI_ACL_ENTRY_ATTR_ACTION_SET_INNER_DST_MAC,
+    SAI_ACL_ENTRY_ATTR_ACTION_END = SAI_ACL_ENTRY_ATTR_ACTION_SET_ECMP_HASH_ALGORITHM,
```
kcudnik, line 3346: maybe we should resign with setting this value explicitly, and just leave it as last attribute to not have to change that explicitly every time

JaiOCP (approved):

PR #2079, opened 2024-09-27
Add ACL actions for setting inner MACs

marian-pritsak:
Add actions SET_INNER_SRC_MAC and SET_INNER_DST_MAC for tunneled packets to be able to override inner
values independently from router based on ACL policy.

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -291,6 +291,12 @@ typedef enum _sai_acl_action_type_t
 
     /** Disable packet trim */
     SAI_ACL_ACTION_TYPE_PACKET_TRIM_DISABLE = 0x00000039,
+
+    /** Set Packet inner Src MAC Address */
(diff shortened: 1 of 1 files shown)


prsunny (approved):

anish-n (approved):

kcudnik (approved):

tjchadaga:
@JaiOCP, @rck-innovium - Could you please help review/sign off on this?

kcudnik (approved):

mclayto:
My understanding is that SET_INNER_* does not apply to a soon-to-be-inner and so is a no-op if the inner field does not exist at the ACL stage where it is applied.

mclayto (approved):

tjchadaga:
@marian-pritsak - could you please help resolve conflicts?

PR #2082, opened 2024-09-30
SAI Egress DSCP action behavior

siqbal1986:
This document des not define any new behavior for the SAI DSCP change capabilty. It only solidifies the existing the behavior of the packet pipelines observed.
This document lists different places in the packet pipeline where the Packet DSCP value can be changed. It further clearifies how changing of the DSCP value at one stage may be overridden by the next stage and shows how changing this value at different stages may have impact on the final packet egressing the pipeline.
Clean commit created.
Orignal PR reviewed https://github.com/opencomputeproject/SAI/pull/2049


siqbal1986:
@helloanandhi The issues raised in orignal proposal have been formalized in the document.
https://github.com/opencomputeproject/SAI/pull/2049

PR #2088, opened 2024-10-09
Update SAI version to v1.15.0

tjchadaga:
SAI version update to v1.15.0


rlhui (approved):

kcudnik:
i wish you could wait with merging that, i have a discussion on todays SAI meeting to discuss whether we want to go with breaking change on sai_port_oper_status_t struct change

PR #2087, opened 2024-10-09
Introduce new extended port oper status notification

kcudnik:
Bring back compatibility with old structure.

So we are actually in luck, since sizeof(sai_port_oper_status_notification_t) == 16, because of port_id being 64 bit aligned, also as sizeof(sai_extended_port_oper_status_notification_t) == 16, which means we maybe don't need this PR

PR #2089, opened 2024-10-11
SAI Proposal TAM stream telemetry

Pterosaur:
Proposal a new streaming method for telemetry


inc/saitam.h
```diff
@@ -1334,6 +1334,30 @@ typedef enum _sai_tam_report_attr_t
...
+
+    /**
+     * @brief Query IPFIX template
+     *
+     * Return the IPFIX template binary buffer
+     *
+     * @type sai_u8_list_t
+     * @flags READ_ONLY
+     */
+    SAI_TAM_REPORT_ATTR_IPFIX_TEMPLATE,
```
r12f, line 1359: for sessions that exceeds the max length of template or data, we need to split the template into multiple templates:

- Modify the template type to sai_u8_list_list_t
- Rename SAI_TAM_REPORT_ATTR_IPFIX_TEMPLATE to SAI_TAM_REPORT_ATTR_IPFIX_TEMPLATE_LIST
- Rename SAI_TAM_REPORT_ATTR_REPORT_IPFIX_TEMPLATE_ID to SAI_TAM_REPORT_ATTR_REPORT_IPFIX_BASE_TEMPLATE_ID

Each template data has the template id in it, so we don't need another attribute or field to specify the template id list.

inc/saitam.h
```diff
@@ -1472,9 +1514,44 @@ typedef enum _sai_tam_telemetry_attr_t
...
+     * @brief Tam telemetry cache size
+     *
+     * If the collector isn't ready to receive the report, this value indicates how many
+     * reports that can be cached. 0 means no cache which is the default behavior.
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_TAM_TELEMETRY_ATTR_CACHE_SIZE,
```
r12f, line 1553: since the ring buffer is created on driver load, we need to move it to sai switch attribute.

inc/saitam.h
```diff
@@ -1426,6 +1450,23 @@ typedef enum _sai_tam_reporting_unit_t
...
+{
+    /**
+     * @brief Report type is time based
+     */
+    SAI_TAM_REPORTING_TYPE_TIME_BASED,
+
+    /**
+     * @brief Report type is count based
+     */
+    SAI_TAM_REPORTING_TYPE_COUNT_BASED,
```
r12f, line 1466: rename to SIZE_BASED to be more explicit.

inc/saitam.h
```diff
@@ -1472,9 +1514,44 @@ typedef enum _sai_tam_telemetry_attr_t
...
+     *
+     * defines the size of reporting chunk, which means TAM will report to the collector every time
+     * if the report count reaches the chunk size.
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 1
+     * @validonly SAI_TAM_TELEMETRY_ATTR_TAM_REPORTING_TYPE == SAI_TAM_REPORTING_TYPE_COUNT_BASED
+     */
+    SAI_TAM_TELEMETRY_ATTR_REPORTING_CHUNK_SIZE,
```
r12f, line 1541: CHUNK_BYTES

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,798 @@
...
+sai_attr_list[0].value.oid = sai_tam_tel_type_obj;
+
+sai_attr_list[1].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_OBJECT_ID;
+sai_attr_list[1].value.oid = port_obj;
+
+sai_attr_list[2].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_STAT_ID;
+sai_attr_list[2].value.oid = SAI_PORT_STAT_IF_IN_OCTETS;
+
+sai_attr_list[3].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_LABEL;
+sai_attr_list[3].value.oid = index; // Element ID of the object in the IPFIX template
```
r12f, line 731: we need to add explicit capability check here instead of leveraging the return value.

for capability query, we need to add a new API in saiobject.h:

```c
sai_status_t sai_query_stats_hft_capability(
        _In_ sai_object_type_t object_type,
        _Inout_ sai_stat_hft_capability_list_t *stats_capability);
```

Then in saitypes.h, add the sai_stat_hft_capability_list_t following the sai_stat_capability_list_t, and we need to add the minimum polling interval in it.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,798 @@
...
+
+attr_count = 4;
```
r12f, line 733: to support the get-and-clear operation on queue watermark counters, we need another attribute to specify the mode.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,798 @@
...
+sai_attr_list[0].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_TEL_TYPE;
+sai_attr_list[0].value.oid = sai_tam_tel_type_obj;
+
+sai_attr_list[1].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_OBJECT_ID;
+sai_attr_list[1].value.oid = port_obj;
+
+sai_attr_list[2].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_STAT_ID;
+sai_attr_list[2].value.oid = SAI_PORT_STAT_IF_IN_OCTETS;
+
+sai_attr_list[3].id = SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_LABEL;
```
Pterosaur, line 730: Update label comment on sai.h

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,798 @@
...
+The existing telemetry solution relies on a process to proactively query stats and counters via the SAI API. This approach causes the process to spend excessive time on SAI communication. The stream telemetry described in this document aims to provide a more efficient method for collecting object stats. The main idea is that selected stats will be proactively pushed from the vendor's driver to the collector via netlink.
+
+## Requirements
+
+- The number of SAI object types should not exceed 32,768 ($2^{15}$). This means the value of SAI_OBJECT_TYPE_MAX should be less than 32,768.
+- The number of SAI object extension types should not exceed 32,768.
+- The number of stats types for a single SAI object type should not exceed 32,768.
+- The number of extension stats types for a single SAI object type should not exceed 32,768.
+- The number of SAI objects of the same type should not exceed 32,768.
+- The vendor SDK should support publishing stats in IPFIX format and its IPFIX template.
```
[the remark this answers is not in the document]
Pterosaur: Because the IPFIX recording is published by the vendor, So, according to the self-consistency principle, the IPFIX template will also be published from the vendor side as well. For example, we don't limit the order of data in the IPFIX , it's acceptable once the order of date is consistent in the IPFIX template or recording.
We expect the generation of all IPFIX template or reocrding is maintained by one side.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,798 @@
...
+- If a polling frequency for stats cannot be supported, the vendor's SDK should report this error.
```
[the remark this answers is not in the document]
Pterosaur: Sure, agree. Will provide a new API to query this capability

inc/saiobject.h
```diff
@@ -294,6 +294,20 @@ sai_status_t sai_query_stats_capability(
...
+/**
+ * @brief Query statistics capability for statistics bound at object level under the stream telemetry mode
+ *
+ * @param[in] switch_id SAI Switch object id
+ * @param[in] object_type SAI object type
+ * @param[inout] stats_capability List of implemented enum values, the statistics modes (bit mask) supported and minimal polling interval per value
+ *
+ * @return #SAI_STATUS_SUCCESS on success, #SAI_STATUS_BUFFER_OVERFLOW if lists size insufficient, failure status code on error
+ */
+sai_status_t sai_query_stats_st_capability(
```
kcudnik, line 306: maybe we should make this more generic ? since very similar function already exists and this is adding only 1 field to structure

r12f: hi Kamil, we discussed internally using the sai_query_stats_capability directly and add the polling interval in it, but the reason we are not using that API is because of 2 reasons:

1. Not all counters support using this way to retrieve, so if the counters are not returned in the list, it means it is not supported.
2. Adding another field in the current stats capability breaks the ABI and making this frequently used function not backward compatible.

So after discussing it, we ended up deciding creating a new API for it.

rck-innovium (changes requested):
SAI already supports streaming telemetry at sub-second intervals. Please refer: https://github.com/opencomputeproject/SAI/pull/1670/files

SAI also supports sending packets to local collector over GENETLINK. Please refer: https://github.com/opencomputeproject/SAI/pull/1986/files

1) Title
The current PR is adding some enhancements to the already supported TAM streaming telemetry to local collector. The title is misleading, it is incorrectly suggesting that  streaming telemetry is so far not supported in SAI. Maybe rename to "TAM Streaming Telemetry Count mode".

2) List of enhancements
It would be good to clearly call out the new SAI attributes and enums being defined, their meaning and why they are needed.
The new attributes/enums and APIs are limited to just the below functionality but this is not clearly called out at the beginning of the document like other SAI specs.
- Bulking based on count
- SAI vendor to publish IPFIX templates
- Switch TAM Cache (motivation and semantics-FIFO/LIFO are unclear)
- Streaming Stats Capability query

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,906 @@
...
+```
+
+- If the number of stats in a group is small, multiple snapshots may be encoded into a single IPFIX message.
+- If the number of stats in a group exceeds 8K, the group must be split across multiple IPFIX messages.
+
+The IPFIX template should be provided by vendors. This document does not restrict how to split or concatenate snapshots, but each separated snapshot must include its own `observationTimeNanoseconds`.
+
+#### Netlink message
+
+We expect all control messages and out-of-band information to be transmitted by the SAI. Therefore, it is unnecessary to read the attribute header of netlink and the message header of Genetlink from the socket. Instead, we can insert a bulk of IPFIX recordings as the payload of the netlink message. The sample code for building the message from the kernel side is as follows:
```
rck-innovium, line 270: Can you explain more on this? When defining new Generic Netlink message formats it is recommended to make use of the Netlink attributes wherever possible. The Netlink attribute mechanism has been carefully designed to allow for future message expansion while preserving backward compatibility. In order to follow the GENETLINK semantics, we should at the least have one GENETLINK attribute say "STEL\_ATTRIBUTE\_IPFIX\_ENTRIES" and define its type, say 'nested'.  The example code is adding the IPFIX entries without using any attributes.

Standardizing the attribute name and type allows the NOS code to operate across different SAI vendors.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,906 @@
...
+     * @brief Tam telemetry cache count
+     *
+     * If the collector isn't ready to receive the report, this value indicates how many
+     * reports that can be cached. 0 means no cache which is the default behavior.
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_ONLY
+     * @default 0
+     */
+    SAI_SWITCH_ATTR_TAM_CACHE_COUNT,
```
rck-innovium, line 451: What is the caching semantics? Say the collector is not ready, and more reports are generated than the cache capacity, are the first 'n' or the last 'n' cached?

Please also explain why this is required?

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,906 @@
...
+sai_create_tam_collector_fn(&sai_tam_collector_obj, switch_id, attr_count, sai_attr_list);
+
+```
+
+#### Creating TAM report object
+
+``` c++
+
+sai_attr_list[0].id = SAI_TAM_REPORT_ATTR_TYPE;
+sai_attr_list[0].value.s32 = SAI_TAM_REPORT_TYPE_IPFIX;
```
rck-innovium, line 569: The report type should be GENETLINK, and the encoding within the GENETLINK is IPFIX. The latter part is already specified using SAI\_HOSTIF\_ATTR\_GENETLINK\_MCGRP\_NAME.

Pterosaur: IMHO, this attribute is for specifying the format of message, Other values of this attribute is more like the format instead of channel.
image

Pterosaur: And I didn't see an attribute for encoding format if not to specify at here.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,906 @@
...
+    {
+        struct ipfix *msg = ring_buffer.pop();
+        if (msg == NULL)
+        {
+            break;
+        }
+        nla_append(skb_out, msg->data, msg->len);
+    }
+
+    genlmsg_multicast(&genl_family, skb_out, 0, 0/* group_id to ipfix group */, GFP_KERNEL);
```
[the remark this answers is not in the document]
Pterosaur: Thanks for your comment. It's a typo. Change it to stel_family.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
Pterosaur: I think it's intentionally designed. Because this genetlink is specific for stream telemetry with IPFIX, we don't want to pass any other information, including IPFIX template, on this Channel. We expect other information should only pass by SAI. If this channel is designed more flexible, we have to define/design another kind of schema out of SAI for it. That's what we want to avoid.

I don't feel it's not standardizing. Because this API is just provided by the netlink library. Why you said it's not standardizing?

Pterosaur:
> SAI already supports streaming telemetry at sub-second intervals. Please refer: https://github.com/opencomputeproject/SAI/pull/1670/files
> 
> SAI also supports sending packets to local collector over GENETLINK. Please refer: https://github.com/opencomputeproject/SAI/pull/1986/files
> 
> 1. Title
>    The current PR is adding some enhancements to the already supported TAM streaming telemetry to local collector. The title is misleading, it is incorrectly suggesting that  streaming telemetry is so far not supported in SAI. Maybe rename to "TAM Streaming Telemetry Count mode".
> 2. List of enhancements
>    It would be good to clearly call out the new SAI attributes and enums being defined, their meaning and why they are needed.
>    The new attributes/enums and APIs are limited to just the below functionality but this is not clearly called out at the beginning of the document like other SAI specs.
> 
> * Bulking based on count
> * SAI vendor to publish IPFIX templates
> * Switch TAM Cache (motivation and semantics-FIFO/LIFO are unclear)
> * Streaming Stats Capability query

Make sense, Thanks for your suggestion. I will change it.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,906 @@
...
+We expect all control messages and out-of-band information to be transmitted by the SAI. Therefore, it is unnecessary to read the attribute header of netlink and the message header of Genetlink from the socket. Instead, we can insert a bulk of IPFIX recordings as the payload of the netlink message. The sample code for building the message from the kernel side is as follows:
+
+``` c
+
+struct genl_multicast_group stel_mcgrps[] = {
+    { .name = "ipfix" },
+};
+
+// Family definition
+static struct genl_family nl_bench_family = {
```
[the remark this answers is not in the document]
Pterosaur: It should be `stel_family` that stands for `stream telemetry family`. nl_bench_family is a typo that I copied from a sample code.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
Pterosaur: Thanks for your comments. Update the comments.

Pterosaur: We hope it's possible to collect the counters during the collector restarts after the collector crash.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,907 @@
...
+192-192: "1"
+193-207: "Element ID = Object index for the stats N"
+208-223: "Field Length = 8 bytes"
+224-255: "Enterprise Number = SAI TYPE ID + SAI STATS ID for the stats N"
+
+```
+
+- For high-frequency counters, the native IPFIX timestamp unit of seconds is insufficient. Therefore, we introduce an additional element, `observationTimeNanoseconds`, for each record to meet our requirements.
+- The enterprise bit is always set to 1 for stats records.
+- The element ID of IPFIX is derived from the object index. For example, for `Ethernet5`, the element ID will be `0x5 | 0x8000 = 0x8005`.
```
rck-innovium, line 167: Please explain where the 0x8000 comes from.

Pterosaur: It indicates the enterprise bit is set to 1. Thanks for your pointing this unclear part. Polish it.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,907 @@
...
+| SAI          | The Switch Abstraction Interface          |
+| IPFIX        | Internet Protocol Flow Information Export |
+| TAM          | Telemetry and Monitoring                  |
+| BW           | Bandwidth                                 |
+
+## Overview
+
+The existing telemetry solution relies on a process to proactively query stats and counters via the SAI API. This approach causes the process to spend excessive time on SAI communication. The stream telemetry described in this document aims to provide a more efficient method for collecting object stats. The main idea is that selected stats will be proactively pushed from the vendor's driver to the collector via netlink.
+
+## Requirements
```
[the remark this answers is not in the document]
Pterosaur: Yes, correct. This template of document is just following [the SONiC hld](https://github.com/sonic-net/SONiC/blob/master/doc/hld_template.md) . As it said, the requirements section includes exemptions (not supported). But I will update the constraints title for this section for more clear.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,906 @@
...
+sai_attr_list[0].value.objlist.list[0] = sai_tam_telemetry_obj;
+
+sai_attr_list[1].id = SAI_TAM_ATTR_TAM_BIND_POINT_TYPE_LIST;
+sai_attr_list[1].value.objlist.count = 2;
+sai_attr_list[1].value.objlist.list[0] = SAI_TAM_BIND_POINT_TYPE_PORT;
+sai_attr_list[1].value.objlist.list[0] = SAI_TAM_BIND_POINT_TYPE_QUEUE;
+
+attr_count = 2;
+sai_create_tam_fn(&sai_tam_obj, switch_id, attr_count, sai_attr_list);
+
```
rck-innovium, line 790: The sequence is incomplete. We need to bind sai\_tam\_obj to port/queue/switch etc. Pls refer SAI\_PORT\_ATTR\_TAM\_OBJECT.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,906 @@
...
+``` c++
+
+sai_attr_list[0].id = SAI_TAM_REPORT_ATTR_TYPE;
+sai_attr_list[0].value.s32 = SAI_TAM_REPORT_TYPE_IPFIX;
+
+sai_attr_list[1].id = SAI_TAM_REPORT_ATTR_REPORT_MODE;
+sai_attr_list[1].value.s32 = SAI_TAM_REPORT_MODE_BULK;
+
+sai_attr_list[2].id = SAI_TAM_REPORT_ATTR_REPORT_INTERVAL;
+sai_attr_list[2].value.u32 = poll_interval; // STREAM_TELEMETRY_PROFILE:profile_name[poll_interval] on Config DB
```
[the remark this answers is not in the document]
Pterosaur: Do you know what the difference is between `SAI_TAM_TELEMETRY_ATTR_REPORTING_INTERVAL` and `SAI_TAM_REPORT_ATTR_REPORT_INTERVAL`?
IMO, `SAI_TAM_REPORT_ATTR_REPORT_INTERVAL` is about the sampling rate under the ASIC and `SAI_TAM_TELEMETRY_ATTR_REPORTING_INTERVAL` is about the frequency of reporting to the user's collector.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
Pterosaur: Yes, it's not a complete list. This list is fully depended on which counters we subscribed. 

That's why I said: 
https://github.com/opencomputeproject/SAI/pull/2089/files#diff-0e0ad14d84f3332634521f6b2d53007b4f7ccec6c53e003139ef77da1a4d9d50R373

BTW, why do you mention `SAI_PORT_ATTR_TAM_OBJECT`? I feel that's another communication mode for TAM. In our approach, we don't need to set this attribute.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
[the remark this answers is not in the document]
rck-innovium: @JaiOCP Can you help clarify the difference `SAI_TAM_TELEMETRY_ATTR_REPORTING_INTERVAL` and `SAI_TAM_REPORT_ATTR_REPORT_INTERVAL?`

The interpretation in this proposal is that the report attribute (`SAI_TAM_REPORT_ATTR_REPORT_INTERVAL` ) is the h/w measuring/sampling interval, and the Telemetry attribute (`SAI_TAM_TELEMETRY_ATTR_REPORTING_INTERVAL)`is the interval between reports (i.e., batching interval).

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,906 @@
...
+packet-beta
+0-15: "Version = 0x000a"
+16-31: "Message Length = (16 + payload) bytes"
+32-63: "Export Timestamp: Second"
+64-95: "Sequence Number = 0, start from 0 and incremental sequence counter modulo 2^32"
+96-127: "Observation Domain ID = 0, always 0"
+
+```
+
+#### IPFIX template
```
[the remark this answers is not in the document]
Pterosaur: We only expect that the IPFIX template must include these stat-IDs that we explicitly configured. But if there are more entries included in the template, I feel we can ignore it.

But I feel there are two concerns:
1. what's the label(element ID) will be assigned? Because we pass the label to vendor's SDK via `SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_STAT_ID`
2. If it includes too much un-required data, the efficiency may be decreased dramatically.

@r12f Please correct me if you have any suggestions.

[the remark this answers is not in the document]
rajkumar38: > 2\. If it includes too much un-required

1. Since label and SAI_TAM_COUNTER_SUBSCRIPTION_ATTR_STAT_ID have fixed encoding, SAI implementation can pass the additional info (that are not subscribed) in the same format and these additional records should be ignored.
2. Agree that un-wanted data will cause performance issues.

In general,  this proposal talks about only SAI_TAM_TELEMETRY_TYPE_COUNTER_SUBSCRIPTION. Do you have plans to extend this proposal to support other telemetry types ?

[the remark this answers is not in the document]
Pterosaur: What's the meaning about other telemetry types?
We have `sai_tam_report_type_t`, which can support 
```c
// ...
    /**
     * @brief Report using IPFIX
     */
    SAI_TAM_REPORT_TYPE_IPFIX,

    /**
     * @brief Report using GPB
     */
    SAI_TAM_REPORT_TYPE_PROTO,

    /**
     * @brief Report using THRIFT
     */
    SAI_TAM_REPORT_TYPE_THRIFT,

    /**
     * @brief Report using JSON
     */
    SAI_TAM_REPORT_TYPE_JSON,

    /**
     * @brief Report using P4 format
     */
    SAI_TAM_REPORT_TYPE_P4_EXTN,

    /**
     * @brief Report using Histogram
     */
    SAI_TAM_REPORT_TYPE_HISTOGRAM,
// ...
```
So, I believe the current SAI is flexible enough to support other telemetry types. But I don't have a short-term plan to discuss other types in this proposal.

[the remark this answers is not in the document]
rajkumar38: By other telemetry type, I meant below. For example, SAI_TAM_TELEMETRY_TYPE_SWITCH can as well fetch all supported statistics from all ports in the switch. But, in this type, LABEL (element_id) cannot be mentioned by NOS. If SAI_TAM_TELEMETRY_TYPE_SWITCH is applicable, can you clarify what should be IPFIX template encoding for element-id, enterprise number ? 

```
    /**
     * @brief Switch silicon TAM
     * All the data relevant to switch
     * e.g. route, port, queue statistics
     */
    SAI_TAM_TELEMETRY_TYPE_SWITCH,

```

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,967 @@
...
+    /**
+     * @brief Set TAM telemetry state change event notification callback function passed to the adapter.
+     *
+     * Use sai_tam_telemetry_state_change_notification_fn as notification function.
+     *
+     * @type sai_pointer_t sai_tam_telemetry_state_change_notification_fn
+     * @flags CREATE_AND_SET
+     * @default NULL
+     */
+    SAI_SWITCH_ATTR_TAM_TELEMETRY_STATE_CHANGE_NOTIFY,
```
r12f, line 513: will this one be better to be put on TAM_TEL_TYPE?

in this way, if we change the state of 1 tel_type (say, queue), we won't impact the state of another type (say, port), in the non-mixed case.

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,967 @@
...
+
+    // ...
+} sai_switch_attr_t;
+
+sai_attr_list[0].id = SAI_SWITCH_ATTR_TAM_BUFFER_SIZE;
+sai_attr_list[0].value.u32 = buffer_size;
+
+sai_attr_list[1].id = SAI_SWITCH_ATTR_TAM_CHUNK_SIZE;
```
r12f, line 521: will BUFFER_CHUNK_BYTES or REPORT_CHUNK_BYTES be better than just CHUNK?

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -0,0 +1,967 @@
...
+    commit --> ready: Vendor set (callback via notification)
+
+    ready --> start: NOS set (New templates will be applied to generate the new record streaming)
+    ready --> stop: NOS set
+
+    stop --> [*]: Object remove
+
+```
+
+#### Initialize TAM cache count for Switch
```
r12f, line 464: looks like we are probably missing the mix-mode capability query here. is it because these 2 modes are being merged together?

Pterosaur: https://github.com/opencomputeproject/SAI/pull/2089/files#diff-0e0ad14d84f3332634521f6b2d53007b4f7ccec6c53e003139ef77da1a4d9d50R680

This one?

Pterosaur: Done

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
Pterosaur: Done

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
Pterosaur: Done

inc/saitam.h
```diff
@@ -910,6 +910,47 @@ typedef enum _sai_tam_telemetry_type_t
...
+     */
+    SAI_TAM_TEL_TYPE_STATE_CONFIG_GENERATING,
+
+    /**
+     * @brief Telemetry type config is ready
+     */
+    SAI_TAM_TEL_TYPE_STATE_CONFIG_READY,
+
+} sai_tam_tel_type_state_t;
+
```
JaiOCP, line 953: Nitpik, names of the attributes are not clear about the intent of the state machine.
I belive what you are trying to do is as follows
Default State:  READY
Once all the subscription is completed, orchagent want to indiate to SAI that it can now start the creation of templates
Next State: CREATE
SAI adapter will do out of sync notification of template creation complete and once orchagent reads the templates then must transition to state to that hw should start streaming now
Next State: START
If there is any change in NOS or streaming need to be stopped then HW wil be instructed to stop streaming.
Next State: STOP
Once SAI adapter cleans its own data stucture etc, NOS will transition the state to ready
Next State: READY

FSM will look like this
READY -> CREATE - > START -> STOP -> READY
There are no intermediate state transitions.

inc/saitam.h
```diff
@@ -1076,6 +1117,41 @@ typedef enum _sai_tam_tel_type_attr_t
...
+     */
+    SAI_TAM_TEL_TYPE_ATTR_STATE,
+
+    /**
+     * @brief Current state of TAM telemetry type
+     *
+     * @type sai_tam_tel_type_state_t
+     * @flags READ_ONLY
+     */
+    SAI_TAM_TEL_TYPE_ATTR_CURRENT_STATE,
```
JaiOCP, line 1143: Not sure why you need this attribute

JaiOCP (approved):

inc/saiswitch.h
```diff
@@ -3084,6 +3084,45 @@ typedef enum _sai_switch_attr_t
...
+     * The total number of bytes in a report should be as close as possible to this value.
+     * We can increase this value to reduce the number of sys calls.
+     * If the type of message is IPFIX, this value should not be less than 65535.
+     * Because we don't expect the IPFIX record to be fragmented.
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_AND_SET
+     * @default 65535
+     */
+    SAI_SWITCH_ATTR_TAM_REPORT_CHUNK_SIZE,
```
bandaru-viswanath, line 3099: Please move this to the TAM_REPORT (or the TAM) object being used. It doesn't make sense as a switch attribute and moving to ta/tam_report also allows different chunk sizes per-report/per-tam if allowed by implementation.

inc/saiswitch.h
```diff
@@ -3084,6 +3084,45 @@ typedef enum _sai_switch_attr_t
...
+     * This value indicates how many chunks of reports that can be restored in the buffer.
+     * If the data structure is a ring buffer, the byte size of ring buffer is chunk count * chunk size.
+     * 1 means that there is only one chunk of report in the buffer.
+     * If the buffer is full, new incoming data will be dropped.
+     *
+     * @type sai_uint32_t
+     * @flags CREATE_ONLY
+     * @default 1
+     */
+    SAI_SWITCH_ATTR_TAM_CHUNK_COUNT,
```
bandaru-viswanath, line 3113: Can this be moved to TAM object as opposed to keeping in SWITCH object ?

inc/saiswitch.h
```diff
@@ -3084,6 +3084,45 @@ typedef enum _sai_switch_attr_t
...
+    /**
+     * @brief Set TAM telemetry type config change event notification callback function passed to the adapter.
+     *
+     * Use sai_tam_tel_type_config_change_notification_fn as notification function.
+     *
+     * @type sai_pointer_t sai_tam_tel_type_config_change_notification_fn
+     * @flags CREATE_AND_SET
+     * @default NULL
+     */
+    SAI_SWITCH_ATTR_TAM_TEL_TYPE_CONFIG_CHANGE_NOTIFY,
```
bandaru-viswanath, line 3124: Can this be moved to TAM object ?

inc/saitam.h
```diff
@@ -2370,6 +2460,8 @@ typedef struct _sai_tam_api_t
     sai_remove_tam_counter_subscription_fn         remove_tam_counter_subscription;
     sai_set_tam_counter_subscription_attribute_fn  set_tam_counter_subscription_attribute;
     sai_get_tam_counter_subscription_attribute_fn  get_tam_counter_subscription_attribute;
+    sai_bulk_object_create_fn                      create_tam_counter_subscriptions;
```
bandaru-viswanath, line 2463: please use a _tam_ string as part of the type for consistency.

bandaru-viswanath (changes requested):

inc/saiswitch.h
Pterosaur: It looks like all notify handlers have to be bound on the switch object. It's the limitation of SAI design.

inc/saiswitch.h
Pterosaur: Actually, It was the previous design.  But we discussed with some vendors, they said they can only support global ring buffer configuration and cannot be changed at runtime.

inc/saitam.h
Pterosaur: I guess it's a convention of SAI to use sai_bulk_object_xxxx.

https://github.com/opencomputeproject/SAI/blob/ad20062eb1f2a24aa79983422e83b921b49c011f/inc/sailag.h#L395-L396
https://github.com/opencomputeproject/SAI/blob/ad20062eb1f2a24aa79983422e83b921b49c011f/inc/sainexthop.h#L335-L338

inc/saitam.h
```diff
@@ -1076,6 +1121,33 @@ typedef enum _sai_tam_tel_type_attr_t
      */
     SAI_TAM_TEL_TYPE_ATTR_COUNTER_SUBSCRIPTION_LIST,
 
+    /**
+     * @brief The mode of TAM telemetry type
+     *
+     * @type sai_tam_tel_type_mode_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_TAM_TEL_TYPE_ATTR_MODE,
```
kcudnik, line 1130: This make this chicane breaking change since this attribute will be mandatory in next release

kcudnik: Should this be create only with some default value ?

kcudnik:
please fix build

Pterosaur:
> please fix build

Hi @kcudnik , it was failed on the graph check. I feel the git graph check has an issue, it shouldn't check the conflict on these commits belong to this PR rather than check the conflict between PR with master branch.
Otherwise, if I want to fix the graph check, I have to rebase the commit history.

./checkancestry.sh
ancestry graph
*   74f9cf8 Merge 4a210e373be4db0c1e7a980150000509ceb4852d into ad20062eb1f2a24aa79983422e83b921b49c011f
|\  
| * 4a210e3 Change the default value for chunk count
| * c449e5d Merge branch 'master' into ipfix_stream_telemetry
|/  
* ad20062 Introduce new extended port oper status notification (#2087)
git rev list from 5f75d99 to HEAD
git checkout work tree commits: 5f75d99 91ef792 97c8ed2 ad20062 c449e5d 4a210e3 74f9cf8
loaded history from ancestry.5f75d99.history

inc/saiswitch.h
bandaru-viswanath: OK. The attribute atleast needs to be renamed to include _ST_ or equivalent since TAM is too generic and there are many other TAM features, and TAM_REPORT_CHUNK_SIZE is too generic.

inc/saiswitch.h
bandaru-viswanath: I don't know if its SAI design limitation, or just that all callback handlers are at switch level currently. @kcudnik please comment.

r12f: Hi Bandaru, yes, in my understanding, this is basically how the notify handler works. The handler will be used by all objects in that type, hence it is added on upper layers, and the best place is the switch, as it will be setup during the initialization.

For example, fob handler will be used by all fdb entries, hence it is not tied to any fdb entry.

This is why the notifications are added on the switch.

kcudnik: > I don't know if its SAI design limitation, or just that all callback handlers are at switch level currently. @kcudnik please comment.

Currently it's forced that all notifications be in switch, this will unify entire design, and it will help update pointers when switch is put into warm boot mode, currently we don't have specific notifications per object and we are not planning to do so, more explanation is like Riff said in above comment

bandaru-viswanath: > Hi Bandaru, yes, in my understanding, this is basically how the notify handler works. The handler will be used by all objects in that type, hence it is added on upper layers, and the best place is the switch, as it will be setup during the initialization.

In this case, all objects (tam_tel_type etc) are all under the single TAM object. That is the container and not switch object. Its not an upper layer, but a container. The comparision with FDB is also incorrect. Please see below.

> 
> For example, fob handler will be used by all fdb entries, hence it is not tied to any fdb entry.

The FDB notification (callback registration) is for all entries and not for a specific entry. There is no object that holds all FDB entries and thus switch is the only way out. 

> 
> This is why the notifications are added on the switch.

kcudnik: is this a problem that all tam objects will be put over a single notification pointer in the switch? you still need to explicitly create "tam" object using switch_id, so it does not matter how internally tam is organizes, it still can use singe swithc notification to push notify and identify object that is affected

bandaru-viswanath: Actually, the call back is specific to the TAM object that is linked to all "streaming telemetry" objects. That TAM object is a single one. However, in a given SAI Adapter instance there may be more than one TAM object, each TAM object serving a different feature. Thus, it doesn't make sense to link the callback to the SWITCH which implies that it is applicable for all TAM objects.

kcudnik: you can return object_id of that tam in notification, and act accordingly, we currently are not planning to decouple pointers from switch to single objects, we generate a lot of metadat which help us organize everygnig, including notification pointers assignments, which are coupled to the switch, at this point it will be up to user to decide differnt action  based on notificaion data

bandaru-viswanath: If we have a restriction in the metadata and other aspects, I think it is OK. I will wait for the name edits that I suggested above, before approving this PR. This callback item is closed from my side.

kcudnik: Those are not hard restrictions, but we're made for convince, and drifting away from current schema would cause a lot of refactoring

bandaru-viswanath: Thats OK, I am not talking about hard restrictions alone. No need to change all the infra for this for now.
 
However, I request you to consider the possibility that the callbacks may - at sometime in future - may be per-object and plan any changes for this.  Thanks @kcudnik

bandaru-viswanath (approved):

inc/saitam.h
```diff
@@ -1076,6 +1121,34 @@ typedef enum _sai_tam_tel_type_attr_t
      */
     SAI_TAM_TEL_TYPE_ATTR_COUNTER_SUBSCRIPTION_LIST,
 
+    /**
+     * @brief The mode of TAM telemetry type
+     *
+     * @type sai_tam_tel_type_mode_t
+     * @flags CREATE_ONLY
+     * @default SAI_TAM_TEL_TYPE_MODE_SINGLE_TYPE
+     */
```
ygdalnaouri, line 1130: For clarity, it should be marked as:
@validonly SAI_TAM_TEL_TYPE_ATTR_TAM_TELEMETRY_TYPE == SAI_TAM_TELEMETRY_TYPE_COUNTER_SUBSCRIPTION

Pterosaur: good idea，thanks for your suggestion.

r12f (approved):

inc/saiswitch.h
kcudnik: there is also another side why all pointers are in switch, during warm boot of the platform, user is calling create_switch, and it can pass new pointers to the switch to receive notifications, this is importnt, since after platform warm boot, new notifications wii most likely have different memory address than previous boot, so notifications would need to be updated. and this could be done in a single call during create switch, if you would have pointers in objects , you would need to go and find that object and assign new notification on it or you would have invalid pointer, or null, depends how platform handles pointers in sdk

Pterosaur:
@kcudnik  Could you please help to merge this PR?

kcudnik (approved):

PR #2091, opened 2024-10-16
Add HA session attribute Bounceback IP

marian-pritsak:
Add bounceback IP which is a different IP from
peer IP, to emulate the same network path for data packets as for the SYN packets.


marian-pritsak:
Depends on https://github.com/opencomputeproject/SAI/pull/2090

KrisNey-MSFT:
@marian-pritsak it looks like 2090 is merged (this one dependent on it).  Ok to merge this one now?

kcudnik (approved):

KrisNey-MSFT:
@kcudnik - is it ok to merge this one?

PR #2093, opened 2024-10-24
[DASH] Update dash flow and dash tunnel

jimmyzhai:
1. Update SAI of dash flow to align with its p4 table in dash-pipeline
2. Update SAI of dash tunnel to align with its p4 table in dash-pipeline


experimental/saitypesextensions.h
```diff
@@ -180,10 +184,32 @@ typedef enum _sai_dash_flow_enabled_key_t
...
+
+    SAI_DASH_FLOW_ACTION_DNAT = 32,
+
+    SAI_DASH_FLOW_ACTION_NAT46 = 64,
+
+    SAI_DASH_FLOW_ACTION_NAT64 = 128,
+
+    SAI_DASH_FLOW_ACTION_SNAT_PORT = 256,
+
+    SAI_DASH_FLOW_ACTION_DNAT_PORT = 512,
```
kcudnik, line 212: mayeb use arrows << like other to easy spot which bit is turned on?

inc/saiobject.h
```diff
@@ -37,6 +37,8 @@
 #include <saisrv6.h>
 
 /* new experimental object type includes */
+#include <saiexperimentaldashtunnel.h>
+#include <saiexperimentaldashha.h>
```
kcudnik, line 41: no need for this

experimental/saitypesextensions.h
zhixiongniu: agree

zhixiongniu (approved):

experimental/saitypesextensions.h
jimmyzhai: fixed

inc/saiobject.h
jimmyzhai: Why not? Here the list has included `saiexperimentaldashappliance.h, saiexperimentaldashflow.h, saiexperimentaldashmeter.h, saiexperimentaldashvip.h`, etc. The list is automatically updated while `make sai-headers` in DASH to have header files of all dash objects.

kcudnik: since no item is used from included files, so including them is redundant

kcudnik (approved):

PR #2094, opened 2024-10-28
[DASH] Mark appliance table local region ID attribute as create-only

mukeshmv:
https://github.com/sonic-net/DASH/pull/634


prsunny:
@kcudnik , @tjchadaga , would you see why the PR is stuck at PR check stage?

PR #2114, opened 2024-11-22
Unify SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_CUSTOM_RANGE_BASE value

kcudnik:
To match all other custom range bases starting from 0x1000000.

All other _CUSTOM_RANGE_BASE enums are indexed at 0x1000000, lets's move this one to the same base to have all vendor custom enums at the same base across entire SAI

### DIFF
--- a/inc/saihostif.h
+++ b/inc/saihostif.h
@@ -712,13 +712,13 @@ typedef enum _sai_hostif_user_defined_trap_type_t
     /** Traps to be associated with TAM collector */
     SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_TAM,
 
-    /** Custom range base */
-    SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_CUSTOM_RANGE_BASE = 0x00001000,
-
     /**
      * @brief End of user defined trap types
      */
     SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_END,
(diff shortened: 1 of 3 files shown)


kcudnik:
Unfortunately this is enum breaking change, fortunately sonic is not using this attribute, and in vendor case, if they have any custom traps, it's easy to workaround it in internal headers:

```c
#define SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_CUSTOM_RANGE_BASE_OLD (0x1000)

if (trapid >= SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_CUSTOM_RANGE_BASE_OLD &&
    trapid <  SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_CUSTOM_RANGE_BASE)
{
    trapid = trapid - SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_CUSTOM_RANGE_BASE_OLD +
                      SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_CUSTOM_RANGE_BASE;
}
```

if this will help, we can introduce depreciated enum in SAI with previous values for convenience

kcudnik:
FYI @JaiOCP @tushar-ty @marian-pritsak @itaibaz please join discussion if you have any comments

PR #2124, opened 2024-12-24
Create new port stat extended list

maunnikr-cisco:


kcudnik:
please describe what is this change for, and current form is invalid, please run make in meta directory locally to verify your change

PR #2126, opened 2025-01-10
[meta] Update check attr condition/validonly

kcudnik:
Add support for condition values that have default values present. Also update check to support enums (s32) and bool values

### DIFF
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -3047,6 +3047,31 @@ void check_attr_mixed_validonly(
     }
 }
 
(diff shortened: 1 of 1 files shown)


baorliu:
hi @kcudnik , can you double commit this PR to SAI 1.15/1.14/1.13? 
I will double commit the PR https://github.com/opencomputeproject/SAI/pull/2127 to the above branch, but need your fix before I double commit it. thanks.

kcudnik:
If we add labels to those branches then it should be pulle dautomatically

PR #2133, opened 2025-01-23
[TAM]: Remove mandatory flag on the attribute TAM DSCP

Pterosaur:
In some transport protocols, such as GENetlink, the attribute DSCP is meaningless.
So, remove the mandatory flag on the attribute TAM DSCP and use the default value 0 to indicate this protocol doesn't support the field DSCP.

### DIFF
--- a/inc/saitam.h
+++ b/inc/saitam.h
@@ -1871,9 +1871,12 @@ typedef enum _sai_tam_collector_attr_t
 
     /**
      * @brief DSCP value
+     * If DSCP value cannot be set on the target transport protocol, such as GENETLINK,
+     * this value will be set to 0 and ignored.
      *
      * @type sai_uint8_t
-     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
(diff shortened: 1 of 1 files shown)


rck-innovium (approved):

r12f:
makes sense! Thanks Ze!

Pterosaur:
@kcudnik Could you please help to  merge this PR?

kcudnik:
was this already discussed on SAI community meeting and approved ?

Pterosaur:
> was this already discussed on SAI community meeting and approved ?

Not yet, but this is proposed by a vendor and we both it's reasonable and straightforward. Because the original API design was already adapted the GENetlink few years ago, and it is obvious that GENetlink doesn't have DSCP field. Right now, vendors need to implement this scenario and this issue is revealed.

kcudnik:
agree, and this change is backward compatible, but i think every change to api should be discussed on team meeting, is this time sensitive to merge this soon ?

kcudnik (approved):

PR #2141, opened 2025-02-19
[TAM] SAI TAM enhancements

rajkumar38:
This proposal extends SAI TAM 2.0 with,
- enhanced reporting modes (sampling, microburst)
- new GRE transport type and transport attributes for header encapsulation
- collector destination port/lag
- new ACL action attribute to associate the TAM object with specific flow

### DIFF
--- a/inc/saitam.h
+++ b/inc/saitam.h
@@ -1289,6 +1289,9 @@ typedef enum _sai_tam_report_mode_t
     /** Report in a bulk mode */
     SAI_TAM_REPORT_MODE_BULK,
 
+    /** Report in a sampling mode, one report is sent for every n reports */
+    SAI_TAM_REPORT_MODE_SAMPLING,
+
 } sai_tam_report_mode_t;
 
 /**
@@ -1415,6 +1418,37 @@ typedef enum _sai_tam_report_attr_t
      * @validonly SAI_TAM_REPORT_ATTR_REPORT_MODE == SAI_TAM_REPORT_MODE_BULK
      */
     SAI_TAM_REPORT_ATTR_REPORT_INTERVAL_UNIT,
+
+    /**
+     * @brief Sampling rate (every 1/sample_rate)
+     *
+     * @type sai_uint32_t
+     * @flags MANDATORY_ON_CREATE | CREATE_AND_SET
+     * @condition SAI_TAM_REPORT_ATTR_REPORT_MODE == SAI_TAM_REPORT_MODE_SAMPLING
+     */
+    SAI_TAM_REPORT_ATTR_SAMPLE_RATE,
+
+    /**
+     * @brief Maximum report rate per second
+     *
+     * Value 0 to no limit.
+     *
+     * @type sai_uint64_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     * @validonly SAI_TAM_REPORT_ATTR_REPORT_MODE == SAI_TAM_REPORT_MODE_ALL
+     */
+    SAI_TAM_REPORT_ATTR_MAX_REPORT_RATE,
+
+    /**
+     * @brief Maximum reports per burst
+     *
+     * @type sai_uint64_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     * @validonly SAI_TAM_REPORT_ATTR_REPORT_MODE == SAI_TAM_REPORT_MODE_ALL
+     */
+    SAI_TAM_REPORT_ATTR_MAX_REPORT_BURST,
 
     /**
      * @brief End of Attributes
@@ -1652,6 +1686,11 @@ typedef enum _sai_tam_transport_type_t
      * @brief Transport MIRROR session
      */
     SAI_TAM_TRANSPORT_TYPE_MIRROR,
+
+    /**
+     * @brief Transport GRE
+     */
+    SAI_TAM_TRANSPORT_TYPE_GRE,
 } sai_tam_transport_type_t;
 
 /**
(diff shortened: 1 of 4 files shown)


kcudnik:
you created object dependency loop

rajkumar38:
> you created object dependency loop

Hi @kcudnik ,

Here Port associated with SAI_TAM_COLLECTOR_ATTR_DESTINATION will be different from TAM bind point. Here TAM DESTINATION refers to the network port reachable to external collector where reports are transmitted out. Can you suggest how to resolve this ?

_WARN: LOOP DETECTED on object type: SAI_OBJECT_TYPE_PORT    **-----> Port where original traffic rx/tx**
WARN:  SAI_OBJECT_TYPE_PORT: SAI_PORT_ATTR_QOS_INGRESS_BUFFER_PROFILE_LIST
WARN:  SAI_OBJECT_TYPE_BUFFER_PROFILE: SAI_BUFFER_PROFILE_ATTR_POOL_ID
WARN:  SAI_OBJECT_TYPE_BUFFER_POOL: SAI_BUFFER_POOL_ATTR_TAM
WARN:  SAI_OBJECT_TYPE_TAM: SAI_TAM_ATTR_TELEMETRY_OBJECTS_LIST
WARN:  SAI_OBJECT_TYPE_TAM_TELEMETRY: SAI_TAM_TELEMETRY_ATTR_COLLECTOR_LIST
WARN:  SAI_OBJECT_TYPE_TAM_COLLECTOR: SAI_TAM_COLLECTOR_ATTR_DESTINATION **--> nw port connected to collector**
WARN:  -> SAI_OBJECT_TYPE_PORT
 ASSERT FAILED (on line 4224): LOOP is detected, we can't have loops in graph, please fix attributes
make: *** [Makefile:93: all] Error 1_

kcudnik:
loop detector is installed to not have loops in dependency graph, not sure how to resolve that, usually to prevent loops you can have objects marked as read_only

that validator is not checking whether port is actually the same port, it's checking that dependency is on port object type,

let's have this discussion on todays SAI meeting, will you be able to join ?

rajkumar38:
> loop detector is installed to not have loops in dependency graph, not sure how to resolve that, usually to prevent loops you can have objects marked as read_only
> 
> that validator is not checking whether port is actually the same port, it's checking that dependency is on port object type,
> 
> let's have this discussion on todays SAI meeting, will you be able to join ?

Unfortunately, I couldn't join the call yesterday. I will request a slot next week (02/27) to discuss the proposal.

inc/saitam.h
```diff
@@ -1289,6 +1289,12 @@ typedef enum _sai_tam_report_mode_t
     /** Report in a bulk mode */
     SAI_TAM_REPORT_MODE_BULK,
 
+    /** Report in a probabilistic sampling mode, one report is sent in n events */
+    SAI_TAM_REPORT_MODE_PROBABILISTIC_SAMPLING,
+
+    /** Report in a microburst mode, specifying report rate and max number of reports per microburst event */
```
ashutosh-agrawal, line 1295: If the objective here is to rate-limit the number of reports, then attaching a policer object to tam_report will be more intuitive.

rajkumar38: Base on last community discussion, all policer attributes may not be applicable and we don't have capability query mechanism based on policer bind point to identify which of these attributes are supported.
We will follow existing SAI attribute names instead of using microburst.

rajkumar38:
> loop detector is installed to not have loops in dependency graph, not sure how to resolve that, usually to prevent loops you can have objects marked as read_only
> 
> that validator is not checking whether port is actually the same port, it's checking that dependency is on port object type,
> 
> let's have this discussion on todays SAI meeting, will you be able to join ?

Hi Kamil,
As discussed in last community meeting, I have added exception in saisanitycheck. Pls review.

tjchadaga:
@rajkumar38 - could you please resolve branch conflicts?

inc/saitam.h
ashutosh-agrawal: Please update the document also to remove references to microburst and also use the updated attribute names.

rajkumar38: Thanks @ashutosh-agrawal . I missed to replace the HLD file. Now I have pushed the changes. Pls review.

rajkumar38:
> @rajkumar38 - could you please resolve branch conflicts?

Done.

inc/saitam.h
```diff
@@ -1416,6 +1419,37 @@ typedef enum _sai_tam_report_attr_t
...
+     * @validonly SAI_TAM_REPORT_ATTR_REPORT_MODE == SAI_TAM_REPORT_MODE_ALL
+     */
+    SAI_TAM_REPORT_ATTR_MAX_REPORT_RATE,
+
+    /**
+     * @brief Maximum reports per burst
+     *
+     * @type sai_uint64_t
+     * @flags CREATE_AND_SET
+     * @default 0
```
ashutosh-agrawal, line 1448: What would be the behavior for a default value of 0? Is the implementation expected to sent out zero reports?

rajkumar38: If value is 0, it means no rate-limit and all reports will be sent out.

ashutosh-agrawal: I understand the behavior for the MAX_REPORT_RATE attribute and there is a comment there as well. But I'm still unclear on a default value of 0 for MAX_REPORT_BURST.

Let's say MAX_REPORT_RATE is set to 1000 reports/second and MAX_REPORT_BUTST is not set so it defaults to 0. What will be the expected behavior? Is the device expected to ignore the max_report_rate and sent out all reports?

rajkumar38: In this case, the device should not ignore the MAX_REPORT_RATE. With MAX_REPORT_BURST set to 0, no bursts are allowed, so the transmission rate of reports will be limited to 1000 reports per second. If a burst of traffic is received, it depends on the vendor's implementation whether the excess reports are dropped or delayed.

rajkumar38: When MAX_REPORT_RATE = 0, rate-limiting is not applicable. With this is, MAX_REPORT_BURST defaulted to value 1 seems incorrect.
To align with existing SAI model (similar to shaper), we can have MAX_REPORT_BURST default as 0 and it's up to the user to configure MAX_REPORT_RATE and MAX_REPORT_BURST with vendor supported values for the rate-limiter to work.

ashutosh-agrawal: Hi Raj, I was asking about the case when MAX_REPORT_RATE is non-zero and MAX_REPORT_BURST is not defined. 

But I'm fine with the user properly defining these two values to make the rate-limiter work properly.

ashutosh-agrawal (approved):

PR #2142, opened 2025-02-24
ERSPAN Mirror Session Enhancements

ppanara-marvell:
This update extends ERSPAN mirroring with added support for ERSPAN-II and ERSPAN-III. The key enhancements include:

- Two new enums introduced to enable proper encapsulation and handling of ERSPAN-II and ERSPAN-III mirrored packets.
- A new attribute to modify the first 16 bits of the GRE header, allowing repurposing of unused fields for additional functionality.
- A new attribute to update the session_id field in ERSPAN-II and ERSPAN-III headers, ensuring unique identification of mirrored traffic flows.

### DIFF
--- a/inc/saimirror.h
+++ b/inc/saimirror.h
@@ -62,6 +62,16 @@ typedef enum _sai_erspan_encapsulation_type_t
      */
     SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_L3_GRE_TUNNEL,
 
+    /**
+     * @brief L3 GRE Tunnel Encapsulation | L2 Ethernet header | IP header | GRE header | ERSPAN (8B) | Original mirrored packet
+     */
+    SAI_ERSPAN_ENCAPSULATION_TYPE_II,
+
+    /**
+     * @brief L3 GRE Tunnel Encapsulation | L2 Ethernet header | IP header | GRE header | ERSPAN (12B) + optional headers | Original mirrored packet
+     */
+    SAI_ERSPAN_ENCAPSULATION_TYPE_III,
+
 } sai_erspan_encapsulation_type_t;
 
 /**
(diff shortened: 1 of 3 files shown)


meta/saisanitycheck.c
```diff
@@ -1753,7 +1754,19 @@ void check_attr_validonly(
...
-            META_MD_ASSERT_FAIL(md, "conditional attribute is also conditional, not allowed");
+            if (md->objecttype == SAI_OBJECT_TYPE_MIRROR_SESSION && 
+			    (md->attrid == SAI_MIRROR_SESSION_ATTR_ERSPAN_SESSION_ID
+			     || md->attrid == SAI_MIRROR_SESSION_ATTR_GRE_HEADER_FIRST_16BIT))
+            {
+                /*
+                 * ERSPAN_SESSION_ID is valid for ERSPAN and when encap_type is ERSPAN_II or ERSPAN_III
+		 * GRE_HEADER_FIRST_16BIT is valid for ERSPAN encap_type GRE
+                 */
+            }
```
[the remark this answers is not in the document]
kcudnik: we probably should not have such dependencies

inc/saimirror.h
```diff
@@ -62,6 +62,16 @@ typedef enum _sai_erspan_encapsulation_type_t
      */
     SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_L3_GRE_TUNNEL,
 
+    /**
+     * @brief L3 GRE Tunnel Encapsulation | L2 Ethernet header | IP header | GRE header | ERSPAN (8B) | Original mirrored packet
+     */
+    SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_L3_GRE_TUNNEL_ERSPAN_TYPE_II,
```
ashutosh-agrawal, line 68: This is a mouthful. Can we simply call it SAI_ERSPAN_ENCAPSULATION_TYPE_II?

ppanara-marvell: Will take care of this.

meta/saisanitycheck.c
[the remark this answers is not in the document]
ppanara-marvell: This is valid case. 
I saw similar check already present for SAI_MIRROR_SESSION_ATTR_VLAN_TPID and few other attributes which depends on conditional attribute VLAN_HEADER_VALID.

Plz suggest how this kind of conditions are usually taken care.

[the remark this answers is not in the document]
kcudnik: similar check should be added: https://github.com/opencomputeproject/SAI/pull/2131

tjchadaga:
@ppanara-marvell - could you please address the review comments?

meta/saisanitycheck.c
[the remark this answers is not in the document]
ppanara-marvell: Taken care as done in mentioned PR.

rck-innovium:
@ashutosh-agrawal  @kcudnik Your comments are addressed. Can you please review.

inc/saimirror.h
```diff
@@ -362,6 +372,28 @@ typedef enum _sai_mirror_session_attr_t
...
+    SAI_MIRROR_SESSION_ATTR_ERSPAN_SESSION_ID,
+
+    /**
+     * @brief The first 16 bits of the GRE header.
+     *
+     * @type sai_uint16_t
+     * @flags CREATE_AND_SET
+     * @isvlan false
+     * @default 0
+     * @validonly SAI_MIRROR_SESSION_ATTR_ERSPAN_ENCAPSULATION_TYPE == SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_L3_GRE_TUNNEL
```
ashutosh-agrawal, line 393: Since ERSPAN II and ERSPAN III also have GRE headers, is there any particular reason behind limiting this to ERSPAN Type I only?

ppanara-marvell: As per IETF draft ERSPAN-I is only type where they mandate to have first 16bits as 0. Where as in case of ERSPAN-II S-bit will be set and ERSPAN-III also can have any other flags(C, K, S) set, that could result in different GRE header length. It would be difficult for a parser to find out exact length of GRE header. Hence this attribute is defined only for ERSPAN-I.

ppanara-marvell: This attribute was added for a customer because they didn't want to add any other overhead on top of mirror packet.  In case of ERSPAN-II and ERSPAN-III we already have headers, hence we added this only for erspan-I. But in SAI, it make sense to make it generic for all types. I will update PR.

kcudnik (approved):

ashutosh-agrawal (approved):
LGTM

PR #2145, opened 2025-03-03
Add back port_error_status to sai_port_oper_status_notification_t

prgeor:
`sai_port_oper_status_notification_t`  notification in SAI version starting from v1.15.0 onwards had the new field `sai_port_error_status`. 

This field got removed by [PR2087](https://github.com/opencomputeproject/SAI/pull/2087) . 

Added the field back to prevent this incompatibility for vendors using SAI v1.15.0 and above

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -141,6 +141,8 @@ typedef struct _sai_port_oper_status_notification_t
     /** Port operational status */
     sai_port_oper_status_t port_state;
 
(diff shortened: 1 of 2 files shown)


kcudnik (approved):

kcudnik:
@prgeor if we don't need to wait for community meeting Thursday we can merge it today

JaiOCP (approved):

PR #2148, opened 2025-03-05
Add switch-level attribute to control PTP for the entire switch

rdsugarmannv:
The SAI API currently only allows PTP to be configured on a individual port basis.  Some switches only support global configuration (on/off) for PTP.  This change adds a switch level SAI attribute to enable PTP for the whole switch.

### DIFF
--- a/doc/SAI-Proposal-Global-PTP-Configuration.md
+++ b/doc/SAI-Proposal-Global-PTP-Configuration.md
@@ -1,0 +1,88 @@
+# Global PTP Configuration in SAI Switch
+
+Title       | Global PTP Configuration in SAI Switch
+------------|----------------------------------------
+Authors     | Open
+Status      | Draft
+Type        | Standards track
+Created     | 2024-03-03
+SAI-Version | 1.15
+
+## Overview
+Introducing a PTP (Precision Time Protocol) attribute to the SAI switch object, allowing for switch-wide PTP mode settings that affect all ports on the switch. 
+
+A switch that does not set this attribute will observe no change in behavior.
+
+## Motivation
+Currently, PTP configuration in SAI is port-based, and many switches only support configuration of PTP on a global basis.
+In addition, many existing switches have a configuration model which allows global configuration of PTP, and then override on a port level.  SAI should follow this flexible example.
+
+## Technical Specification
+
+### Updated description of existing PORT PTP type - the main comment is new
+```c
+/**
+ * @brief PTP mode
+ * 
+ * These modes can be used at the port and switch level.
+ * All ports use the value set at the switch level unless explicitly configured
+ * at the port level to a value other than SAI_PORT_PTP_MODE_NONE.
+ */
+typedef enum _sai_port_ptp_mode_t
+{
+    /** No special processing for PTP packets */
+    SAI_PORT_PTP_MODE_NONE,
+
+    /** Single-step Timestamp mode for the PTP packets */
+    SAI_PORT_PTP_MODE_SINGLE_STEP_TIMESTAMP,
+
+    /** Two-step Timestamp mode for the PTP packets */
+    SAI_PORT_PTP_MODE_TWO_STEP_TIMESTAMP,
+
+} sai_port_ptp_mode_t;
+
+
+### New Switch Attribute
+```c
+typedef enum _sai_switch_attr_t
+{
+    // ... existing attributes ...
+
+    /**
+     * @brief Global PTP mode configuration
+     *
+     * Global PTP mode configuration for the switch.
+     * Applies to all ports unless overridden by port-specific settings.
+     *
+     * @type sai_port_ptp_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PORT_PTP_MODE_NONE
+     */
+    SAI_SWITCH_ATTR_PORT_PTP_MODE,
+
+    // ... existing attributes ...
+} sai_switch_attr_t;
+```
+
+## Usage Example
+```c
+// Set global PTP configuration at switch level
+sai_attribute_t attr;
+attr.id = SAI_SWITCH_ATTR_PORT_PTP_MODE;
+attr.value.s32 = SAI_PORT_PTP_MODE_SINGLE_STEP_TIMESTAMP;
+
+sai_status_t status = sai_switch_api->set_switch_attribute(
+    switch_id,
+    &attr);
+
+// Get global PTP configuration
+attr.id = SAI_SWITCH_ATTR_PORT_PTP_MODE;
+status = sai_switch_api->get_switch_attribute(
+    switch_id,
+    1,
+    &attr);
+```
+
+## References
+1. IEEE 1588-2008 Standard for a Precision Clock Synchronization Protocol for Networked Measurement and Control Systems
+2. IEEE 1588-2019 Standard for a Precision Clock Synchronization Protocol for Networked Measurement and Control Systems 
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -395,6 +395,9 @@ typedef enum _sai_port_priority_flow_control_mode_t
 
 /**
  * @brief PTP mode
+ * These modes can be used at the port and switch level.
+ * All ports use the value set at the switch level unless explicitly configured
+ * at the port level to a value other than SAI_PORT_PTP_MODE_NONE.
  */
 typedef enum _sai_port_ptp_mode_t
 {
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -3198,6 +3198,18 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_SHARED_BUFFER_CELL_SIZE,
 
     /**
+     * @brief Global PTP mode configuration
+     *
+     * Global PTP mode configuration for the switch.
+     * Applies to all ports unless overridden by port-specific settings.
+     *
+     * @type sai_port_ptp_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PORT_PTP_MODE_NONE
+     */
+    SAI_SWITCH_ATTR_PORT_PTP_MODE,
+
+    /**
      * @brief End of attributes
      */
     SAI_SWITCH_ATTR_END,


inc/saiswitch.h
```diff
@@ -3197,6 +3197,18 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_SHARED_BUFFER_CELL_SIZE,
 
+/**
```
kcudnik, line 3200: fix comment ident

rdsugarmannv: Done.

kcudnik (approved):

tjchadaga:
@rdsugarmannv - as discussed please fix the meta checker errors

WARNING: first line doxygen comment should be with '/**': saiswitch.h: /** 
WARNING: line ends in whitespace saiswitch.h 643: /** 
WARNING: line ends in whitespace saiswitch.h 656:     
WARNING: Header doesn't meet style requirements (most likely ident is not 4 or 8 spaces) saiswitch.h 656:

tjchadaga:
@rajkumar38, @rck-innovium, @JaiOCP - please help review

rdsugarmannv:
> @rdsugarmannv - as discussed please fix the meta checker errors
> 
> WARNING: first line doxygen comment should be with '/**': saiswitch.h: /** WARNING: line ends in whitespace saiswitch.h 643: /** WARNING: line ends in whitespace saiswitch.h 656: WARNING: Header doesn't meet style requirements (most likely ident is not 4 or 8 spaces) saiswitch.h 656:

Hopefully fixed now

tjchadaga:
@rdsugarmannv - still failing
WARNING: Both enums have the same value SAI_SWITCH_ATTR_PTP_MODE and SAI_SWITCH_ATTR_GLOBAL_PTP_CONFIG = 0x000000f6

you may need to squash your commits and force push

rdsugarmannv:
> @rdsugarmannv - still failing WARNING: Both enums have the same value SAI_SWITCH_ATTR_PTP_MODE and SAI_SWITCH_ATTR_GLOBAL_PTP_CONFIG = 0x000000f6
> 
> you may need to squash your commits and force push

done

rajkumar38 (approved):

inc/saiswitch.h
```diff
@@ -634,6 +634,22 @@ typedef enum _sai_packet_trim_queue_resolution_mode_t
 
 } sai_packet_trim_queue_resolution_mode_t;
 
+/**
+ * @brief Global PTP configuration
+ */
+typedef enum _sai_switch_ptp_mode_t
```
rck-innovium, line 640: We should create a new SAI PTP mode that can be used for instead of duplicating sai_port_ptp_mode_t as sai_switch_ptp_mode_t.

inc/saiswitch.h
```diff
@@ -634,6 +634,22 @@ typedef enum _sai_packet_trim_queue_resolution_mode_t
...
+{
+    /** None - per-port configuration can be used to switch PTP on for a port */
+    SAI_SWITCH_PTP_MODE_NONE,
+
+    /** One-step - all ports use one-step PTP mode */
```
rck-innovium, line 645: This comment is misleading. We can still have one or more ports with  SAI_PORT_ATTR_PTP_MODE == SAI_PORT_PTP_MODE_TWO_STEP_TIMESTAMP. These ports will NOT use one-step PTP mode.

inc/saiswitch.h
```diff
@@ -634,6 +634,22 @@ typedef enum _sai_packet_trim_queue_resolution_mode_t
...
+    SAI_SWITCH_PTP_MODE_SINGLE_STEP_TIMESTAMP,
+
+    /** Two-step - all ports use two-step PTP mode */
```
rck-innovium, line 648: Same as above. We can still have one or more ports with  SAI_PORT_ATTR_PTP_MODE == SAI_PORT_PTP_MODE_ONE_STEP_TIMESTAMP. These ports will NOT use two-step PTP mode.

rck-innovium (changes requested):
1. The existing SAI_PORT_ATTR_PTP_MODE=SAI_PORT_PTP_MODE_NONE meaning was to disable special PTP processing on the port. Now we are changing this to mean that SAI_SWITCH_ATTR_PTP_MODE should take effect. However, the comments do not bring this out clearly.
2. What does the below config mean?
SAI_SWITCH_ATTR_PTP_MODE = SAI_SWITCH_PTP_MODE_SINGLE_STEP_TIMESTAMP, and port P1's SAI_PORT_ATTR_PTP_MODE=SAI_PORT_PTP_MODE_NONE.  
 If we want to enable PTP single step timestamp on all ports, and disable PTP processing on port P1, what should be the config?

rdsugarmannv:
> 1. The existing SAI_PORT_ATTR_PTP_MODE=SAI_PORT_PTP_MODE_NONE meaning was to disable special PTP processing on the port. Now we are changing this to mean that SAI_SWITCH_ATTR_PTP_MODE should take effect. However, the comments do not bring this out clearly.

I think you're asking that I change the PORT comments to reference the global ones so this behavior is clear?   I'll do that.

> 2. What does the below config mean?
>    SAI_SWITCH_ATTR_PTP_MODE = SAI_SWITCH_PTP_MODE_SINGLE_STEP_TIMESTAMP, and port P1's SAI_PORT_ATTR_PTP_MODE=SAI_PORT_PTP_MODE_NONE.

It means global/SWITCH-level config takes effect.  The port would use SINGLE_STEP mode.   I'll resolve this by updating the comments for the PORT attributes (same action as above).

>    If we want to enable PTP single step timestamp on all ports, and disable PTP processing on port P1, what should be the config?

Today, before my PR, you'd do this by using the per-port API.  You'd set all ports to SINGLE_STEP, and then set the ports you wanted to disable to NONE.  

With my PR, this doesn't change.  You wouldn't set the SWITCH attribute at all, and it would default to NONE, and you'd get the same bahvior.

In future, if you wanted to use the SWITCH API to enable PTP for all ports, and then switch it off on a per-port basis, you could add a new PORT-level attribute called "OFF" (or similar).  But I don't want to add that as part of this change, since there's no requirement for it yet, and I don't want to add attributes that no-one plans to use.  Sound OK?

inc/saiswitch.h
rdsugarmannv: > This comment is misleading. We can still have one or more ports with SAI_PORT_ATTR_PTP_MODE == SAI_PORT_PTP_MODE_TWO_STEP_TIMESTAMP. These ports will NOT use one-step PTP mode.

Agree - will fix to say "unless over-ridden on a per-port basis".

inc/saiswitch.h
JaiOCP: +    /**
+     * @brief Global PTP mode configuration
+     *
+     * Global PTP mode configuration for the switch.
+     * Applies to all ports unless overridden by port-specific settings.
+     *
+     * @type sai_port_ptp_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PORT_PTP_MODE_NONE
+     */
+    SAI_SWITCH_ATTR_PORT_PTP_MODE,
+

JaiOCP: There is no reason to create a separate switch level ptp enums. What is supported either at switch level or port can be queried by the capability query

rdsugarmannv: > There is no reason to create a separate switch level ptp enums. What is supported either at switch level or port can be queried by the capability query

Latest commit removes the SWITCH level enum and uses just the PORT level enum as requested.

inc/saiport.h
```diff
@@ -395,6 +395,9 @@ typedef enum _sai_port_priority_flow_control_mode_t
 
 /**
  * @brief PTP mode
+ * These modes can be used at the port and switch level.
+ * All ports use the value set at the switch level unless explicitly configured
+ * at the port level to a value other than SAI_PORT_PTP_MODE_NONE.
```
rck-innovium, line 400: @kcudnik  We want SAI_SWITCH_ATTR_PTP_MODE (new) and SAI_PORT_ATTR_PTP_MODE (existing) to use the same enums. Today,  SAI_PORT_ATTR_PTP_MODE is using sai_port_ptp_mode_t.

Is there a way to deprecate sai_port_ptp_mode_t and instead use a new sai_ptp_mode_t? The enum values will be the same.

kcudnik: You can change enum in switch attributes in new one, and add in comments that this is deprecated, it will introduce issues in serialization and deserializstion in sonic sairedis where we would need to add manual conversion for both of those enums, we can discuss this on sai community meeting today

Is there a reason that ptp mode was introduced in switch if it applies to ports anyway ? Or you just want to have anum without port I name? If values will be the same then it seems like one of those enums is redundant and should not be added on the first place

kcudnik: And it seems like switch attributes should be named SAI SWITCH ATTR PORT PTP MODE maybe

I think description of this attributes explains it, but also it seems like this attributes is redundant, if this only applies to ports, then this attribute will n3ver have effect since each time port ptp mode will be different, port always overrides this, not sure why this was added at the first place, let's discuss on sai meeting g

rdsugarmannv: > @kcudnik We want SAI_SWITCH_ATTR_PTP_MODE (new) and SAI_PORT_ATTR_PTP_MODE (existing) to use the same enums. Today, SAI_PORT_ATTR_PTP_MODE is using sai_port_ptp_mode_t.
> 
> Is there a way to deprecate sai_port_ptp_mode_t and instead use a new sai_ptp_mode_t without breaking backward compatibility? The enum values will be the same.

I don't see an obvious way to deprecate the old names that doesn't introduce confusion.  There are other examples that share enums the way I've done here.  For example, see how sai_tunnel_encap_ecn_mode_t is used by both SAI_SWITCH_TUNNEL_ATTR_TUNNEL_ENCAP_ECN_MODE and SAI_TUNNEL_ATTR_ENCAP_ECN_MODE.

rdsugarmannv: > And it seems like switch attributes should be named SAI SWITCH ATTR PORT PTP MODE maybe
> 
> I think description of this attributes explains it, but also it seems like this attributes is redundant, if this only applies to ports, then this attribute will n3ver have effect since each time port ptp mode will be different, port always overrides this, not sure why this was added at the first place, let's discuss on sai meeting g

I'll be at the SAI meeting today.

kcudnik (approved):

doc/SAI-Proposal-Global-PTP-Configuration.md
```diff
@@ -0,0 +1,81 @@
...
+## Motivation
+Currently, PTP configuration in SAI is port-based, and many switches only support configuration of PTP on a global basis.
+In addition, many existing switches have a configuration model which allows global configuration of PTP, and then override on a port level.  SAI should follow this flexible example.
+
+## Technical Specification
+
+### New Switch PTP Configuration Type
+```c
+typedef struct _sai_switch_ptp_mode_t
+{
```
rck-innovium, line 25: All the comments that are addressed in the C header files needs to be reflect in this MD file as well.

kcudnik:
please fix errors:
WARNING: not expected number of spaces after * (1,4,8) saiport.h 400: * 
WARNING: line ends in whitespace saiport.h 400:  *

eddyk-nvidia (approved):

inc/saiswitch.h
```diff
@@ -3197,6 +3197,18 @@ typedef enum _sai_switch_attr_t
...
+     * @brief Global PTP mode configuration
+     *
+     * Global PTP mode configuration for the switch.
+     * Applies to all ports unless overridden by port-specific settings.
+     *
+     * @type sai_port_ptp_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PORT_PTP_MODE_NONE
+     */
+    SAI_SWITCH_ATTR_PTP_MODE,
```
rck-innovium, line 3210: The sai meta checkers would expect this attribute to be named as SAI_SWITCH_ATTR_PORT_PTP_MODE or the enum to be named as sai_ptp_mode_t.

rck-innovium (approved):

tjchadaga:
@rdsugarmannv - please squash your commits

rdsugarmannv:
> @rdsugarmannv - please squash your commits

done

PR #2154, opened 2025-03-18
Add SAI port attribute to fetch PAM4 eye height values

prgeor:
SAI port attribute to fetch PAM4 eye heights and widths of the lower, middle and upper eye of a PAM4 signal.

Eye height in mV
Eye width in psecs. If an implementation does not support then, -1

A narrower eye is more sensitive to jitter and smaller eye has lower SNR.

A good eye height and width threshold is implementation defined based upon the desired SNR, jitter tolerance etc for the application or deployment scenario

SAI_PORT_ATTR_PAM4_EYE_VALUES - Is the attribute introduced for PAM4 SerDes lanes.

SAI_PORT_ATTR_EYE_VALUES - Can be used for NRZ SerDes lanes

image

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -2677,6 +2677,14 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_PORT_STAT_EXTENDED,
 
     /**
+     * @brief List of port's PAM4 lanes eye values
+     *
+     * @type sai_port_pam4_eye_values_list_t
+     * @flags READ_ONLY
+     */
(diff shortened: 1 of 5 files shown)


prgeor:
@kcudnik can you review?

kcudnik:
```
ERROR: FATAL: sai_attribute_value_t members were removed on commit 93d80b9, NOT ALLOWED!
```

seems like you are not on the master, can you rebase your change to master, squash and force push here again to satisfy validator

kcudnik (approved):

PR #2156, opened 2025-03-18
Add custom serdes attributes for saiport.h

longhuan-cisco:
**Problem description:**

Today, [saiport.h](https://github.com/opencomputeproject/SAI/blob/ee6f49bde8100e3797f3540ebf362d27f13d96a2/inc/saiport.h#L3842-L4116) and [SONiC](https://github.com/sonic-net/sonic-swss/blob/7a965caf4c7211afca5303191cf731858c791bcd/orchagent/portsorch.cpp#L366-L455) only supports 10+ NPU SI serdes attributes, but there are below situations for certain vendors/platforms:
1. The total number of attributes can be way more than 10+ for certain platforms/vendors (e.g. 40+ attrs), and the number can keep growing even more for future platforms and new ASIC board versions for existing platforms.
2. Attributes can be vendor/platform specific. Different platforms may have totally different sets of attributes (or only very few attributes are common). And SONiC’s orchagent (which is platform agnostic and independently built) should not include a platform-specific extension header file containing those attributes enums.
3. Attributes can be proprietary, which means original names for those attributes cannot be exposed to public, they cannot be upstreamed as it is, unless as dummy/alias attributes.

**Proposal:**
(credit to @kcudnik for his idea https://github.com/opencomputeproject/SAI/pull/2156#issuecomment-2758536612)

Add one single attribute, whose value is a string that includes a JSON object with key-value pairs. Each pair's key represents the serdes attribute name, and each pair's value is a list of values for every serdes lane. This allows vendor-specific serdes attributes to be forwarded in a JSON string without the sender needing to know the details. The sender simply passes along the data, vendor-defined rules (e.g. for SONiC, it's [media_settings.json](https://github.com/sonic-net/sonic-platform-daemons/blob/master/sonic-xcvrd/tests/media_settings.json) on [platform side](https://github.com/sonic-net/sonic-platform-daemons/blob/0b0ea3b2a3ed60ef00f25305631a30314f59a6fe/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py#L35)) determine which key-value pairs to include, and the vendor SDK interprets the JSON accordingly.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -4116,6 +4116,45 @@ typedef enum _sai_port_serdes_attr_t
     SAI_PORT_SERDES_ATTR_RX_PRECODING,
 
     /**
+     * @brief A collection of custom serdes attributes
+     *
+     * The value is of type sai_json_t, which can include multiple custom serdes
+     * attributes. This allows vendor-specific serdes attributes to be forwarded
+     * in a JSON string without the sender needing to know the details. The
+     * sender simply passes along the data, vendor-defined rules determine which
+     * attributes and values to include in different situations, and the vendor
+     * SDK interprets the JSON accordingly.
+     *
+     * Example of the JSON object:
+     * {
+     * "attributes": [
+     * {
+     *    "attr_xyz": {
+     *        "sai_metadata": {
+     *        "sai_attr_value_type": "SAI_ATTR_VALUE_TYPE_INT32_LIST"
+     *        },
+     *        "value": [10, 10, 10, 10]
+     *    }
+     * },
+     * {
+     *    "attr_abc": {
+     *        "sai_metadata": {
+     *        "sai_attr_value_type": "SAI_ATTR_VALUE_TYPE_INT32_LIST"
+     *        },
+     *        "value": [20, 20, 20, 20]
+     *    }
+     * },
+     * ...
+     * ]
+     * }
+     *
+     * @type sai_json_t
+     * @flags CREATE_AND_SET
+     * @default internal
+     */
+    SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION,
+
+    /**
      * @brief End of attributes
      */
     SAI_PORT_SERDES_ATTR_END,


kcudnik:
custom attributes are per vendor, and are not meant to be public, please remove those, and each vendor should implement its own custom attributes, please reffer to this PR: https://github.com/opencomputeproject/SAI/pull/2122 and take a look at its documentation

longhuan-cisco:
cc @prgeor

longhuan-cisco:
> custom attributes are per vendor, and are not meant to be public, please remove those, and each vendor should implement its own custom attributes, please reffer to this PR: #2122 and take a look at its documentation


Thanks for the pointer. 
PR https://github.com/opencomputeproject/SAI/pull/2122 seems to introduce a way to support custom attributes by allowing vendors to put vendor specific headers (containing custom attributes) under `SAI/custom` folder.

But the situation here is: The SAI requestor side, such as SONiC’s orchagent (which is public, platform agnostic and independently built as separate docker), might not be able to include vendor specific headers.
We would need a way to support custom attributes without putting vendor specific headers at `SAI/custom` folder of the requestor side (e.g. SONiC orchagent). Thus raised the proposal of adding public dummy/placeholder serdes attributes in saiport.h to allow vendors to use them for whatever attributes if needed.

If today's enum range `SAI_PORT_SERDES_ATTR_CUSTOM_RANGE_START` is reserved only for the custom attributes supposed to be included in `SAI/custom/<headers>`.  Then, I'm wondering if we could add another enum range (e.g. `SAI_PORT_SERDES_ATTR_PUBLIC_CUSTOM_RANGE_START = 0x20000000`) dedicated for those dummy/placeholder serdes attributes?

kcudnik:
your assumption is false, if OA will require custom headers then it will need to include them as metadata generated from headers is per vendor, so if 1st vendor will have custom attribute let say int, and 2nd vendor the same attribute as oid list, you will get crash right away.

if OA want to use custom headers, it need to include them per vendor, and it needs to be compiled per vendor.

Adding unnamed custom attributes make no sense, and it's unsolvable. that's why #2122 is introduced

longhuan-cisco:
> your assumption is false, if OA will require custom headers then it will need to include them as metadata generated from headers is per vendor, so if 1st vendor will have custom attribute let say int, and 2nd vendor the same attribute as oid list, you will get crash right away.
> 
> if OA want to use custom headers, it need to include them per vendor, and it needs to be compiled per vendor.
> 
> Adding unnamed custom attributes make no sense, and it's unsolvable. that's why #2122 is introduced

In this particular case here (sai_port_serdes_attr_t), the values of SAI_PORT_SERDES_ATTR are always of type sai_s32_list_t where the count is number lanes in a port and the list specifies list of values to be applied to each lane, regardless of vendors. (I updated the diff by adding comments indicating the value type as `sai_s32_list_t` for all the new attributes, and making the enum range starting from `SAI_PORT_SERDES_ATTR_PUBLIC_CUSTOM_RANGE_START = 0x20000000` to avoid conflicts with https://github.com/opencomputeproject/SAI/pull/2122)

And OA is platform agnostic, it would not be able to include any vendor specific headers.

Any particular reasons to not be able to introduce dummy serdes attributes with the same value type? Is this a deviation from SAI design?


Add @prgeor for comments also

kcudnik:
Custom range is for each vendor to use, and they are not compatoble, for public interface attributes should be added to main range. Lepsze discuss this on todays sai community meeting

longhuan-cisco:
> Custom range is for each vendor to use, and they are not compatoble,

I understood the [`SAI_PORT_SERDES_ATTR_CUSTOM_RANGE_START = 0x10000000`](https://github.com/opencomputeproject/SAI/blob/b8fbbcc8458a95dddabf9455541dbac08152d4a3/inc/saiport.h#L4113) is reserved for each vendor to use. That's why I updated my diff to put all the [public custom serdes attributers](https://github.com/opencomputeproject/SAI/blob/b8fbbcc8458a95dddabf9455541dbac08152d4a3/inc/saiport.h#L4132C5-L5419) in a separate range starting from [`SAI_PORT_SERDES_ATTR_PUBLIC_CUSTOM_RANGE_START = 0x20000000`](https://github.com/opencomputeproject/SAI/blob/b8fbbcc8458a95dddabf9455541dbac08152d4a3/inc/saiport.h#L4119). 

And, unlike most other types of attributes, a port serdes attribute is always inherently meant to be a list of integers (each integer represents a value for each serdes lane of the specific port), regardless of whether it's common or vendor-specific. That's why I predefined the type of all [public custom serdes attributers](https://github.com/opencomputeproject/SAI/blob/b8fbbcc8458a95dddabf9455541dbac08152d4a3/inc/saiport.h#L4132C5-L5419) as `sai_s32_list_t`, where I didn't see compatibility problem. 

On the other hand, in today's SONiC, the serdes attributes are applied to SAI/SDK by OA on a per-platform basis: each platform/vendor defines its own [attribute abbreviation](https://github.com/sonic-net/sonic-platform-daemons/blob/29e65fe7f6d736841bb140150adea03a3e125132/sonic-xcvrd/tests/media_settings.json#L5)->[attribute value](https://github.com/sonic-net/sonic-platform-daemons/blob/29e65fe7f6d736841bb140150adea03a3e125132/sonic-xcvrd/tests/media_settings.json#L6-L9) in `/usr/share/sonic/device/<platform>/media_settings.json`, and only the ones defined in this file will get applied.
With these proposed [public custom serdes attributers](https://github.com/opencomputeproject/SAI/blob/b8fbbcc8458a95dddabf9455541dbac08152d4a3/inc/saiport.h#L4132C5-L5419), platform/vendor would be able to easily support large number of newly added proprietary vendor-specific serdes attributes by just updating `media_settings.json` (plus corresponding SDK support), without having to expose the original real names of the attributes to public, without having to change SONiC and SAI.

> for public interface attributes should be added to main range. Lepsze discuss this on todays sai community meeting

(Not sure if I'm understanding correctly) Or were you suggesting to put all the [public custom serdes attributers](https://github.com/opencomputeproject/SAI/blob/b8fbbcc8458a95dddabf9455541dbac08152d4a3/inc/saiport.h#L4132C5-L5419) at main range right after [`SAI_PORT_SERDES_ATTR_RX_PRECODING`](https://github.com/opencomputeproject/SAI/blob/b8fbbcc8458a95dddabf9455541dbac08152d4a3/inc/saiport.h#L4105)?

inc/saiport.h
```diff
@@ -4113,7 +4113,1313 @@ typedef enum _sai_port_serdes_attr_t
     SAI_PORT_SERDES_ATTR_CUSTOM_RANGE_START = 0x10000000,
 
     /** End of custom range base */
-    SAI_PORT_SERDES_ATTR_CUSTOM_RANGE_END
+    SAI_PORT_SERDES_ATTR_CUSTOM_RANGE_END,
+
+    /** Public custom range base value */
+    SAI_PORT_SERDES_ATTR_PUBLIC_CUSTOM_RANGE_START = 0x20000000,
```
kcudnik, line 4119: this range is reserved for DASH

kcudnik (changes requested):
lets have a live discussion since i dont see point of introducing arbitrary incompatible attributes for all vendors

tjchadaga:
@kcudnik, @longhuan-cisco - do you want to discuss this in this week's SAI community meeting?

longhuan-cisco:
sure, let's discuss, could you please send me (longhuan@cisco.com) the link and include @prgeor also.

inc/saiport.h
longhuan-cisco: Changed to `0x30000000` to avoid conflict with DASH.

And will discuss more in the meeting.

kcudnik:
my proposal:
```c
    /**
     * @brief custom vendor intersface
     *
     * vendor have 3 attributes: custom_1, custom_2, custom_3
     *
     * we carete json/ini/or vendor format:
     *
     * std::string s1 =
     * {
     * "custom_1": "value1"
     * }
     *
     *
     *
     * sai_attribute_t a;
     * a.id = SAI_PORT_SERDES_ATTR_CUSTOM_INTERFACE;
     * a.value.s32list.count = s1.lenght();
     * a.value.s32list.data = s1.data()
     * portserdesapi->set_port_serteds_attribute(&a); // this will set custom_1
     *
     *
     *
     * std::string s2 =
     * {
     * "custom_1": "value1" // string, int, bool
     * "custom_2": "value2"
     * }
     *
     * foo=bar // old ini file format
     *
     *
     * a.value.s32list.count = s2.lenght();
     * a.value.s32list.data = s2.data()
     * portserdesapi->set_port_serteds_attribute(&a); // this will set custom_1 and custom_@
     *
     * @type sai_s32_list_t
     * @flags CREATE_AND_SET
     * @default internal
     */
    SAI_PORT_SERDES_ATTR_CUSTOM_INTERFACE, // etc

```

longhuan-cisco:
Thanks @kcudnik for joining today's discussion and the proposal.
I updated the diff as what you proposed, and modified the PR description to reflect the latest.

BTW. Since it's your idea, just feel free to close mine, in case if you feel more comfortable to raise your own PR,

kcudnik:
> Thanks @kcudnik for joining today's discussion and the proposal. I updated the diff as what you proposed, and modified the PR description to reflect the latest.
> 
> BTW. Since it's your idea, just feel free to close mine, in case if you feel more comfortable to raise your own PR,

im ok for you to drive this change

inc/saiport.h
```diff
@@ -4104,6 +4104,30 @@ typedef enum _sai_port_serdes_attr_t
...
+     * include, and the vendor SDK interprets the JSON accordingly.
+     *
+     * Example of the JSON object:
+     *  {
+     *     "attr1": ["lane1_value", "lane2_value", "lane3_value", "lane4_value"],
+     *     "attr2": ["lane1_value", "lane2_value", "lane3_value", "lane4_value"],
+     *      ...
+     *  }
+     *
+     * @type sai_u8_list_t
```
harshitgulati18, line 4125: Should we use sai_json_t here instead?
https://github.com/opencomputeproject/SAI/blob/master/inc/saitypes.h#L1134-L1160

longhuan-cisco: If we use `sai_json_t`, does it enforce the JSON string to follow the format below (including metadata validation), or are we free to define any format for its contents?
https://github.com/opencomputeproject/SAI/blob/46e839113f19afff070c9ce5c277324bcbba398a/inc/saitypes.h#L1135-L1148

Additionally, here, the `"value"` should be of type `sai_s32_list_t`. Should we then represent it as JSON array, something like `"value": [10, 11, 12, 13]`?

kcudnik: i think attribute format should be vendor specific

longhuan-cisco: Updated type to `sai_json_t`.

And according to [HLD](https://github.com/opencomputeproject/SAI/blob/master/doc/SAI_Proposal_Generic_Extensions.md#sai_json_t), `grammar` of this JSON is predefined, so I followed the same `grammar` to provide the example.
https://github.com/opencomputeproject/SAI/blob/46e839113f19afff070c9ce5c277324bcbba398a/doc/SAI_Proposal_Generic_Extensions.md?plain=1#L52

kcudnik (approved):

lguohan:
@prgeor , can you take a look?

inc/saiport.h
```diff
@@ -4136,7 +4136,7 @@ typedef enum _sai_port_serdes_attr_t
      * }
      *
      * @type sai_json_t
-     * @flags CREATE_AND_SET
+     * @flags CREATE_ONLY
```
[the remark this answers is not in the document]
longhuan-cisco: All today's serdes attributes are `CREATE_ONLY`, so just aligned with it.

Normally, e.g. in SONiC, serdes attributes are applied via [`sai_port_api->create_port_serdes`](https://github.com/sonic-net/sonic-swss/blob/492a85aef6d23196a78e17590cb6e7abaecd8aba/orchagent/portsorch.cpp#L8953C14-L8953C46)
, and before `create`, [`sai_port_api->remove_port_serdes`](https://github.com/sonic-net/sonic-swss/blob/492a85aef6d23196a78e17590cb6e7abaecd8aba/orchagent/portsorch.cpp#L8927) gets called to remove existing ones. 

Add @prgeor if any opinions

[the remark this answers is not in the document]
prgeor: @longhuan-cisco can you please make this `CREATE_AND_SET`, so that we have the flexibility to change the attribute at run time.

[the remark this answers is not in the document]
longhuan-cisco: sure, updated to `CREATE_AND_SET` for future flexibility

prgeor (commented):
can we hold the PR merge. @longhuan-cisco will sync with you offline.

longhuan-cisco:
Hi @prgeor,
Just a gentle follow-up — we went over it last week; just checking in on the merge when you get a chance.

prgeor (approved):

harshitgulati18 (approved):

kcudnik:
please fix build isse:
```

WARNING: not expected number of spaces after * (1,4,8) saiport.h 4133:     *         "sai_metadata": {
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4134:     *           "sai_attr_value_type": "SAI_ATTR_VALUE_TYPE_INT32_LIST"
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4135:     *         },
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4136:     *         "value": [10, 10, 10, 10]
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4137:     *       },
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4138:     *       "attr_abc": {
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4139:     *         "sai_metadata": {
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4140:     *           "sai_attr_value_type": "SAI_ATTR_VALUE_TYPE_INT32_LIST"
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4141:     *         },
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4142:     *         "value": [20, 20, 20, 20]
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4143:     *       },
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4144:     *       ...
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4145:     *     }
WARNING: not expected number of spaces after * (1,4,8) saiport.h 4146:     *   ]
```

longhuan-cisco:
Committed [change](https://github.com/opencomputeproject/SAI/pull/2156/commits/65d304c853d0f8ba881d84d1054bf4b29e0fb281) to: 
1.  Fix Doxygen complaint by adjusting the indentations and encapsulating JSON example with `verbatim` notation.
2.  Fix the JSON example to exactly follow the `sai_json_t` grammar defined in `SAI_Proposal_Generic_Extensions.md`, by having each attribute as separate object under the array of the mandatory top-level key `attributes`
https://github.com/opencomputeproject/SAI/blob/a56c38ed33aab0f81ec1c1e9f716ff722dc94f96/doc/SAI_Proposal_Generic_Extensions.md?plain=1#L179-L196

longhuan-cisco:
I saw [CodeQL run ](https://github.com/opencomputeproject/SAI/actions/runs/14482565056/job/40622251950?pr=2156)got canceled due to below two errors, which doesn't seem to be related to my change
```
1. This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, 
2. GitHub Actions has encountered an internal error when running your job.
```
Wondering if there's a way to re-run it?

kcudnik:
yes, you need to force push squashed

longhuan-cisco:
I did force-push squashed. 
It's weird, [build](https://dev.azure.com/mssonic/build/_build/results?buildId=831598&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=1f6d5e21-f0ce-508d-13df-d1c50f7ea7b9&l=260) still complained about below error, even though my commit simply adds one enum, it doesn't remove anything.
```
ERROR: FATAL: sai_attribute_value_t members were removed on commit 03a9a17, NOT ALLOWED!
make: *** [Makefile:90: all] Error 1
```

kcudnik:
you will need to squash your commits and rebase to master

longhuan-cisco:
> you will need to squash your commits and rebase to master and then force pushed, since it seems like you are behind master

Thanks, rebased to latest master

kcudnik (approved):

PR #2162, opened 2025-04-07
MACSec and IPSec FIPS Compliance

JaiOCP:
This PR brings support for running POST on an IPSec and MACSec engine mainly for FIPS compliance


inc/saimacsec.h
```diff
@@ -67,6 +67,24 @@ typedef enum _sai_macsec_max_secure_associations_per_sc_t
...
+    SAI_MACSEC_POST_STATUS_UNKNOWN,
+
+    /** Pass */
+    SAI_MACSEC_POST_STATUS_PASS,
+
+    /** In Progress */
+    SAI_MACSEC_POST_STATUS_IN_PROGRESS,
+
+    /** Fail */
+    SAI_MACSEC_POST_STATUS_FAIL,
```
kcudnik, line 85: unnecessary comments, not introducing any meaningful information, please remove

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,397 @@
...
+2. MACSec/IPSec object create
+3. MACSec/IPSec port object create
+4. Once the ports are created and SA association and other attributes are set, the port is ready to rx/tx traffic.
+
+This document introduces a stage called as Pre-Operational Self-Test "POST" before the port is operational to rx/tx traffic.
+
+FIPS-103 compliance requires that POST is executed on each port and only if POST completes with ‘success’, port must be enabled for admitting traffic, else port will remain in down state.
+
+There are two variations of MACSec/IPSec engine and ports
+1. Each MACSec/IPSec emgine serving ‘n’ number of ports (1::n)
```
judyjoseph, line 35: typo "engine"

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,397 @@
...
+        _In_ sai_macsec_post_status_t macsec_post_status);
+
+/**
+ * @brief Attribute Id in sai_set_switch_attribute() and
+ * sai_get_switch_attribute() calls.
+ */
+typedef enum _sai_switch_attr_t
+{
+    /**
+     * @brief Callback for completion status of all the MACSEC ports serviced by this MACSEC engine
```
judyjoseph, line 284: does the notification callback gets registered at the sai switch level or at the macsec engine level

skeesara-nokia:
Can we consider support for running POST - where it can be requested at the switch level - preferably also be able to request it as part of switch_create.

The most common use case for systems with FIPS is - they are running in a mode where FIPS is enabled or not enabled. It isn't typical for FIPS compliance to be required on some MACSEC ports and not on others. For this, the application implementation is simplified significantly, if it could request POST at initialization time and proceed on the knowledge of whether that was successful for all MACSEC engines in hardware or not. 

If an implementation does not support requesting POST at this global/swich level - it would indicate that the attribute is not supported. In that case the application will need to implement code to request it whenever it creates individual MACSEC ports as described in this spec. That can be a more sophisticated application implementation that does not have be done right away.

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,483 @@
...
+SAI_OBJECT_TYPE_SWITCH -> SAI_SWITCH_ATTR_IPSEC_OBJECT_LIST[x] -> SAI_IPSEC_ATTR_SUPPORTED_PORT_LIST[] -> SAI_IPSEC_PORT_ATTR_PORT_ID
+```
+
+### 2.1 MACSec
+Following sections describes the attributes for MACSec engine
+
+#### 2.1.1 MACSec Engine
+Each MACSec engine is represented by a MACSec SAI object id. Single MACSec object id can be serving one or more MACSec ports.
+
+New attribute is introduced to trigger POST at the MACSec engine granularity.
```
dipankar-nexthop, line 58: Typically the entire system is FIPS certified and it requires POST before MACsec or IPsec is enabled.  AFAIK, existing systems often use a simple implementation which invokes POST  for all security engines in the system at the time of system initialization.  For SAI, that would requires only static boolean controls (CREATE_ONLY attributes) SAI_SWITCH_ATTR_MACSEC_POST_ENABLED and SAI_SWITCH_ATTR_IPSEC_POST_ENABLED.  With that approach, each ASIC vendor decides when to run POST within their system initialization sequence and removes the responsibility of correctly triggering POST from the NOS.  Can we take this approach and remove all SAI attributes that trigger POST (SAI_IPSEC_ATTR_ENABLE_POST, SAI_MACSEC_ATTR_ENABLE_POST, SAI_SWITCH_ATTR_IPSEC_ENABLE_POST, SAI_SWITCH_ATTR_MACSEC_ENABLE_POST)?

inc/saiswitch.h
```diff
@@ -3277,6 +3305,102 @@ typedef enum _sai_switch_attr_t
...
+    SAI_SWITCH_ATTR_MACSEC_ENABLE_POST,
+
+    /**
+     * @brief Setting the value to true will start the post on all IPSEC engines
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_SWITCH_ATTR_IPSEC_ENABLE_POST,
```
JaiOCP, line 3364: Change the flags CREATE_ONLY so as to suppress runtime change of this attribute

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,483 @@
...
+1. Switch create
+2. MACSec/IPSec object create
+3. MACSec/IPSec port object create
+4. Once the ports are created and SA association and other attributes are set, the port is ready to rx/tx traffic.
+
+This document introduces a trigger called as Pre-Operational Self-Test "POST" before the MACSec/IPSec engine enables the MACSec/IPSec port for forwarding traffic.
+
+FIPS-103 compliance requires that POST is executed on each MACSec/IPSec port and only if POST completes with ‘success’, the MACSec/IPSec port must be enabled for admitting traffic.
+
+There are two variations of MACSec/IPSec engine and ports
```
dipankar-nexthop, line 36: After 4/17/25 SAI meeting discussion, I went back and reexamined the definition of SAI_MACsec and SAI_IPsec objects.  As per https://github.com/opencomputeproject/SAI/blob/master/doc/macsec-gearbox/SAI_MACsec_API_Proposal-v1.4.docx, each SAI ASIC is is expected to have only 2 instances of SAI_MACSEC - one for Rx (decryption) and another for Tx (encryption).  The SAI IPsec documentation is less prescriptive, but AFAIR was based on the same principle.  All MACsec/IPsec enabled ports of a SAI_SWITCH are associated to one instance of such objects.  That is why there is no attribute for object count or associativity with ports.

Unless there is some important POST feature loss, the simplest would be to adopt this definiton and not increase the number of SAI_MACsec and SAI_IPsec objects.  Otherwise we have to make signficant changes to ensure:
1. Compatibility with existing SAI drivers
2. Introduce the concept of multiple Security Engine per SAI_SWITCH and allow SAI_MACSEC and SAI_IPSEC instances per Security Engine
3. Explain additional POST features enabled by allowing multiple Security Engines.
4. Expose Security Engine count and port binding (typically each port is bound to one Security Engine)

kcudnik:
please resolve conflicts

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,483 @@
...
+```
+### 3.0 Capability Query
+Switch can be queried to find out if the system is IPSec and/or MACSec compliant. This is done using simple GET APIs for attributes SAI_SWITCH_ATTR_IPSEC_FIPS_COMPLIANT and SAI_SWITCH_ATTR_MACSEC_FIPS_COMPLIANT.
+
+Query API is used to determine if POST is supported at switch level or engine level granuarity.
+```
+    sai_status_t status = SAI_STATUS_SUCCESS;
+    sai_attr_capability_t post_capability={0}; 
+
+    // Query if MACSec POST is supported at Switch level
```
wumiaont, line 413: I have some concern here about how to trigger POST. As my understanding I need to take SAI_SWITCH_ATTR_MACSEC_ENABLE_POST as one of the attribute and then finally trigger sai_switch_api->create_switch() to create switch. the gSwitchId is the output of create_switch(). Before I pass in the SAI_SWITCH_ATTR_MACSEC_ENABLE_POST I need to check if the switch support POST or not. Which the query needs gSwitchId. 
SAI_SWITCH_ATTR_MACSEC_ENABLE_POST will be CREATE_ONLY based on the comments below. Wondering how would I do the query before switch_create().

dipankar-nexthop: Perhaps we could unconditionally allow setting SAI_SWITCH_ATTR_MACSEC_ENABLE_POST and the switch could return a POST_NOT_SUPPORTED status?

doc/fips/SAI-Proposal-FIPS-Compliance.md
kartik-arista: This can work as long as there is no assumption that POST completes synchronously. We cannot guarantee that every implementation is able to implement a synchronous POST.   We need to retain the notification API that the NOS can hook to, so it can act when POST completes.

Can you please confirm you agree with this @dipankar-nexthop ?

dipankar-nexthop: I agree POST is likely not going to complete synchronously and the notification API would be useful.

inc/saimacsec.h
```diff
@@ -313,6 +327,24 @@ typedef enum _sai_macsec_attr_t
...
+
+    /**
+     * @brief Setting the value to true will start the post on all the ports serviced by this MACSEC engine
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_MACSEC_ATTR_ENABLE_POST,
+
```
JaiOCP, line 347: MACSec and IPSec object level POST enable is mainly for diagnostics purposes.
Add comments in documentation and attribute clarifying this.

doc/fips/SAI-Proposal-FIPS-Compliance.md
JaiOCP: We need to handle the case where FIPS module is present and not present. Running POST always will be not honoring the configuration. POST should run based on the NOS configuration.

kcudnik:
please resolve conflicts

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,467 @@
...
+ sai_macsec_api->create_macsec(
+     &macsec_obj,
+     switch_id,
+     attrs.size(),
+     attrs.data());
+```
+
+**Step 4: ** 
+Once the POST is completed on all the ports hosted by the MACSec engine, registered post macsec callback will be called by the SAI adapter.
+If the status is SAI_MACSEC_POST_STATUS_FAIL, NOS MUST read the post status of all the ports hosted by the engine to find out which port POST has failed.
```
dipankar-nexthop, line 445: Why is NOS required to read the POST status of all ports?  Can't it simply disable MACsec for the whole switch?
Also, the example code below is incorrect/mispeading.  SAI_MACSEC_ATTR_SUPPORTED_PORT_LIST is a list of SAI_PORT objects, not SAI_MACSEC_PORT objects.

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,467 @@
...
+        _In_ sai_macsec_post_status_t macsec_post_status);
+
+/**
+ * @brief Attribute Id in sai_set_switch_attribute() and
+ * sai_get_switch_attribute() calls.
+ */
+typedef enum _sai_switch_attr_t
+{
+    /**
+     * @brief Callback for completion status of all the MACSEC ports serviced by this MACSEC engine
```
dipankar-nexthop, line 176: This should for all MACsec capable ports of the SAI_SWITCH object.

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,467 @@
...
+
+**Step 4: ** 
+Once the POST is completed on all the ports hosted by the MACSec engine, registered post macsec callback will be called by the SAI adapter.
+If the status is SAI_MACSEC_POST_STATUS_FAIL, NOS MUST read the post status of all the ports hosted by the engine to find out which port POST has failed.
+```
+for all macsec_port_obj in macsec_obj->SAI_MACSEC_ATTR_SUPPORTED_PORT_LIST:
+   read(macsec_port_obj->SAI_MACSEC_PORT_ATTR_POST_STATUS)
+```
+**Step 5: **
+Set the POST enable flag in the macsec object to false. This is mainly for the hw where POST can be triggered runtime after the initialization.
```
dipankar-nexthop, line 451: This cannot be done for a CREATE_ONLY attribute.

inc/saiswitch.h
```diff
@@ -3316,6 +3344,86 @@ typedef enum _sai_switch_attr_t
      */
     SAI_SWITCH_ATTR_PACKET_TRIM_DSCP_RESOLUTION_MODE,
 
+    /**
+     * @brief Callback for completion status of all the MACSEC ports serviced by this MACSEC engine
```
dipankar-nexthop, line 3348: This should be "all MACsec capble ports of the switch", not related to a specific MACsec engine.

inc/saiswitch.h
```diff
@@ -3316,6 +3344,86 @@ typedef enum _sai_switch_attr_t
...
+     * Use sai_macsec_post_status_notification_fn as notification function.
+     *
+     * @type sai_pointer_t sai_macsec_post_status_notification_fn
+     * @flags CREATE_AND_SET
+     * @default NULL
+     */
+    SAI_SWITCH_ATTR_MACSEC_POST_STATUS_NOTIFY,
+
+    /**
+     * @brief Callback for completion status of all the IPSEC ports serviced by this IPSEC engine
```
dipankar-nexthop, line 3359: This should be "all IPsec capble ports of the switch", not related to a specific IPsec engine.

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,467 @@
...
+
+} sai_macsec_port_attr_t;
+```
+
+#### 2.1.3 POST Completion Callback
+Single aggregate callback function is provided to return the status of POST status for the entire MACSec engine. 
+
+If the engine is servicing a single port then this callback essentially becomes a per port callback.
+
+If the engine is servicing multiple ports then each port needs to be queried for its POST status using the READ only attribute of the port.
```
dipankar-nexthop, line 155: Please make this unconditional. A single instance of SAI_MACSEC object is expected to support all MACsec capable ports of a an SAI_SWITCH.

doc/fips/SAI-Proposal-FIPS-Compliance.md
```diff
@@ -0,0 +1,467 @@
...
+    
+    . . .
+
+} sai_macsec_port_attr_t;
+```
+
+#### 2.1.3 POST Completion Callback
+Single aggregate callback function is provided to return the status of POST status for the entire MACSec engine. 
+
+If the engine is servicing a single port then this callback essentially becomes a per port callback.
```
dipankar-nexthop, line 153: This is incorrect.  See below.

dipankar-nexthop (changes requested):

JaiOCP:
https://github.com/opencomputeproject/SAI/pull/2167 new PR

PR #2166, opened 2025-04-23
Next hop group attribute to specify if weights are achieved by repeating members or via the weighting support in the hardware..

erohsik:
Add new next hop group attribute to specify if a next hop group is programmed by repeating the next hop members in proportion to their weight or by using the intrinsic next hop group weight support in the hardware.

### DIFF
--- a/inc/sainexthopgroup.h
+++ b/inc/sainexthopgroup.h
@@ -300,6 +300,17 @@ typedef enum _sai_next_hop_group_attr_t
     SAI_NEXT_HOP_GROUP_ATTR_LABEL,
 
     /**
+     * @brief Weighted multi path configuration mode.
+     * false: Nexthop group is programmed with repeated member entries proportional to their weight
+     * true: Nexthop group is programmed with switch native configuration
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_NEXT_HOP_GROUP_ATTR_NATIVE_WCMP,
+
+    /**
      * @brief End of attributes
      */
     SAI_NEXT_HOP_GROUP_ATTR_END,


kcudnik (approved):

inc/sainexthopgroup.h
```diff
@@ -299,6 +299,17 @@ typedef enum _sai_next_hop_group_attr_t
...
     SAI_NEXT_HOP_GROUP_ATTR_LABEL,
 
+    /**
+     * @brief Weighted multi path configuration mode.
+     * false: Nexthop group is programmed with repeated member entries proportional to their weight
+     * true: Nexthop group is programmed with switch native configuration
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
```
rck-innovium, line 309: The default value of "false" will NOT be backward compatible for devices that support Native WCMP.

The name of this attribute must be SAI_NEXT_HOP_GROUP_ATTR_DISABLE_NATIVE_WCMP with default value as False.

rck-innovium (commented):
Need more understanding so as to standardize the expected behavior.

inc/sainexthopgroup.h
```diff
@@ -299,6 +299,17 @@ typedef enum _sai_next_hop_group_attr_t
      */
     SAI_NEXT_HOP_GROUP_ATTR_LABEL,
 
+    /**
+     * @brief Weighted multi path configuration mode.
+     * false: Nexthop group is programmed with repeated member entries proportional to their weight
```
ashutosh-agrawal, line 304: When the SAI implementation repeats next hop group member entries, should the SAI_SWITCH_ATTR_AVAILABLE_NEXT_HOP_GROUP_MEMBER_ENTRY attribute be updated to reflect the count of the repeated entries, or should it be based on the number of unique members in the ASIC_DB?

tjchadaga:
@erohsik - could you please help address the comments?

inc/sainexthopgroup.h
JaiOCP: @erohsik Kishore, Can you please add comment that NOS should query this attribute as supported or not.
And only if supported then we should make the default as true i.e. native WECP.

If platform doesn't support this attribute then WECMP works just the way today it is.

JaiOCP (approved):
Approving with the suggested comments

PR #2171, opened 2025-05-13
Add switch attributes for CPU port default buffer pools

james-jra:
Adds read-only switch attributes to allow users to query SAI for default
SAI-owned buffer pools for use with CPU ports.

This follows the existing pattern for exposing internal default objects, like default virtual router ID (SAI_SWITCH_ATTR_DEFAULT_VIRTUAL_ROUTER_ID).

Vendors may handle traffic to and from CPU port with specific
internally-managed buffer pools, present from start-of-day with some
default configuration. The new attributes introduced in this change allow
SAI to expose those default buffer pools to the user so they can be
configured and used in QoS configuration.

This change is flexible:
- The attributes of the default buffer pools are implementation
  specific.
- If no such default buffer pool exists, SAI simply returns a Null
  object ID, and the user may create and manage buffer pools as normal.
- Vendors are not required to support both ingress and egress
  attributes. They can support both, one, or neither.

I've not added a `.md`, but if you'd like one for more description of the context and use-case, I'm happy to provide one.

### DIFF
--- a/doc/SAI-Proposal-Default-Buffer-Pools.md
+++ b/doc/SAI-Proposal-Default-Buffer-Pools.md
@@ -1,0 +1,147 @@
+# Configurable Default CPU Port Buffer Pools
+
+Title       | Configurable Default CPU Port Buffer Pools
+------------|--------------------------------------------
+Authors     | Nvidia
+Status      | Draft
+Type        | Standards track
+Created     | 2025-05-13
+SAI-Version | 1.16
+
+## Overview
+
+Introducing read-only attributes to the SAI switch object which allow the user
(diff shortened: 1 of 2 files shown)


kcudnik (approved):

tjchadaga:
@vmittal-msft , @marian-pritsak, @bandaru-viswanath - could you please help review?

rajkumar38 (approved):

james-jra:
@tjchadaga  does this still need more reviewers before it can be merged?

If so, @vmittal-msft  @marian-pritsak  @bandaru-viswanath, could you please review?

PR #2183, opened 2025-07-10
KV for managing the Hierarchical NHG

JaiOCP:
New KV parameter is introduced to manage the system init value of SAI_NEXT_HOP_GROUP_ATTR_HIERARCHICAL_NEXTHOP

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -3694,6 +3694,8 @@ typedef enum _sai_switch_stat_t
 
 /**
  * @def SAI_KEY_NHG_HIERARCHICAL_NEXTHOP
+ * true: Nexthop groups consists of tunnel and IP nexthop
+ * false: Nexthop groups consists of IP nexthop only
  */
 #define SAI_KEY_NHG_HIERARCHICAL_NEXTHOP          "SAI_NHG_HIERARCHICAL_NEXTHOP"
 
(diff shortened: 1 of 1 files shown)


r12f (approved):

prsunny (approved):

inc/saiswitch.h
```diff
@@ -3692,6 +3692,11 @@ typedef enum _sai_switch_stat_t
  */
 #define SAI_KEY_INIT_CONFIG_FILE                  "SAI_INIT_CONFIG_FILE"
 
+/**
+ * @def SAI_KEY_NHG_HIERARCHICAL_NEXTHOP
+ */
```
j-bos, line 3699: For this one, please add a comment on usage, similar to the SAI_KEY_BOOT_TYPE. For example if 0 means hierarchical nhg are disabled and 1 means hierarchical nhg are enabled.

Can take a follow-up item to check on means to read it back.

JaiOCP: Thank You Jason for the review.

I have added comments to clarify the semantics of the Key.
(https://github.com/opencomputeproject/SAI/pull/2183/commits/4557ba517e1ffecef07d91e824ec3ea7faebd8c7)

We will take the reading of KV as an independent PR. This is a good idea.

JaiOCP:
@j-bos Can you please approve the PR.
Tags are getting pulled tomorrow and I would like this PR to be in it.

PR #2188, opened 2025-07-18
Update DWRR scheduler weight range 

kperumalbfn:
Current DWRR weight range is from 1-100, increase to 1-255 as some ASIC supports higher value.


kperumalbfn:
@JaiOCP could you please review

kperumalbfn:
@eddyk-nvidia could you please review

inc/saischeduler.h
```diff
@@ -71,7 +71,7 @@ typedef enum _sai_scheduler_attr_t
     /**
      * @brief Scheduling algorithm weight
      *
-     * Range [1 - 100].
+     * Range [1 - 255].
```
JaiOCP: Please update the comments reflecting the 8 bit value range  0..255, where 0 is invalid value

JaiOCP (approved):
Approving with the assumption that review comment would be taken care off

inc/saischeduler.h
kperumalbfn: @JaiOCP Updated the range to 0...255 and 0 as invalid weight

tjchadaga:
@rajkumar38, @ashutosh-agrawal - please help get this reviewed

inc/saischeduler.h
```diff
@@ -71,7 +71,7 @@ typedef enum _sai_scheduler_attr_t
     /**
      * @brief Scheduling algorithm weight
      *
-     * Range [1 - 100].
+     * Range [0 - 255], 0-invalid weight
```
r12f, line 74: hi Kumaresh, will you also update the sonic-mgmt test to cover the higher values?

r12f, line 74: for better supporting tests, do we need to add capability to return what would be the max value that ASIC supports?

kperumalbfn: Agree @r12f, let us have the capability to query the DWRR weight in the new PRs.

kperumalbfn:
@ashutosh-agrawal @itaibaz Could you please review and approve the PR

itaibaz:
I think there was already a suggestion to add a new switch attribute, dwrr weight range, uint8 range, that will reflect the HW capabilities, for example [1..100] or [1..255] and would allow the application to query that

kperumalbfn:
> I think there was already a suggestion to add a new switch attribute, dwrr weight range, uint8 range, that will reflect the HW capabilities, for example [1..100] or [1..255] and would allow the application to query that

Thanks @itaibaz Max DWRR weight using switch capability will be added in the later PRs. This PR is a minor change to only update the comment.

r12f (approved):
LGTM. Thanks Kumaresh!

ashutosh-agrawal (approved):
LGTM.

Issue #2185, opened 2025-07-17
There is no custom range for SAI_ACL_ENTRY_ATTR_ACTION


iamjustadd:
Hi SAI team,

There is no custom range for SAI_ACL_ENTRY_ATTR_ACTION, only SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_START.
Is there any workaround for generating metadata?

Do you have plan to add thus range?

Thanks

iamjustadd:
Hi team,

I need to update the issue, it seems like it is a bug in saisanitycheck.c that there is not check for SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_START, we have a internal work around.
`
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -561,6 +561,12 @@ bool sai_metadata_is_acl_field_or_action(
             return true;
         }

\+        if (metadata->attrid >= SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_START &&
\+                metadata->attrid <= SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_END)
\+        {
\+            return true;
\+        }
\+
         if (metadata->isextensionattr)
         {
             return true;
`

Thanks

kcudnik:
can you make a PR ? or it is alraedy fixed ?

iamjustadd:
Got it, will do it soon.

PR #2190, opened 2025-07-21
There is no check for custom range of SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE…

iamjustadd:
…_START (#2185)

    1. Fix bug that there is no check for range SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_START to SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_END

### DIFF
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -26,6 +26,7 @@
 #include <string.h>
 #include <alloca.h>
 #include <sai.h>
+#include <saicustom.h>
 #include <saiversion.h>
 #include "saimetadatautils.h"
 #include "saimetadata.h"
@@ -557,6 +558,18 @@ bool sai_metadata_is_acl_field_or_action(
 
         if (metadata->attrid >= SAI_ACL_ENTRY_ATTR_ACTION_START &&
                 metadata->attrid <= SAI_ACL_ENTRY_ATTR_ACTION_END)
+        {
+            return true;
+        }
+
+        if (metadata->attrid >= SAI_ACL_ENTRY_ATTR_ACTION_CUSTOM_RANGE_START &&
+            metadata->attrid <= SAI_ACL_ENTRY_ATTR_ACTION_CUSTOM_RANGE_END)
+        {
+            return true;
+        }
+
+        if (metadata->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_CUSTOM_RANGE_START &&
+            metadata->attrid <= SAI_ACL_ENTRY_ATTR_FIELD_CUSTOM_RANGE_END)
         {
             return true;
         }
@@ -2204,8 +2217,10 @@ void check_attr_acl_fields(
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT8_LIST:
 
             if (md->objecttype == SAI_OBJECT_TYPE_ACL_ENTRY &&
-                    md->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_START &&
-                    md->attrid  <= SAI_ACL_ENTRY_ATTR_FIELD_END)
+                    ((md->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_START &&
+                    md->attrid  <= SAI_ACL_ENTRY_ATTR_FIELD_END) ||
+                    (md->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_CUSTOM_RANGE_START &&
+                    md->attrid <= SAI_ACL_ENTRY_ATTR_FIELD_CUSTOM_RANGE_END)))
             {
                 break;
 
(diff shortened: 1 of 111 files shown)


meta/saisanitycheck.c
```diff
@@ -561,6 +561,12 @@ bool sai_metadata_is_acl_field_or_action(
             return true;
         }
 
+        if (metadata->attrid >= SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_START &&
+            metadata->attrid <= SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_END)
+        {
+            return true;
+        }
+
```
kcudnik: actually im not sure if this is correct, since this check is doing acl entry action and field, custom attributes are not in that range

iamjustadd: To me it is proper, custom acl action/field is also acl action/field, or whereelse we can put this judge?

iamjustadd: Or maybe create another function sai_metadata_is_custom_acl_field_or_action? All logic 
if (sai_metadata_is_acl_field_or_action()) change to 
if (sai_metadata_is_acl_field_or_action() || sai_metadata_is_custom_acl_field_or_action())?

kcudnik: not all custom attributes are field or action, since there are other attributes in acl as well, and this is the problem

iamjustadd: Then I recommand to add 
SAI_ACL_ENTRY_ATTR_FIELD_CUSTOM_RANGE_START = SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_START + 0x00001000

SAI_ACL_ENTRY_ATTR_ACTION_CUSTOM_RANGE_START = SAI_ACL_ENTRY_ATTR_CUSTOM_RANGE_START + 0x00002000

SAI_ACL_TABLE_ATTR_FIELD_CUSTOM_RANGE_START = SAI_ACL_TABLE_ATTR_CUSTOM_RANGE_START + 0x00001000


and do related changes in sanitycheck.

How about this?

iamjustadd:
@kcudnik 
I made some changes, 

1. Delete all custom_range_start/end definitions in sai*.h and replace them with
SAI_*_ATTR_CUSTOM_RANGE_BASE = 0x10000000
Reason: All SAI_.*ATTR_CUSTOM_RANGE_START/END definitions have been in sai.*\.h for approximately nine years. The community only defined the custom folder six months ago. From my perspective, nine years ago, the community expected custom attribute developers to insert all custom attribute definitions between CUSTOM_RANGE_START and CUSTOM_RANGE_END, so that the enumeration value of CUSTOM_RANGE_END would be equal to the enumeration value of the last defined custom attribute + 1. Later, the community introduced the custom folder and expected users to define custom attributes in sai*custom.h. This practice actually had no effect on the enumeration value of CUSTOM_RANGE_END. So CUSTOM_RANGE_START/END should not at previous place.

2. For ACL_TABLE_ATTR and ACL_ENTRY_ATTR , I defined range for normal attribute from  0x10000000 -> 0x10001000, and field attribute from 0x10001000 -> 0x10002000, action attribute from 0x10002000 -> 0x10003000, also change the test.md and saisanitycheck.c to compitable with them.

kcudnik:
custom ranges are designed for internal use for vendors, to define custom attributes that are not meant to public use, hence start and end are after each other and not populated since headers design

i added custom directory about half a year ago for broadcom ask, since they wanted to add custom attributes without modifying existing headers.

kcudnik:
please read custom headers documentation and examples here: https://github.com/opencomputeproject/SAI/tree/master/doc/custom-headers

your PR introduces all custom headers at once when they are not needed at all

iamjustadd:
>custom ranges are designed for internal use for vendors, to define custom attributes that are not meant to public use, hence start and end are after each other and not populated since headers design
>
>i added custom directory about half a year ago for broadcom ask, since they wanted to add custom attributes without modifying existing headers.

Ok, that's my bad. From your description, I think that instead of creating custom attributes in the custom folder, the community expects users to add attributes directly to the defined range_start/end.

Anyway, we have chosen to use create in folder. From us, It's very hard to declare custom attributes in custom folder, because most of custom ranges is already defined and not extensible. Even not decalre all custom headers and move the range, we'd better to change them to range_base instead of range_start/end.

Let's get back to original problem. For ACL part, we have to define something for sanitycheck and test.

how about we define:
 ACL_ENTRY_ATTR_CUSTOM_RANGE_START = 0x10000000,
 ACL_ENTRY_ATTR_CUSTOM_RANGE_END,
 ACL_ENTRY_ATTR_FIELD_CUSTOM_RANGE_START= 0x10010000,
 ACL_ENTRY_ATTR_FIELD_CUSTOM_RANGE_END,
 ACL_ENTRY_ATTR_ACTION_CUSTOM_RANGE_START= 0x10020000,
 ACL_ENTRY_ATTR_ACTION_CUSTOM_RANGE_END

Do some work around in sanitycheck and test.pm, for example:
next if $enum =~ /ACL_(ENTRY|TABLE)_ATTR_FIELD_;
next if $enum =~ /ACL_(ENTRY|TABLE)_ATTR_ACTION_;

PR #2201, opened 2025-08-29
Fix bug of creating metadata for custom attribute SAI_ACL_ENTRY/TABLE_ATTR_(FIELD/ACTION)

iamjustadd:
1. Fix bug in saisanitycheck that there is no check for range
       SAI_ACL_ENTRY_ATTR(/_FIELD/_ACTION)_CUSTOM_RANGE_START to
       SAI_ACL_ENTRY_ATTR(/_FIELD/_ACTION)_CUSTOM_RANGE_END
    2. Remove all SAI_*_CUSTOM_RANGE_START/END, using SAI_*_CUSTOM_RANGE_BASE instead.
    3. Adjust metadata generator to compitable with above changes.

### DIFF
--- a/meta/saisanitycheck.c
+++ b/meta/saisanitycheck.c
@@ -92,8 +92,8 @@ defined_attr_t* defined_attributes = NULL;
 
 /* custom ranges start are the same for all objects */
 
-#define CUSTOM_ATTR_RANGE_START SAI_PORT_ATTR_CUSTOM_RANGE_START
 #define CUSTOM_RANGE_START (0x10000000)
+#define CUSTOM_ATTR_RANGE_START CUSTOM_RANGE_START
 #define EXTENSION_RANGE_START (0x20000000)
 #define EXTENSION_OBJECT_TYPE_COUNT (SAI_OBJECT_TYPE_EXTENSIONS_RANGE_END - SAI_OBJECT_TYPE_EXTENSIONS_RANGE_START)
 /* #define CUSTOM_OBJECT_TYPE_COUNT (SAI_OBJECT_TYPE_CUSTOM_RANGE_END - SAI_OBJECT_TYPE_CUSTOM_RANGE_START) */
@@ -107,6 +107,8 @@ defined_attr_t* defined_attributes = NULL;
 #ifndef SAI_OBJECT_TYPE_CUSTOM_RANGE_END
 #define SAI_OBJECT_TYPE_CUSTOM_RANGE_END (CUSTOM_RANGE_START + SAI_METADATA_CUSTOM_OBJECT_COUNT)
 #endif
+
+#define SUB_CUSTOM_RANGE_SIZE 0x1000
 
 bool is_extensions_enum(
         _In_ const sai_enum_metadata_t* emd)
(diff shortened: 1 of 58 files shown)


iamjustadd:
@kcudnik , it's a new commit.
1. Didn't create any custom headers, only have a example custom ACL header in doc/ .
2. Replace CUSTOM_RANGE_START/END with CUSTOM_BASE.
Reason: In custom headers, create new attribute won't modify the value of CUSTOM_RANGE_END. So it's better to define START/END at custom headers for customer's need.
3. Create ACL_ENTRY_ACTION/FIELD_CUSTOM_RANGE_BASE = ACL_ENTRY_CUSTOM_RANGE_BASE + 0x1000 to align with non-custom attributes range.
4. Modify saisanitycheck.c and test.pm to compitable with above changes.

meta/saisanitycheck.c
```diff
@@ -2204,8 +2216,10 @@ void check_attr_acl_fields(
         case SAI_ATTR_VALUE_TYPE_ACL_FIELD_DATA_UINT8_LIST:
 
             if (md->objecttype == SAI_OBJECT_TYPE_ACL_ENTRY &&
-                    md->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_START &&
-                    md->attrid  <= SAI_ACL_ENTRY_ATTR_FIELD_END)
+                    ((md->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_START &&
+                    md->attrid  <= SAI_ACL_ENTRY_ATTR_FIELD_END) ||
+                    (md->attrid >= SAI_ACL_ENTRY_ATTR_FIELD_CUSTOM_RANGE_BASE &&
+                    md->attrid <= SAI_ACL_ENTRY_ATTR_FIELD_CUSTOM_RANGE_BASE + 0x1000)))
```
kcudnik: since you are using 0x1000 many times, declare it as define

meta/saisanitycheck.c
```diff
@@ -5064,6 +5078,50 @@ void check_acl_entry_actions()
         enum_index++;
     }
 
+    for (; meta_acl_entry[index] != NULL; index++)
+    {
+        if (meta_acl_entry[index]->attrid == SAI_ACL_ENTRY_ATTR_ACTION_CUSTOM_RANGE_BASE)
+        {
+            break;
+        }
+    }
```
kcudnik, line 5097: this is pointless, nothing happens here

meta/saisanitycheck.c
```diff
@@ -5064,6 +5078,50 @@ void check_acl_entry_actions()
...
+
+    while (meta_acl_entry[index] != NULL &&
+           meta_acl_entry[index]->attrid >= SAI_ACL_ENTRY_ATTR_ACTION_CUSTOM_RANGE_BASE &&
+           meta_acl_entry[index]->attrid < SAI_ACL_ENTRY_ATTR_ACTION_CUSTOM_RANGE_BASE + 0x1000) {
```
kcudnik: bracket at the new line {

meta/test.pm
```diff
@@ -124,31 +124,6 @@ sub CreateSwitchIdTest
     WriteTest "}";
 }
 
-sub CreateCustomRangeTest
```
kcudnik, line 127: why tis is removed?

meta/saisanitycheck.c
iamjustadd: Why? I think here is to move the index to the first custom attribute of SAI_ACL_ENTRY_ATTR_ACTION_*

meta/test.pm
iamjustadd: Because no matter there is SAI_*_CUSTOM_RANGE_START/END, If there is SAI_OBJECT_TYPE_* defined, then it will check the range. This make we have to pre-define all custom_range_start/end.

iamjustadd: Besides there is CreateCustomRangeAllTest, which will test the range for exactly defined custom_start/end, I think that function is enough.

meta/saisanitycheck.c
kcudnik: add a comment then whats happening, and check if index gets at the last element then log error

iamjustadd: I added comment for this loop.

This function is to check whether the number of attributes acl_entry_action_* is matched with sai_acl_action_type , in another word, assert that user defined attribute acl_entry_action which must have a correspond acl_action_tyoe.
At line 5136 have this assertion:
    META_ASSERT_TRUE(enum_index == sai_metadata_enum_sai_acl_action_type_t.valuescount,
            "number of acl entry action mismatch vs number of enums in sai_acl_action_type_t");


If User not create custom attribute of ACL_ENTRY_ACTION, then here will loop till the 
meta_acl_entry[index] == NULL(hit the last).
This is not an ERROR.

iamjustadd: @tjchadaga @kcudnik Hi gentlemen, Any other comments?

iamjustadd:
@tjchadaga @kcudnik  Hi team, can we make a progress?

meta/saisanitycheck.c
iamjustadd: @tjchadaga Hi Tejaswini, please kindly review this PR?

PR #2207, opened 2025-09-15
Added port serdes attributes for various taps

prgeor:
Added SAI attribute for TX FIRs, RX FFE, RX DFE and VGA.

These taps are attributed in such a way it does not specify the TAP name or the number of TAPs to keep any vendor specific Serdes details


inc/saiport.h
```diff
@@ -4277,6 +4277,76 @@ typedef enum _sai_port_serdes_attr_t
...
+     */
+    SAI_PORT_SERDES_ATTR_RX_FFE_COUNT,
+
+    /**
+     * @brief Total number of TX FFE taps supported
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_PORT_SERDES_ATTR_TX_FFE_COUNT,
```
prgeor: FIR

PR #2217, opened 2025-10-09
Support OTN attenuator and amplifier

jjin62:
1. Add a new precision attribute flag for optical analog values (e.g., gain, tilt, power), instead of using double/float point type which would introduce NaN issue.
2. Add support for optical attenuator in experimental extensions, derived from openconfig-optical-attenuator.yang.
3. Add support for optical amplifier in experimental extensions, derived from openconfig-optical-amplifier.yang.


jjin62:
1) Introducing two new SAI objects OTN-OA and OTN-ATTENUATOR using SAI experimental extension mechanism. OA and ATTENUATOR are optical modules which can form a simplest optical amplifier.
2) Gauged/Analog values (optical power, gain, attenuation etc..) are usually represented in a float point format (double or float). However, double/float data type is prohibited in SAI to avoid NaN issue (see meta/Makefile compile flag "-Wfloat-equal"). Therefore, a @precision tag is introduced to represent the number of digits after the decimal point. Any integer type with a @precision tag can be used to represent a decimal type.
Ref #2216

meta/parse.pl
```diff
@@ -2353,6 +2365,21 @@ sub ProcessNextRelease
...
 }
 
+sub ProcessPrecision {
+    my ($stat, $precision) = @_;
+
+    # Default 0 if precision is not provided
+    return 0 unless defined $precision;
+
+    # Must be an integer >= 0
+    if ($precision =~ /^\d+$/) {
```
kcudnik: { to new line, and add empty libe before return same in line 434

jjin62: done.

meta/parse.pl
```diff
@@ -2477,6 +2508,7 @@ sub ProcessSingleObjectType
         WriteSource ".iscustom                      = ($attr >= 0x10000000) && ($attr < 0x20000000),";
         WriteSource ".apiversion                    = $apiversion,";
         WriteSource ".nextrelease                   = $nextrelease,";
+        WriteSource ".valueprecision                = $precision,";
```
kcudnik, line 2515: there should be test in saisanitycheck.c for that value if it's in reasonable range, like 0..18 etc since you can put number like 23423542542 and it will be cast as int, truncated, to some absurd value which will make no sense

jjin62: good point. check precision range [0..18], 18 is the max precision used in transponder.

image

kcudnik (approved):

tjchadaga:
@yawei-yin - please help sign-off

yawei-yin (approved):

qiluo-msft:
@jjin62 Could you check build failure?
```
ERROR: FATAL: sai_qos_map_params_t members were removed on commit fbbd7e4, NOT ALLOWED!
```

jjin62:
I am doing another rebase to the latest master.

jjin62:
Rebased. @tjchadaga, please start the pipeline again, /azp run. Thanks.

tjchadaga:
Thanks, @jjin62 - could you also squash your commits and force push?

jjin62:
@tjchadaga squash done.

tjchadaga (approved):

PR #2229, opened 2025-12-05
add SAI definitions for optical circuit switch

wni0812:
Add API definitions for Optical Circuit Switch, mainly including:
1. Creation and remove of OCS cross-connects, including both individual switching and simultaneous switching with multiple connections.
2. OCS ports operation, including the ability to set a state to any specified group of ports.
3. Retrieval of current operational status on OCS port.
4. Retrieval of hardware-specific OCS module data, such as port physical mapping information and factory insertion loss values etc.


jjin62:
Need to add new SAI objects and types for OCS in SAI extension header files. Without the following change, the SAI OCS object/api will not be processed by parser to generating meta data.

_saiextensions.h_
````
/* OCS extension */
#include "saiexperimentalocs.h"

...
   SAI_API_DASH_APPLIANCE,

   SAI_API_OCS_CROSS_CONNECT,

   SAI_API_OCS_PORT,

    /* Add new experimental APIs above this line */

   SAI_API_EXTENSIONS_RANGE_END

} sai_api_extensions_t;

#endif /* __SAIEXTENSIONS_H_ */
````
_saitypesextensions.h‎_

````
    SAI_OBJECT_TYPE_DASH_APPLIANCE,

    SAI_OBJECT_TYPE_OCS_CROSS_CONNECT,

    SAI_OBJECT_TYPE_OCS_PORT,

    /* Add new experimental object types above this line */

    SAI_OBJECT_TYPE_EXTENSIONS_RANGE_END

} sai_object_type_extensions_t
````

File: experimental/saiexperimentalocs.h [hunk 1]
```diff
@@ -0,0 +1,453 @@
...
+
+    /**
+     * @brief B side port
+     *
+     * @type sai_object_id_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_OCS_PORT
+     */
+    SAI_OCS_CROSS_CONNECT_ATTR_B_SIDE_PORT_ID,
+
```
jjin62, line 65: should we add operational status so user know if a connection is up or down, without looking at the corresponding port status?

````
/**
     * @brief Cross connect operation status
     *
     * @type sai_ocs_cross_connect_oper_status_t
     * @flags READ_ONLY
     */
    SAI_OCS_CROSS_CONNECT_ATTR_OPER_STATUS,
````

File: experimental/saiexperimentalocs.h [hunk 2]
```diff
@@ -0,0 +1,453 @@
...
+/**
+ * @brief Attribute data for OCS port status parameter
+ */
+typedef enum _sai_ocs_port_status_t
+{
+    /**
+     * @brief Default status for any port which is with no configuration
+     */
+    SAI_OCS_PORT_STATUS_BLOCKED,
```
jjin62, line 213: This default value (SAI_OCS_PORT_STATUS_BLOCKED) conflicts to the default of override_state_t,  SAI_OCS_PORT_OVERRIDE_STATE_NORMAL??

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,453 @@
...
+     * @type sai_u8_list_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     */
+    SAI_OCS_PORT_ATTR_NAME = SAI_OCS_PORT_ATTR_START,
+
+    /**
+     * @brief Overrides the status imposed by the programmed cross-connects
+     *
+     * @type sai_ocs_port_override_state_t
+     * @flags CREATE_AND_SET
```
jjin62, line 253: In latest SAI version, for CREATE_AND_SET attributors, it is required to have default value flag:
`@default SAI_OCS_PORT_OVERRIDE_STATE_NORMAL`
Otherwise parser and compile will be error-out.  All CREATE_AND_SET attributes need to add `@default`

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,453 @@
...
+ */
+typedef enum _sai_ocs_port_override_state_t
+{
+    /**
+     * @brief First element that the beam hits is powered off, for testing
+     */
+    SAI_OCS_PORT_OVERRIDE_STATE_POWERED_OFF,
+
+    /**
+     * @brief Default, state to be determined by presence of cross-connect
```
jjin62: how is port state NORMAL is determined connection present or not?

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,453 @@
...
+
+/**
+ * @brief List of OCS cross connect attributes
+ */
+typedef enum _sai_ocs_cross_connect_attr_t
+{
+    /**
+     * @brief Start of attributes
+     */
+    SAI_OCS_CROSS_CONNECT_ATTR_START,
```
jjin62, line 46: All new acronyms (non-English words), such as `OCS` need to be add the in the meta dictionary `meta/aspell.en.pws`. compile will fail without this.

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,453 @@
...
+     */
+    SAI_OCS_CROSS_CONNECT_FACTORY_DATA_ATTR_B_SIDE_PORT_NAME,
+
+    /**
+     * @brief Center frequency at which insertion loss is measured, e.g., 229.1 (O-band) 193.5, use int32 for representation of decimal value
+     *
+     * List of center frequency values. Use int32 for representation of decimal value with 3 fraction digits
+     *
+     * @type sai_s32_list_t
+     * @flags READ_ONLY
```
jjin62, line 458: To convert sai_s32_list_t -> decimal, how does the application code (ex. SWSS) know that it should have 3 decimal digits, without hard code to  SAI_OCS_CROSS_CONNECT_FACTORY_DATA_ATTR_FREQUENCY_THZ? For decimal -> sai_s32_list_t, how do we validate that the decimal string is 3 precision digits in redis/NBI?

The decimal precision info should be part of the meta data, so the conversion can done generically.  One way is to use `sai_s32_t` and add a new tag `@precision n` and so the application can retrieve the number of decimal digits from meta data. The proposal to introduce a new type `sai_double_t` is dropped due to NAN issue.

Same comments for all decimal representation below.

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,453 @@
...
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
+     * @objects SAI_OBJECT_TYPE_OCS_PORT
+     */
+    SAI_OCS_CROSS_CONNECT_ATTR_B_SIDE_PORT_ID,
+
+    /**
+     * @brief End of attributes
+     */
+    SAI_OCS_CROSS_CONNECT_ATTR_END,
+
```
jjin62, line 70: Add custom rang for vendor extension (see DASH SAI). same for port attributes.
````
/** Custom range base value */
    SAI_OCS_CROSS_CONNECT_ATTR_CUSTOM_RANGE_START = 0x10000000,

    SAI_OCS_CROSS_CONNECT_ATTR_CUSTOM_RANGE_END
````

File: experimental/saiexperimentalocs.h [hunk 8]
```diff
@@ -0,0 +1,453 @@
...
+{
+    /**
+     * @brief Start of device port attributes
+     */
+    SAI_OCS_PORT_ATTR_START,
+
+    /**
+     * @brief Unique identifier for a port
+     *
+     * @type sai_u8_list_t
```
jjin62, line 244: for port name why not using `@type char`?

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,453 @@
...
+ *
+ *    Microsoft would like to thank the following companies for their review and
+ *    assistance with these files: Intel Corporation, Mellanox Technologies Ltd,
+ *    Dell Products, L.P., Facebook, Inc., Marvell International Ltd.
+ *
+ * @file    saiexperimentalocs.h
+ *
+ * @brief   This module defines SAI OCS
+ */
+
```
jjin62, line 26: please add the following to compatible with SAI extension common practice:
````
 *
 * @warning This module is a SAI experimental module.
 */
````

File: experimental/saiexperimentalocs.h [hunk 10]
```diff
@@ -0,0 +1,453 @@
...
+{
+    /**
+     * @brief Start of attributes
+     */
+    SAI_OCS_CROSS_CONNECT_ATTR_START,
+
+    /**
+     * @brief A side port
+     *
+     * @type sai_object_id_t
```
jjin62, line 51: Not sure why you don't use port name (a string) in the port attributes? `sai_object_id `is auto-generated when SAI object is created. Do you want to enforce creating cross-connect after creating ports? This would need extra logic in SWSS? In any case, I think port name A and B are useful info.

experimental/saiexperimentalocs.h
wni0812: Agreed, updated the code. Thanks

experimental/saiexperimentalocs.h
wni0812: Using SAI object id as type is for reference of OCS ports in OCS cross-connect, it can also enforce that OCS cross-connect must be provisioned between existing OCS ports.  OCS ports can be created by swss based device specific on template/profile.

experimental/saiexperimentalocs.h
wni0812: Done, thanks

experimental/saiexperimentalocs.h
wni0812: The OCS port override state is a configurable attribute of OCS port. it is defined as CREATE_AND_SET. It can be configured independently, no matter whether cross-connect exists on the port.

doc/OCS/SAI-Proposal-OCS.md
```diff
@@ -0,0 +1,307 @@
...
+
+![OCS Port Operational Status](images/port_status.png)
+# 4. OCS Cross-Connects
+OCS cross-connects are dynamically created and removed via SAI APIs.    
+## 4.1. Object Relationships
+* The SAI object IDs of OCS ports serve as the SAI attributes for both the ingress and egress ports in OCS cross-connect SAI object.
+* OCS ports are referenced to OCS cross-connect by their SAI object IDs in cross-connect attributes.
+## 4.2. Cross-Connect Configuration 
+* To have fast switching performance, SAI bulk create/remove APIs are used for management of OCS cross-connect configurations.
+* OCS cross-connect can not be provisioned between two A side ports or two B side ports.
```
mdemerchant: This requirement restricts OCS technology choices.  While most OCS devices available today are limited to A to B connections, not all devices have this limitation.  If using bidi optics the ability to connect true any to any port seems like it would be quite valuable, ideally the SAI API should be generalized to support that functionality.

doc/OCS/SAI-Proposal-OCS.md
```diff
@@ -0,0 +1,307 @@
...
+* One OCS port can only involved in one OCS cross-connect; when switching existing OCS cross-connect, e.g. switch from 1A-1B to 1A-2B, the upstream application need to use SAI API to remove 1A-1B and then create 1A-2B.
```
mdemerchant: The described behavior seems like it would be performance limiting if you are trying to make bulk changes.  You would first have to bulk delete existing cross connects, then bulk create the new desired cross connects which requires two operations.  If the behavior is that the switching engine simply replaces the existing connection if you re-use a port then it would be able to do it in one operation.

doc/OCS/SAI-Proposal-OCS.md
```diff
@@ -0,0 +1,307 @@
...
+* The SAI object IDs of OCS ports serve as the SAI attributes for both the ingress and egress ports in OCS cross-connect SAI object.
+* OCS ports are referenced to OCS cross-connect by their SAI object IDs in cross-connect attributes.
+## 4.2. Cross-Connect Configuration 
+* To have fast switching performance, SAI bulk create/remove APIs are used for management of OCS cross-connect configurations.
+* OCS cross-connect can not be provisioned between two A side ports or two B side ports.
+* One OCS port can only involved in one OCS cross-connect; when switching existing OCS cross-connect, e.g. switch from 1A-1B to 1A-2B, the upstream application need to use SAI API to remove 1A-1B and then create 1A-2B. 
+# 5. OCS Cross-Connects Factory Data
+OCS cross-connect factory data are read-only information about insertion loss measurements of all possible cross-connects. 
+* Factory data are measured during the manufacturing calibration phase.
+* The measurement is for every possible optical path (cross-connect) with an optical spectrum (central frequency) in specific environment (temperature).
```
mdemerchant: There is no definition given to standardize what data is actually present, only the format of how the data is presented.  Is it useful without some standard expectation of what measurements are actually made?  What if one vendor decides to measure in O Band and one in C Band.  Or one vendor makes a single wavelength measurement and one decides to measure at ten different wavelengths on each path?

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,464 @@
...
+
+/**
+ * @brief Attribute data for OCS port override state parameter
+ */
+typedef enum _sai_ocs_port_override_state_t
+{
+    /**
+     * @brief First element that the beam hits is powered off, for testing
+     */
+    SAI_OCS_PORT_OVERRIDE_STATE_POWERED_OFF,
```
mdemerchant: I don't think POWERED_OFF is a good choice of state name.  It implies specific OCS technology.  Wouldn't something more generic like DISABLED be better?  It seems the only operational impact of the port being in this state is that you just can't use it to create a cross connection.
What is the actual use case for wanting this state anyway as distinct from the FORCE_BLOCKED state?  What this will do optically seems undefined and will vary from device to device.

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,464 @@
...
+
+    /**
+     * @brief Insertion loss is >0.5dB of target
+     */
+    SAI_OCS_PORT_STATUS_TUNING,
+
+    /**
+     * @brief Insertion loss is within 0.5dB of target
+     */
+    SAI_OCS_PORT_STATUS_TUNED,
```
mdemerchant, line 223: Defining this as loss within 0.5dB of target seems like a strange definition and potentially complex to implement for vendors.  Why not just define it as when the device has finished switching?  For devices that don't have on-going control loops this is simple and for devices that do have on-going control loops they can implement their own threshold of when it's close enough to guarantee their optical performance specs.

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,464 @@
...
+
+    /**
+     * @brief If there is a hardware failure
+     */
+    SAI_OCS_PORT_STATUS_FAILED
```
mdemerchant, line 228: Is this state intended only to represent a permanent HW failure or simply some kind of failure to set up the last requested connection for this port?  What about failure mechanisms where a particular cross connect path connect be achieved but other paths using the port still work fine?  The object definition ties failures to ports but not all failure mechanisms are port related, depending on the OCS technology.

I think representing failures in a technology agnostic way needs something considerably more sophisticated in order to convey useful information to the upper layer SW.

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,464 @@
...
+     */
+    SAI_OCS_PORT_ATTR_OPER_STATUS,
+
+    /**
+     * @brief Mapping between face plate port and OCS physical element
+     *
+     * @type sai_u8_list_t
+     * @flags READ_ONLY
+     */
+    SAI_OCS_PORT_ATTR_PHYSICAL_MAPPING,
```
mdemerchant, line 272: Need to add something to the markdown file or comments here documenting what this attribute actually is.  It doesn't seem to be in the port creation example in the markdown file either.

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,464 @@
...
+ */
+typedef void (*sai_ocs_port_state_change_notification_fn)(
+        _In_ uint32_t count,
+        _In_ const sai_ocs_port_status_notification_t *data);
+
+/**
+ * @brief List of OCS cross connect factory data attributes. Inventory data for all possible cross-connects,
+ * factory insertion loss measurements
+ */
+typedef enum _sai_ocs_cross_connect_factory_data_attr_t
```
mdemerchant, line 429: Are there examples of comparable data structures used for other types of switches in SAI?  Maybe I'm misunderstanding the proposed flow but it seems like the SAI implementation wouldn't actually do anything with this data, not even read it from hardware.  Some higher level platform specific SW would have to read it from device storage and parse it, then pass it into SAI to create all the objects at initialization.  Then later higher level management SW can get the data via those SAI objects.  I can see it makes that second access device independent but it just seems kind of an odd implementation.  It seems very different from how general device inventory information is handled.

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,464 @@
...
+ */
+typedef sai_status_t (*sai_get_ocs_cross_connect_factory_data_attribute_fn)(
+        _In_ sai_object_id_t ocs_cross_connect_factory_data_id,
+        _In_ uint32_t attr_count,
+        _Inout_ sai_attribute_t *attr_list);
+
+/**
+ * @brief OCS methods table retrieved with sai_api_query()
+ */
+typedef struct _sai_ocs_api_t
```
mdemerchant, line 560: I understand that SAI has singular and bulk operations because historically it started with singular and added bulk later.  But if the OCS is a greenfield application is it necessary/useful to include the singular operations?

mdemerchant (commented):
I think this proposal needs discussion with more OCS vendors to make sure it is not too technology specific.  It would be very useful to have a discussion within the OCP OCS sub-project since a large number of vendors are participating.

experimental/saiexperimentalocs.h
wni0812: Yes, the custom range can be defined for vendor specific extensions, updated the code.

experimental/saiexperimentalocs.h
wni0812: In SAI attribute value, char data is defined as: char chardata[32], to support port name with length > 32, sai_u8_list_t is used instead.

experimental/saiexperimentalocs.h
wni0812: For SAI layer as the physical information, cross-connect is more about configuration and the configurational result success/failure. For upper layer, like swss, if necessary, it can use port status and other alarm information to have a logic result about cross-connect operational status.

experimental/saiexperimentalocs.h
wni0812: SAI_OCS_PORT_STATUS_BLOCKED is a value of OCS port operational status while SAI_OCS_PORT_OVERRIDE_STATE_NORMAL is a value of OCS port configuration. When OCS port is created, the default override state configuration is SAI_OCS_PORT_OVERRIDE_STATE_NORMAL. OCS port operational status is updated by OCS driver/SDK, dynamically based on rules (by checking port override state configuration and OCS cross-connect on the port)..

experimental/saiexperimentalocs.h
wni0812: Agreed, updated the code.

experimental/saiexperimentalocs.h
jjin62: > SAI_OCS_PORT_STATUS_BLOCKED is a value of OCS port operational status while SAI_OCS_PORT_OVERRIDE_STATE_NORMAL is a value of OCS port configuration. When OCS port is created, the default override state configuration is SAI_OCS_PORT_OVERRIDE_STATE_NORMAL. OCS port operational status is updated by OCS driver/SDK, dynamically based on rules (by checking port override state configuration and OCS cross-connect on the port)..

got it.

experimental/saiexperimentalocs.h
jjin62: understood. But the description seems not accurate.

experimental/saiexperimentalocs.h
jjin62: 32 should be long enough for a port name? char is much easy to use than u8_list.

experimental/saiexperimentalocs.h
jjin62: This is certain valid design choice. However, In config DB cross-connect uses port name. so we need logic (SWSS) to find our the port's sai_object_id. As sai_object_id is dynamically generated at run time, so it is different on each restart. This make debug a little hard.

wni0812: Understood the point, yes, using this proposal, SONiC SWSS can maintain the association between port sai_object_id and port name. Port name is defined as a SAI OCS port attribute. so in ASIC DB port name can be read via port's port sai_object_id. In OCS vendor SAI lib, association of port sai_object_id and port name can also be  maintained (during the processing of SAI creating OCS ports.

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,471 @@
...
+    sai_get_ocs_port_attribute_fn                              get_ocs_port_attribute;
+    sai_bulk_object_create_fn                                  create_ocs_ports;
+    sai_bulk_object_remove_fn                                  remove_ocs_ports;
+    sai_bulk_object_set_attribute_fn                           set_ocs_ports_attribute;
+    sai_bulk_object_get_attribute_fn                           get_ocs_ports_attribute;
+    sai_create_ocs_cross_connect_fn                            create_ocs_cross_connect;
+    sai_remove_ocs_cross_connect_fn                            remove_ocs_cross_connect;
+    sai_set_ocs_cross_connect_attribute_fn                     set_ocs_cross_connect_attribute;
+    sai_get_ocs_cross_connect_attribute_fn                     get_ocs_cross_connect_attribute;
+    sai_bulk_object_create_fn                                  create_ocs_cross_connects;
```
swapnaNann, line 580: For bulk operations, what execution modes are supported for OCS? Example shows STOP_ON_ERROR. Will additional modes such as IGNORE_ON_ERROR, ROLLBACK_ON_ERROR or ATOMIC be supported when setting up cross-connects?

doc/OCS/SAI-Proposal-OCS.md
wni0812: A and B in the YANG model (and SAI) should denote the two endpoints of cross-connect, they do not necessarily need to be on the A or B sides (think of it as a-end and z-end). Agreed that this line in the documentation isn’t accurate, and the YANG model is meant to support cross-connects like ‘1A-2A' or ‘1-5’ too if the hardware supports it. Feasible cross-connects is platform/hardware dependent can be provided as part of YANG model device capabilities.
SAI definition on cross-connect and ports as A point to B point are string type. It is quite flexible to use any port to any port without limitation, like 1A-2A etc. The description is not accurate in the markdown.
Updated the markdown.

doc/OCS/SAI-Proposal-OCS.md
wni0812: The logic under SAI from vendor can also implement one bulk operation for cross connects change to OCS module or hardware, instead of deletion and then creation.
In current SAI specification, the supported operations/bulk operations on SAI objects are get, set, create, remove. Different operations are not allowed to be mixed in one bulk API call. Remove is needed to delete SAI related objects from logic view, instead of operations to module or hardware. The description is not accurate in the markdown.
Updated the markdown.

doc/OCS/SAI-Proposal-OCS.md
wni0812: The intention here is to provide a model for presenting this data in a generic way that matches the physics of loss measurement, not to enforce completeness of test coverage. Perhaps we can remove the “all/every” cross-connect terminology in this description and replace it with factory beginning-of-life measurements that serve as a baseline reference that users can compare with. The model includes measurement conditions like laser source frequency band or temperature to support various test conditions.
Updated the markdown.

experimental/saiexperimentalocs.h
wni0812: Agreed that POWERED_OFF state is causing more ambiguity in implementation. Considering that this state was meant for optical DVT testing, we can remove it from the YANG model. Regarding NORMAL/FORCE_BLOCKED vs ENABLED/DISABLED, the NORMAL/FORCE_BLOCKED follows the actual optical behavior more explicitly. For example, it is fair to assume a user considers NORMAL as the default state but in ENABLED/DISABLED it is less clear what the default would be and whether this is a logic default state of a port or actual optical blocking state. POWERED_OFF can be removed but keeping the NORMAL/FORCE_BLOCKED states.
POWERED_OFF can be removed from SAI definition header file. If in the future any new technical solution, like pluggable transceiver with laser, then, new enum values for different operations can be added.
Updated the definition of this header file.

experimental/saiexperimentalocs.h
wni0812: Agreed that we can change the description. This 0.5dB was meant as general guidance on what finished switching.
Updated the description of this header file.

experimental/saiexperimentalocs.h
wni0812: This is not a configuration or permanent failure flag. It is meant as a status in the port tuning state machine, for example, when software fails to apply a beam steering state to the DACs due to some communication issue. It is up to the vendor to determine what constitutes a failure in port tuning. Note that the “failed” status is not an all-inclusive status but only used at times that the software knows of a certain failure in tuning.
Permanent failure assignments (for either individual ports or cross-connects) are not modeled in the YANG model currently, those could instead be tracked in an off-box database that control plane has access to.
Updated the description of this header file.

experimental/saiexperimentalocs.h
wni0812: This is the “internal port” mapping to faceplate port.
Updated the description of this header file.

experimental/saiexperimentalocs.h
wni0812: Agree that factory data is read-only. Similar as other devices, ASIC specific data can be queried via SAI interface. If some OCS devices do not support it, SAI definition is there and it is optional. As SAI definition rule, it requires that create, remove, set, get APIs need be defined there, even APIs are not used. Otherwise, the compiling will fail. To ensure that the upper applications are not allowed to create/set/remove the read-only data, vendors can implement these APIs with directly returning ERROR or NOT_IMPLEMENTED.

experimental/saiexperimentalocs.h
wni0812: Based on current understanding, SAI definition rule requires both singular and bulk APIs to be there, even APIs are not used, like other existing SAI saifdb.h or saiport.h. Otherwise, the compile will fail (e.g. in SONiC sairedis build). So, SAI definition need have both singular and bulk APIs, though current implement only uses bulk API. Vendors can implement bulk API for functions and implement singular API with returning ERROR or NOT_IMPLEMENTED.

experimental/saiexperimentalocs.h
wni0812: Two error modes (STOP_ON_ERROR and IGNORE_ERROR) are defined:
`typedef enum _sai_bulk_op_error_mode_t
{
    /**  
     * @brief Bulk operation error handling mode where operation stops on the first failed creation
     *
     * Rest of objects will use SAI_STATUS_NON_EXECUTED return status value.
     */
    SAI_BULK_OP_ERROR_MODE_STOP_ON_ERROR,

    /**  
     * @brief Bulk operation error handling mode where operation ignores the failures and continues to create other objects
     */
    SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR,                                                                                                                                                                                                    
} sai_bulk_op_error_mode_t;`

experimental/saiexperimentalocs.h
wni0812: Understood the point, but in current SAI, "precision" tag/meta data is not available.

experimental/saiexperimentalocs.h
wni0812: 32 may not be long enough and to have more flexibility,  better to use sai_u8_list_t as data type of port name.

experimental/saiexperimentalocs.h
wni0812: Updated the description of this in header file.

jjin62 (approved):
All my comments are addressed.

inc/saiswitch.h
```diff
@@ -3373,6 +3373,17 @@ typedef enum _sai_switch_attr_t
...
+    /**
+     * @brief Set Switch OCS port state change event notification callback function passed to the adapter.
+     *
+     * Use sai_ocs_port_state_change_notification_fn as notification function.
+     *
+     * @type sai_pointer_t sai_ocs_port_state_change_notification_fn
+     * @flags CREATE_AND_SET
+     * @default NULL
+     */
+    SAI_SWITCH_ATTR_OCS_PORT_STATE_CHANGE_NOTIFY,
```
jjin62: to minimize the change to the packet switch code, you can move this to /experimental/saiswitchextensions.h

jjin62 (approved):

mdemerchant (approved):
All my comments have been addressed.

experimental/saiexperimentalocs.h
richardjyu, line 213: What status would we use to indicate that a port has not been intialized to any position, but also does not have an enforced blocking mechanism applied? If a port is in the "blocked" state, it would imply that any light entering the port is isolated from the entire dataplane network within the OCS. However, if a device implements this as an active state, it may not be guaranteed during initialization or if power is lost.

doc/OCS/SAI-Proposal-OCS.md
```diff
@@ -0,0 +1,307 @@
...
+if (_port_key_to_saiobjid_map.end() != it) 
+{
+    sai_attribute_t attr;
+    sai_status_t status;
+    attr.id = SAI_OCS_PORT_ATTR_OPER_STATUS;
+
+    status = sai_ocs_api->get_ocs_port_attribute(it->second, 1, &attr);
+}
+```
+## 8.4. Usage Example: Get All OCS Port Operational Status
```
richardjyu, line 207: Has there been any consideration for a push mechanism for port status changes? Since these changes can impact traffic, it may be useful to have a push mechanism instead of relying on polling.

experimental/saiexperimentalocs.h
wni0812: Port should be blocked from physical layer as default. It will be updated with further operations from the upper layer.

doc/OCS/SAI-Proposal-OCS.md
wni0812: SAI already defines the OCS port status notification: sai_ocs_port_status_notification_t and upper layer can use both ways to manage the port status.

lguohan (approved):

inc/saiswitch.h
JaiOCP: Agreed, This need to be moved to switchextension just like other experimental features

andriy-kokhan:
SAI has SAI_SWITCH_ATTR_PORT_LIST  that allows a control plain to retrieve the list of ASIC ports. Should we introduce something similar for SAI_OBJECT_TYPE_OCS_PORT ?

experimental/saiexperimentalocs.h
```diff
@@ -0,0 +1,466 @@
...
+{
+    /**
+     * @brief Start of device port attributes
+     */
+    SAI_OCS_PORT_ATTR_START,
+
+    /**
+     * @brief Unique identifier for a port
+     *
+     * @type sai_u8_list_t
```
andriy-kokhan, line 244: SAI port object uses HW lane list as a key. Shouldn't we use something similar here? Also, with sai_u8_list_t  its not clear whether to should be null terminated or not.

kcudnik: in lists you always get lenght/count, so it should not be numlterminated

inc/saiswitch.h
wni0812: Yes, the notification definition has been moved to saiswitchextensions.h

experimental/saiexperimentalocs.h
```diff
@@ -223,21 +214,29 @@ typedef enum _sai_ocs_port_attr_t
...
+ *
+ * @flags ranges
+ */
+typedef enum _sai_ocs_port_stat_t
+{
+    /** OCS port stat range start */
+    SAI_OCS_PORT_STAT_START,
+
+    /** OCS port stat unknown, it is used to make build pass, more stats can be added later. */
+    SAI_OCS_PORT_STAT_UNKNOWN,
```
kcudnik, line 292: = SAI_OCS_PORT_STAT_START

doc/OCS/SAI-Proposal-OCS.md
```diff
@@ -0,0 +1,310 @@
...
+    sai_bulk_object_remove_fn                                  remove_ocs_ports;
+    sai_bulk_object_set_attribute_fn                           set_ocs_ports_attribute;
+    sai_bulk_object_get_attribute_fn                           get_ocs_ports_attribute;
+    sai_get_ocs_port_stats_fn                                  get_ocs_port_stats;
+    sai_get_ocs_port_stats_ext_fn                              get_ocs_port_stats_ext;
+    sai_clear_ocs_port_stats_fn                                clear_ocs_port_stats;
+    sai_create_ocs_cross_connect_fn                            create_ocs_cross_connect;
+    sai_remove_ocs_cross_connect_fn                            remove_ocs_cross_connect;
+    sai_set_ocs_cross_connect_attribute_fn                     set_ocs_cross_connect_attribute;
+    sai_get_ocs_cross_connect_attribute_fn                     get_ocs_cross_connect_attribute;
```
r12f, line 109: will be good to add stats functions for cross connects too.

but feel free to leave the stats list empty as well.

wni0812: Updated OCS SAI definitions and markdown to add stats functions for cross connects.

r12f (approved):

JaiOCP (approved):

tjchadaga (approved):

PR #2264, opened 2026-03-20
[meta] Force enums with strict flags to have zero enum defined

kcudnik:
This could be handy for serialization purposes if no flags are defined in serialized enum. But actual purpose of having zero value could have no meaning (for exaple stats mode)

### DIFF
--- a/experimental/saitypesextensions.h
+++ b/experimental/saitypesextensions.h
@@ -141,6 +141,8 @@ typedef enum _sai_dash_tunnel_dscp_mode_t
  */
 typedef enum _sai_dash_routing_actions_t
 {
+    SAI_DASH_ROUTING_ACTIONS_NONE = 0,
+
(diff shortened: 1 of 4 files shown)


tjchadaga (approved):

saksarav-nokia:
@kcudnik  , can we cherry-pick this to msft-2601?

kcudnik:
yes, can you make PR ?

saksarav-nokia:
> yes, can you make PR ?

@kcudnik , since this is SAI submodule PR, do you know which SAI submodule version will be used by msft-202601 sonic-sairedis?. I see this commit in both master and V1.18.1?. If it uses different SAI version, then we need to cherry-pick to that version,

kcudnik:
no clue sorry, ask @tjchadaga or @rlhui

PR #2265, opened 2026-03-30
SAI API Performance Monitoring

JaiOCP:
This PR brings in the performance monitoring of the SAI API. This proposal was presented in OCP 2023.


deepak-singhal0408 (commented):
Review comments on `sai_perfdata_t` struct — precision, capacity, and completeness for real-world SAI performance monitoring.

inc/saitypes.h
```diff
@@ -1568,6 +1569,18 @@ typedef struct _sai_port_err_status_list_t
...
+typedef struct _sai_perfdata_t
+{
+    /** Time when metrics is collected */
+    sai_timespec_t timespec;
+
+    /** Latency measure in milliseconds */
+    sai_uint32_t latency;
+
+    /** Number of objects handled in this measure */
+    sai_uint16_t num_objects;
```
deepak-singhal0408, line 1591: **`sai_uint16_t` overflows at 65535 — too small for bulk operations.**

Bulk route operations commonly handle >65K objects per batch. For example, DPU configurations use batch sizes of 65536, which hits exactly the uint16 boundary and wraps to 0.

**Suggestion:** Use `sai_uint32_t num_objects` to safely handle bulk operations at any scale.

```c
/** Number of objects handled in this measure */
sai_uint32_t num_objects;
```

inc/saitypes.h
```diff
@@ -1568,6 +1569,18 @@ typedef struct _sai_port_err_status_list_t
...
+} sai_perfdata_t;
```
deepak-singhal0408, line 1592: **Missing cumulative counters for true per-route average latency.**

With the current struct, `AVERAGE_LATENCY` returns a per-API-call average. But bulk SAI operations have variable batch sizes — e.g., programming 5000 routes at batch size 1024 produces calls of 1024, 1024, 1024, 1024, and 904 objects. Averaging across these calls weights small batches equally with large ones, giving a **per-batch** average rather than a **per-route** average — which is what operators actually need for performance tuning.

The struct returns `num_objects` for `INST_LATENCY` (last call only), but doesn't expose the cumulative counters needed to compute meaningful averages.

**Suggestion:** Add cumulative fields so consumers can compute accurate per-route averages:

```c
typedef struct _sai_perfdata_t
{
    sai_timespec_t timespec;
    sai_uint64_t latency_us;       /* microseconds */
    sai_uint32_t num_objects;      /* objects in this measure */
    sai_uint64_t cum_latency_us;   /* cumulative latency since last reset */
    sai_uint32_t total_objects;    /* cumulative objects since last reset */
} sai_perfdata_t;
```

With `cum_latency_us` and `total_objects`, the consumer computes:
`per_route_avg = cum_latency_us / total_objects`

This is meaningful regardless of batch size variation and enables accurate performance baselining across different hardware.

inc/saitypes.h
```diff
@@ -1568,6 +1569,18 @@ typedef struct _sai_port_err_status_list_t
     sai_port_err_status_t *list;
 } sai_port_err_status_list_t;
 
+typedef struct _sai_perfdata_t
+{
+    /** Time when metrics is collected */
+    sai_timespec_t timespec;
+
+    /** Latency measure in milliseconds */
+    sai_uint32_t latency;
```
deepak-singhal0408, line 1588: Millisecond precision may be too coarse for performance monitoring. Modern ASICs can program thousands of routes in <1ms — with ms granularity, fast batches would report `latency = 0`.

**Suggestion:** Consider using microseconds for meaningful precision:

```c
/** Latency measure in microseconds */
sai_uint32_t latency_us;
```

inc/saitypes.h
JaiOCP: Thanks for explaining the requirement. I will capture the details in the md file as well.
Change the num objects to uint32 and added cumlatency and total object fields in the structure.

inc/saitypes.h
JaiOCP: change it to usec.

inc/saitypes.h
JaiOCP: changed it to uint32

deepak-singhal0408:
thanks @JaiOCP .. Could you push the updated PR. and resolve the conflicts?

JaiOCP:
> thanks @JaiOCP .. Could you push the updated PR. and resolve the conflicts?

I have to open a new PR. PR branch is diverged too much from the origin

PR #2272, opened 2026-04-03
saiport: Add counters for 1519_TO_MAX_OCTETS packet size range

AnandhiDhanabalan:
Add a new port statistic for ASICs that expose a single aggregate jumbo packet counter instead of the detailed SAI jumbo histogram buckets (1519–2047, 2048–4095, 4096–9216, 9217–16383).

The following counters are introduced:
- SAI_PORT_STAT_ETHER_STATS_PKTS_1519_TO_MAX_OCTETS
- SAI_PORT_STAT_ETHER_IN_PKTS_1519_TO_MAX_OCTETS
- SAI_PORT_STAT_ETHER_OUT_PKTS_1519_TO_MAX_OCTETS

The new counter reports packets with frame length greater than 1518 octets up to the maximum frame size supported by the port hardware (MAX_OCTETS). This enables platforms that provide only a cumulative jumbo packet counter to expose the statistic through SAI.

If the underlying ASIC provides packet length histogram counters, vendors may derive and expose this cumulative counter by aggregating the corresponding histogram buckets.

From a user perspective, a cumulative jumbo packet counter provides a simple view of jumbo traffic without requiring consumers to interpret or sum multiple histogram counters.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -4042,13 +4042,13 @@ typedef enum _sai_port_stat_t
     /** SAI port stat credits freed update messages rx */
     SAI_PORT_STAT_CBFC_NUM_CF_UPDATE_MESSAGES_RX,
 
-    /** SAI port stat ether stats packets with frame size 1519 octets to max MTU supported by the port */
+    /** Number of packets with frame size 1519 octets to max MTU supported by the port */
     SAI_PORT_STAT_ETHER_STATS_PKTS_1519_TO_MAX_OCTETS,
 
-    /** SAI port stat ether RX packets with frame size 1519 octets to max MTU supported by the port */
+    /** Number of received packets with frame size 1519 octets to max MTU supported by the port */
     SAI_PORT_STAT_ETHER_IN_PKTS_1519_TO_MAX_OCTETS,
 
-    /** SAI port stat ether TX packets with frame size 1519 octets to max MTU supported by the port */
+    /** Number of transmitted packets with frame size 1519 octets to max MTU supported by the port */
     SAI_PORT_STAT_ETHER_OUT_PKTS_1519_TO_MAX_OCTETS,
 
     /** Port stat in drop reasons range start */
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -4042,6 +4042,15 @@ typedef enum _sai_port_stat_t
     /** SAI port stat credits freed update messages rx */
     SAI_PORT_STAT_CBFC_NUM_CF_UPDATE_MESSAGES_RX,
 
+    /** SAI port stat ether stats packets with frame size 1519 octets to max MTU supported by the port */
+    SAI_PORT_STAT_ETHER_STATS_PKTS_1519_TO_MAX_OCTETS,
+
+    /** SAI port stat ether RX packets with frame size 1519 octets to max MTU supported by the port */
+    SAI_PORT_STAT_ETHER_IN_PKTS_1519_TO_MAX_OCTETS,
+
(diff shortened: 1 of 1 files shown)


inc/saiport.h
```diff
@@ -3835,6 +4032,23 @@ typedef enum _sai_port_stat_t
...
+    SAI_PORT_STAT_CBFC_NUM_CC_UPDATE_MESSAGES_TX,
+
+    /** SAI port stat credits freed update messages tx */
+    SAI_PORT_STAT_CBFC_NUM_CF_UPDATE_MESSAGES_TX,
+
+    /** SAI port stat credits consumed update messages rx */
+    SAI_PORT_STAT_CBFC_NUM_CC_UPDATE_MESSAGES_RX,
+
+    /** SAI port stat credits freed update messages rx */
+    SAI_PORT_STAT_CBFC_NUM_CF_UPDATE_MESSAGES_RX,
```
kcudnik, line 4043: i think that "SAI port stat" is redundant, could be "Number of ..."

AnandhiDhanabalan: @kcudnik , The naming convention for 'SAI_PORT_STAT_ETHER_STATS' has been followed to be inline with the existing ETHER_STATS. 
Do let me know if you still think this needs to be changed

AnandhiDhanabalan: @kcudnik , is you comment for the ETHER_STATS or for CBFC ?

kcudnik: my comment is for comments in each added new enum, that string seems redundant since each enum starts with SAI_PORT_STAT_xxx

AnandhiDhanabalan: @kcudnik , they are 3 different Ether stats as shown below and are inline with existing ether stats
1. **Tx+Rx** - SAI_PORT_STAT_ETHER_STATS_PKTS_1519_TO_MAX_OCTETS
2. **Rx only**  - SAI_PORT_STAT_ETHER_IN_PKTS_1519_TO_MAX_OCTETS
3. **Tx Only** - SAI_PORT_STAT_ETHER_OUT_PKTS_1519_TO_MAX_OCTETS

kcudnik: i think you are not getting it, i say that there is no need to repeat string "SAI port stat" in comment lines like 4050, 4047 4044 and so on

AnandhiDhanabalan: @kcudnik , I intended to follow the same format as the existing Ethernet statistics. Quoting a couple of examples below for reference:
```c
    /** SAI port stat ether stats pkts 65 to 127 octets */
    SAI_PORT_STAT_ETHER_STATS_PKTS_65_TO_127_OCTETS,

    /** SAI port stat ether stats pkts 128 to 255 octets */
    SAI_PORT_STAT_ETHER_STATS_PKTS_128_TO_255_OCTETS,
```

If there is a need to adjust the format for the new enums, would the following be acceptable?
```c
    /** Packets with frame size 1519 octets to max MTU supported by the port */
    SAI_PORT_STAT_ETHER_STATS_PKTS_1519_TO_MAX_OCTETS,
```
Please let me know your thoughts, and I can update it accordingly.

kcudnik: check for example SAI_PORT_STAT_LLR_RX_ACK_CTL_OS

AnandhiDhanabalan: @kcudnik , Changed the comments based on your review. Please check and approve.

kcudnik (approved):

AnandhiDhanabalan:
@tjchadaga Can you please help merge this PR?

tjchadaga (approved):

PR #2274, opened 2026-04-10
  Add ACL Mirror Copy Type and Mirror Session ID match fields

mobinmohan:
This PR introduces two new ACL match fields to the SAI ACL headers:
  
  1. `SAI_ACL_TABLE_ATTR_FIELD_ACL_MIRROR_COPY_TYPE`: Allows ACL rules to match packets based on whether they are mirror copies and their specific type (Ingress, Egress, or Ingress Or Egress).
  
  2. `SAI_ACL_TABLE_ATTR_FIELD_ACL_MIRROR_SESSION_ID`: Allows ACL rules to match based on a specific mirror session ID.


JaiOCP:
> _ No description provided. _

Pleas provide description of the PR

JaiOCP:
@mobinmohan Is it an ingress mirror or egress mirror copy match condition.

mobinmohan:
> @mobinmohan Is it an ingress mirror or egress mirror copy match condition.

The attribute was to match both ingress and egress mirror copies. However, I can change the type from `bool` to `sai_acl_mirror_copy_type_t` to allow selective matching, if the underlying hardware supports it. Please let me know your recommendation.

```
+/**
+ * @brief Mirror copy type for ACL matching
+ */
+typedef enum _sai_acl_mirror_copy_type_t
+{
+    SAI_ACL_MIRROR_COPY_TYPE_NONE,
+    SAI_ACL_MIRROR_COPY_TYPE_INGRESS,
+    SAI_ACL_MIRROR_COPY_TYPE_EGRESS,
+    SAI_ACL_MIRROR_COPY_TYPE_ANY
+} sai_acl_mirror_copy_type_t;
+
+
```

JaiOCP:
> > @mobinmohan Is it an ingress mirror or egress mirror copy match condition.
> 
> The attribute was to match both ingress and egress mirror copies. However, I can change the type from `bool` to `sai_acl_mirror_copy_type_t` to allow selective matching, if the underlying hardware supports it. Please let me know your recommendation.
> 
> ```
> +/**
> + * @brief Mirror copy type for ACL matching
> + */
> +typedef enum _sai_acl_mirror_copy_type_t
> +{
> +    SAI_ACL_MIRROR_COPY_TYPE_NONE,
> +    SAI_ACL_MIRROR_COPY_TYPE_INGRESS,
> +    SAI_ACL_MIRROR_COPY_TYPE_EGRESS,
> +    SAI_ACL_MIRROR_COPY_TYPE_ANY
> +} sai_acl_mirror_copy_type_t;
> +
> +
> ```

I don't think we need to have type_t.
If the ACL table stage is ingress then the match action would apply to ingress mirror.
If the ACL table stage is egress then the match action would apply to egress mirror.

Typically there is no case where ingress mirror copy needs an egress ACL rule for match action as the mirror copy is taken out to the mirror port directly but I am not very sure about it. Is that a use case for you?

inc/saiacl.h
```diff
@@ -2825,10 +2834,19 @@ typedef enum _sai_acl_entry_attr_t
...
+     * @flags CREATE_AND_SET
+     * @default disabled
+     */
+    SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY = SAI_ACL_ENTRY_ATTR_FIELD_START + 0x167,
+
     /**
      * @brief End of Rule Match Fields
      */
-    SAI_ACL_ENTRY_ATTR_FIELD_END = SAI_ACL_ENTRY_ATTR_FIELD_CSIG_D_BIT,
+    SAI_ACL_ENTRY_ATTR_FIELD_END = SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY,
```
kcudnik, line 2849: i'm wondering whether we should make those as last item, not equal to last valid item, all _END attributes in all SAI are defined after last valid enum item, except those 4:
```c
saiacl.h:    SAI_ACL_TABLE_ATTR_FIELD_END = SAI_ACL_TABLE_ATTR_FIELD_CSIG_D_BIT,
saiacl.h:    SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_END = SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_DST_IPV6,
saiacl.h:    SAI_ACL_ENTRY_ATTR_FIELD_END = SAI_ACL_ENTRY_ATTR_FIELD_CSIG_D_BIT,
saiacl.h:    SAI_ACL_ENTRY_ATTR_ACTION_END = SAI_ACL_ENTRY_ATTR_ACTION_TAM_OBJECT,
```

which requires alywas to set change those after adding new values

we could redefine those 4 and not have to touch them every time, then we could have consistency across all SAI

proble could be for 
```c

1753     SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_END = SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_DST_IPV6,
1754 
1755     /**
1756      * @brief End of ACL Table attributes
1757      */
1758     SAI_ACL_TABLE_ATTR_END,
1759 

3480     SAI_ACL_ENTRY_ATTR_ACTION_END = SAI_ACL_ENTRY_ATTR_ACTION_TAM_OBJECT,
3481 
3482     /**
3483      * @brief End of ACL Entry attributes
3484      */
3485     SAI_ACL_ENTRY_ATTR_END,
3486 
```

since if we made it like this:
```c
SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_END,
SAI_ACL_TABLE_ATTR_END


SAI_ACL_ENTRY_ATTR_ACTION_END,
SAI_ACL_ENTRY_ATTR_END
```

then SAI_ACL_TABLE_ATTR_END and SAI_ACL_ENTRY_ATTR_END will not point to last valid attribute but rather SAI_ACL_TABLE_ATTR_FIELD_VALID_BITS_END and SAI_ACL_ENTRY_ATTR_ACTION_END which are just markers

JaiOCP: We can take this up as a separate PR. 
Quick question: Except that we have to change  _END attribute whenever we add new attribute, functionally there is no impact. Is that understanding correct?

kcudnik: there should be no impact, some of the vendors may use _END attribute to check if range is valid, so in those 4 cases could mean to change <= to <

mobinmohan:
> > _ No description provided. _
> 
> Pleas provide description of the PR

Added description.

mobinmohan:
> > > @mobinmohan Is it an ingress mirror or egress mirror copy match condition.
> > 
> > 
> > The attribute was to match both ingress and egress mirror copies. However, I can change the type from `bool` to `sai_acl_mirror_copy_type_t` to allow selective matching, if the underlying hardware supports it. Please let me know your recommendation.
> > ```
> > +/**
> > + * @brief Mirror copy type for ACL matching
> > + */
> > +typedef enum _sai_acl_mirror_copy_type_t
> > +{
> > +    SAI_ACL_MIRROR_COPY_TYPE_NONE,
> > +    SAI_ACL_MIRROR_COPY_TYPE_INGRESS,
> > +    SAI_ACL_MIRROR_COPY_TYPE_EGRESS,
> > +    SAI_ACL_MIRROR_COPY_TYPE_ANY
> > +} sai_acl_mirror_copy_type_t;
> > +
> > +
> > ```
> 
> I don't think we need to have type_t. If the ACL table stage is ingress then the match action would apply to ingress mirror. If the ACL table stage is egress then the match action would apply to egress mirror.
> 
> Typically there is no case where ingress mirror copy needs an egress ACL rule for match action as the mirror copy is taken out to the mirror port directly but I am not very sure about it. Is that a use case for you?

Sounds good. We can continue with the default(bool).

Gnanapriya27:
> _sai_acl_mirror_copy_type_t

The current bool definition still leaves ambiguity on what exactly is being matched.
Interpreting this based on the stage as per ACL table,
Ingress ACL stage → match ingress-generated mirror copies only
Egress ACL stage → match egress-generated mirror copies only

then SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY=true in an egress ACL table, does it mean only egress mirror copies, not any mirrored packet ? please clarify.

Modeling this stage/direction explicitly using sai_acl_mirror_copy_type_t (as proposed in this conversation) would make the semantics unambiguous and future-proof the design, enum capability query allows the vendor to announce their capabilities.

inc/saiacl.h
```diff
@@ -1647,10 +1647,19 @@ typedef enum _sai_acl_table_attr_t
...
     SAI_ACL_TABLE_ATTR_FIELD_CSIG_D_BIT = SAI_ACL_TABLE_ATTR_FIELD_START + 0x166,
 
+    /**
+     * @brief Match on a mirror copy packet
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_MIRROR_COPY = SAI_ACL_TABLE_ATTR_FIELD_START + 0x167,
```
Gnanapriya27: Can you clarify whether SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY matches mirror copies generated by any mirror session (i.e. a generic “mirrored packet” qualifier), or whether there is intent to support classification of mirror copies generated by a specific mirror session?

A potential use case is applying different ACL treatment to traffic mirrored by different mirror sessions (for example, Mirror session A vs session B). With the current boolean qualifier, it is not clear how per-session discrimination would be achieved.

JaiOCP:
> > _sai_acl_mirror_copy_type_t
> 
> The current bool definition still leaves ambiguity on what exactly is being matched. Interpreting this based on the stage as per ACL table, Ingress ACL stage → match ingress-generated mirror copies only Egress ACL stage → match egress-generated mirror copies only
> 
> then SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY=true in an egress ACL table, does it mean only egress mirror copies, not any mirrored packet ? please clarify.
> 
> Modeling this stage/direction explicitly using sai_acl_mirror_copy_type_t (as proposed in this conversation) would make the semantics unambiguous and future-proof the design, enum capability query allows the vendor to announce their capabilities.

I agree and take my earlier statement back.
Just providing the stage is not enough. Lets use the type_t

mobinmohan:
> > _sai_acl_mirror_copy_type_t
> 
> The current bool definition still leaves ambiguity on what exactly is being matched. Interpreting this based on the stage as per ACL table, Ingress ACL stage → match ingress-generated mirror copies only Egress ACL stage → match egress-generated mirror copies only
> 
> then SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY=true in an egress ACL table, does it mean only egress mirror copies, not any mirrored packet ? please clarify.
> 
> Modeling this stage/direction explicitly using sai_acl_mirror_copy_type_t (as proposed in this conversation) would make the semantics unambiguous and future-proof the design, enum capability query allows the vendor to announce their capabilities.

I received a similar feedback during the weekly review meeting. I agree using sai_acl_mirror_copy_type_t makes the design future proof. I have updated the PR to use the enum type, though proper distinction will depend on the ASIC support.

mobinmohan:
> > > _sai_acl_mirror_copy_type_t
> > 
> > 
> > The current bool definition still leaves ambiguity on what exactly is being matched. Interpreting this based on the stage as per ACL table, Ingress ACL stage → match ingress-generated mirror copies only Egress ACL stage → match egress-generated mirror copies only
> > then SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY=true in an egress ACL table, does it mean only egress mirror copies, not any mirrored packet ? please clarify.
> > Modeling this stage/direction explicitly using sai_acl_mirror_copy_type_t (as proposed in this conversation) would make the semantics unambiguous and future-proof the design, enum capability query allows the vendor to announce their capabilities.
> 
> I agree and take my earlier statement back. Just providing the stage is not enough. Lets use the type_t

Thanks for reviewing this during the weekly meeting. I have update the PR now.

JaiOCP (approved):

mikeberesford (approved):

inc/saiacl.h
```diff
@@ -1647,10 +1665,19 @@ typedef enum _sai_acl_table_attr_t
...
     SAI_ACL_TABLE_ATTR_FIELD_CSIG_D_BIT = SAI_ACL_TABLE_ATTR_FIELD_START + 0x166,
 
+    /**
+     * @brief Match on mirror copy type
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_ACL_TABLE_ATTR_FIELD_ACL_MIRROR_COPY_TYPE = SAI_ACL_TABLE_ATTR_FIELD_START + 0x167,
```
Gnanapriya27, line 1675: Below comment was outdated, as the patch was modified after it was made. Re-adding it here for the author’s review and response.

Can you clarify whether SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY matches mirror copies generated by any mirror session (i.e. a generic “mirrored packet” qualifier), or whether there is intent to support classification of mirror copies generated by a specific mirror session?

A potential use case is applying different ACL treatment to traffic mirrored by different mirror sessions (for example, Mirror session A vs session B). With the current boolean qualifier, it is not clear how per-session discrimination would be achieved.

mobinmohan: > The current bool definition still leaves ambiguity on what exactly is being matched. Interpreting this based on the stage as per ACL table, Ingress ACL stage → match ingress-generated mirror copies only Egress ACL stage → match egress-generated mirror copies only

@Gnanapriya27 In the latest commit SAI_ACL_ENTRY_ATTR_FIELD_ACL_MIRROR_COPY_TYPE was changed from bool to sai_acl_mirror_copy_type_t type. This definition resolves the ambiguity regarding ingress vs egress mirror copies.

> Can you clarify whether SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY matches mirror copies generated by any mirror session (i.e. a generic “mirrored packet” qualifier), or whether there is intent to support classification of mirror copies generated by a specific mirror session?

This is an interesting use case, we can consider introducing a new match field ACL_MIRROR_SESSION_ID. I would recommend tracking this as a separate PR.

Gnanapriya27: As you propose, introducing a match field such as ACL_MIRROR_SESSION_ID would make this proposal more complete and practically usable.

Consider a scenario where multiple mirror sessions (e.g., local/SPAN and ERSPAN) are active and generating mirrored traffic. In such cases, it may be necessary to selectively classify or apply policy only to mirror copies generated by a specific session (for example, an ERSPAN session used for telemetry or security analysis).

With the current generic qualifier (SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY), classification is limited to identifying whether a packet is ingress/egress mirrored or not, without the ability to distinguish which mirror session originated the copy

inc/saiacl.h
```diff
@@ -2825,10 +2861,29 @@ typedef enum _sai_acl_entry_attr_t
...
     SAI_ACL_ENTRY_ATTR_FIELD_CSIG_D_BIT = SAI_ACL_ENTRY_ATTR_FIELD_START + 0x166,
 
+    /**
+     * @brief Match on mirror copy type
+     *
+     * @type sai_acl_field_data_t sai_acl_mirror_copy_type_t
+     * @flags CREATE_AND_SET
+     * @default disabled
+     */
+    SAI_ACL_ENTRY_ATTR_FIELD_ACL_MIRROR_COPY_TYPE = SAI_ACL_ENTRY_ATTR_FIELD_START + 0x167,
```
Gnanapriya27, line 2871: Semantics related:
Remove ACL from the new attr to be inlined with existing attr SAI_ACL_ENTRY_ATTR_FIELD_ACL_MIRROR_COPY_TYPE  to
SAI_ACL_ENTRY_ATTR_FIELD_MIRROR_COPY_TYPE

Same for other attr as well

inc/saiacl.h
```diff
@@ -587,6 +587,24 @@ typedef enum _sai_acl_table_group_member_attr_t
 
 } sai_acl_table_group_member_attr_t;
 
+/**
+ * @brief Mirror copy type for ACL matching
+ */
+typedef enum _sai_acl_mirror_copy_type_t
+{
+    /** Match packets that are not mirror copies */
+    SAI_ACL_MIRROR_COPY_TYPE_NONE,
```
Gnanapriya27, line 596: what does NONE signifies ?
MIRROR_COPY is a match criteria that is specifically meant for matching on the mirrored copy of the packet alone and not the original packet.

`NONE' description indicated not mirror copies
NONE enum description and the primary intent of this qualifier still raises ambiguity

inc/saiacl.h
```diff
@@ -2825,10 +2861,29 @@ typedef enum _sai_acl_entry_attr_t
...
+     * @brief Match on mirror copy type
+     *
+     * @type sai_acl_field_data_t sai_acl_mirror_copy_type_t
+     * @flags CREATE_AND_SET
+     * @default disabled
+     */
+    SAI_ACL_ENTRY_ATTR_FIELD_ACL_MIRROR_COPY_TYPE = SAI_ACL_ENTRY_ATTR_FIELD_START + 0x167,
+
+    /**
+     * @brief Match on mirror session ID
```
Gnanapriya27: it would be better to brief it like Match on mirror copies generated by this mirror session ID

inc/saiacl.h
mobinmohan: The SAI metadata checker enforces a convention for attributes with custom enum type. It requires the attribute name suffix match the enum type name (excluding the sai_ prefix and _t suffix). Here the type name is sai_acl_mirror_copy_type_t. Removing acl keyword from the type_t would not be preferred here.

The naming is also similar to SAI_ACL_ENTRY_ATTR_FIELD_ACL_IP_TYPE and SAI_ACL_ENTRY_ATTR_FIELD_ACL_IP_FRAG.

Gnanapriya27: Thanks for the details, Agree.

inc/saiacl.h
mobinmohan: > what does NONE signifies ?

Match on original packet is specified using NONE.

> MIRROR_COPY is a match criteria that is specifically meant for matching on the mirrored copy of the packet alone and not the original packet.

I do not think that aligns with the intended behavior. If a user shares the same ACL group for multiple features, they need control to differentiate between original vs mirror traffic. If a feature should not apply to mirror copies, they need a way to specify that. NONE will help to specify the rule should not match on mirror copies.

inc/saiacl.h
mobinmohan: Sure. I have updated it.

Gnanapriya27 (approved):
Looks good to me

tjchadaga (approved):

PR #2276, opened 2026-04-19
Added new attribute SAI_MIRROR_SESSION_ATTR_GUARANTEE_RATE

milosmoskovljevic:
Introduce a new mirror session attribute SAI_MIRROR_SESSION_ATTR_GUARANTEE_RATE to support guaranteed sampling for mirror sessions.
The attribute defines a sample rate in packets per second (pps) below which all traffic is sampled. Above that rate, sampling falls back to the existing statistical sampling controlled by SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE.

### DIFF
--- a/inc/saimirror.h
+++ b/inc/saimirror.h
@@ -404,6 +404,22 @@ typedef enum _sai_mirror_session_attr_t
     SAI_MIRROR_SESSION_ATTR_LABEL,
 
     /**
+     * @brief Guarantee sample rate in packets per second (pps)
+     *
+     * While guarantee rate is not breached, all packets will be sampled (100% sampling).
+     * Once breached, packets will be sampled based on the mirror statistics (fallback to
+     * statistical sampling based on SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE).
+     *
+     * This is expected to enable guaranteed sampling for very low pps flows.
+     * A value of 0 disables the guarantee rate feature.
+     *
+     * @type sai_uint64_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_MIRROR_SESSION_ATTR_GUARANTEE_RATE,
+
+    /**
      * @brief End of attributes
      */
     SAI_MIRROR_SESSION_ATTR_END,
--- a/inc/saimirror.h
+++ b/inc/saimirror.h
@@ -406,12 +406,16 @@ typedef enum _sai_mirror_session_attr_t
     /**
      * @brief Guarantee sample rate in packets per second (pps)
      *
-     * While guarantee rate is not breached, all packets will be sampled (100% sampling).
-     * Once breached, packets will be sampled based on the mirror statistics (fallback to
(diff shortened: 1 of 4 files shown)


inc/saimirror.h
```diff
@@ -403,6 +403,22 @@ typedef enum _sai_mirror_session_attr_t
      */
     SAI_MIRROR_SESSION_ATTR_LABEL,
 
+    /**
+     * @brief Guarantee sample rate in packets per second (pps)
```
Gnanapriya27, line 421: Newly introduced attr - SAI_MIRROR_SESSION_ATTR_GUARANTEE_RATE  is in packet-rate threshold(pps) and existing SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE statistical sampling (1:N statistical ratio)  - Since these attribute operate in different domain - can you define the interaction semantics in this proposal.

milosmoskovljevic: I have added the documentation, it should be much clearer now.

inc/saimirror.h
```diff
@@ -403,6 +403,26 @@ typedef enum _sai_mirror_session_attr_t
...
+     * A value of 0 disables the guarantee rate feature.
+     *
+     * Rate measurement and enforcement (including behavior on rate transitions,
+     * burst handling, and reset) are implementation-defined.
+     *
+     * @type sai_uint64_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_MIRROR_SESSION_ATTR_GUARANTEE_RATE,
```
Gnanapriya27, line 438: This PR adds SAI_MIRROR_SESSION_ATTR_GUARANTEE_RATE to complement the existing SAI_MIRROR_SESSION_ATTR_SAMPLE_RATE on mirror sessions object. Similar 1-in-N statistical sampling concept (SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE) exists independently on SAI_OBJECT_TYPE_SAMPLEPACKET in saisamplepacket.h as well

SAI defines two sampling architectures:
Method A: Mirror session owns the sampling rate 
Method B : Sample session owns 1-in-N sampling decision;  in such cases, the mirror session handles encapsulation only. In method B, there is no way to express a guarantee rate. 

Does this PR scopes to extend the saisamplepacket.h as well ?

milosmoskovljevic: Thanks for raising this, good point.
You are right that SAI supports both paths: sampling can be owned by the mirror session, or by the samplepacket object. The initial scope of this PR was mirror-session-only, which left no way to express guarantee rate on the samplepacket path.

Added SAI_SAMPLEPACKET_ATTR_GUARANTEE_RATE in saisamplepacket.h, paired with SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE, with the same semantics and formula as on the mirror session.
Updated the proposal document with a short note in the overview and an API reference entry for the samplepacket attribute.

milosmoskovljevic:
@JaiOCP @Gnanapriya27
Could you please review this when you have time?

Gnanapriya27 (approved):

PR #2285, opened 2026-05-14
Optimized Forwarding Headers

JaiOCP:
This PR brings a generic framework to support multiple activities going on for defining the Optimized Forwarding Header like IEEE compressed Header, OCP ESUN Header, UEC Unified Forwarding Header and more.


inc/saidebugcounter.h
```diff
@@ -331,6 +331,21 @@ typedef enum _sai_in_drop_reason_t
     /** IPv4 or IPv6 Routing table (LPM) unicast miss */
     SAI_IN_DROP_REASON_LPM_MISS = 0x00000039,
 
+    /** OFH packet TTL expired */
+    SAI_IN_DROP_REASON_OFH_TTL,
```
JaiOCP, line 335: Keep it consistent with IP or not ?

inc/saiswitch.h
```diff
@@ -3598,6 +3598,24 @@ typedef enum _sai_switch_attr_t
...
+     *
+     * @type sai_object_list_t
+     * @flags CREATE_AND_SET
+     * @objects SAI_OBJECT_TYPE_OFH
+     * @default empty
+     */
+    SAI_SWITCH_ATTR_OFH,
+
+    /**
+     * @brief OFH header max size
```
JaiOCP, line 3631: Improve comments to explian the use of max size supported by HW.

kcudnik:
@JaiOCP validator is not handling merges, you will need to rebase your change on master and force push

JaiOCP:
Sounds Good. I will recreate the PR as well. Too many changes have gone in
mainline.

kcudnik:
you dont need to recreate, you can jsut rebase and force push to your branch

inc/saiofhroute.h
```diff
@@ -0,0 +1,336 @@
...
+    /** READ-ONLY */
+
+    /**
+     * @brief OFH Route entry type
+     *
+     * @type sai_ip_addr_family_t
+     * @flags READ_ONLY
+     * @isresourcetype true
+     */
+    SAI_OFH_ROUTE_ENTRY_ATTR_IP_ADDR_FAMILY,
```
rbudhia-upscaleai: Should this be: SAI_OFH_ROUTE_ENTRY_ATTR_**OFH**_ADDR_FAMILY?

inc/saitypes.h
```diff
@@ -471,6 +473,51 @@ typedef struct _sai_ip_prefix_list_t
...
+    sai_mac_t afh_t2;
+
+    /** @validonly ofh_addr_family == SAI_OFH_ADDR_FAMILY_UFH_T1 */
+    sai_uint32_t ufh_t1;
+
+    /** @validonly ofh_addr_family == SAI_OFH_ADDR_FAMILY_UFH_T2 */
+    sai_uint16_t ufh_t2;
+} sai_ofh_addr_t;
+
+typedef struct _sai_ofh_addr_and_mask_t
```
rbudhia-upscaleai, line 503: Suggest renaming this to _sai_ofh_prefix_t to be consistent with IP addressing

kcudnik:
you need to squash and force push

inc/saiofh.h
```diff
@@ -0,0 +1,404 @@
...
+    SAI_OFH_ATTR_TYPE = SAI_OFH_ATTR_START,
+
+    /**
+     * @brief OFH ether type identifier
+     *
+     * @type sai_uint16_t
+     * @flags CREATE_AND_SET
+     * @isvlan false
+     * @default 0
+     */
```
rbudhia-upscaleai, line 140: Ethertype is only valid for certain OFH types. Need to add comment to specify this.

inc/saiofh.h
```diff
@@ -0,0 +1,404 @@
...
+
+    /**
+     * @brief OFH Version Field Bit Offset
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     * @validonly SAI_OFH_ATTR_TYPE == SAI_OFH_TYPE_AFH_TYPE_1 or SAI_OFH_ATTR_TYPE == SAI_OFH_TYPE_AFH_TYPE_2
+     */
+    SAI_OFH_ATTR_VER_OFFSET,
```
rbudhia-upscaleai, line 151: Some headers have their format defined by the version field. While the _ofh_type_ may be the same, the actual header format can differ depending on the version.
For example, the next version of ESUN will likely share the same Ethertype as the current version, but the header format will be determined by the value of the version (RevisionNumber) field.
To handle these cases, add a version enum attribute. This attribute specifies which version of the OFH is being defined.

inc/saiofh.h
```diff
@@ -0,0 +1,404 @@
...
+     * @default 0
+     */
+    SAI_OFH_ATTR_ETHERTYPE,
+
+    /**
+     * @brief OFH Version Field Bit Offset
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
```
rbudhia-upscaleai, line 148: For all Field Offset definitions in this file, update the comments to indicate the offset's starting point - either the start of the frame or the start of the header.

inc/saiofh.h
```diff
@@ -0,0 +1,404 @@
...
+    SAI_OFH_ATTR_CNM_WIDTH,
+
+    /**
+     * @brief OFH header size
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_OFH_ATTR_HDR_SIZE,
```
rbudhia-upscaleai, line 367: Option discussed at the SAI meeting: API validator should be able to verify that the header length specified in this attribute equals the sum of header field widths. This requires defining offsets and widths for all reserved fields as well.

inc/saiofhroute.h
JaiOCP: fixed

inc/saiofh.h
JaiOCP: done

inc/saiofh.h
JaiOCP: As we discussed I will add details on layering in comments.

inc/saiofh.h
JaiOCP: I will add comments to that intent:
Header size includes all the fields in OFH header including reserved fields

inc/saitypes.h
JaiOCP: reason not to use prefix was to not give any implications that ofh address space need to be ip prefix like. It may or may not be. Given deployment may program it as linear index lookup like LID in IB.

inc/saiofh.h
JaiOCP: Added

inc/saidebugcounter.h
BYGX-wcr: If we assume that the switch will only accept a single type of Header for now, I think we should use the original TTL reason for OFH TTL=0.

BYGX-wcr:
Hi @JaiOCP,

I have a few questions regarding the abstract forwarding behavior of OFH below:
1. How do we specify if the switch should do L2 forwarding (exact match) or L3 forwarding (LPM) on the OFH packet? Looks like we assume that the switch uses a single OFH route table for every packet in the current model?
2. If we want to define the switch to do L3 forwarding on OFH packets, is the SAI Router Interface Object still relevant here? If we still use SAI Router Interface Object, the router mac of the RIF object will be different than the MAC DA of a transit OFH packet. In this scenario, is the switch still able to do routing on the packet?

inc/saiofh.h
```diff
@@ -0,0 +1,472 @@
...
+    /**
+     * @brief OFH Version number
+     *
+     * @type sai_ofh_ver_t
+     * @flags CREATE_AND_SET
+     * @default SAI_OFH_VER_1
+     * @validonly SAI_OFH_ATTR_TYPE == SAI_OFH_TYPE_AFH or SAI_OFH_ATTR_TYPE == SAI_OFH_TYPE_ESUN
+     */
+    SAI_OFH_ATTR_VER,
+
```
rbudhia-upscaleai, line 131: As discussed in the meeting, the version enum needs to be decoupled from the value in the version field. Version enum indicates to the SAI adapter which fields are present or absent for a given header version. The version number, however, represents the actual value  stored in the header's version field for that enum.
Example: if specifying IP, IPv4 and IPv6 could use version enum values 0 and 1, respectively, while the corresponding values in the header’s version field would be 4 and 6..

inc/saiofhroute.h
```diff
@@ -0,0 +1,336 @@
...
+     * @brief Next hop or next hop group id for the packet, or a router interface
+     * in case of directly reachable route, or the CPU port in case of IP2ME route
+     *
+     * The next hop id is only effective when the packet action is one of the
+     * following: FORWARD, COPY, LOG, TRANSIT.
+     *
+     * The next hop id can be a generic next hop object, such as next hop, next
+     * hop group. Directly reachable routes are the IP subnets that are
+     * directly attached to the router. For such routes, fill the router
+     * interface id to which the subnet is attached. IP2ME route adds a local
```
BYGX-wcr, line 80: We still want the OFH route to support IP2ME route semantics because we are thinking of running routing protocols such as BGP across the SUE switches. Without the support of IP2ME route, the switches can't exchange OFH packets carrying routing protocol payload unless there is another handy alternative.

inc/sairouterinterface.h
```diff
@@ -436,6 +436,15 @@ typedef enum _sai_router_interface_attr_t
      */
     SAI_ROUTER_INTERFACE_ATTR_QOS_TC_AND_COLOR_TO_DOT1P_MAP,
 
+    /**
+     * @brief Admin OFH state
+     *
```
rbudhia-upscaleai, line 441: Add a comment indicating that the MYMAC check is not applicable for OFH traffic.

inc/saiofhroute.h
```diff
@@ -0,0 +1,336 @@
...
+     * @objects SAI_OBJECT_TYPE_HOSTIF_USER_DEFINED_TRAP
+     * @allownull true
+     * @default SAI_NULL_OBJECT_ID
+     */
+    SAI_OFH_ROUTE_ENTRY_ATTR_USER_TRAP_ID,
+
+    /**
+     * @brief Next hop or next hop group id for the packet, or a router interface
+     * in case of directly reachable route, or the CPU port in case of IP2ME route
+     *
```
rbudhia-upscaleai, line 73: Update text for OFH. IP2ME still applies, but in the context of OFH.

inc/sairouterinterface.h
JaiOCP: added the comment

inc/saiofhroute.h
JaiOCP: updated

inc/saiofhroute.h
JaiOCP: updated the text

JaiOCP:
Base branch has diverged too much with lot of baseline conflict resolution.
Opening a new PR and closing this one.
https://github.com/opencomputeproject/SAI/pull/2297

PR #2084, opened 2024-10-01
EVPN Multi Home Support

JaiOCP:
Host may be reachable via multiple VTEPs in the EVPN topology. In this case load balancing need to happen across VTEPs.

This is achieved using a new bridge port type SAI_BRIDGE_PORT_TYPE_BRIDGE_PORT_NEXT_HOP_GROUP that is going to use SAI_BRIDGE_PORT_ATTR_BRIDGE_PORT_NEXT_HOP_GROUP_ID for nexthop selection. Nexthop wll be the one of the available VTEP tunnels.

There can be an associated protection NHG configured SAI_BRIDGE_PORT_ATTR_BRIDGE_PORT_PROTECTION_NEXT_HOP_GROUP_ID, as a backup path in case the bridge port goes down.

This PR is open in lieu of
https://github.com/opencomputeproject/SAI/pull/2058

PR #2286, opened 2026-05-27
EVPN VXLAN multihoming: split-horizon for P2MP tunnels

james-jra:
This change adds a new bridge port type to enable split-horizon filtering for P2MP VXLAN tunnels. This provides parity with the existing P2P tunnel model without requiring vendor-specific ACL logic in the NOS.

The existing spec for EVPN VXLAN multihoming ([SAI-Proposal-EVPN-Multihoming.md](https://github.com/opencomputeproject/SAI/blob/master/doc/tunnel/SAI-Proposal-EVPN-Multihoming.md) provides an example for configuring split-horizon which targets only P2P tunnels. Vendors using only P2MP tunnels have no equivalent path within the current API.

Split-horizon requires the ability to attach isolation groups at per-remote-VTEP granularity. With P2P tunnels this falls out naturally since each remote VTEP already has its own tunnel object and bridge port. With P2MP tunnels there's a single shared tunnel object, so we need something else to hang the isolation group on.

This change introduces a new bridge port type, `SAI_BRIDGE_PORT_TYPE_TUNNEL_TERM_PEER`, with an associated attribute `SAI_BRIDGE_PORT_ATTR_TUNNEL_TERM_PEER_IP`. This bridge port represents decapsulated tunnel-terminated traffic from a specific remote VTEP. It is explicitly rx-only and cannot be used for encap/tx.

A dedicated bridge port type makes it clear that these objects exist solely to deumultiplex inbound decapsulated traffic by source VTEP, giving the NOS a per-peer object to attach isolation groups to. This avoids any ambiguity with the existing `SAI_BRIDGE_PORT_TYPE_TUNNEL`, which is used for encap (via FDB entries, L2MC groups, next hops), and may still be used for decap in scenarios which do not require per-remote VTEP demux.

The flow for configuring split-horizon with P2MP tunnels then mirrors the P2P case:

1. Create a P2MP tunnel object.
2. Create a `TUNNEL_TERM_PEER` bridge port per remote VTEP, specifying the peer IP.
3. Attach an isolation group to any peer bridge port whose remote VTEP shares an Ethernet Segment with the local device.
4. Add the relevant local access ports as members of that isolation group.

## Alternative approach - User-managed ACLs

The alternative is to push this into NOS-managed ACLs. This approach was proposed in [EVPN Multi Home Support #2084](https://github.com/opencomputeproject/SAI/pull/2084#discussion_r1836429887), and uses ACLs to match on `SAI_ACL_ENTRY_FIELD_TUNNEL_TERMINATED` and `SAI_ACL_ENTRY_ATTR_FIELD_SRC_IP` to identify traffic from an ES peer, then use `SAI_ACL_ENTRY_ATTR_ACTION_SET_ISOLATION_GROUP` to block forwarding to access ports representing any ES shared with that ES peer.

ACL support is very varied between vendors, with differing stage support for fields, field/action combinations, and resource constraints. Requiring the NOS to manage ACLs pushes this vendor-specific complexity into the NOS. For example: the match fields here (source IP, tunnel terminated) are properties of the ingress/decap path, while the action (set isolation group) is an egress filtering operation. A vendor whose hardware doesn't support both at the same pipeline stage would require the NOS to use two ACL tables: one ingress ACL to identify ES-peer traffic and stamp metadata (`SAI_ACL_ENTRY_ATTR_ACTION_SET_ACL_META_DATA`), and one egress ACL to match the metadata and apply the isolation group. The NOS becomes responsible for metadata value allocation as well as ACL table/entry and isolation group management. Other vendors may have different restrictions resulting in vendor-specific code in the NOS.

This approach is also inconsistent with the level of abstraction of other multi-homing scenarios. Specifically for non-designated forwarder, the existing API has attributes like `SAI_BRIDGE_PORT_ATTR_TUNNEL_TERM_BUM_TX_DROP` which assume the SAI implementation internally understands whether traffic arriving at an egress bridge port was tunnel-terminated. Requiring the NOS to separately manage ACLs that detect and mark tunnel-terminated traffic is a different level of abstraction to this.

The proposed bridge port type allows the vendor to encapsulate any pipeline-specific logic within their SAI implementation and gives the NOS a uniform model that closely mirrors the P2P split-horizon flow.

### DIFF
--- a/doc/tunnel/SAI-Proposal-EVPN-Multihoming.md
+++ b/doc/tunnel/SAI-Proposal-EVPN-Multihoming.md
@@ -32,6 +32,7 @@
 | ---- | ---------  | ------------------------------------------------ | ------------------ |
 | 0.1  | Sep 23' 2024   | Jai Kumar, Rajesh Sankaran                   | Initial draft  |
 | 0.2  | Oct 23' 2024   | Rajesh Sankaran                              | Changed DF, single active attributes |
+| 0.3  | May 18' 2026   | James Andrew                                 | Added split-horizon support for P2MP tunnels |
 
 
 # 1.0  Introduction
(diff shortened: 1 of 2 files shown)


james-jra:
@tjchadaga can you please prompt someone to review this change?

yanjundeng (approved):

eddyk-nvidia:
@tjchadaga - what should we do to progress with this HLD ? It was reviewed by the SAI community and approved by Cisco. Is something else required to have this HLD merged ?

@itaibaz

jeflo (approved):

tjchadaga:
@r12f , @JaiOCP, @rck-innovium - please help review this PR

rck-innovium (approved):

james-jra:
@tjchadaga / @kcudnik this now has 2 separate approvals, you ok to run the pipeline and merge?

james-jra:
@kcudnik I fixed the metadata error, can you please rerun?

JaiOCP (approved):

james-jra:
@kcudnik can I please get a CI re-run?

james-jra:
@tjchadaga this is all green, can you merge?

james-jra:
@kcudnik @tjchadaga all green and approved, ok to merge?

tjchadaga (approved):

PR #2290, opened 2026-05-28
Packet Drop Monitoring

JaiOCP:
Packet drops are critical in AI fabrics and need to be monitored closely.
This PR brings in changes to support packet drop monitoring.


JaiOCP:
New PR is opened
https://github.com/opencomputeproject/SAI/pull/2298

PR #2293, opened 2026-06-11
Add per-port MACsec PFC mode attribute

darius-nexthop:
Define a new per-port attribute to control whether PFC frames on a MACsec-enabled port are handled as unencrypted (clear) or encrypted (secure) PFC frames.

Per-port control is needed because mixed rollout scenarios are possible on a single device, and a global knob does not scale well for dynamic port management and staged deployment.

The default behavior is set to unencrypted PFC.

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -408,6 +408,22 @@ typedef enum _sai_port_priority_flow_control_mode_t
 } sai_port_priority_flow_control_mode_t;
 
 /**
+ * @brief MACsec PFC mode
+ *
+ * Controls whether PFC frames on a MACsec-enabled port are handled as
+ * normal clear PFC frames or as encrypted/secure PFC frames.
+ */
+typedef enum _sai_port_macsec_pfc_mode_t
+{
+    /** Default behavior: clear PFC */
+    SAI_PORT_MACSEC_PFC_MODE_UNENCRYPTED,
+
+    /** Secure / encrypted PFC */
+    SAI_PORT_MACSEC_PFC_MODE_ENCRYPTED,
+
+} sai_port_macsec_pfc_mode_t;
+
+/**
  * @brief PTP mode
  * These modes can be used at the port and switch level.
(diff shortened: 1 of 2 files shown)


inc/saiport.h
```diff
@@ -1508,6 +1524,18 @@ typedef enum _sai_port_attr_t
...
+     * @brief Configure per-port MACsec PFC mode
+     *
+     * When MACsec is not enabled on the port, implementations may ignore this
+     * attribute and behave as SAI_PORT_MACSEC_PFC_MODE_UNENCRYPTED.
+     *
+     * @type sai_port_macsec_pfc_mode_t
+     * @flags CREATE_AND_SET
+     * @default SAI_PORT_MACSEC_PFC_MODE_UNENCRYPTED
+     */
+    SAI_PORT_ATTR_MACSEC_PFC_MODE,
```
Gnanapriya27: Why is the SAI_PORT_ATTR_MACSEC_PFC_MODE placed in saiport.h rather than as a SAI_MACSEC_PORT_ATTR_* in saimacsec.h on SAI_OBJECT_TYPE_MACSEC_PORT? All other MACsec port config id defined there. Placing it on the generic port creates GET/SET ambiguity when MACSEC is not enabled on this port.

inc/saiport.h
```diff
@@ -1508,6 +1524,18 @@ typedef enum _sai_port_attr_t
...
+
```
Gnanapriya27: New attr is placed in-between the existing attr - pl note that this placement changes the enum integer values of subsequent port attributes, which is an API break. The SAI community convention is to add new attributes at the end of the list, immediately before SAI_XXX_ATTR_END

darius-nexthop: Moved to the end, thanks

inc/saiport.h
darius-nexthop: Moved to saimacsec.h, thanks

PR #2296, opened 2026-06-17
Update Observation Domain ID description in TAM streaming proposal

MaksymPry:
In case there are several streaming sessions configured at the same time we can distinguish them by Observation Domain ID. Sequence Number is monotonically incremented per such Observation Domain ID.

### DIFF
--- a/doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
+++ b/doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
@@ -132,7 +132,7 @@ For more information on IPFIX, refer to the following resources:
 
 #### IPFIX header
 
-The `Version` and `Observation Domain ID` fields of the IPFIX header are identical for each IPFIX message.
+The `Version` field of the IPFIX header is identical for each IPFIX message. `Observation Domain ID` is defined by vendor implementation. `Sequence Number` starts from 0 and is monotonically incremented per `Observation Domain ID` modulo 2^32.
 
 ``` mermaid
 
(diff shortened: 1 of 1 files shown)


doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
```diff
@@ -132,7 +132,7 @@ For more information on IPFIX, refer to the following resources:
 
 #### IPFIX header
 
-The `Version` and `Observation Domain ID` fields of the IPFIX header are identical for each IPFIX message.
+The `Version` field of the IPFIX header is identical for each IPFIX message. `Observation Domain ID` is a unique value for each streaming session. `Sequence Number` starts from 0 and is monotonically incremented modulo 2^32.
```
Pterosaur: I think requiring a unique Observation Domain ID for each streaming session is too restrictive. Multiple streaming sessions may share the same Observation Domain ID, such as the original design, or each session may use a distinct one, depending on the vendor implementation. SAI should not mandate a one-to-one mapping between streaming sessions and Observation Domain IDs.

The key requirement should be that, for the same Observation Domain ID, the Sequence Number is monotonically incremented modulo 2^32.

Pterosaur (approved):

doc/TAM/SAI-Proposal-TAM-stream-telemetry.md
MaksymPry: Thank you,
Updated

PR #2298, opened 2026-06-22
Packet Drop Monitoring

JaiOCP:
This PR introduces packet drop monitoring where monitoring can be enabled for a specific set of drop reasons for a specific stage of pipeline like ingress, egress and buffer. If this is done with hw learning enabled for deduplcation then this form packet drop monitoring is considered stateful.

This PR introduces framework for both stateful and stateless packet drop monitoring.

Accidentally closed PR
https://github.com/opencomputeproject/SAI/pull/2290

### DIFF
--- a/doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
+++ b/doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
@@ -563,6 +563,8 @@ ACL table is created with fields enabled that will be used to install the flow e
 
 This table can later be queried by NOS for learned entries using the GET API. Learned entries can be aged out using the SAI_TAM_EVENT_LEARN_ENTRY_ATTR_ACL_ENTRY object.
 
+NOS can also age out the entries using sai_tam_event_learn_entry_attr_t/SAI_OBJECT_TYPE_TAM_EVENT_LEARN_ENTRY object. In such case SAI adapter will dereference the hardware index of the learned entry for aging.
+
 Conceptually there is no SET on the ACL table entries as they are created by SAI adapter.
 SAI adapter should fail any SET operation if done by the NOS.
 
--- a/doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
+++ b/doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
@@ -1,0 +1,634 @@
+# Mirror on Drop SAI Specification
+-------------------------------------------------------------------------------
+ Title       | SAI support for Packet Drop Monitoring
+:-------------|:-----------------------------------------------------------------
+ Authors     | Jai Kumar, Broadcom Inc
+ Status      | In review
+ Type        | Standards track
+ Created     | 03/18/20206: Initial Draft
+ SAI-Version | 1.18
+-------------------------------------------------------------------------------
+
+
+## 1.0  Introduction
+Packet drops in networks, whether intentional due to policy or unintentional due to hardware errors and misconfigurations, adversely impact application performance and SLAs. Common causes include hardware errors (CRC, SER), policy-based drops (ACLs), forwarding decisions (invalid routes, TTL expiry), and network congestion.
+
+Monitoring these drops can place a significant load on the collector subsystem. Depending on the requirements, a collector may need to monitor all drops, unique drops, or specific types of drops for certain flows.
+
+TAM provides a common framework for monitoring these occurrences. This document outlines the SAI API steps required to configure packet drop monitoring for a flow or flow group based on specific drop reasons.
+
+
+## 2.0 Terms and Acronyms
+
+| Term| Description | 
+|:---|:---|
+| Drop Report | Typically IPFIX format packet carrying drop information along with other metadata |
+| TAM | Telemetry And Monitoring. A SAI object for anything telemetry |
+
+
+
+
+## 3.0 Overview
+Packet drops in the switch are critical events. Reporting these drops, along with specific drop reasons and metadata, is essential for the fabric controller to analyze the cause and take corrective action.
+
+Drop reports can be generated directly by the switch hardware directred twowards an external collector or local CPU or can be sent as notifications to the local CPU. 
+
+Drop Report is the packetized information about the drop send to the collector or to the host typically encapsulated in IPFIX format. TAM framework supports different types of encapsulation and can be supported for drop report based on the underlying hardware support.
+Minimaly drop report Should contain
+
+- Trimmed Packet
+- One or more drop reasons
+- Timestamp
+- Ingress port
+
+## 4.0 Goals
+
+Following principles are maintained regarding the drop monitoring infrastructure:
+
+1. Hardware Independence: The APIs do not make assumptions about the hardware's drop monitoring capabilities. Silicon vendors may provide various capabilities and support for vendor specific drop reasons.
+
+2. TAM Hierarchy: The NOS must use the TAM hierarchy  (including collectors, reports, events, and drop types) for configuring drop monitoring.
+
+3. Flow Definitions: Flow definitions must use ACLs or equivalent infrastructure within the NOS, as these are basic match-and-action constructs.
+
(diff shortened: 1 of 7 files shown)


inc/saitam.h
```diff
@@ -2575,6 +2707,323 @@ typedef void (*sai_tam_event_notification_fn)(
...
+     * @objects SAI_OBJECT_TYPE_TAM_EVENT_LEARN_ENTRY
+     */
+    sai_attribute_t *attr;
+
+    /**
+     * @brief TAM Learn entry ID
+     *
+     * @objects SAI_OBJECT_TYPE_TAM_EVENT_LEARN_ENTRY
+     */
+    sai_object_id_t le_id;
```
kcudnik: does it make sense to make this object first in the structure?

JaiOCP: The attributes need a reference to the object hence I have kept it here in structure.
Any other suggestions?

kcudnik: im just saying make it first field

JaiOCP: Just for curiosity. How does it help.

Alignment ??

kcudnik: it does not change alignement, just from order perspective it make sense to have object id first and then it;s attributes, for example the same idea is when you create object or call set api, object id is first

inc/saidebugcounter.h
```diff
@@ -76,6 +76,31 @@ typedef enum _sai_debug_counter_bind_method_t
...
+typedef enum _sai_buffer_drop_reason_t
+{
+    /** Start of buffer drop reasons */
+    SAI_BUFFER_DROP_REASON_START,
+
+    /** Any buffer drop */
+    SAI_BUFFER_DROP_REASON_ANY = SAI_BUFFER_DROP_REASON_START,
+
+    /** IPG packet drops */
+    SAI_BUFFER_DROP_REASON_IPG,
```
JaiOCP, line 91: Create two sections using comments for "BUFFER_IN_DROP"
Ingress Priority Group Buffer
Queue Buffer Drops

inc/saiswitch.h
```diff
@@ -3616,6 +3616,42 @@ typedef enum _sai_switch_attr_t
...
+    /**
+     * @brief List of supported out drop reasons
+     *
+     * @type sai_s32_list_t sai_out_drop_reason_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_OUT_DROP_REASON_LIST,
+
+    /**
+     * @brief List of supported buffer drop reasons
```
johevans, line 3636: In both Ingress and Egress, there are two kinds of drops: drops before packet processing and drops after packet processing. For example, most architectures have an early drop mechanism to handle oversubscription. Based only on packet header attributes (such as the dscp or ieee priority bits), low priority packets are dropped before even entering the pipeline. This is done to conserve pipeline resources and allow control packets through even during heavy, ASIC-wide congestion such as oversubscription. However, since the packets have not been processed, the drop decision is rather crude and without any exceptions. Ideally, you want to drop packets after packet processing, which could remark the priority of the packet or a set a pipeline attribute such as "drop precedence". Most architectures have a set of virtual statistics queues (VSQs) for this purpose where buffer limits can be set based on priority, class, filter results, etc. Note: the drop decision may be probabilistic such as WRED, but fundamentally it is the same mechanism. Both kinds of drops, pre and post packet processing, could be considered "buffer drops" or "mmu drops" since they're based on buffer thresholds. Therefore, it might be useful to separate Ingress admission drops from Ingress processing drops. In addition, there are also rare, hardware related buffer drops such as corrupted dram reads.

The situation at Egress is similar. In a fabric architecture, packets could be dropped before they are even assembled, e.g. cells dropped due to CRC errors. Also, packets could be dropped before entering the pipeline based on various thresholds, such as the global buffer limits for unicast/multicast, the drop precedence carried from Ingress, packet buffer and descriptor limits per QoS, multicast replication resource limits, etc. Once in the Egress pipeline, packets may be dropped because the destination queue is full or an Egress filter has decided to drop the packet. And again, there also are esoteric hardware specific reasons.

In summary, it would be useful to separate Ingress/Egress admission (pre pipeline) drops from post-processing drops.

doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
```diff
@@ -0,0 +1,617 @@
...
+	attr_count, 
+	sai_attr_list);
+   
+```
+
+### 6.3 Create ACL table for learning 
+
+ACL table is created with fields enabled that will be used to install the flow entry when the leraning happens. In this example flow is learned for the qset[ip proto, src ip, dst ip, sport, dport, ingress port].
+
+This table can later be queried by NOS for learned entries using the GET API. Conceptually there is no SET or REMOVE on this table and SAI adapter should fail any set/remove operation if done by the NOS.
```
rbudhia-upscaleai: If aging is performed by the NOS, it will need a mechanism to remove aged entries. A remove operation may be required for this.

doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
```diff
@@ -0,0 +1,617 @@
...
+```
+/**
+ * @brief Enum defining learn event.
+ */
+typedef enum _sai_tam_event_learn_entry_attr_t
+{
+    /**
+     * @brief Start of Attributes
+     */
+    SAI_TAM_EVENT_LEARN_ENTRY_ATTR_START,
```
rbudhia-upscaleai, line 263: Include the ACL entry ID in the learn notification to allow the NOS to correlate the learned entry with the table entry. This is especially needed for NOS-managed aging.

inc/saitam.h
JaiOCP: ok, just from flow point of view. First object, then atribute count and then attributes.

Accepted the change.

doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
JaiOCP: Accepted the change

doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
JaiOCP: Agreed, Updated the document

JaiOCP:
@rbudhia-upscaleai @j-bos Please approve. All the comments are taken care off.

rbudhia-upscaleai (approved):

tjchadaga:
@johevans - could you please complete the review and sign-off on this?

doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
```diff
@@ -0,0 +1,634 @@
...
+	attr_count, 
+	sai_attr_list);
+   
+```
+
+### 6.3 Create ACL table for learning 
+
+ACL table is created with fields enabled that will be used to install the flow entry when the leraning happens. In this example flow is learned for the qset[ip proto, src ip, dst ip, sport, dport, ingress port].
+
+This table can later be queried by NOS for learned entries using the GET API. Learned entries can be aged out using the SAI_TAM_EVENT_LEARN_ENTRY_ATTR_ACL_ENTRY object.
```
bandaru-viswanath, line 564: Can we also support an workflow where aging can be performed by deleting the HW-Learn-Entry object ?

bandaru-viswanath (commented):
Could you please see the comment ?

doc/DropMonitor/SAI-Proposal-Packet-Drop-Monitoring-Spec.md
JaiOCP: Yes, That is possible as well. I will update the document with the text. There is no separate workflow just a text indicating that aging can be done using ACL entry object or hw learn object.

JaiOCP:
> Could you please see the comment ?

Done. Can you please approve the PR

https://github.com/opencomputeproject/SAI/pull/2298/changes/2527a0e979f5c6f3c1832aec342bfbf4616bd531

bandaru-viswanath (approved):

tjchadaga:
@JaiOCP - could you please squash your commits?

JaiOCP:
https://github.com/opencomputeproject/SAI/pull/2328 is opened

PR #2299, opened 2026-06-22
saithriftv2: link VPP libraries for platform=vpp + bookworm/py3 build fixes

nicholasching:
**Context / motivation**

This PR is part of the **SAIVPP unit-test framework**: a Docker harness (`docker-sai-test-vpp`) that runs the upstream OCP `sai_test` PTF suite against the real VPP SAI backend (`libsaivs`) in one container — VPP + `saiserver` + PTF + veth/AF_PACKET topology. That work is documented in our devdocs under `sonic-sairedis/.azure-pipelines/docker-sai-test-vpp/devdocs/` (see `progress.md`).

To run the suite, `saiserver` (saithriftv2) must build and link against the VPP SAI backend on a bookworm/python3 toolchain. Upstream `saithriftv2` only has a `vs` link line and a python2.7 install path, so it does not build for the `vpp` platform on bookworm. This is **Phase 1, Task 2** of the SAIVPP UT HLD (the saithriftv2 Makefile work).

**What this change does**

Build/packaging only, no test logic:

- **`test/saithriftv2/Makefile`** — add a `platform=vpp` link line that pulls in the 5 VPP libraries `SaiVppXlate.c` requires (`-lvlib -lvlibapi -lvppapiclient -lvlibmemoryclient -lvppinfra`) plus `-lswsscommon`; also add `-lswsscommon` to the `vs` line.
- **`test/saithriftv2/src/saiserver.cpp`** — enable SWSS debug logging at startup (`swss::Logger` to STDOUT), so the harness captures a high-level SAI RPC trace in `saiserver.log` (used throughout our debugging).
- **`debian/python-saithrift.install`** — install the Python bindings from the python3 `dist-packages` path so the package builds on bookworm (was hardcoded to `python2.7/site-packages`).
- **`test/saithriftv2/convert_header.py`** — fix an enum `__str__` that did `split(".")[1]` and threw an IndexError on some enum names; use `self.name` instead.

**Scope / risk**

- **4 files, +11/−3** — build flags, a packaging path, a logging init, and a one-line stringification fix.
- **No change to SAI semantics, RPC surface, or test logic.**
- The `vpp` link line is new (additive); the `vs` line only gains `-lswsscommon`.

PR #2300, opened 2026-06-22
saiserver: default switch_id on neighbor/route entries when omitted 

nicholasching:
**Context / motivation**

Part of the **SAIVPP unit-test framework** (see PR 1 / our `docker-sai-test-vpp/devdocs/`). When running the OCP `sai_test` suite against the VPP SAI backend, neighbor- and route-create tests failed at create time with `SAI_STATUS_INVALID_PARAMETER (-5)`:

```
ERROR: meta_generic_validation_create: switch id is NULL for SAI_OBJECT_TYPE_NEIGHBOR_ENTRY
```

**Root cause**

`sai_thrift_neighbor_entry_t` / `sai_thrift_route_entry_t` carry a `switch_id`, but many OCP tests build these entries with only `rif_id` (+ `ip_address` / `destination`) and **omit `switch_id`**. The saithriftv2 RPC server copied the thrift `switch_id` straight through, so meta validation received `switch_id == 0` and rejected the create. (Tests that pass through the common-config path, which sets `switch_id` explicitly, succeeded — only entries that omit it failed.)

**What this change does**

- **`meta/templates/sai_rpc_server_helper_functions.tt`** — custom `sai_thrift_parse_neighbor_entry()` / `sai_thrift_parse_route_entry()` that, when the thrift `switch_id` is `0`, fall back to the RPC-global `switch_id` (the OID returned by `sai_thrift_create_switch`) and then `gSwitchId`. (The auto-generated struct copies for these two entry types are skipped in favor of these.)
- **`test/saithrift/src/switch_sai.thrift`, `switch_sai_rpc_server.cpp`** — the same `switch_id` field + fallback in the legacy v1 thrift server, for parity.
- **`test/sai_test/config/route_configer.py`** — pass `switch_id=self.dut.switch_id` explicitly on the route entries the common config builds, so the persisted T0 config is consistent regardless of the fallback.

This mirrors how a real ASIC SAI shim behaves (the switch is unambiguous), and unblocks the neighbor/route-create-on-RIF tests.

**Scope / risk**

- **4 files, +79/−18** — RPC-server parsing for two entry types (+ v1 parity) and explicit `switch_id` in the test common config.
- **No change to SAI behavior for callers that already pass `switch_id`** — the fallback only triggers when it is `0`/omitted.
- Server-side only; no packet-path or test-logic change.

**Dependencies**
None. Related to #2299, however, their edits are isolated and each target `master`.

### DIFF
--- a/meta/templates/sai_rpc_server_helper_functions.tt
+++ b/meta/templates/sai_rpc_server_helper_functions.tt
@@ -136,7 +136,6 @@ void sai_thrift_parse_[% struct.short_name %](const [% struct.thrift_name %] &th
 [%- ######################################################################## -%]
 
 [%- BLOCK special_helper_functions -%]
-extern sai_object_id_t gSwitchId;
 extern sai_object_id_t switch_id;
 
 void sai_thrift_parse_buffer(const std::string &thrift_buffer,
@@ -152,13 +151,11 @@ void sai_thrift_parse_neighbor_entry(const sai_thrift_neighbor_entry_t &thrift_n
     }
 
     // OCP sai_test often omits switch_id on neighbor_entry; fall back to the
-    // RPC-global switch_id (set by sai_thrift_create_switch) then gSwitchId.
+    // RPC-global switch_id set by sai_thrift_create_switch.
     if (thrift_neighbor_entry.switch_id != 0) {
         neighbor_entry->switch_id = (sai_object_id_t)thrift_neighbor_entry.switch_id;
     } else if (switch_id != SAI_NULL_OBJECT_ID) {
         neighbor_entry->switch_id = switch_id;
-    } else if (gSwitchId != SAI_NULL_OBJECT_ID) {
-        neighbor_entry->switch_id = gSwitchId;
     } else {
         neighbor_entry->switch_id = SAI_NULL_OBJECT_ID;
     }
@@ -179,8 +176,6 @@ void sai_thrift_parse_route_entry(const sai_thrift_route_entry_t &thrift_route_e
         route_entry->switch_id = (sai_object_id_t)thrift_route_entry.switch_id;
     } else if (switch_id != SAI_NULL_OBJECT_ID) {
         route_entry->switch_id = switch_id;
-    } else if (gSwitchId != SAI_NULL_OBJECT_ID) {
-        route_entry->switch_id = gSwitchId;
     } else {
         route_entry->switch_id = SAI_NULL_OBJECT_ID;
     }
--- a/meta/templates/sai_rpc_server_helper_functions.tt
+++ b/meta/templates/sai_rpc_server_helper_functions.tt
@@ -31,7 +31,8 @@
 }
             [%- END -%]
             [%- FOREACH struct IN apis.$api.structs %]
-
+                [%- NEXT IF struct.short_name == 'neighbor_entry' %]
+                [%- NEXT IF struct.short_name == 'route_entry' %]
 
                 [%- PROCESS struct_helper_function_header %] {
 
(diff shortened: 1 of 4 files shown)


nicholasching:
@tjchadaga please run azure pipeline when possible; this PR is needed for an upstream feature at Cisco ASAP, thanks!

test/saithrift/src/switch_sai_rpc_server.cpp
```diff
@@ -241,7 +241,13 @@ class switch_sai_rpcHandler : virtual public switch_sai_rpcIf {
   }
 
   void sai_thrift_parse_neighbor_entry(const sai_thrift_neighbor_entry_t &thrift_neighbor_entry, sai_neighbor_entry_t *neighbor_entry) {
-      neighbor_entry->switch_id = gSwitchId;
```
yue-fred-gao, line 244: I don't quite understand about this change. From my reading of the code, saiserver stores switch_id in gSwitchId (https://github.com/opencomputeproject/SAI/blob/c0203c03249ae378c2a447d2b83bd8dddb9abf73/test/saithrift/src/saiserver.cpp#L438). Subsequent calls can use the gSwitchId if it is not in the RPC request, such as neighbor entries. Is there a case that gSwitchId is invalid?

nicholasching: Yes, that is correct; for v1, `gSwitchId` is always valid so the change is not needed here. I made it to maintain feature parity w/`saithriftv2`.

Edits to `saithriftv2` were needed because `gSwitchId` in `saiserver.cpp` is only declared and not assigned on the RPC path.

https://github.com/opencomputeproject/SAI/blob/9e90412102040a2bb85db34cf61b7850fa2e2102/test/saithriftv2/src/saiserver.cpp#L43

Instead, `sai_thrift_create_switch()` stores the OID in the RPC-global `switch_id`.

So when tests created a neighbor or route over RPC, and did not send `switch_id`, `saithriftv2` would pass `switch_id=0` to `SAI`, resulting in `switch id is NULL` error (-5). The change to `saithriftv2` adds a fallback for when `switch_id` is not passed in (0): first to the RPC path `switch_id` set by `create_switch` (which resolves the -5), then to `gSwitchId` as backup.

Should I keep this change for feature parity or remove to avoid touching v1?

test/saithrift/src/switch_sai.thrift
```diff
@@ -198,8 +198,9 @@ struct sai_thrift_route_entry_t {
 }
 
 struct sai_thrift_neighbor_entry_t {
-    1: sai_thrift_object_id_t rif_id;
-    2: sai_thrift_ip_address_t ip_address;
+    1: sai_thrift_object_id_t switch_id;
```
yue-fred-gao: I think this is not needed. We should keep the original design. let sai rpc server assign switch_id use the stored RPC scoped switch_id variable. Then the changes in route_configer.py are also not needed.

meta/templates/sai_rpc_server_helper_functions.tt
```diff
@@ -135,8 +136,57 @@ void sai_thrift_parse_[% struct.short_name %](const [% struct.thrift_name %] &th
...
+    }
+
+    // OCP sai_test often omits switch_id on neighbor_entry; fall back to the
+    // RPC-global switch_id (set by sai_thrift_create_switch) then gSwitchId.
+    if (thrift_neighbor_entry.switch_id != 0) {
+        neighbor_entry->switch_id = (sai_object_id_t)thrift_neighbor_entry.switch_id;
+    } else if (switch_id != SAI_NULL_OBJECT_ID) {
+        neighbor_entry->switch_id = switch_id;
+    } else if (gSwitchId != SAI_NULL_OBJECT_ID) {
+        neighbor_entry->switch_id = gSwitchId;
```
yue-fred-gao: same as above, we probably can remove this line.

meta/templates/sai_rpc_server_helper_functions.tt
```diff
@@ -135,8 +136,57 @@ void sai_thrift_parse_[% struct.short_name %](const [% struct.thrift_name %] &th
 [%- ######################################################################## -%]
 
 [%- BLOCK special_helper_functions -%]
+extern sai_object_id_t gSwitchId;
```
yue-fred-gao: In thrift v2, gSwitchId is never set, right? We probably can remove this line.

test/saithrift/src/switch_sai_rpc_server.cpp
yue-fred-gao: I see. Didn't know there are 2 versions of saithrift. I think we can undo the change. Leave v1 as is

test/saithrift/src/switch_sai.thrift
nicholasching: I agree, since issue was fixed by RPC-scoped `switch_id` fallback, no need for tests to pass in `switch_id`

meta/templates/sai_rpc_server_helper_functions.tt
nicholasching: Yes, that is correct; `gSwitchId` is never set in v2. I will remove the line.

test/saithrift/src/switch_sai_rpc_server.cpp
nicholasching: Will do!

nicholasching:
Hi @yue-fred-gao! I've made the requested changes on latest; now only fallback to RPC-scoped switch_id. Changes for this PR are much cleaner now, only touching 1 file; thanks!

meta/templates/sai_rpc_server_helper_functions.tt
```diff
@@ -135,8 +136,52 @@ void sai_thrift_parse_[% struct.short_name %](const [% struct.thrift_name %] &th
...
+void sai_thrift_parse_neighbor_entry(const sai_thrift_neighbor_entry_t &thrift_neighbor_entry,
+                                     sai_neighbor_entry_t *neighbor_entry)
+{
+    if (neighbor_entry == NULL) {
+        return;
+    }
+
+    // OCP sai_test often omits switch_id on neighbor_entry; fall back to the
+    // RPC-global switch_id set by sai_thrift_create_switch.
+    if (thrift_neighbor_entry.switch_id != 0) {
```
yue-fred-gao, line 155: does sai_thrift_neighbor_entry_t have switch_id?

nicholasching: Yes, the`sai_thrift_neighbor_entry_t` in `thriftv2` is generated from the SAI header `sai_neighbor_entry_t`, of which `switch_id` is a member.

https://github.com/opencomputeproject/SAI/blob/c0203c03249ae378c2a447d2b83bd8dddb9abf73/inc/saineighbor.h#L185-L192

yue-fred-gao: I see. The previous change to sai_thrift_neighbor_entry_t is actually for v1. v2 always generate sai thrift from SAI headers.

yue-fred-gao (approved):

nicholasching:
@kcudnik, could you please invoke azure pipelines to run checks, please. I do not have the permissions to do so.

nicholasching:
Hi @kcudnik, the changes in this PR have been reviewed by @yue-fred-gao and are passing checks. When you have a chance, could you please take a quick look and help us merge into master for upstream feature in `sonic-sairedis`? Thanks!

kcudnik (approved):

PR #2301, opened 2026-06-22
sai_test config: drop redundant bridge_id, env-gated port bring-up, fix v4/v6 NHG port-list aliasing

nicholasching:
**Context / motivation**

Part of the **SAIVPP unit-test framework** (see PR 1 / our `docker-sai-test-vpp/devdocs/`, esp. the 6-19 entry). Three independent correctness/robustness fixes in the OCP `sai_test` config helpers that the suite needs when run against the VPP SAI backend.

**What this change does**

- **`test/sai_test/config/port_configer.py` — drop `bridge_id` on `create_bridge_port`.** Passing `bridge_id` to `sai_thrift_create_bridge_port` caused a create failure in our backend; the default 1Q bridge is used, so the argument is redundant. Removing it lets bridge-port creation succeed.

- **`test/sai_test/config/port_configer.py` — env-gated, bounded port bring-up wait.** `turn_up_and_get_checked_ports()` waited **per port, serially** (`retries × sleep`) for oper-status UP. On a 32-port topology where oper-status is slow to settle, this is ~60s+ of dead time per common-config build. The wait is now tunable via env (`SAI_PORT_UP_RETRIES`, `SAI_PORT_UP_POLL_INTERVAL`, `SAI_PORT_UP_SHARED_WAIT`) and can poll **all ports together** in one bounded window. **Defaults preserve the original behavior** (per-port wait), so real-HW/other OCP consumers are unchanged unless they opt in; only our harness sets the fast values.

- **`test/sai_test/config/route_configer.py` — give v4/v6 NHGs independent `member_port_indexs`.** `create_nexthop_group_by_nexthops()` constructed the v4 and v6 `NexthopGroup` objects sharing **one** Python list object for `member_port_indexs`. A mutation on one group (member remove/re-add tests) then corrupted the other's port list (`ValueError: list.remove(x): x not in list`). Each group now gets its own `list(...)` copy. (Latent aliasing bug, independent of any harness specifics.)

**Scope / risk**

- **2 files, +64/−25** — test-config helpers only; no change to SAI/backend code or packet semantics.
- The port-up change is **opt-in via env with original defaults**, so it does not alter timing for existing consumers.
- The `bridge_id` removal relies on the default 1Q bridge (already how these tests are used).

**Dependencies**
None. Related to #2299 and #2300, however, their edits are isolated and each target `master`.

### DIFF
--- a/test/sai_test/config/port_configer.py
+++ b/test/sai_test/config/port_configer.py
@@ -18,6 +18,7 @@
 #
 #
 
+import os
 from collections import OrderedDict
 from ptf import config
 from sai_utils import *  # pylint: disable=wildcard-import; lgtm[py/polluting-import]
@@ -497,7 +498,22 @@ class PortConfiger(object):
         '''
 
         # For brcm devices, need to init and setup the ports at once after start the switch.
-        retries = 2
+        #
+        # Waiting strategy is configurable via environment, defaulting to the original
+        # behavior (retries=2, 1s poll interval). The original loop waited PER PORT
+        # serially (retries x interval seconds for each of N ports), which on a large
+        # port count where oper-status is slow to settle costs N * retries * interval
+        # seconds (e.g. 32 ports x 2 x 1s ~= 64s). The shared-wait path below issues
+        # admin-up to every port once, then polls ALL ports together for at most
+        # (retries x interval) seconds total, re-asserting admin-up on stragglers each
+        # round. Semantics are preserved: every port is set admin-up, oper-status is
+        # polled, and down_port_list / the returned list are identical; only the total
+        # wall-clock wait changes. Real-HW/ASIC consumers that do not set these env
+        # vars keep the exact original timing.
+        retries = int(os.environ.get('SAI_PORT_UP_RETRIES', '2'))
+        poll_interval = float(os.environ.get('SAI_PORT_UP_POLL_INTERVAL', '1'))
+        # Opt-in: poll all ports together in one bounded loop instead of per-port.
+        shared_wait = os.environ.get('SAI_PORT_UP_SHARED_WAIT', '0') == '1'
         down_port_list = []
         test_port_list:List[Port] = []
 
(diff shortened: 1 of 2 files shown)


test/sai_test/config/port_configer.py
```diff
@@ -120,7 +121,6 @@ def create_bridge_ports(self, bridge_id, port_list: List['Port']):
         for index, item in enumerate(port_list):
             port_bp = sai_thrift_create_bridge_port(
                 self.client,
-                bridge_id=bridge_id,
```
yue-fred-gao, line 123: How will this change affect other platforms? Can this be fixed in vpp platform?

nicholasching: In the SAI spec, `SAI_BRIDGE_PORT_ATTR_BRIDGE_ID` is not valid for `SAI_BRIDGE_PORT_TYPE_PORT`. 

https://github.com/opencomputeproject/SAI/blob/0aba2360f0dbb7b869628fd9aca6c4ffd1d401a8/inc/saibridge.h#L182-L190

For this type, the bridge port is implicitly associated with the switch’s default 1Q bridge and `bridge_id` should not be passed on create.

VPP (via saiserver) was using the `sonic-sairedis` meta, enforcing the `@condition` above and rejected the extra `bridge_id` for `PORT` type. This was causing issues during the tests where create failed at SAI API validation. Since `bridge_id` should not be included for `PORT` type according to `saibridge.h`, I believe VPP's enforcement is correct and the fix would be relevant to other platforms as well; they were likely ignoring the extra attribute before.

yue-fred-gao: got it. I see create_bridge_ports has only one caller and the type is PORT. We can safely remove bridge id.

yue-fred-gao (approved):

nicholasching:
Hi @kcudnik, the changes in this PR have been reviewed by @yue-fred-gao and are passing checks. When you have a chance, could you please take a quick look and help us merge into master for upstream feature in sonic-sairedis? Thanks!

nicholasching:
Thank you for your help, @kcudnik!

PR #2302, opened 2026-06-22
NHG with Unequal Cost Members

JaiOCP:
This PR brings in support for high performant fabric where NHG is created with prior information about the members with weights.

NOS can use sai_query_attribute_capability() to find out if the platform supports such behavior or not. Legacy implementation will always return the status as not supported attribute.

### DIFF
--- a/inc/sainexthopgroup.h
+++ b/inc/sainexthopgroup.h
@@ -142,6 +142,22 @@ typedef struct _sai_next_hop_group_hw_protection_switchover_notification_data_t
 } sai_next_hop_group_hw_protection_switchover_notification_data_t;
 
 /**
+ * @brief Next hop group is configured with weights or not
+ */
+typedef enum _sai_next_hop_group_members_weight_t
+{
+    /** This is for legacy platforms where this attribute is don't care */
+    SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_DONT_CARE,
+
+    /** Next hop group members are weighted */
+    SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_WEIGHTED,
+
+    /** Next hop group members are unweighted */
+    SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_UNWEIGHTED,
+
+} sai_next_hop_group_members_weight_t;
+
+/**
  * @brief Attribute id for next hop
  */
 typedef enum _sai_next_hop_group_attr_t
@@ -360,17 +376,13 @@ typedef enum _sai_next_hop_group_attr_t
     SAI_NEXT_HOP_GROUP_ATTR_ADMIN_ROLE,
 
     /**
-     * @brief This attribute indicates the nexthop group members weights
+     * @brief This attribute indicates if all the members are with weights or are unweighted
      *
-     * NOS must query this attribute for the support.
-     * False: All members are with equal cost
-     * True: Members are configured with unequal cost
-     *
-     * @type bool
+     * @type sai_next_hop_group_members_weight_t
      * @flags CREATE_ONLY
-     * @default false
+     * @default SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_DONT_CARE
      */
-    SAI_NEXT_HOP_GROUP_ATTR_MEMBERS_WITH_UNEQUAL_WEIGHTS,
+    SAI_NEXT_HOP_GROUP_ATTR_MEMBERS_WEIGHT,
 
     /**
      * @brief End of attributes
--- a/inc/sainexthopgroup.h
+++ b/inc/sainexthopgroup.h
@@ -360,6 +360,19 @@ typedef enum _sai_next_hop_group_attr_t
     SAI_NEXT_HOP_GROUP_ATTR_ADMIN_ROLE,
 
     /**
+     * @brief This attribute indicates the nexthop group members weights
+     *
+     * NOS must query this attribute for the support.
+     * False: All members are with equal cost
+     * True: Members are configured with unequal cost
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_NEXT_HOP_GROUP_ATTR_MEMBERS_WITH_UNEQUAL_WEIGHTS,
+
+    /**
      * @brief End of attributes
      */
     SAI_NEXT_HOP_GROUP_ATTR_END,


inc/sainexthopgroup.h
```diff
@@ -359,6 +359,19 @@ typedef enum _sai_next_hop_group_attr_t
...
+     *
+     * NOS must query this attribute for the support.
+     * False: All members are with equal cost
+     * True: Members are configured with unequal cost
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_NEXT_HOP_GROUP_ATTR_MEMBERS_WITH_UNEQUAL_WEIGHTS,
```
sathk-marvell: @JaiOCP, what is the additional benefit of querying this capability?

If NOS is configured to program unequal paths, NOS can directly add those members to the NHG with the required weights, right? I believe we don't need any additional hardware capability/support to handle unequal-cost paths in this case.

Please let me know if I am missing any specific use case where this capability is required.

JaiOCP:
There is a case where NHG need to allocated differently for ECMP and UCMP for highly optimized fabrics.This PR addresses that case.On Jun 22, 2026, at 8:26 PM, sathk-marvell ***@***.***> wrote:﻿
@sathk-marvell commented on this pull request.



In inc/sainexthopgroup.h:
> @@ -359,6 +359,19 @@ typedef enum _sai_next_hop_group_attr_t
      */
     SAI_NEXT_HOP_GROUP_ATTR_ADMIN_ROLE,
 
+    /**
+     * @brief This attribute indicates the nexthop group members weights
+     *
+     * NOS must query this attribute for the support.
+     * False: All members are with equal cost
+     * True: Members are configured with unequal cost
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_NEXT_HOP_GROUP_ATTR_MEMBERS_WITH_UNEQUAL_WEIGHTS,

@JaiOCP, what is the additional benefit of querying this capability?
If NOS is configured to program unequal paths, NOS can directly add those members to the NHG with the required weights, right? I believe we don't need any additional hardware capability/support to handle unequal-cost paths in this case.
Please let me know if I am missing any specific use case where this capability is required.

inc/sainexthopgroup.h
JaiOCP: For the legacy implementation, yes.
For newer implementation where there is a dedicated resource for weight NHG vs unweighted NHG, we need to differentiate the NHG during NHG create.
Now you can make an argument that why not a new NHG type instead of a bool. Problem with that approach is that NHG may be of any type with members as wieghted or unweighted. Hence the introduction of a new bool attribute.

inc/sainexthopgroup.h
```diff
@@ -359,6 +359,19 @@ typedef enum _sai_next_hop_group_attr_t
...
+     * NOS must query this attribute for the support.
+     * False: All members are with equal cost
+     * True: Members are configured with unequal cost
+     *
+     * @type bool
+     * @flags CREATE_ONLY
+     * @default false
+     */
+    SAI_NEXT_HOP_GROUP_ATTR_MEMBERS_WITH_UNEQUAL_WEIGHTS,
+
```
eddyk-nvidia, line 386: Default for SAI_NEXT_HOP_GROUP_ATTR_MEMBERS_WITH_UNEQUAL_WEIGHTS is False (all members are with equal cost).  It means that the existing code not setting a new SAI attribute will not be able to use SAI_NEXT_HOP_GROUP_MEMBER_ATTR_WEIGHT to set a weight per member ? Shouldn't TRUE be used as default  ?

rck-innovium: I agree with @eddyk-nvidia that with the current description and using "false" as the default value affects legacy platforms. 

@JaiOCP What I can infer from the explanations is that we want this control on hardware that has dedicated resources for UCMP.  Maybe model this as a bool attribute "SAI_NEXT_HOP_GROUP_ATTR_NATIVE_UCMP" with default value as false. The application will do a capability query on this attr and ignore it on unsupported platforms. In supported platforms, the application will explicitly set SAI_NEXT_HOP_GROUP_ATTR_NATIVE_UCMP to true on NHGs where it wants to leverage the h/w's dedicated UCMP functionality.  _Basically, model this attribute as a hint to the SDK rather than a mandate_.

rck-innovium (changes requested):

inc/sainexthopgroup.h
JaiOCP: introduced an enum to be clear on the intent.
DONT_CARE default means that this attribute can be ignored by legacy platforms or SONIC.

JaiOCP:
@eddyk-nvidia @rck-innovium @tjchadaga Changes made as per the review.
Please take a look and let me know if anything else is needed

inc/sainexthopgroup.h
```diff
@@ -360,17 +376,13 @@ typedef enum _sai_next_hop_group_attr_t
     SAI_NEXT_HOP_GROUP_ATTR_ADMIN_ROLE,
 
     /**
-     * @brief This attribute indicates the nexthop group members weights
-     *
-     * NOS must query this attribute for the support.
-     * False: All members are with equal cost
-     * True: Members are configured with unequal cost
+     * @brief This attribute indicates if all the members are with weights or are unweighted
```
rck-innovium, line 379: 1. Please add some more text to bring out that this attribute is an hint. For e.g., a NOS can set the type as SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_WEIGHTED and still not give any weights to the members. On the other hand, if the NOS sets the type as SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_UNWEIGHTED, the NOS MUST NOT give any weights to the members.

    _"This attribute is for the NOS to provide a hint that members of this group will be created with weights. When this is set as SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_UNWEIGHTED, it allows the SAI implementation to optimize hardware resources for ECMP."_

2. We need to describe the relationship with SAI_NEXT_HOP_GROUP_ATTR_TYPE.
Example: 
`@validonly SAI_NEXT_HOP_GROUP_ATTR_TYPE == SAI_NEXT_HOP_GROUP_TYPE_DYNAMIC_UNORDERED_ECMP or SAI_NEXT_HOP_GROUP_ATTR_TYPE == SAI_NEXT_HOP_GROUP_TYPE_DYNAMIC_ORDERED_ECMP`

JaiOCP: Hi Ravi,

1) Your first sentence is what we have today. Instead of adding detail for each type in comment (and go out of space) I have added comments in the sai_next_hop_group_members_weight_t stating what each enum means. We can extend that if you think it needs more details.

2) I though about it and this attribute will be valid for almost all the NHG types. For example PROTECTION NHG can be weighted or unweighted as well. Today we do not distinguish for any NHG type that if it can be weighted or unweighted. So I felt that adding a valid would break the backward compatibility.

JaiOCP:
@eddyk-nvidia @rck-innovium Please approve. This PR is next release candidate

inc/sainexthopgroup.h
```diff
@@ -141,6 +141,22 @@ typedef struct _sai_next_hop_group_hw_protection_switchover_notification_data_t
     sai_object_list_t failed_next_hop_groups;
 } sai_next_hop_group_hw_protection_switchover_notification_data_t;
 
+/**
+ * @brief Next hop group is configured with weights or not
+ */
+typedef enum _sai_next_hop_group_members_weight_t
+{
+    /** This is for legacy platforms where this attribute is don't care */
+    SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_DONT_CARE,
```
kcudnik, line 150: maybe change to SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_IGNORED ? or NOT_APPLICABLE or UNSPECIFIED:? this souds more user friendly :)

JaiOCP:
Fair enough :) will change it to unspecified.I was still living in my karnaugh map :)On Aug 12, 2026, at 9:24 PM, Kamil Cudnik ***@***.***> wrote:﻿
@kcudnik commented on this pull request.



In inc/sainexthopgroup.h:
> @@ -141,6 +141,22 @@ typedef struct _sai_next_hop_group_hw_protection_switchover_notification_data_t
     sai_object_list_t failed_next_hop_groups;
 } sai_next_hop_group_hw_protection_switchover_notification_data_t;
 
+/**
+ * @brief Next hop group is configured with weights or not
+ */
+typedef enum _sai_next_hop_group_members_weight_t
+{
+    /** This is for legacy platforms where this attribute is don't care */
+    SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_DONT_CARE,

maybe change to SAI_NEXT_HOP_GROUP_MEMBERS_WEIGHT_IGNORED ? or NOT_APPLICABLE or UNSPECIFIED:? this souds more user friendly :)

inc/sainexthopgroup.h
JaiOCP: Will change it as suggested

JaiOCP:
Opened a new PR
https://github.com/opencomputeproject/SAI/pull/2330

PR #2306, opened 2026-06-25
SAI Switch Temperature Attributes for Sub-degree Precision

bp899348:
SAI Switch Temperature Attributes for Sub-degree Precision

`SAI_SWITCH_ATTR_TEMP_LIST` returns temperature values rounded to whole degrees Celsius. This discards sub-degree precision that is available from the hardware, making it unsuitable for applications that require finer thermal granularity.

### DIFF
--- a/inc/saiswitch.h
+++ b/inc/saiswitch.h
@@ -3617,6 +3617,32 @@ typedef enum _sai_switch_attr_t
     SAI_SWITCH_ATTR_PERFMON_LIST,
 
     /**
+     * @brief List of raw (non-rounded) temperature readings from all sensors.
+     *
+     * Values should be interpreted along with SAI_SWITCH_ATTR_TEMP_BASE_TEN_PRECISION.
+     * Companion to SAI_SWITCH_ATTR_TEMP_LIST (which returns rounded Celsius).
+     *
+     * @type sai_s32_list_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_TEMP_LIST_RAW,
+
+    /**
+     * @brief Power-of-10 exponent for raw temperature precision.
+     *
+     * Defines the scale of values in SAI_SWITCH_ATTR_TEMP_LIST_RAW.
+     * Formula: actual_celsius = raw_value * (10 ^ precision)
+     * Examples:
+     *    precision = 0 => values already in whole Celsius (same as TEMP_LIST)
+     *    precision = -1 => values in 0.1 degrees C steps (445 = 44.5 degrees C)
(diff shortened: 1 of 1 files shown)


inc/saiswitch.h
```diff
@@ -3616,6 +3616,32 @@ typedef enum _sai_switch_attr_t
...
 
+    /**
+     * @brief List of raw (non-rounded) temperature readings from all sensors.
+     *
+     * Values should be interpreted along with SAI_SWITCH_ATTR_TEMP_BASE_TEN_PRECISION.
+     * Companion to SAI_SWITCH_ATTR_TEMP_LIST (which returns rounded Celsius).
+     *
+     * @type sai_s32_list_t
+     * @flags READ_ONLY
+     */
```
jjin62, line 3627: [The commit](https://github.com/opencomputeproject/SAI/commit/37fa50ffe0bbc406b8c161a74da68841c65d93f9) added a new tag @precision. You can define the precision in SAI_SWITCH_ATTR_TEMP_LIST_RAW and no need SAI_SWITCH_ATTR_TEMP_BASE_TEN_PRECISION. You can get the precision from the meta data attr->valuePrecision.
See the examples in https://github.com/opencomputeproject/SAI/blob/master/experimental/saiexperimentalotnoa.h

bp899348: Thank you for the suggestion. The @precision tag is a compile-time constant in the metadata — it is hardcoded at attribute definition time and cannot change at runtime.

In our case, the temperature resolution is hardware and platform dependent and must be queried at runtime. SAI_SWITCH_ATTR_TEMP_BASE_TEN_PRECISION is a READ_ONLY attribute that the driver reports dynamically, allowing consumers to adapt without any code or metadata change as platforms evolve.

So yes, @precision would still be hardcoded — just in a different place.

jjin62: I see. But we can always put a conservative @precision, like 6 to cover all the vendors. Two attributes for one value implies much semantics and is error prone.  Just my 2 cents.

bp899348: I understand the point, but a conservative fixed @precision like 6 means every consumer always multiplies by 10^-6 - even when the hardware only reports tenths of a degree. This would silently produce incorrect values on platforms that don't match that precision, with no way to detect the mismatch at runtime.
With SAI_SWITCH_ATTR_TEMP_BASE_TEN_PRECISION, the driver reports exactly what the hardware delivers — no guessing, no silent error. The cost is one extra attribute query; the benefit is guaranteed correctness across all platforms without any consumer-side assumptions.

JaiOCP (approved):

Issue #1570, opened 2022-08-16
Saistatus.h: saistatus range macros do not handle negative statuses (Linux) well


yohadd:
When using saistatuses that are in range (SAI_STATUS_*_0 - SAI_STATUS_*_MAX) ,

on Linux setup (negative saistatus), the helper MACROs SAI_STATUS_IS_*(x), returns False for statuses in range except for the SAI_STATUS_*_0).

It seems that they are not handling negative statuses well.

For example (in saistatus.h):
#define SAI_STATUS_CODE(_S_)    (-(_S_))

/**
 * @brief Invalid attribute value
 *
 * Range from 0x00020000L to 0x0002FFFFL.
 */
#define SAI_STATUS_INVALID_ATTR_VALUE_0             SAI_STATUS_CODE(0x00020000L)
/**
 * @brief End of invalid attribute value range
 */
#define SAI_STATUS_INVALID_ATTR_VALUE_MAX           SAI_STATUS_CODE(0x0002FFFFL)

/**
 * @brief Is invalid attribute value helper
 */
#define SAI_STATUS_IS_INVALID_ATTR_VALUE(x)     (((x) & (~0xFFFF)) == SAI_STATUS_INVALID_ATTR_VALUE_0)

SAI_STATUS_IS_INVALID_ATTR_VALUE(x) will return False for all negative numbers between  -0x00020001L to -0x0002FFFFL

Thanks in advance.

kcudnik:
this problem was already discussed here: https://github.com/opencomputeproject/SAI/pull/1235 but not addressed

PR #2308, opened 2026-07-02
Fix SAI_STATUS_IS_* attribute-range helpers for negative status codes

### LINKED ISSUE #2307: SAI_STATUS_IS_* attribute range helpers: wrong for negative status codes on Linux, and IS_UNKNOWN_ATTRIBUTE checks the wrong base (all platforms)
There are two separate bugs in the `SAI_STATUS_IS_*` helpers in `inc/saistatus.h`:

- Bug 1 (Linux / non-Windows only): the range mask is wrong for negative status codes.
- Bug 2 (all platforms): `SAI_STATUS_IS_UNKNOWN_ATTRIBUTE` compares against the wrong base.

## Bug 1 - range mask is wrong for negative status codes (Linux only)

`inc/saistatus.h` defines the attribute range helpers by masking the raw status value, e.g.:

```c
#define SAI_STATUS_IS_INVALID_ATTR_VALUE(x)  (((x) & (~0xFFFF)) == SAI_STATUS_INVALID_ATTR_VALUE_0)
```

On non-Windows platforms `SAI_STATUS_CODE(_S_)` is `-(_S_)`, so status codes are negative. Masking the raw negative value is wrong: for a non-zero attribute index the two's-complement borrow crosses the 16-bit boundary, so the masked value matches the wrong range.

Example, `INVALID_ATTR_VALUE` with attribute index 3:

```
code             = SAI_STATUS_CODE(0x20000 + 3) = -0x20003
-0x20003 & ~0xFFFF = -0x30000                   = SAI_STATUS_ATTR_NOT_IMPLEMENTED_0
```

So `SAI_STATUS_IS_INVALID_ATTR_VALUE(-0x20003)` returns false, and `SAI_STATUS_IS_ATTR_NOT_IMPLEMENTED(-0x20003)` returns true. Only attribute index 0 is classified correctly. (On Windows `SAI_STATUS_CODE` is identity, so codes are positive and the original mask works there - this bug is Linux-only.)

## Bug 2 - IS_UNKNOWN_ATTRIBUTE checks the wrong base (all platforms)

`SAI_STATUS_IS_UNKNOWN_ATTRIBUTE` compares against the wrong base, so it never identifies `UNKNOWN_ATTRIBUTE` and aliases `INVALID_ATTRIBUTE` (this is wrong on Windows and Linux alike):

```c
/* current - compares against INVALID_ATTRIBUTE_0 */
#define SAI_STATUS_IS_UNKNOWN_ATTRIBUTE(x)  (((x) & (~0xFFFF)) == SAI_STATUS_INVALID_ATTRIBUTE_0)
```

## Fix (both bugs)

Apply `SAI_STATUS_CODE(x)` first to get the positive magnitude (on Windows it is a no-op; on Linux it flips the sign back to positive), then do the range mask - and fix the unknown-attribute base:

```c
#define SAI_STATUS_IS_INVALID_ATTRIBUTE(x)      ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_INVALID_ATTRIBUTE_0))
#define SAI_STATUS_IS_INVALID_ATTR_VALUE(x)     ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_INVALID_ATTR_VALUE_0))
#define SAI_STATUS_IS_ATTR_NOT_IMPLEMENTED(x)   ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_ATTR_NOT_IMPLEMENTED_0))
#define SAI_STATUS_IS_UNKNOWN_ATTRIBUTE(x)      ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_UNKNOWN_ATTRIBUTE_0))
#define SAI_STATUS_IS_ATTR_NOT_SUPPORTED(x)     ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_ATTR_NOT_SUPPORTED_0))
```

Impact: any consumer that classifies attribute errors (or derives the attribute index from the offset) gets wrong results whenever the offending attribute is not at index 0.

MoriLevinzon:
Fixes #2307

Two fixes in the `SAI_STATUS_IS_*` helpers in `inc/saistatus.h`:

- Compare on the magnitude (`SAI_STATUS_CODE(x)`) before masking, so the range
  check is correct for negative status codes on Linux. Previously a non-zero
  attribute index borrowed across the 16-bit boundary and matched the wrong
  range (e.g. `INVALID_ATTR_VALUE` index 3 classified as `ATTR_NOT_IMPLEMENTED`).
  No-op on Windows, where `SAI_STATUS_CODE` is identity.
- Fix `SAI_STATUS_IS_UNKNOWN_ATTRIBUTE`, which compared against
  `SAI_STATUS_INVALID_ATTRIBUTE_0` instead of `SAI_STATUS_UNKNOWN_ATTRIBUTE_0`
  (wrong on all platforms).

### DIFF
--- a/inc/saistatus.h
+++ b/inc/saistatus.h
@@ -264,27 +264,27 @@
 /**
  * @brief Is invalid attribute helper
  */
-#define SAI_STATUS_IS_INVALID_ATTRIBUTE(x)      (((x) & (~0xFFFF)) == SAI_STATUS_INVALID_ATTRIBUTE_0)
+#define SAI_STATUS_IS_INVALID_ATTRIBUTE(x)      ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_INVALID_ATTRIBUTE_0))
 
 /**
  * @brief Is invalid attribute value helper
  */
-#define SAI_STATUS_IS_INVALID_ATTR_VALUE(x)     (((x) & (~0xFFFF)) == SAI_STATUS_INVALID_ATTR_VALUE_0)
+#define SAI_STATUS_IS_INVALID_ATTR_VALUE(x)     ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_INVALID_ATTR_VALUE_0))
 
 /**
  * @brief Is attribute not implemented helper
  */
-#define SAI_STATUS_IS_ATTR_NOT_IMPLEMENTED(x)   (((x) & (~0xFFFF)) == SAI_STATUS_ATTR_NOT_IMPLEMENTED_0)
+#define SAI_STATUS_IS_ATTR_NOT_IMPLEMENTED(x)   ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_ATTR_NOT_IMPLEMENTED_0))
 
 /**
  * @brief Is unknown attribute helper
  */
-#define SAI_STATUS_IS_UNKNOWN_ATTRIBUTE(x)      (((x) & (~0xFFFF)) == SAI_STATUS_INVALID_ATTRIBUTE_0)
+#define SAI_STATUS_IS_UNKNOWN_ATTRIBUTE(x)      ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_UNKNOWN_ATTRIBUTE_0))
 
 /**
  * @brief Is attribute not supported helper
  */
-#define SAI_STATUS_IS_ATTR_NOT_SUPPORTED(x)     (((x) & (~0xFFFF)) == SAI_STATUS_ATTR_NOT_SUPPORTED_0)
+#define SAI_STATUS_IS_ATTR_NOT_SUPPORTED(x)     ((SAI_STATUS_CODE(x) & (~0xFFFF)) == SAI_STATUS_CODE(SAI_STATUS_ATTR_NOT_SUPPORTED_0))
 
 /**
  * @}
--- a/meta/aspellcheck.pl
+++ b/meta/aspellcheck.pl
@@ -193,6 +193,7 @@ for my $file (@files)
         next if defined $spellExceptions{$word};
 
         next if $word =~ /_/;
+        next if $word =~ /^0[xX][0-9a-fA-F]+$/; # hexadecimal numbers
         next if $word =~ /xYYY+/;
         next if $word =~ /fe\d+/;
         next if $word =~ /ebe\d+/;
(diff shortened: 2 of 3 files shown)


kcudnik:
Related: https://github.com/opencomputeproject/SAI/issues/1570 https://github.com/opencomputeproject/SAI/pull/1235

kcudnik (changes requested):
if you  are changing this, add unittests in sanitycheck for corner cases like suggested here: https://github.com/opencomputeproject/SAI/pull/1235

MoriLevinzon:
Added check_status_attr_ranges() in meta/saisanitycheck.c (called from main), covering the corner cases from #1235 for all five attribute-error ranges:

Encoding value: SAI_STATUS_<X>_0 + SAI_STATUS_CODE(index) == SAI_STATUS_CODE(base + index).
Classification:  each indexed code matches its own SAI_STATUS_IS_* helper and no other (this also covers the SAI_STATUS_IS_UNKNOWN_ATTRIBUTE wrong-base fix).
Boundaries: indices 0, 1, _MAX and _MAX-1 (the MAX-2 case from), plus the codes just below _0 and just above _MAX must not match.
Index round-trip: the decoded attribute index equals the input.
Sentinels must not match: SAI_STATUS_SUCCESS, SAI_STATUS_FAILURE, SAI_STATUS_NOT_EXECUTED, 1, -1.

The test passes with the fixed macros and fails against the current ones, so it guards the fix. Verified it compiles under the meta flags (-Wconversion -Werror).

kcudnik (approved):

kcudnik:
update aspell:

```
WARNING: Word 'xFFFF' is misspelled ./saisanitycheck.c
```

MoriLevinzon:
> update aspell:
> 
> ```
> WARNING: Word 'xFFFF' is misspelled ./saisanitycheck.c
> ```
Added xFFFF to meta/aspell.en.pws

kcudnik:
we could update aspell to automatically exclude all hexadecimal numbers instead of adding each one somewhere here https://github.com/opencomputeproject/SAI/blob/master/meta/aspellcheck.pl#L190

MoriLevinzon:
> we could update aspell to automatically exclude all hexadecimal numbers instead of adding each one somewhere here https://github.com/opencomputeproject/SAI/blob/master/meta/aspellcheck.pl#L190

Updated meta/aspellcheck.pl to skip C-style hexadecimal literals and removed the one-off xFFFF entry from meta/aspell.en.pws, so future hex constants won't need dictionary additions

kcudnik (approved):

kcudnik:
thanks

MoriLevinzon:
Thanks @kcudnik! All checks are green and it's approved — please merge when you can

PR #2309, opened 2026-07-06
Update saidebugcounter.h comments on SRV6 drop counters

benzeerbava:
We believe the comment in 3rd line needs update. {Next Header is not SRH while local SID is configured for packet DA}
    In SRV6 packet, SRH header need not be always present. For example
    a) MicroSID cases
    b) In reduced mode with only one SID


benzeerbava:
resubmitting

PR #2318, opened 2026-07-20
[ACL] Add support for REDIRECT_ORIGINAL_PACKET action

mobinmohan:
Add SAI_ACL_ENTRY_ATTR_ACTION_REDIRECT_ORIGINAL_PACKET action to redirect packets to the destination without modification. This action is used to redirect the original copy of packet to a front panel port/lag.

#### How is this different from SAI_ACL_ACTION_TYPE_REDIRECT ?
Existing SAI_ACL_ACTION_TYPE_REDIRECT allows packets to go through egress editing(like L2/L3 rewrites) after ACL lookup. However, this new action sends packets to a Port or LAG without modifying them. It skips the egress stages like L2/L3 rewrite and VLAN editing, so the headers stay exactly the same as when they were first received by the switch.

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -324,6 +324,9 @@ typedef enum _sai_acl_action_type_t
     /** Bind a TAM object */
     SAI_ACL_ACTION_TYPE_TAM_OBJECT = 0x0000003d,
 
+    /** Set Redirect Original Packet */
+    SAI_ACL_ACTION_TYPE_REDIRECT_ORIGINAL_PACKET = 0x0000003e,
+
     /** Custom range base value */
     SAI_ACL_ACTION_TYPE_CUSTOM_RANGE_BASE = 0x10000000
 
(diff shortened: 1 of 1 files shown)


inc/saiacl.h
```diff
@@ -324,6 +324,9 @@ typedef enum _sai_acl_action_type_t
     /** Bind a TAM object */
     SAI_ACL_ACTION_TYPE_TAM_OBJECT = 0x0000003d,
 
+    /** Set Redirect Original Packet */
+    SAI_ACL_ACTION_TYPE_REDIRECT_ORIGINAL_PACKET = 0x0000003e,
```
AnandhiDhanabalan, line 328: How is this different from `SAI_ACL_ACTION_TYPE_REDIRECT` ?
Would this attr also use `SAI_ACL_ACTION_TYPE_REDIRECT_LIST` as the destination for redirected packets ?

mobinmohan: Existing SAI_ACL_ACTION_TYPE_REDIRECT allows packets to go through egress editing(like L2/L3 rewrites) after ACL lookup. However, this new action sends packets to a Port or LAG without modifying them. It skips the egress stages like L2/L3 rewrite and VLAN editing, so the headers stay exactly the same as when they were first received by the switch.

I have updated PR summary with these details.

##### Would this attr also use SAI_ACL_ACTION_TYPE_REDIRECT_LIST as the destination for redirected packets ?
I found the below lines in saiacl.h for REDIRECT_LIST. Hence a _LIST equivalent is not added for the new action.
```
     * This action is deprecated and will be removed in future release. In order to achieve the functionality,
     * please use redirect action with next hop group or L2MC group or IPMC group.
```

rck-innovium (approved):

inc/saiacl.h
AnandhiDhanabalan: Please capture the action resolution for  parallel lookups and sequential lookups. Also if this is applicable only for ingress, should post ingress be skipped ?

JaiOCP: Can you update the comments.
I believe intent is that ingress packet is not modified at all. L3 as well as L2 data is intact i.e. no TTL changes and/or L2 rewrite.

PR #2319, opened 2026-07-21
Support for running firmware in embedded controllers

JaiOCP:
Modern switches have embedded controller to run firmware for low latency applications.
This PR bring in support for installing and monitoring firmware based on the hardware support.

### DIFF
--- a/doc/fw/SAI-FW-API.md
+++ b/doc/fw/SAI-FW-API.md
@@ -1,0 +1,689 @@
+# Firmware Instances SAI Specification
+-------------------------------------------------------------------------------
+ Title       | SAI Firmware API
+:-------------|:-----------------------------------------------------------------
+ Authors     | Jai Kumar, Broadcom Inc
+ Status      | In review
+ Type        | Standards track
+ Created     | 06/24/2026: Initial Draft
+ SAI-Version | 1.19
+-------------------------------------------------------------------------------
+
+
+## 1.0 Introduction
+Modern switches support one or more embedded controllers for low latency performance sensitive functions. These functions are implemented as embedded software modules aka firmware running on these controllers. There can be one or more set of firmwares running on one or more controllers. This specification introduces set of APIs to manage the firmware running on the controller.
+
+
+## 2.0 Overview
+The SAI Firmware API provides a standardized mechanism for managing firmware instances associated with a switch object. The API supports:
+
+* Firmware object creation
+* Firmware lifecycle management including start and stop of firmware operations
+* Firmware state monitoring
+* Runtime firmware configuration
+
+The API is exposed through the `sai_fw_api_t` method table, using a new FW object.
+
+### 2.1 Firmware Object Lifecycle
+
+There are two life cycles that need to be supported.
+
+#### 2.1.1 Firmware loaded during switch create
+Depending on the functionality provided, a firmware may need to be loaded during the switch creation process itself.
+
+A create only switch attribute SAI_SWITCH_ATTR_FW_LIST is added to specify the details of firmware during switch create time.
+
+NOS will specify the list of firmwares in sai_fw_list_t data structure. Each element in a list specifies
+- name of the firmware with fully the qualified path
+- logfile name with the fully qualified path
+- core id on which it is running. IDs are linear integer name space 0, 1,2 and so on
+- administrative state the firmware
+
+```
+    /**
+     * @brief Firmware list
+     *
+     * @type sai_fw_list_t
+     * @flags CREATE_ONLY
+     * @default internal
+     */
+    SAI_SWITCH_ATTR_FW_LIST,
+
+    /**
+     * @brief Firmware enabled on the switch
+     *
+     * @type sai_object_list_t
+     * @flags READ_ONLY
+     * @objects SAI_OBJECT_TYPE_FW
+     */
+    SAI_SWITCH_ATTR_FW,
+
+    /**
+     * @brief Maximum number of cores supported
+     *
+     * @type sai_uint8_t
+     * @flags READ_ONLY
+     */
+    SAI_SWITCH_ATTR_MAX_FW_CORES,
+```
+
+New data type specifies the necessary information for loading and starting the firware.
+```
+/**
+ * @brief SAI firmware administrative state
+ */
+typedef enum _sai_fw_admin_state_t
+{
+    /** Firmware admin state is automatic loading and running */
+    SAI_FW_ADMIN_STATE_AUTO,
+
+    /** Firmware admin state to start the firmware */
+    SAI_FW_ADMIN_STATE_START_FW,
+
+    /** Firmware admin state is stop the firmware */
+    SAI_FW_ADMIN_STATE_STOP_FW,
+
+    /** Firmware admin state is load the firmware */
+    SAI_FW_ADMIN_STATE_LOAD_FW,
+
+    /** Firmware admin state is unload the firmware */
+    SAI_FW_ADMIN_STATE_UNLOAD_FW,
+} sai_fw_admin_state_t;
+
+/**
+ * @brief Defines a firmware instance
+ */
+typedef struct _sai_fw_inst_t
+{
+    /** Firmware path */
+    sai_s8_list_t fw_path_name;
+
+    /** Firmware log file path */
+    sai_s8_list_t log_path_name;
+
+    /** Firmware core id */
+    uint8_t  core_id;
+
+    /** Firmware admin state */
+    sai_fw_admin_state_t admin_state;
+} sai_fw_inst_t;
+
+/**
+ * @brief Defines a list of firmware instances
+ */
+typedef struct _sai_fw_list_t
+{
+    /** Number of firmware instances */
+    uint32_t count;
+
+    /** List of firmware instances */
+    sai_fw_inst_t *list;
+} sai_fw_list_t;
+```
+
+Sequence of sets to load and start the firmware during switch create:
+
+```text
+Set SAI_SWITCH_ATTR_FW_LIST attribute
+          |
+          v
+    [sai_s8_list_t fw_path_name]
+    [sai_s8_list_t log_path_name]
+    [uint8_t       core_id]
+    [sai_fw_admin_state_t admin_state]
+          |
+          v
+Set Other Switch Attributes
+          |
+          v
+Create Switch Object
+          |
+          v
+SDK Start Firmware based on configured admin state during switch create
+          |
+          v
++---------+---------+---------+---------+
+|         |         |         |         |
+v         v         v         v         v
+Auto    Start      Stop      Load       UnLoad
+ 
+```
+
+> In this workflow SAI allocates the new instance of SAI_OBJECT_TYPE_FW object, based on the attribute SAI_SWITCH_ATTR_FW_LIST. 
+
+#### 2.1.2 Runtime loading of firmware 
+This is the case where system is up and running and a firmware need to be loaded on one of the controllers.
+
+New object type SAI_OBJECT_TYPE_FW is introduced to manage the loading or unloading of firnware instances. NOS must create a SAI_OBJECT_TYPE_FW for each firmware. Based on the admin state configured SAI adapter will either just load the firmware or start it.
+
+```
+typedef enum _sai_fw_attr_t
+{
+    /**
+     * @brief Start of Attributes
+     */
+    SAI_FW_ATTR_START,
+
+    /**
+     * @brief Firmware Log File and Path Name [char[SAI_FW_NAME_SIZE]]
+     *
+     * The maximum number of characters for the name is SAI_FW_NAME_SIZE - 1 since
+     * it needs the terminating null byte ('\0') at the end.
+     *
+     * @type char
+     * @flags CREATE_AND_SET
+     * @default ""
+     */
+    SAI_FW_ATTR_LOG_FILE_AND_PATH_NAME = SAI_FW_ATTR_START,
+
+    /**
+     * @brief Firmware File and Path Name [char[SAI_FW_NAME_SIZE]]
+     *
+     * The maximum number of characters for the name is SAI_FW_NAME_SIZE - 1 since
+     * it needs the terminating null byte ('\0') at the end.
+     *
+     * @type char
+     * @flags CREATE_AND_SET
+     * @default ""
+     */
+    SAI_FW_ATTR_FW_FILE_AND_PATH_NAME,
+
+    /**
+     * @brief Firmware to be loaded on the core
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_FW_ATTR_CORE_ID,
+
+    /**
+     * @brief Firmware admin state
+     *
+     * @type sai_fw_admin_state_t
+     * @flags CREATE_AND_SET
+     * @default SAI_FW_ADMIN_STATE_AUTO
+     */
+    SAI_FW_ATTR_FW_ADMIN_STATE,
+
+    /**
+     * @brief Firmware operational state
+     *
+     * @type sai_fw_op_state_t
+     * @flags READ_ONLY
+     */
+    SAI_FW_ATTR_FW_OP_STATE,
+
+    /**
+     * @brief Firmware major version
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_FW_ATTR_FW_MAJOR_VERSION,
+
+    /**
+     * @brief Firmware minor version
+     *
+     * @type sai_uint32_t
+     * @flags READ_ONLY
+     */
+    SAI_FW_ATTR_FW_MINOR_VERSION,
+
+    /**
+     * @brief End of Performance Monitoring attributes
+     */
+    SAI_FW_ATTR_END,
+
+    /** Custom range base value */
+    SAI_FW_ATTR_CUSTOM_RANGE_START = 0x10000000,
+
+    /** End of custom range base */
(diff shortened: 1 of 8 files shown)


inc/saifw.h
```diff
@@ -0,0 +1,217 @@
...
+     */
+    SAI_FW_ATTR_LOG_FILE_AND_PATH_NAME = SAI_FW_ATTR_START,
+
+    /**
+     * @brief Firmware File and Path Name [char[SAI_FW_NAME_SIZE]]
+     *
+     * The maximum number of characters for the name is SAI_FW_NAME_SIZE - 1 since
+     * it needs the terminating null byte ('\0') at the end.
+     *
+     * @type char
```
kcudnik, line 90: char is refering to sai_attribute_value_t::chardata which is 32 bytes, if you want longer string, use sai_s8_list_t

meta/parse.pl
```diff
@@ -1751,6 +1751,10 @@ sub ProcessDefaultValue
     {
         WriteSource "$val = { .chardata = { 0 } };";
     }
+    elsif ($default =~ /^0$/ and $type =~ /^(sai_fw_inst_t)/)
```
kcudnik, line 1754: type eq "sai_fw_inst_t, you can merge this with 
```"elsif ($default =~ /^0$/ and $type =~ /^(sai_timespec_t)/)"
elsif ($default =~ /^0$/ and $type =~ /^(sai_timespec_t|sai_fw_inst_t)$/)
```

inc/saifw.h
kcudnik: same SAI_FW_ATTR_LOG_FILE_AND_PATH_NAME

kcudnik:
you have only SAI_OBJECT_TYPE_FW as readable variable you cant assign it anywhre, its possible to create any of thos objects ? is that by design ?

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+* Firmware object creation
+* Firmware lifecycle management including start and stop of firmware operations
+* Firmware state monitoring
+* Runtime firmware configuration
+
+The API is exposed through the `sai_fw_api_t` method table, using a new FW object.
+
+### 2.1 Firmware Object Lifecycle
+
+There are two life cycles that need to be supported.
```
gananthateerta, line 29: If firmware crashes, the only way NOS finds out is by polling SAI_FW_ATTR_FW_OP_STATE and noticing it changed to ERROR.

A notification callback, consistent with how SAI handles other async events, is neeed.

Following functionality can be added:
Crash reason?
Core dump?
Auto-restart?

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+     * @brief Firmware File and Path Name [char[SAI_FW_NAME_SIZE]]
+     *
+     * The maximum number of characters for the name is SAI_FW_NAME_SIZE - 1 since
+     * it needs the terminating null byte ('\0') at the end.
+     *
+     * @type char
+     * @flags CREATE_AND_SET
+     * @default ""
+     */
+    SAI_FW_ATTR_FW_FILE_AND_PATH_NAME,
```
AnandhiDhanabalan, line 189: Should this be CREATE_ONLY | MANDATORY ?
When multiple firmware objects are available, how will the NOS be able to map the firmware object to a specific firmware ?

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+
+    /**
+     * @brief Firmware to be loaded on the core
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_FW_ATTR_CORE_ID,
```
AnandhiDhanabalan, line 198: For CORE_ID to be changed - what is the allowed oper and admin state ? How do we establish this requirment of specific admin and operstates when CORE_ID is changed ?

Ideally this can be changed only when SAI_FW_ADMIN_STATE_UNLOAD_FW, other admin state should not allow it. Is this understanding correct ?

**Some additional thoughts..**
The @flags for SAI_FW_ATTR_CORE_ID should be CREATE_ONLY, not CREATE_AND_SET. Changing the core assignment of a firmware object while it is in RUNNING or LOADED state is not meaningful — the F/W is already committed to that core's memory space. To move firmware to a different core, remove and recreate the FW object. If dynamic reassignment is intended (e.g., for load balancing) a new admin state and associated semantics must be defined.

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+     * @default ""
+     */
+    SAI_FW_ATTR_FW_FILE_AND_PATH_NAME,
+
+    /**
+     * @brief Firmware to be loaded on the core
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
```
AnandhiDhanabalan, line 196: How will NOS know supported core-ids ? Should we have a default read-only core-id list ?
Also should we support a `default_internal` option instead of `default`, if NOS wants to go with default option ?

doc/fw/SAI-FW-API.md
AnandhiDhanabalan, line 198: Given core_id can have only single f/w object or multiple ?
If Single - how do you upgrade a f/w from say V1 to V2 - Can you give the flow ? 
Should core_id be kept as key?

The design must address co-running FW constraints. Some implementations compile multiple features into a single F/W image. 

The design doc should also clarify: when two F/W objects share Asic h/w resource, what is the resource arbitration contract? Is it implementation-defined or must it be modeled in SAI?

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+
+## 2.0 Overview
+The SAI Firmware API provides a standardized mechanism for managing firmware instances associated with a switch object. The API supports:
+
+* Firmware object creation
+* Firmware lifecycle management including start and stop of firmware operations
+* Firmware state monitoring
+* Runtime firmware configuration
+
+The API is exposed through the `sai_fw_api_t` method table, using a new FW object.
```
AnandhiDhanabalan, line 25: How will the existing f/w switch attributes co-exist with this new infra ?

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+
+    /** Firmware admin state is stop the firmware */
+    SAI_FW_ADMIN_STATE_STOP_FW,
+
+    /** Firmware admin state is load the firmware */
+    SAI_FW_ADMIN_STATE_LOAD_FW,
+
+    /** Firmware admin state is unload the firmware */
+    SAI_FW_ADMIN_STATE_UNLOAD_FW,
+} sai_fw_admin_state_t;
```
AnandhiDhanabalan, line 91: During warm boot how do we reconnect to the f/w ?
Should we have  a additional admin state like `sync` to indicate warm boot ?

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+
+    /** Firmware log file path */
+    sai_s8_list_t log_path_name;
+
+    /** Firmware core id */
+    uint8_t  core_id;
+
+    /** Firmware admin state */
+    sai_fw_admin_state_t admin_state;
+} sai_fw_inst_t;
```
AnandhiDhanabalan, line 109: Can we have a vendor specific read-only string added to sai_fw_inst_t ? To identify the f/w, type etc or aA firmware role/type attribute is needed to bind a FW object to its consuming SAI subsystem. 

Could you specify how the relationship between SAI_OBJECT_TYPE_FW and any other object in the sai system that uses this firmware object, say SAI_OBJECT_TYPE_TAM or SAI_OBJECT_TYPE_TAM_INT is established ?

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+Once firmware is loaded, NOS need to know its operational state. Operational state is queried using the attribute SAI_FW_ATTR_FW_OP_STATE which is of the data type sai_fw_op_state_t.
+Operational state indicates the firmware state.
+
+| State                      | Description                     |
+| -------------------------- | ------------------------------- |
+| SAI_FW_OP_STATE_LOADED     | Firmware loaded but not started |
+| SAI_FW_OP_STATE_NOT_LOADED | Firmware unavailable and not loaded|
+| SAI_FW_OP_STATE_RUNNING    | Firmware loaded and running     |
+| SAI_FW_OP_STATE_STOPPED    | Firmware loaded but stopped     |
+| SAI_FW_OP_STATE_ERROR      | Firmware loaded and encountered an error   |
```
AnandhiDhanabalan, line 279: For ASICs with async FW loading, transitional states are needed:

```
SAI_FW_OP_STATE_LOADING,   // image transfer in progress 
SAI_FW_OP_STATE_STOPPING,  // graceful shutdown in progress

```
Without these, the NOS has no way to distinguish "FW load not started" from "FW load in progress" when polling SAI_FW_ATTR_FW_OP_STATE. Also consider adding a notification callback (SAI_SWITCH_ATTR_FW_STATE_CHANGE_NOTIFY) so the NOS doesn't need to poll.

doc/fw/SAI-FW-API.md
JaiOCP: Existing Firmware attributes are very specific to Gearbox with MDIO/EEPROM load methods and can support only single firmware instance.
These attributes are not deprecated and stay such.

NOS can choose existing or the new infra based on the capability query for the device and both can be enabled at the same time if supported.

doc/fw/SAI-FW-API.md
JaiOCP: Accepted the feedback. Will add a callback

doc/fw/SAI-FW-API.md
JaiOCP: I am not clear what reconnect means? After warmboot NOS will query the admin state and op state to figure out the state of FW.

If reconnect means some kind of RPC channel then note that NOS do not open any RPC channel. If there is one it is internal to the chip vendor.

doc/fw/SAI-FW-API.md
JaiOCP: Vendor specific request was discussed in meeting. If there is a need to specify any kind of hint for SDK to pick on for some vendor specific operations that can be encoded in the FW file name.


SAI_OBJECT_TYPE_FW  has a naroow focus of manaigng the life cycle of the firmware. What kind of functionality it exposes is not part of this PR.

Typically any function either by the FW or the native switch silicon is abstracted independently as a SAI API. For example TAM functions can be implemented using a firmware. In this firmware life cycle management is done by the FW object and function is exposed using the TAM object. There is no coupling between two within the SAI framework.

Good question though :)

doc/fw/SAI-FW-API.md
JaiOCP: I was thinking if NOS need to modify the file name for some new encoding we should allow it run time and hence not mandatory.

doc/fw/SAI-FW-API.md
JaiOCP: There is a readonly attribute for MAX cores. NOS can ping a FW on an abstracted core no of 0 to max-1. This can internally be mapped by the SAI adapter.

Reason I kept it this way is mainly to provide flexibility to silicon vendor to internally id the cores as applicable. Note that cores may not be in single domain space as we talked about.

doc/fw/SAI-FW-API.md
JaiOCP: This is a good point. Lets discuss this in meeting

JaiOCP: FW is an abstracted entity and may support one or more features. As I mentioned that features are exposed via separate SAI APIs and have no correlation to FW object.

If FW need to migrated to different core, then it is a destructive event. Ideally FW object should be deleted and recreated with different core id (this is where your suggestion of CREATE_ONLY for core id is valid). And this will impact all the features supported by that instance of FW.

doc/fw/SAI-FW-API.md
JaiOCP: I agree.

inc/saifw.h
```diff
@@ -0,0 +1,217 @@
...
+    SAI_FW_OP_STATE_STOPPED,
+
+    /** Firmware operational state is in error */
+    SAI_FW_OP_STATE_ERROR,
+} sai_fw_op_state_t;
+
+/**
+ * @brief Defines maximum length of firmware and log file name
+ */
+#define SAI_FW_NAME_SIZE 64
```
eladnachman, line 60: Why is the limited set at 64 bytes? According to the PR documentation this is used for both the path and the file name. Under complex Linux mounting, this limit might not be enough. Consider enlarging to at least 128 if not 256.

inc/saifw.h
JaiOCP: length is char[SAI_FW_NAME_SIZE] and is used elsewhere as well to specify managed strings of specific lengths.

meta/parse.pl
JaiOCP: accepted

JaiOCP:
> you have only SAI_OBJECT_TYPE_FW as readable variable you cant assign it anywhre, its possible to create any of thos objects ? is that by design ?

This is by design. SAI adapter will create these objects that should be readable.

doc/fw/SAI-FW-API.md
JaiOCP: Agreed, Changing the flag the CREATE_ONLY will make things cleaner where if FW need to be spawned on a new core id then the NOS need to do first clean up i.e. stop the fw, delete the fw object and create a new fw object with new core id.

inc/saifw.h
JaiOCP: Agree. I will change the size to 256.

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+     */
+    SAI_FW_ATTR_START,
+
+    /**
+     * @brief Firmware Log File and Path Name [char[SAI_FW_NAME_SIZE]]
+     *
+     * The maximum number of characters for the name is SAI_FW_NAME_SIZE - 1 since
+     * it needs the terminating null byte ('\0') at the end.
+     *
+     * @type char
```
crameshk, line 173: Should this be sai_s8_list_t  like line 99 above for fw_path_name variable.
```
    /** Firmware path */
    sai_s8_list_t fw_path_name;
```

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+     */
+    SAI_FW_ATTR_LOG_FILE_AND_PATH_NAME = SAI_FW_ATTR_START,
+
+    /**
+     * @brief Firmware File and Path Name [char[SAI_FW_NAME_SIZE]]
+     *
+     * The maximum number of characters for the name is SAI_FW_NAME_SIZE - 1 since
+     * it needs the terminating null byte ('\0') at the end.
+     *
+     * @type char
```
crameshk, line 185: Same: Should this be sai_s8_list_t  like line 102 above for fw_path_name variable.
```
    /** Firmware log file path */
    sai_s8_list_t log_path_name;
```

doc/fw/SAI-FW-API.md
```diff
@@ -0,0 +1,635 @@
...
+    SAI_SWITCH_ATTR_MAX_FW_CORES,
+```
+
+New data type specifies the necessary information for loading and starting the firware.
+```
+/**
+ * @brief SAI firmware administrative state
+ */
+typedef enum _sai_fw_admin_state_t
+{
```
AvivAgam10, line 76: Is the expectation that every SAI implementation must support all admin-state transitions (LOAD/UNLOAD/START/STOP) independently, or is atomic load-and-run the only mandatory one and the rest optional? 
If the latter, the spec needs a capability attribute to advertise which transitions a platform supports, plus a standard "not supported" status — otherwise a NOS can't tell an unsupported mode from a real failure.

doc/fw/SAI-FW-API.md
JaiOCP: I will bring this change via new PR given that this well contained change.
As discussed plan is to use the SAI_SWITCH_ATTR_SWITCH_ASIC_SDK_HEALTH_EVENT_NOTIFY callback.

doc/fw/SAI-FW-API.md
JaiOCP: Added a section talking about existing and new fw infratrcutrue.

doc/fw/SAI-FW-API.md
JaiOCP: As we discussed in meeting. Any kind of RPC connection if need to be maintained is maintained/abstracted within the SDK and is not exposed outside.

doc/fw/SAI-FW-API.md
JaiOCP: Its a char array. Increased the size of array to 256 bytes as suggested in other review comment

doc/fw/SAI-FW-API.md
JaiOCP: Please see previous comment

doc/fw/SAI-FW-API.md
JaiOCP: sai_query_attribute_enum_values_capability() can be used to query what values are support admin state. I will add this to documentation.

JaiOCP:
https://github.com/opencomputeproject/SAI/pull/2331 is opened.

PR #2320, opened 2026-07-21
Add label attribute for ACL entry

shri-khare:
Summary:

This is similar to #1158, #1407, $1430.

Those PRs added label attribute for LAG/virtual router, counter, ACL Counter. This PR adds similar label attribute for ACL entries.

ACL entries have PRIORITY attribute, but since multiple ACL entries can be configured with same priority, PRIRORITY attribute cannot be used to uniquely identlfy an ACL entry within a given ACL Table.

Adding a label attribute that can be used to uniquely identify ACL entry object during warmboot. This attribute is consider as user data attached to the object.

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -1850,6 +1850,15 @@ typedef enum _sai_acl_entry_attr_t
      */
     SAI_ACL_ENTRY_ATTR_ADMIN_STATE,
 
+    /**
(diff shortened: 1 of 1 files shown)


shri-khare:
Discarding in favor of https://github.com/opencomputeproject/SAI/pull/2322

PR #2322, opened 2026-07-24
Add label attribute for ACL entry

zechengh09:
Summary:

This is similar to #1158, #1407, #1430.

Those PRs added label attribute for LAG/virtual router, counter, ACL Counter. This PR adds similar label attribute for ACL entries.

Adding a label attribute that can be used to uniquely identify ACL entry object during warmboot. This attribute is considered as user data attached to the object.

### DIFF
--- a/inc/saiacl.h
+++ b/inc/saiacl.h
@@ -1850,6 +1850,15 @@ typedef enum _sai_acl_entry_attr_t
      */
     SAI_ACL_ENTRY_ATTR_ADMIN_STATE,
 
+    /**
+     * @brief Attribute used to uniquely identify ACL entry.
+     *
+     * @type sai_s8_list_t
+     * @flags CREATE_AND_SET
+     * @default empty
+     */
+    SAI_ACL_ENTRY_ATTR_LABEL,
+
     /*
      * Match fields [sai_acl_field_data_t]
      * - Mandatory to pass at least one field during ACL Rule creation.


kcudnik (approved):

inc/saiacl.h
```diff
@@ -1850,6 +1850,15 @@ typedef enum _sai_acl_entry_attr_t
...
 
+    /**
+     * @brief Attribute used to uniquely identify ACL entry.
+     *
+     * @type sai_s8_list_t
+     * @flags CREATE_AND_SET
+     * @default empty
+     */
+    SAI_ACL_ENTRY_ATTR_LABEL,
+
```
JaiOCP, line 1861: Why the data type is specified as s8_list_t instead of char?

shri-khare: The SAI spec limits char datatype to 32-bytes. In practice, we have seen cases where the label length may exceed this limit. For example, we may want to capture the use case for the ACL entry in the label name itself, and that could exceed 32-bytes. 

sai_s8_list_t keeps it generic, so NOS can choose the desired label length without being constrained by the 32-byte limits.

kcudnik: 32 bytes for label is i think long enough, so far almost all labels are chars, and this should be a label not poem

JaiOCP (approved):

tjchadaga (approved):

PR #2324, opened 2026-07-24
Feature: PTP Peer-to-Peer (P2P) Path Delay Support

gurprem-singh:
Adds the headers and design specification for the PTP Pdelay configuration object, expanding SAI capabilities to configure localized hardware measurement profiles for link propagation delays.


meta/saiserialize.c
```diff
@@ -1355,3 +1355,49 @@ int sai_deserialize_attribute(
...
+
+    SAI_META_LOG_WARN("failed to deserialize '%.*s' as PTP clock-id", MAX_CHARS_PRINT, buffer);
+    return SAI_SERIALIZE_ERROR;
+}
+
+int sai_serialize_ptp_l2_header(
+        _Out_ char *buffer,
+        _In_ const sai_ptp_l2_header_t header)
+{
+    return 0;
```
kcudnik: missing serialization method

meta/saiserialize.c
```diff
@@ -1355,3 +1355,49 @@ int sai_deserialize_attribute(
...
+}
+
+int sai_deserialize_ptp_l2_header(
+        _In_ const char *buffer,
+        _Out_ sai_ptp_l2_header_t header)
+{
+    return 0;
```
kcudnik: missing serialization method

meta/saiserialize.c
```diff
@@ -1355,3 +1355,49 @@ int sai_deserialize_attribute(
 
     return (int)(buf - buffer);
 }
+
+int sai_serialize_ptp_clock_id(
```
kcudnik: if you are adding serialize and dserialize, then add also unittest for that

meta/saiserialize.c
```diff
@@ -1355,3 +1355,49 @@ int sai_deserialize_attribute(
...
+            &arr[0], &arr[1], &arr[2], &arr[3], &arr[4], &arr[5], &arr[6], &arr[7], &read);
+
+    if (n == 8 && sai_serialize_is_char_allowed(buffer[read]))
+    {
+        for (n = 0; n < 8; n++)
+        {
+            id[n] = (uint8_t)arr[n];
+        }
+
+        return 8;
```
kcudnik: it should return "read" variable, number of consumed characters, take a look at sai_deserialize_mac and sai_deserialize_object_id

meta/saisanitycheck.c
```diff
@@ -6321,7 +6339,7 @@ void check_struct_and_union_size()
     CHECK_STRUCT_SIZE(sai_acl_action_parameter_t, 24);
     CHECK_STRUCT_SIZE(sai_acl_field_data_data_t, 16);
     CHECK_STRUCT_SIZE(sai_acl_field_data_mask_t, 16);
-    CHECK_STRUCT_SIZE(sai_attribute_value_t, 40);
+    CHECK_STRUCT_SIZE(sai_attribute_value_t, 64);
```
kcudnik: this is breaking (not backward compatible) change as attribute value size is changing

kcudnik (changes requested):
address comments

meta/saisanitycheck.c
gurprem-singh: fixed to use sai_u8_list_t instead

meta/saiserialize.c
gurprem-singh: custom serializer removed. applies to other comments as well.

gurprem-singh:
Thanks for review @kcudnik. Latest patch is updated so both attributes now use the standard sai_u8_list_t instead. This keeps the union at 40 bytes, and since it's a natively-supported type, it no longer needs any custom serialize/deserialize methods or their unit tests.

inc/saiport.h
```diff
@@ -3332,6 +3332,69 @@ typedef enum _sai_port_attr_t
...
+     *
+     * This attribute configures the link delay to be used while computing the correction
+     * field in the incoming PTP packet. It can either be computed by the peer delay
+     * mechanism or by NOS.
+     *
+     * @type sai_int64_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_PORT_ATTR_LINK_DELAY,
```
JaiOCP, line 3398: Please specify the unit in comments

inc/saiptppdelay.h
```diff
@@ -0,0 +1,262 @@
...
+     * @default SAI_PTP_PROTOCOL_IEEE8023
+     */
+    SAI_PTP_PDELAY_ATTR_PTP_PROTOCOL = SAI_PTP_PDELAY_ATTR_START,
+
+    /**
+     * @brief MAC address of the port, to be used as source address in peer delay packet
+     *
+     * @type sai_mac_t
+     * @flags CREATE_AND_SET
+     * @default vendor
```
JaiOCP, line 88: Add comments explaining how default as vendor is useful

inc/saiptppdelay.h
```diff
@@ -0,0 +1,262 @@
...
+    SAI_PTP_PDELAY_ATTR_SRC_MAC,
+
+    /**
+     * @brief Peer delay L2 header length.
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_PTP_PDELAY_ATTR_L2_HEADER_LEN,
```
JaiOCP, line 99: in Bytes or Words ?

inc/saiptppdelay.h
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * @brief Peer delay L2 header.
+     *
+     * @type sai_u8_list_t
+     * @flags CREATE_AND_SET
+     * @default empty
+     */
+    SAI_PTP_PDELAY_ATTR_PTP_L2_HEADER,
```
JaiOCP, line 111: Add details that this header length is specified by the L2_HEADER_LEN attribute

inc/saiswitch.h
```diff
@@ -3616,6 +3616,34 @@ typedef enum _sai_switch_attr_t
...
+
+    /**
+     * @brief Maximum number of ports enabled for peer delay.
+     *
+     * @type sai_uint16_t
+     * @flags CREATE_AND_SET
+     * @isvlan false
+     * @default 0
+     */
+    SAI_SWITCH_ATTR_PTP_PDELAY_MAX_PORTS,
```
JaiOCP, line 3672: Should there be a READ attribute for platform supporting max number of ports that can be enabled for peer delay?

doc/PTP/SAI-Proposal-PTP-Pdelay.md
```diff
@@ -0,0 +1,267 @@
...
+
+### SAI_PORT_ATTR_PDELAY_INSTANCE_ID
+
+- Value Type: sai_object_id_t
+- Flags: CREATE_AND_SET
+- Description: Associates the port with a specific configuration profile.
+  Setting this to `SAI_NULL_OBJECT_ID` detaches the profile and disables P2P
+  processing on the physical interface.
+
+### SAI_PORT_ATTR_LINK_DELAY
```
gmuthukrishn, line 146: We already have SAI_PORT_ATTR_PTP_PEER_MEAN_PATH_DELAY. How would this and PDELAY_LINK_DELAY below differ from the existing one ?.

inc/saiport.h
```diff
@@ -3332,6 +3332,69 @@ typedef enum _sai_port_attr_t
...
+     * @brief Raw Link Delay, as calculated by the peer delay mechanism
+     *
+     * This attribute is applicable only when PTP peer delay functionality is enabled.
+     * When PTP peer delay functionality is disabled, 0 should be returned.
+     *
+     * @type sai_int64_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_PORT_ATTR_PDELAY_LINK_DELAY,
```
eddyk-nvidia, line 3360: how is it related to SAI_PORT_ATTR_LINK_DELAY below ?

inc/saiport.h
```diff
@@ -3332,6 +3332,69 @@ typedef enum _sai_port_attr_t
...
+     * This attribute configures the link delay to be used while computing the correction
+     * field in the incoming PTP packet. It can either be computed by the peer delay
+     * mechanism or by NOS.
+     *
+     * @type sai_int64_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
+    SAI_PORT_ATTR_LINK_DELAY,
+
```
eddyk-nvidia, line 3399: how is it related to SAI_PORT_ATTR_PDELAY_LINK_DELAY ?

inc/saiptppdelay.h
```diff
@@ -0,0 +1,262 @@
...
+     */
+    SAI_PTP_PDELAY_ATTR_INGRESS_VLAN_ID,
+
+    /**
+     * @brief Peer delay Packets TTL value
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 1
+     */
```
eddyk-nvidia, line 140: add validation for UDP only ?

inc/saiptppdelay.h
```diff
@@ -0,0 +1,262 @@
...
+    SAI_PTP_PDELAY_ATTR_TTL,
+
+    /**
+     * @brief Peer delay Packets IP DSCP value
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     */
```
eddyk-nvidia, line 150: add validation for UDP only ?

inc/saiswitch.h
```diff
@@ -3616,6 +3616,34 @@ typedef enum _sai_switch_attr_t
...
     SAI_SWITCH_ATTR_PERFMON_LIST,
 
+    /**
+     * @brief Default PTP clock identity
+     *
+     * @type sai_u8_list_t
+     * @flags CREATE_AND_SET
+     * @default empty
+     */
+    SAI_SWITCH_ATTR_CLOCK_ID,
```
eddyk-nvidia, line 3662: what should be set here ? How is it going to be used ? In PTP header ?

tvpradeep-mrvl (changes requested):
https://github.com/gurprem-singh/SAI/blob/d7ba67347b04da77b741bd933730674184837c74/doc/PTP/SAI-Proposal-PTP-Pdelay.md?plain=1#L190
Ideally not the global clock instance, but specific to peer delay messages. Description can be updated accordingly. There is already SAI_SWITCH_ATTR_PORT_PTP_MODE towards global clock operation for PTP event messages.

https://github.com/gurprem-singh/SAI/blob/d7ba67347b04da77b741bd933730674184837c74/doc/PTP/SAI-Proposal-PTP-Pdelay.md?plain=1#L167
What is the difference between SAI_PORT_ATTR_PDELAY_LINK_DELAY and SAI_PORT_ATTR_PDELAY_NEIGHBOR_PROPAGATION_DELAY

https://github.com/gurprem-singh/SAI/blob/d7ba67347b04da77b741bd933730674184837c74/doc/PTP/SAI-Proposal-PTP-Pdelay.md?plain=1#L130
To specify how this inter operates with SAI_HOSTIF_TRAP_TYPE_PTP_PEER_DELAY. If Pdelay measurement is offloaded to hardware then the packet should be terminated and trap should be disabled for that port.

inc/saiport.h
gurprem-singh: SAI_PORT_ATTR_PDELAY_LINK_DELAY is a READ_ONLY attribute providing the raw, instantaneous link delay measurement computed by the hardware peer-delay engine from the 4 timestamps exchange.

SAI_PORT_ATTR_LINK_DELAY was intended as the CREATE_AND_SET attribute used by the control plane/ userspace application to program the filtered steady-state link delay back into the port datapath pipeline to adjust the Correction Field (CF) on transiting PTP event frames.

inc/saiport.h
gurprem-singh: as above ^

inc/saiswitch.h
gurprem-singh: This attribute configures the switch's 8-octet IEEE 1588 ClockIdentity (in network byte order).
It is populated into the sourcePortIdentity.clockIdentity field of all transmitted PTP peer delay frames (Pdelay_Req, Pdelay_Resp, and Pdelay_Resp_Follow_Up) emitted by the hardware peer-delay engine. Updated the docstring to clarify.

kcudnik (approved):

doc/PTP/SAI-Proposal-PTP-Pdelay.md
gurprem-singh: SAI_PORT_ATTR_LINK_DELAY is specifically defined as sai_int64_t for alignment with IEEE 1588:

Standard Alignment: IEEE 1588  defines TimeInterval as a signed 64-bit quantity, and the peer mean path delay (peerMeanPathDelay in 1588-2008, meanLinkDelay in 1588-2019) uses that type. sai_int64_t matches its signedness and width. Although this attribute is in whole nanoseconds, not TimeInterval's 2^-16 ns units, the width just leaves headroom to define fractional scaling later without an ABI break.

An asymmetry-corrected link delay can legitimately be negative, which sai_uint32_t cannot represent.

On the apparent inconsistency with SAI_PORT_ATTR_PDELAY_NEIGHBOR_PROPAGATION_DELAY being sai_uint32_t  these differ deliberately: the propagation delay is a measured magnitude, non-negative by construction, while LINK_DELAY is a control-plane input that may carry an operator-supplied asymmetry correction.

Recommend keeping SAI_PORT_ATTR_LINK_DELAY (sai_int64_t) to represent the IEEE-compliant signed 64-bit delay.

gurprem-singh:
> https://github.com/gurprem-singh/SAI/blob/d7ba67347b04da77b741bd933730674184837c74/doc/PTP/SAI-Proposal-PTP-Pdelay.md?plain=1#L190 Ideally not the global clock instance, but specific to peer delay messages. Description can be updated accordingly. There is already SAI_SWITCH_ATTR_PORT_PTP_MODE towards global clock operation for PTP event messages.
> 
> https://github.com/gurprem-singh/SAI/blob/d7ba67347b04da77b741bd933730674184837c74/doc/PTP/SAI-Proposal-PTP-Pdelay.md?plain=1#L167 What is the difference between SAI_PORT_ATTR_PDELAY_LINK_DELAY and SAI_PORT_ATTR_PDELAY_NEIGHBOR_PROPAGATION_DELAY
> 
> https://github.com/gurprem-singh/SAI/blob/d7ba67347b04da77b741bd933730674184837c74/doc/PTP/SAI-Proposal-PTP-Pdelay.md?plain=1#L130 To specify how this inter operates with SAI_HOSTIF_TRAP_TYPE_PTP_PEER_DELAY. If Pdelay measurement is offloaded to hardware then the packet should be terminated and trap should be disabled for that port.

Response: 
Updated the doc string for the first two points. 
On SAI_HOSTIF_TRAP_TYPE_PTP_PEER_DELAY: When hardware P2P offload is active on a port (SAI_PORT_ATTR_PDELAY_INSTANCE_ID is confiugred), incoming peer delay frames are processed and terminated directly by the hardware engine, and trapping to host CPU via SAI_HOSTIF_TRAP_TYPE_PTP_PEER_DELAY is suppressed on that port.
If offload is disabled (PDELAY_INSTANCE_ID is SAI_NULL_OBJECT_ID), packets can follow standard host interface pipeline processing and can be trapped to the CPU.

JaiOCP (approved):
LGTM

gurprem-singh:
> https://github.com/gurprem-singh/SAI/blob/d7ba67347b04da77b741bd933730674184837c74/doc/PTP/SAI-Proposal-PTP-Pdelay.md?plain=1#L190 Ideally not the global clock instance, but specific to peer delay messages. Description can be updated accordingly. There is already SAI_SWITCH_ATTR_PORT_PTP_MODE towards global clock operation for PTP event messages.
> 
> https://github.com/gurprem-singh/SAI/blob/d7ba67347b04da77b741bd933730674184837c74/doc/PTP/SAI-Proposal-PTP-Pdelay.md?plain=1#L167 What is the difference between SAI_PORT_ATTR_PDELAY_LINK_DELAY and SAI_PORT_ATTR_PDELAY_NEIGHBOR_PROPAGATION_DELAY
> 
> https://github.com/gurprem-singh/SAI/blob/d7ba67347b04da77b741bd933730674184837c74/doc/PTP/SAI-Proposal-PTP-Pdelay.md?plain=1#L130 To specify how this inter operates with SAI_HOSTIF_TRAP_TYPE_PTP_PEER_DELAY. If Pdelay measurement is offloaded to hardware then the packet should be terminated and trap should be disabled for that port.

@tvpradeep-mrvl, thanks for the review! Here are clarifications:

1. Clock Identity scoping: as per IEEE 1588-2019 (defaultDS.clockIdentity), clockIdentity belongs to the PTP Instance rather than an individual port. It is populated into the sourcePortIdentity.clockIdentity field of all transmitted peer delay messages (Pdelay_Req / Resp / Resp_Follow_Up) originated by the hardware engine.

2. PDELAY_LINK_DELAY vs. PDELAY_NEIGHBOR_PROPAGATION_DELAY:

SAI_PORT_ATTR_PDELAY_LINK_DELAY: Raw, instantaneous one-shot calculation directly from the latest timestamp exchange: [(t4 - t1) - (t3 - t2)] / 2.
SAI_PORT_ATTR_PDELAY_NEIGHBOR_PROPAGATION_DELAY: Filtered/smoothed propagation delay incorporating the calculated Neighbor Rate Ratio (NRR), providing steady-state delay for datapath consumption.

I've added a subsection to the design doc covering how these two relate to each other and to SAI_PORT_ATTR_LINK_DELAY.

doc/PTP/SAI-Proposal-PTP-Pdelay.md
```diff
@@ -0,0 +1,315 @@
...
+
```
eddyk-nvidia, line 147: Please consider splitting it to TX and RX link delay. Can be different

doc/PTP/SAI-Proposal-PTP-Pdelay.md
gurprem-singh: Make it explicit about the use of attributes [when PDELAY is enabled] in design doc

AnandhiDhanabalan:
@gurprem-singh, As discussed in the meeting, kindly update the HLD to clarify that the existing model is not deprecated. The new model is applicable only when SAI_PORT_ATTR_PDELAY_INSTANCE_ID is defined; otherwise, the existing PTP model will continue to be used.
Additionally, please explicitly specify which switch- and port-level attributes apply to both models, and clearly identify any attributes that will not take effect when the new model is enabled.

gurprem-singh:
> @gurprem-singh, As discussed in the meeting, kindly update the HLD to clarify that the existing model is not deprecated. The new model is applicable only when SAI_PORT_ATTR_PDELAY_INSTANCE_ID is defined; otherwise, the existing PTP model will continue to be used.

Thanks @AnandhiDhanabalan Section 2.5 of the design doc has been updated to explicitly clarify that the existing PTP model is not deprecated and remains active when SAI_PORT_ATTR_PDELAY_INSTANCE_ID is not configured.

doc/PTP/SAI-Proposal-PTP-Pdelay.md
gurprem-singh: updated the design doc as per comments during discussion

AnandhiDhanabalan:
> Thanks @AnandhiDhanabalan Section 2.5 of the design doc has been updated to explicitly clarify that the existing PTP model is not deprecated and remains active when SAI_PORT_ATTR_PDELAY_INSTANCE_ID is not configured.

Additionally, please explicitly specify which switch- and port-level attributes apply to both models, and clearly identify any attributes that will not take effect when the new model is enabled.

gurprem-singh:
> > Thanks @AnandhiDhanabalan Section 2.5 of the design doc has been updated to explicitly clarify that the existing PTP model is not deprecated and remains active when SAI_PORT_ATTR_PDELAY_INSTANCE_ID is not configured.
> 
> Additionally, please explicitly specify which switch- and port-level attributes apply to both models, and clearly identify any attributes that will not take effect when the new model is enabled.

Please check the latest udpate @AnandhiDhanabalan. Thanks for pointing it out, let me know if further details need to be added.

gurprem-singh:
@AnandhiDhanabalan @JaiOCP @eddyk-nvidia 
Gently checking in on the review status of the PR. Please let us know if there are any remaining questions or if this looks good to go.

tjchadaga:
@tvpradeep-mrvl - please help complete the review on this.

PR #2325, opened 2026-07-26
Added support for supported loopback mode for port object

pdhruv-marvell:
This PR introduces a new port attribute: SAI_PORT_ATTR_SUPPORTED_LOOPBACK_MODE

This attribute provides a list of loopback modes supported by the underlying hardware for a given port.
These values represent the loopback modes that can be configured on a port, such as no loopback, PHY loopback, MAC loopback, or remote loopback modes.

Motivation
----------
Different ports on the same ASIC may support different loopback modes based on hardware and SerDes configuration. sai_query_attribute_enum_values_capability() cannot represent this accurately because it is switch-scoped and takes switch_id, not a port OID. Therefore, a per-port READ_ONLY attribute is needed to report the loopback modes supported by each port.
This enhancement allows applications to:

- Avoid configuration failures due to unsupported loopback modes

- Enable runtime validation before applying SAI_PORT_ATTR_LOOPBACK_MODE

Example
-------
A port may report:
[SAI_PORT_LOOPBACK_MODE_NONE,
SAI_PORT_LOOPBACK_MODE_PHY,
SAI_PORT_LOOPBACK_MODE_MAC,
SAI_PORT_LOOPBACK_MODE_PHY_REMOTE,
SAI_PORT_LOOPBACK_MODE_MAC_REMOTE,],
indicating all loopback modes are supported.

or
[SAI_PORT_LOOPBACK_MODE_NONE,
SAI_PORT_LOOPBACK_MODE_PHY,
SAI_PORT_LOOPBACK_MODE_MAC],
indicating remote loopback is not supported on that port.


inc/saiport.h
```diff
@@ -3481,6 +3481,14 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_APSU_ISL_READY,
 
+   /**
+    * @brief Query list of supported port loopback modes for the port
+    *
+    * @type sai_s32_list_t sai_port_loopback_mode_t
+    * @flags READ_ONLY
+    */
+    SAI_PORT_ATTR_SUPPORTED_LOOPBACK_MODE,
```
JaiOCP, line 3490: How would the local port know the remote end behavior?
    /** Port loopback at PHY remote end */
    SAI_PORT_LOOPBACK_MODE_PHY_REMOTE,

    /** Port loopback at MAC remote end */
    SAI_PORT_LOOPBACK_MODE_MAC_REMOTE

pdhruv-marvell: @JaiOCP 
The remote loopback mode is configured on the local port, not on the remote device. When enabled, the local port loops back traffic received from the remote link partner back toward that same link partner, effectively emulating loopback behavior at the remote end.
For additional context, this behavior was introduced as part of the original remote loopback mode contribution: https://github.com/opencomputeproject/SAI/pull/1234

tjchadaga:
@JaiOCP - could you please help sign off on this?

PR #2326, opened 2026-08-04
Add VLAN-aware MLD Snooping support

AnandhiDhanabalan:
Adds VLAN-level MLD snooping support to SAI by introducing `SAI_VLAN_ATTR_MLD_SNOOPING_ENABLE`. This enables independent control of IPv6 multicast (MLD) snooping alongside existing IGMP (IPv4) snooping.

### DIFF
--- a/doc/SAI-Proposal-MLD-Snooping.md
+++ b/doc/SAI-Proposal-MLD-Snooping.md
@@ -1,0 +1,184 @@
+ VLAN MLD Snooping Support
+-------------------------------------------------------------------------------
+ Title       | VLAN MLD Snooping Support
+:-------------|:-----------------------------------------------------------------
+ Authors     | Anandhi Dhanabalan, Marvell Technology Inc 
+ Status      | In review
+ Type        | Standards track
+ Created     | 08/04/2026: Initial Draft
+ -------------------------------------------------------------------------------
+
+
+## 1. Overview
+
+This enhancement adds support for MLD (Multicast Listener Discovery)
+snooping control at the VLAN level. MLD is the IPv6 equivalent of IGMP,
+used for managing IPv6 multicast group memberships. As networks transition
+to IPv6 and dual-stack deployments, the ability to control MLD snooping
+independently is essential for efficient multicast traffic management.
+
+
+
+## 2. Motivation
+
+### Problem
+
+- No independent IPv6 multicast snooping control.
+
+### Current SAI State
+
+- Existing SAI VLAN attribute provides per-VLAN IGMP snooping control.
+- No corresponding per-VLAN MLD snooping control exists.
+- IPv4 and IPv6 multicast snooping cannot be configured independently.
+
+This proposal addresses these limitations by introducing independent MLD
+snooping control.
+
+
+
+# 3. Proposed SAI Attribute Changes
+
+## 3.1 New VLAN Attribute
+
+```c
+    /**
+     * @brief MLD Snooping enable or disable control for VLAN
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_VLAN_ATTR_MLD_SNOOPING_ENABLE,
+```
+
+# 4. Behavioral Model
+
+## 4.1 MLD Snooping Operation
+
+When VLAN MLD snooping is enabled, MLD protocol messages are delivered to the CPU according to
+the configured MLD traps (see Section 4.1.1). The control plane learns multicast
+listener membership and programs multicast forwarding entries in hardware.
+The interaction between VLAN MLD snooping and global MLD HOSTIF traps is
+described in Section 4.2.
+
+### 4.1.1 MLD Traps
+
+In this document, "MLD traps" refers to the following HOSTIF trap types:
+
+- `SAI_HOSTIF_TRAP_TYPE_IPV6_MLD_V1_V2`
+- `SAI_HOSTIF_TRAP_TYPE_IPV6_MLD_V1_REPORT`
+- `SAI_HOSTIF_TRAP_TYPE_IPV6_MLD_V1_DONE`
+- `SAI_HOSTIF_TRAP_TYPE_MLD_V2_REPORT`
+
+### **When MLD Snooping is ENABLED (per VLAN):**
+```
+Control Traffic:
+  MLD Protocol Messages (ICMPv6 Types 130, 131, 132, 143)
+    └─ Delivered to the CPU via configured MLD traps
+          → Control plane learns multicast listener membership
+
+Data Traffic:
+  IPv6 Multicast Data (FF00::/8)
+    └─ Forwarding behavior is determined by multicast forwarding entries
+       programmed by the control plane and the configured unknown multicast
+       handling policy (see Section 4.4).
+```
+
+> **Note on Unregistered Group Handling:**
+> The behavior for unregistered IPv6 multicast groups is not defined by `SAI_VLAN_ATTR_MLD_SNOOPING_ENABLE`. It follows the platform's existing unregistered multicast handling policy, which may vary across implementations. Refer to Section 4.4
+
+### **When MLD Snooping is DISABLED (per VLAN):**
+```
+Control Traffic:
+  MLD Protocol Messages (ICMPv6 Types 130, 131, 132, 143)
+    └─ Not delivered to the CPU via configured MLD traps
+         → No multicast listener membership learning
+
+Data Traffic:
+  IPv6 Multicast Data (FF00::/8)
+    └─ Existing multicast forwarding entries, if any, continue to apply.
+       No multicast forwarding entries are learned through MLD snooping.
+       Unknown multicast traffic follows the configured unknown multicast
+       handling policy (see Section 4.4).
+```
+
+
+
+## 4.2 Interaction with Global MLD HOSTIF Traps
+
+The VLAN-level MLD snooping attribute only controls whether MLD protocol
+messages on the VLAN participate in multicast listener learning. It does
+not create MLD HOSTIF traps by itself.
+
+Both of the following conditions must be satisfied:
+
+- At least one global MLD HOSTIF trap is configured.
+- VLAN MLD snooping is enabled.
+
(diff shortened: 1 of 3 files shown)


AnandhiDhanabalan:
@kcudnik , This PR also has enum alias added.  The failure seems to be related to this. Can you help fix it ?

```c
  SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE,

    /** @ignore - for backward compatibility */
    SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE = SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE,
```

kcudnik:
squash commits and force push

AnandhiDhanabalan:
@tjchadaga could you please assign a reviewer for this PR

AnandhiDhanabalan:
@kcudnik, the failure appears to be due to the enum alias usage. Should I take any action on this, or will you be handling it? Please advice.

```
**2026-08-07T17:46:29.4352920Z WARNING: Both enums have the same value SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE and SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE = 0x00000014**
2026-08-07T17:46:29.4353576Z processing commit 9611265
2026-08-07T17:46:29.4353918Z ERROR: please correct all 0 error(s) and all 1 warnings before continue
2026-08-07T17:46:29.4354361Z make: *** [Makefile:89: all] Error 1
2026-08-07T17:46:29.4367116Z 
2026-08-07T17:46:29.4409303Z ##[error]Bash exited with code '2'.
2026-08-07T17:46:29.4538788Z ##[section]Finishing: Metadata check
```

kcudnik:
i think problem is taht SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE is ignored, which is legitimate enum in previous commits, make @ignore enum you are adding, and assign it by SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE

```
 /** @ignore - for backward compatibility */
    SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE = SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE
```

you will need to squash commits and force push, check locally before you do that ,

AnandhiDhanabalan:
@kcudnik, just to confirm my understanding: with the following definition, `SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE` is the enum being marked as @ignore, correct?
 ```
/** @ignore - for backward compatibility */
    SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE = SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE 
```
In my case, `SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE` is the existing attribute that I want to deprecate/ignore, while `SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE` is the new attribute. My understanding from the existing @ignore usage is that the @ignore annotation should be placed on the enum that needs to be deprecated. Could you please confirm the intended definition?

**Comment by kcudnik** [comment]
no, @ignore is placed on attribute that have the same enum ID as other exdisting attribute, if you have both wih the same value, none of them can be depreacated

**Comment by AnandhiDhanabalan** [comment]
@kcudnik , I’m seeing the same meta error irrespective of how the enum is defined. Capturing the error below:
```
WARNING: Both enums have the same value SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE and SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE = 0x00000014
ERROR: please correct all 0 error(s) and all 1 warnings before continue
make: *** [Makefile:89: all] Error 1
```
**Tried both variants:**
```
    /**
     * @brief IGMP Snooping enable or disable control for VLAN
     *
     * IGMP Snooping enable control for VLAN. Default is
     * disabled
     *
     * @type bool
     * @flags CREATE_AND_SET
     * @default false
     */
    SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE,

    /** @ignore - for backward compatibility */
    SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE = SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE,
```
**and**

```
    /**
     * @brief IGMP Snooping enable or disable control for VLAN
     *
     * IGMP Snooping enable control for VLAN. Default is
     * disabled
     *
     * @type bool
     * @flags CREATE_AND_SET
     * @default false
     */
    SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE,

    /** @ignore - for backward compatibility */
    SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE = SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE,
```

In both cases, meta flags duplicate enum values.

I also noticed similar @ignore usage is already present (e.g. in sainexthopgroup.h):
```
typedef enum _sai_next_hop_group_type_t
{
    /** Next hop group is ECMP, with a dynamic number of members, unordered */
    SAI_NEXT_HOP_GROUP_TYPE_DYNAMIC_UNORDERED_ECMP,

    /** @ignore - for backward compatibility */
    SAI_NEXT_HOP_GROUP_TYPE_ECMP = SAI_NEXT_HOP_GROUP_TYPE_DYNAMIC_UNORDERED_ECMP,
```
**Could you please confirm:**

- whether enum aliasing like this is still allowed for new additions, or
- if meta checks have been tightened and we should avoid introducing aliases?

If aliases are not allowed, I’ll retain only `SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE` and avoid adding the new name. Please advise what’s the preferred approach.

**Comment by AnandhiDhanabalan** [comment]
@kcudnik , Gentle reminder — awaiting your inputs.

**Comment by kcudnik** [comment]
I don't remember exact conditions, @ignore was added specifically for this purpose, please analyse the parse metadata code to see which line generates that error, and make sure your commit is on top of master, if you are behind master, or you have multiple commits in the PR that go back and forth for same enum then this problem will happen, 

please also consider using git blame to check how other ignore was indroduced such as https://github.com/opencomputeproject/SAI/pull/655

also as friendly reminder, im just doing community service here, and i don't have always time to take a look at everything what is happening on this repo

**Comment by AnandhiDhanabalan** [comment]
@kcudnik , Thanks for the guidance — the issue was in the metadata ancestry handling.

The fix is applied in BuildCommitHistory (ancestry.pl), where @ignore enums were not being properly respected during history reconstruction, causing the duplicate-value warning for the alias (SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE).
Do review and approve the change

**Validation**
Rebased on latest master
make -C meta passes cleanly (no warnings/errors)
Verified behavior with alias enums

`as friendly reminder, im just doing community service here, and i don't have always time to take a look at everything what is happening on this repo`
Thanks for your time and continued support to the SAI community—it’s really appreciated.

File: meta/ancestry.pl [hunk 1]
```diff
@@ -199,7 +199,7 @@ sub BuildCommitHistory
                 {
                     # ok this is the same enum in history
                 }
-                elsif (defined $IGNORED{$enumName})
+                elsif (defined $IGNORED{$enumName} or defined $IGNORED{$HISTORY{$enumTypeName}{$enumValue}})
```
kcudnik: you are using "$HISTORY{$enumTypeName}" but this path is executed when "elsif (not defined $HISTORY{$enumTypeName}" in line 188, so this is invalid! and you should not check for enumValue! since enum value can be a number etc not actual enum or enum plus some value thsi is invalid

kcudnik (changes requested):

kcudnik:
@AnandhiDhanabalan here you go https://github.com/kcudnik/SAI/commit/9e6d4e84792f7cb7f4f94a8051d82617c80b98f3

this is only commit you need on master, and you don't need to touch ancestry.pl, it works you can cherrypick it and rebase on master in your branch

AnandhiDhanabalan:
@kcudnik, although the code below is syntactically valid, it is semantically incorrect. It implies that `SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE` is to be ignored, whereas in reality `SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE` should be ignored.

```
    SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE,

    /** @ignore - for backward compatibility */
    SAI_VLAN_ATTR_IGMP_SNOOPING_ENABLE = SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE,
```
I still feel `ancestry.pl` may need to be revisited to handle @ignore correctly. However, given the current release timeline, I will drop the changes related to `SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE` for now and take this up separately later.

I’ll update my PR accordingly and keep you posted.

kcudnik:
For me it seems like those values shuld not be equal to eachother, still everything is fine even in this configuration, ignore attribute is introduce just to not cause confusion for metadata parser, still values, parser, serialization and deserialization will work correctly, in your approach you are removeing item that was introduced at the beginning, which is not ok, and making this or ancestry will require some more complex re-work. 

note that ancestry.pl is using history file which only uses history and ignored values is not saved in %HISTORY hash, bneacuse it don't need to be, only latest one from commit are validated, and by adding ignored to existing value, and then introducing new one, will always result in error that both values have the same value, since previous walue will be always in HISTORY hash, ignored values are not. 

as said this change will require some more complex rework, to not break existing logic, in your previous attempt you used undefinded values to compare which resulted in bad logic causing errors, thats unfortunate mistake :/

AnandhiDhanabalan:
@tjchadaga can you approve this PR ?

kcudnik:
please add meaningful description

AnandhiDhanabalan:
> please add meaningful description

Done

kcudnik (approved):

AnandhiDhanabalan:
@tjchadaga can you approve this PR ?

PR #2327, opened 2026-08-05
Add wide label attribute for objects with a label attribute

zechengh09:
Summary:

This is similar to #1158, #1407, #1430, #2322.

SAI label attributes are of type char, which maps to sai_attribute_value_t::chardata[32]. That is 32 bytes including the NUL terminator, so 31 usable characters. That can be too small for real identifiers.

Add a LABEL_WIDE attribute of type sai_s8_list_t alongside each existing char label attribute. A list keeps the label length open, so the NOS can choose the length it needs instead of being constrained by an arbitrary limit in the SAI spec. Each new attribute is appended at the end of its attribute enum so that no existing attribute is renumbered.

Objects covered: counter, ACL counter, LAG, mirror session, prefix compression table, policer, scheduler, next hop group, virtual router, UDF group and router interface.

The existing char attributes are marked deprecated. Exactly one of the two attributes may be set to a non-default value; setting both is invalid.


inc/saiacl.h
```diff
@@ -3625,9 +3616,23 @@ typedef enum _sai_acl_counter_attr_t
...
+     *
+     * Replaces #SAI_ACL_COUNTER_ATTR_LABEL, which is limited to 32 bytes.
+     * Exactly one of the two attributes may be set to a non-default
+     * value; setting both is invalid.
+     *
+     * @type sai_s8_list_t
+     * @flags CREATE_AND_SET
+     * @default empty
+     */
+    SAI_ACL_COUNTER_ATTR_LABEL_EXTENDED,
```
JaiOCP: Just a nit:

Instead of SAI_ACL_COUNTER_ATTR_LABEL_EXTENDED, can we use
SAI_ACL_COUNTER_ATTR_LABEL_WIDE

inc/saicounter.h
```diff
@@ -120,6 +121,19 @@ typedef enum _sai_counter_attr_t
...
+     *
+     * Replaces #SAI_COUNTER_ATTR_LABEL, which is limited to 32 bytes.
+     * Exactly one of the two attributes may be set to a non-default
+     * value; setting both is invalid.
+     *
+     * @type sai_s8_list_t
+     * @flags CREATE_AND_SET
+     * @default empty
+     */
+    SAI_COUNTER_ATTR_LABEL_EXTENDED,
```
JaiOCP: Same nit:

Instead of SAI_COUNTER_ATTR_LABEL_EXTENDED, can we use
SAI_COUNTER_ATTR_LABEL_WIDE

inc/saiacl.h
zechengh09: Thanks - updated to use SAI_ACL_COUNTER_ATTR_LABEL_WIDE

inc/saicounter.h
zechengh09: Updated to use SAI_COUNTER_ATTR_LABEL_WIDE

JaiOCP (approved):
Looks Good

rajkumar38 (approved):

inc/sairouterinterface.h
```diff
@@ -436,6 +437,19 @@ typedef enum _sai_router_interface_attr_t
      */
     SAI_ROUTER_INTERFACE_ATTR_QOS_TC_AND_COLOR_TO_DOT1P_MAP,
 
+    /**
+     * @brief Wide label attribute used to uniquely identify router interface.
+     *
+     * Replaces #SAI_ROUTER_INTERFACE_ATTR_LABEL, which is limited to 32 bytes.
+     * Exactly one of the two attributes may be set to a non-default
+     * value; setting both is invalid.
```
kcudnik, line 445: from my perspective it could be valid, and more over they could be even different at the same time

j-bos (approved):

inc/sairouterinterface.h
zechengh09: Thanks for the review - the wide label attribute is added to replace the current label attribute, which is marked as deprecated. It is kept for backward compatibility, but our goal is for both NOS and SDK starting using the new attribute in the new spec.

Therefore, only one attribute should be set at a given time. They could be different only if NOS starting to use the wide label with more than 32 characters.

tjchadaga (approved):

PR #2335, opened 2026-08-19
[meta] Add opt-in saithrift RPCs for draining SAI notifications

nicholasching:
**Context / motivation**
Part of the SAIVPP unit-test framework. SAI delivers port state and BFD session state asynchronously through function pointers written into `SAI_SWITCH_ATTR_PORT_STATE_CHANGE_NOTIFY` and `SAI_SWITCH_ATTR_BFD_SESSION_STATE_CHANGE_NOTIFY`. A Thrift client cannot supply a function pointer, so there is currently no way for any saithrift-based test suite to observe that a notification was delivered. This blocks testing the notification path on any target, not just VPP.

**What this change does**
* **`meta/templates/sai_thrift_utils.tt`, `meta/templates/sai.thrift.tt`.** Declare `sai_thrift_enable_notifications()` and `sai_thrift_drain_notifications()`, plus a `sai_thrift_notification_event_t` record carrying notification type, object id and state.
* **`meta/templates/sai_rpc_server_functions.tt`.** Add the two method names to the `sai_utils_functions` regex so the generator emits stub bodies instead of trying to bind them to SAI metadata.
* **`meta/sai_rpc_frontend.cpp`.** Implement both methods in `sai_rpcHandlerFrontend`, following the existing `sai_thrift_object_type_get_availability` pattern. The enable RPC installs server-owned callbacks and is idempotent; the drain RPC is non-blocking and atomically empties a mutex-protected bounded queue.
* **Post-create registration.** The callbacks are installed on the already-created switch via `set_switch_attribute`, not at create time. `SwitchStateBase::set_switch_default_attributes()` writes NULL into both attributes during `initialize_default_objects`, so a pointer supplied at create time is silently discarded. Both attributes are `CREATE_AND_SET` and the senders re-read them when an event fires, so a post-create set is correct, simpler, and additionally suppresses the forced port-UP notifications raised during host-interface creation.

**Scope / risk**
* **Default-preserving:** behaviour is unchanged unless a test explicitly calls the enable RPC. No existing RPC is modified.
* **No function pointer crosses Thrift.** The callbacks live entirely inside `saiserver`; only notification type, object id and state are serialized.
* **Bounded:** the queue holds 256 events and drops the oldest on overflow, so a client that never drains cannot grow memory without limit.
* **RAII containers and standard synchronization only.** No new dependencies, no unsafe C memory or string calls, no credentials or certificates.
* No SAI headers, metadata definitions, or backend behaviour change.
* **Validated behavior:** `gensairpc.pl` completes and the generated service contains both methods; `saithriftv2` and `saiserver` build and link; enable followed by drain from a throwaway client returns an empty list without crashing. End-to-end validation is in the notification test PR.

**Dependencies**
None. This is a prerequisite for the notification test cases, #2336.

### DIFF
--- a/meta/templates/sai.thrift.tt
+++ b/meta/templates/sai.thrift.tt
@@ -70,6 +70,12 @@ struct sai_thrift_attr_capability_t {
     2: bool boolset_implemented;
     3: bool get_implemented;
 }
+
+struct sai_thrift_notification_event_t {
+    1: i32 notification_type;
+    2: sai_thrift_object_id_t object_id;
+    3: i32 state;
+}
 [% END -%]
 
 [%- ######################################################################## -%]
--- a/meta/templates/sai.thrift.tt
+++ b/meta/templates/sai.thrift.tt
@@ -71,8 +71,13 @@ struct sai_thrift_attr_capability_t {
     3: bool get_implemented;
 }
 
+enum sai_thrift_notification_type_t {
+    SAI_THRIFT_NOTIFICATION_TYPE_PORT_STATE_CHANGE = 0;
+    SAI_THRIFT_NOTIFICATION_TYPE_BFD_SESSION_STATE_CHANGE = 1;
+}
+
 struct sai_thrift_notification_event_t {
-    1: i32 notification_type;
+    1: sai_thrift_notification_type_t notification_type;
     2: sai_thrift_object_id_t object_id;
     3: i32 state;
 }
--- a/meta/sai_rpc_frontend.cpp
+++ b/meta/sai_rpc_frontend.cpp
@@ -897,15 +897,17 @@ static void sai_thrift_nat_type_t_parse(
 
 namespace
 {
-    constexpr std::int32_t kPortStateChangeNotification = 0;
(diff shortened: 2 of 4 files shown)


nicholasching:

meta/templates/sai.thrift.tt
```diff
@@ -70,6 +70,12 @@ struct sai_thrift_attr_capability_t {
     2: bool boolset_implemented;
     3: bool get_implemented;
 }
+
+struct sai_thrift_notification_event_t {
+    1: i32 notification_type;
```
yue-fred-gao: can we define enum type for notification type?

nicholasching:

yue-fred-gao (approved):
lgtm

aaronber0614 (approved):
The notification bridge is opt-in and preserves existing behavior unless explicitly enabled. The callback signatures, bounded synchronized queue, Thrift types, and generated RPC integration look correct for the intended standalone saiserver use case. I found no blocking code issues in the current change.

_Co-authored with CoPilot_

PR #2336, opened 2026-08-19
[sai_test] Add opt-in port and BFD notification test cases

nicholasching:
**Context / motivation**
Part of the SAIVPP unit-test framework. `sai_test` currently has no coverage for asynchronous SAI notifications: nothing exercises `SAI_SWITCH_ATTR_PORT_STATE_CHANGE_NOTIFY` or `SAI_SWITCH_ATTR_BFD_SESSION_STATE_CHANGE_NOTIFY`. This adds the first tests for that path, using the saithrift notification RPCs added separately.

**What this change does**
* **`test/sai_test/sai_notification_test.py`.** New module with five classes: `PortStateChangeTest`, `PortStateRecoveryTest`, `BfdSessionUpTest`, `BfdSessionDownTest` and `BfdMultihopTest`.
* **Opt-in.** Every class skips unless the harness passes the `vpp_notification_test` PTF parameter; the BFD classes additionally require `SIMULATE_SONIC=1`. With neither set the module is inert.
* **Port cases.** Flap the SAI host-interface netdev and wait for the matching port OID and oper status, polling every 0.5s within a 5s budget. The wire-side veth peer is deliberately not used: VPP binds it with an AF_PACKET socket whose driver does not watch the underlying netdev carrier, so Linux reports `NO-CARRIER` while `vppctl show hardware-interfaces` still reports the link up and no interface event is ever raised.
* **BFD cases.** Anchor the session on a LAG router interface that already has a connected local address, create the peer neighbor so the dataplane can build the L2 header, and drive the far end with a small scapy responder. The responder follows the RFC 5880 state machine (a peer in Down only leaves that state on receiving Down or Init), targets UDP 3784/4784 in both directions with an ephemeral source port, and listens on every LAG member because the flow is hashed onto one of them.
* **Independent corroboration.** Assertions are checked against `vppctl` output rather than only re-reading the SAI attribute that the notification path itself writes. All SAI objects and link state are restored in `finally`/`tearDown`.

**Scope / risk**
* **One new file.** No existing test, helper, SAI header, or backend code is modified.
* **Default-off by construction,** so other consumers (real ASICs, other harnesses) are unaffected.
* BFD waits use a longer timeout than the port cases: RFC 5880 holds the control-packet interval at one second until a session is up, and measured bring-up against the responder is around 5.5s.
* **Validated behavior:** on a VPP bench all five classes passed three consecutive module runs, plus a full five-module 90-selector matrix run at `PORT_COUNT=32` with per-test isolation, which reported 90 observed selectors and zero regressions against the existing stable baseline.

**Dependencies**
Requires the saithrift notification bridge PR, #2335; must be merged before this PR can be merged.

### DIFF
--- a/test/sai_test/sai_notification_test.py
+++ b/test/sai_test/sai_notification_test.py
@@ -42,6 +42,7 @@ NOTIFICATION_POLL_INTERVAL = 0.5
 # the notification still has to clear the VPP event poll.
 BFD_NOTIFICATION_TIMEOUT = 25.0
 VPPCTL_TIMEOUT = 10
+SNIFFER_START_TIMEOUT = 5.0
 BFD_EPHEMERAL_SRC_PORT = 49152
 BFD_STATE_INIT = 2
 BFD_STATE_UP = 3
@@ -206,32 +207,44 @@ class BfdResponder:
         self.source_macs = {
             name: get_if_hwaddr(name) for name in self.interface_names
         }
-        self.stop_event = threading.Event()
-        self.threads = [
-            threading.Thread(target=self._run, args=(name,), daemon=True)
-            for name in self.interface_names
-        ]
+        self.sniffers = []
 
     def start(self):
-        for thread in self.threads:
-            thread.start()
+        import functools
+
+        from scapy.all import AsyncSniffer
+
+        for interface_name in self.interface_names:
+            ready = threading.Event()
+            sniffer = AsyncSniffer(
+                iface=interface_name,
+                filter="udp dst port {} and src host {}".format(
+                    self.udp_port, self.local_ip
+                ),
+                store=False,
+                prn=functools.partial(self._respond, interface_name),
+                started_callback=ready.set,
+            )
+            sniffer.start()
+            self.sniffers.append(sniffer)
+
+            if not ready.wait(timeout=SNIFFER_START_TIMEOUT):
+                self.stop()
+                raise AssertionError(
+                    "BFD responder failed to start capture on {}".format(
+                        interface_name
+                    )
+                )
 
     def stop(self):
-        self.stop_event.set()
-        for thread in self.threads:
-            thread.join(timeout=2)
-
-    def _run(self, interface_name):
-        from scapy.all import sniff
-
-        while not self.stop_event.is_set():
-            sniff(
-                iface=interface_name,
-                filter="udp",
-                timeout=0.5,
-                store=False,
-                prn=lambda packet: self._respond(interface_name, packet),
-            )
+        for sniffer in self.sniffers:
+            try:
+                sniffer.stop()
+            except Exception:
+                # stop() raises if the sniffer never reached its run loop, which
+                # happens when start() aborts partway and unwinds through here.
+                pass
+        self.sniffers = []
 
     def _respond(self, interface_name, packet):
         from scapy.all import Ether, IP, UDP, sendp
--- a/test/sai_test/sai_notification_test.py
+++ b/test/sai_test/sai_notification_test.py
@@ -9,11 +9,9 @@
 #    LIMITATION ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS
 #    FOR A PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
 
-"""Opt-in SAI notification tests with a VPP-specific BFD fixture."""
+"""Opt-in SAI notification tests with a provider-backed BFD fixture."""
 
-import re
-import os
-import threading
+import importlib
 import time
 
 from ptf.testutils import test_params_get
@@ -21,7 +19,6 @@ from unittest import SkipTest
 
 from sai_thrift.sai_adapter import *
 from sai_test_base import T0TestBase
-from sai_utils import sai_ipaddress, sai_ipprefix
 
 
 PORT_NOTIFICATION_TYPE = (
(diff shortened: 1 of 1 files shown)


nicholasching:

nicholasching:

test/sai_test/sai_notification_test.py
```diff
@@ -0,0 +1,532 @@
...
+
+from ptf.testutils import test_params_get
+from unittest import SkipTest
+
+from sai_thrift.sai_adapter import *
+from sai_test_base import T0TestBase
+from sai_utils import sai_ipaddress, sai_ipprefix
+
+
+PORT_NOTIFICATION_TYPE = 0
```
yue-fred-gao: This should come from the enum defined in PR 2335

test/sai_test/sai_notification_test.py
```diff
@@ -0,0 +1,532 @@
...
+                sai_thrift_remove_router_interface(self.client, self.lag_rif)
+        finally:
+            super().tearDown()
+
+
+    def bring_session_up(self):
+        self.start_session()
+        self.bfd_event(SAI_BFD_SESSION_STATE_UP)
+        self.assert_bfd_state(SAI_BFD_SESSION_STATE_UP)
+        self.assert_vpp_bfd_state("up")
```
yue-fred-gao: Please check on asic type before calling vpp specific check.

test/sai_test/sai_notification_test.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def stop(self):
+        self.stop_event.set()
+        for thread in self.threads:
+            thread.join(timeout=2)
+
+    def _run(self, interface_name):
+        from scapy.all import sniff
+
+        while not self.stop_event.is_set():
```
yue-fred-gao, line 222: AI points out a gap between sniff calls that may cause loss of bfd message because between sniff calls, there is no BPF socket to receive the bfd packet. It suggests to use AsyncSniffer.
    def start(self):
        import functools

        from scapy.all import AsyncSniffer

        for name in self.interface_names:
            ready = threading.Event()
            sniffer = AsyncSniffer(
                iface=name,
                # Narrow the match in BPF rather than in Python. The filter is
                # now compiled once instead of twice a second, and the responder
                # never sees its own replies looped back by AF_PACKET.
                filter="udp dst port {} and src host {}".format(
                    self.udp_port, self.local_ip
                ),
                store=False,
                # functools.partial, not a lambda: a lambda closing over the
                # loop variable would bind every sniffer to the last interface.
                prn=functools.partial(self._respond, name),
                started_callback=ready.set,
            )
            sniffer.start()
            self.sniffers.append(sniffer)

            # start() returns before the capture socket is bound. Without this
            # the first control packets are missed exactly as before.
            if not ready.wait(timeout=SNIFFER_START_TIMEOUT):
                self.stop()
                raise AssertionError(
                    "BFD responder failed to start capture on {}".format(name)
                )

    def stop(self):
        for sniffer in self.sniffers:
            try:
                sniffer.stop()
            except Exception:
                # stop() raises if the sniffer never reached the run loop.
                pass
        self.sniffers = []

test/sai_test/sai_notification_test.py
```diff
@@ -0,0 +1,532 @@
...
+        return result.stdout.decode("utf-8", "replace")
+
+    def log_vpp_state(self, *command):
+        output = self.vppctl(*command)
+        print("vppctl {}:\n{}".format(" ".join(command), output))
+        return output
+
+    def setUp(self):
+        params = test_params_get() or {}
+        if params.get(PLATFORM_PARAM) != VPP_PLATFORM:
```
yue-fred-gao: do we need this check? what is required for another platform to enable notification test? There is also notification test opt-in flag. If it is not supported in certain flag, they can opt it out.

Our goal is making this test suitable for other platforms. They can add platform specific check. But without it, it still works.

test/sai_test/sai_notification_test.py
nicholasching: I agree, fixed in [2c4ea58](https://github.com/opencomputeproject/SAI/pull/2336/commits/2c4ea58b23dbfbf96475f1de00b7bc32886a5671).

test/sai_test/sai_notification_test.py
nicholasching: Good catch; thanks Fred. All done in [f7c395d](https://github.com/opencomputeproject/SAI/pull/2336/commits/f7c395dc3337779fea12af790cb5a46a8e3593a4), now check `self.platform == VPP_PLATFORM` before call VPP checks.

test/sai_test/sai_notification_test.py
nicholasching: Oh that is very true! Implemented in [2a0aa35](https://github.com/opencomputeproject/SAI/pull/2336/commits/2a0aa3530226cfd9d7ebde08f786909a48a41eb4), BFD responder now uses one continuously running AsyncSniffer per peer interface and waits for capture startup before creating the session, avoiding packet-loss windows between polling sniff calls.

test/sai_test/sai_notification_test.py
nicholasching: > do we need this check? what is required for another platform to enable notification test? 

Hi Fred, yes, unfortunately, unlike the port tests, the current BFD test implementation is VPP-specific. Since it constructs the complete working BFD environment inside the VPP test container, it uses the VPP LAG topology, VPP-specific addressing and routing, Linux peer interfaces, a Scapy BFD responder, and VPP state inspection. Other platforms would need equivalent platform-specific BFD fixtures, which have not been implemented yet.

> Our goal is making this test suitable for other platforms. They can add platform specific check. But without it, it still works.

I've done some more research; it is possible to move the VPP-specific topology and observability from the shared SAI test suite into the SAI VPP UT harness. For a final structure that looks something like this:
```
SAI PR #2336:
    Generic notification tests and BFD fixture interface

sonic-sairedis Phase 4 PR:
    VPP BFD fixture
    VPP Docker image injection
    VPP test parameters
    VPP-specific CI invocation
```

I'm making this change now. This should make the test fully generic and require the platform under test to provide their platform-specific fixture implementation instead.

nicholasching: Hi Fred, I’ve pushed a couple more commits to generalize the notification tests. The harness is now responsible for supplying the platform-specific BFD fixture to SAI notification tests, moving the VPP-specific code from `SAI` into `sonic-sairedis`. Other platforms will be responsible for providing their own fixture when enabling the BFD cases.

test/sai_test/sai_notification_test.py
```diff
@@ -0,0 +1,510 @@
...
+            self.bfd_session,
+            state=True,
+        )
+        self.assertEqual(attributes["state"], state)
+
+    def bring_session_up(self):
+        self.start_session()
+        self.bfd_event(SAI_BFD_SESSION_STATE_UP)
+        self.assert_bfd_state(SAI_BFD_SESSION_STATE_UP)
+        self.assert_vpp_bfd_state("up")
```
yue-fred-gao: Let's move the platform check to the caller. If other platforms are added to this test, we don't want this function calls assert_vpp_bfd_state, assert_cisco_8000_bfd_state etc. It should call the function corresponding to the platform.

yue-fred-gao (approved):

PR #2337, opened 2026-08-19
[submodule] Update test/ptf to a Python 3.12+ compatible revision

nicholasching:
**Context / motivation**
`test/ptf` is pinned at `7494366` from December 2022. That revision's `ptf` runner does `import imp` at module scope and uses `imp.load_module` in `load_test_modules`. The `imp` module was removed in Python 3.12, so on any modern distribution PTF fails before running a single test:

```
File "ptf", line 19, in <module>
    import imp
ModuleNotFoundError: No module named 'imp'
```

This affects every `sai_test` module rather than any particular feature. It was found while bringing up a Debian Trixie based test image, which ships Python 3.13.

**What this change does**
* Moves `test/ptf` from `7494366` to `d587084`, the current head of `p4lang/ptf` main. The change that matters is `771a452` ("Replace use of deprecated package imp with current importlib"), present in the new pin and absent from the old.

**Scope / risk**
* **Submodule pointer only.** No SAI source changes.
* The range spans roughly four years of upstream PTF work, including a uv/pyproject project layout, the nnpy to pynng transition, and scapy 2.7 support, so consumers pinning older Python or older scapy should check their environment.
* `771a452` is the minimum revision that unblocks Python 3.12+, if a smaller step is preferred.
* **Validated behavior:** the test image was rebuilt against the new pin and a 90-selector matrix completed with no infrastructure failures.

**Dependencies**
None functionally, but as a submodule bump this should be the last of the SAIVPP Phase 4 SAI changes to merge.


nicholasching:

nicholasching:

yue-fred-gao (approved):

PR #2343, opened 2026-09-01
ACL Table IP Type Attribute

AnandhiDhanabalan:
This PR introduces `SAI_ACL_TABLE_ATTR_IP_TYPE`, a new `CREATE_ONLY` ACL table attribute that allows the NOS to explicitly declare the packet domain (ANY/IP/IPV4/IPV6/NON_IP) of an ACL table at creation time. This enables implementations to select an optimal TCAM key profile upfront rather than inferring the packet domain from entry match fields.

### DIFF
--- a/doc/SAI-Proposal-ACL-Table-Type.md
+++ b/doc/SAI-Proposal-ACL-Table-Type.md
@@ -1,0 +1,319 @@
+# ACL Table IP Type Attribute
+
+## Revision History
+
+| Version | Date       | Author             | Description   |
+| ------- | ---------- | ------------------ | ------------- |
+| 0.1     | 2026-09-1  | Anandhi Dhanabalan | Initial draft |
+
+---
+
+# 1. Overview
+
+SAI currently defines the packet type of an ACL entry through `SAI_ACL_ENTRY_ATTR_FIELD_ACL_IP_TYPE`, but does not provide an ACL-table-level attribute to declare the intended packet domain of the table.
+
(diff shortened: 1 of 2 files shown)


AnandhiDhanabalan:
@j-bos , Question asked during discussion if SONIC uses any of the `SAI_ACL_IP_TYPE_xxx` enums, especially for ARP

In `orchagent/aclorch.cpp, AclOrch::initDefaultTableTypes()` builds each default table type with an explicit list of supported SAI match fields. `SAI_ACL_TABLE_ATTR_FIELD_ACL_IP_TYPE` is registered for: `L3`, `L3V6`, `L3V4V6`, `MCLAG`, `MIRROR`, and `MIRRORV6`. A rule in these tables is allowed to carry an IP_TYPE match — it's part of that table type's declared capability. SONiC does not use it for any of the default rules.

MCLAG port-isolation ACL rules, auto-generated by `mclagsyncd` (not user-authored). In `mclaglink.cpp:343`, whenever an MCLAG peer link goes down and orphan ports need isolating, `mclagsyncd` programs the ACL rule with `IP_TYPE = ANY`

Also notable: `cfgmgr/natmgr.cpp` validates an IP_TYPE field on NAT-related ACL rules, restricting it to only `IP or IPV4ANY` `(natmgr.cpp:7938-7946)` — this is for user/NAT-orchestrator-authored ACL rules that block/forward NAT'd traffic, not a fully automatic default, but it's another consumer of the same IP_TYPE/SAI_ACL_IP_TYPE_xxx plumbing.

AnandhiDhanabalan:
@tjchadaga, could you approve this PR?

tjchadaga:
@j-bos, @JaiOCP, @marian-pritsak, @eddyk-nvidia - could you please help review?

PR #2344, opened 2026-09-01
 Add User defined trap types for route,ipmc,l2mc,mysid

nipunjoy:
**Motivation :**

- SAI already supported USER_TRAP_ID attributes on FDB and Neighbor entries, allowing packets matching those entries to be trapped/logged/copied to the host CPU via user-defined traps. However, the same capability was missing for IP Multicast (IPMC), L2 Multicast (L2MC), and SRv6 My SID entries. This gap prevented uniform UDT handling across all forwarding entry types
- SAI_IPMC_ENTRY_ATTR_OUTPUT_GROUP_ID identifies the multicast output group used for packet replication. Today, the attribute is only valid when:
```c
SAI_IPMC_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_FORWARD
```
However, this does not fully reflect the behavior of all supported IPMC packet actions.
For  SAI_PACKET_ACTION_COPY, SAI_PACKET_ACTION_LOG, SAI_PACKET_ACTION_TRANSIT and SAI_PACKET_ACTION_DONOTDROP packet forwarding to the multicast output group remains intact. 
Specifically:
* SAI_PACKET_ACTION_COPY forwards the packet to the configured output group and sends an additional copy to the CPU.
* SAI_PACKET_ACTION_LOG forwards the packet to the configured output group and sends a copy to the CPU for logging purposes.
* SAI_PACKET_ACTION_TRANSIT is a combination of COPY_CANCEL and FORWARD.
* SAI_PACKET_ACTION_DONOTDROP does not drop the packet.
Since the output group is still used in both cases, restricting SAI_IPMC_ENTRY_ATTR_OUTPUT_GROUP_ID to SAI_PACKET_ACTION_FORWARD unnecessarily limits valid IPMC configurations.

**Changes :**

- inc/saihostif.h : Added four new SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_* enum values: ROUTE, IPMC, L2MC, and MY_SID
- inc/saiipmc.h : Added SAI_IPMC_ENTRY_ATTR_USER_TRAP_ID attribute
- inc/sail2mc.h : Added SAI_L2MC_ENTRY_ATTR_USER_TRAP_ID attribute
- inc/saisrv6.h : Added SAI_MY_SID_ENTRY_ATTR_USER_TRAP_ID attribute
- inc/sairoute.h : Updated SAI_ROUTE_ENTRY_ATTR_USER_TRAP_ID documentation to include copy alongside trap/log actions
- inc/saineighbor.h : Updated SAI_NEIGHBOR_ENTRY_ATTR_USER_TRAP_ID documentation to include copy alongside trap/log actions

All new attributes follow the established pattern from saifdb.h / saineighbor.h: an sai_object_id_t referencing SAI_OBJECT_TYPE_HOSTIF_USER_DEFINED_TRAP, with create-and-set capability and a null default.

- Extend the validonly condition of SAI_IPMC_ENTRY_ATTR_OUTPUT_GROUP_ID to also allow SAI_PACKET_ACTION_COPY, SAI_PACKET_ACTION_LOG, SAI_PACKET_ACTION_TRANSIT and SAI_PACKET_ACTION_DONOTDROP:

```c
@validonly SAI_IPMC_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_FORWARD ||
           SAI_IPMC_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_COPY ||
           SAI_IPMC_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_LOG ||
           SAI_IPMC_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_TRANSIT || 
           SAI_IPMC_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_DONOTDROP
```

**Impact :**

- API surface: Extends the SAI specification uniformly — UDT is now available on Route, FDB, Neighbor, IPMC, L2MC, and My SID entries, providing consistent trap/log/copy-to-CPU behavior across all major forwarding entry types.
- aligns the specification with the existing forwarding semantics of IPMC packet actions and allows valid multicast forwarding scenarios where packets are both replicated to the output group and copied to the CPU.


nipunjoy:
@j-bos, please review my changes.

inc/saihostif.h
```diff
@@ -727,6 +730,22 @@ typedef enum _sai_hostif_user_defined_trap_type_t
     /** Traps to be associated with TAM collector */
     SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_TAM,
 
+    /**
+     * @brief Route traps
+     *
+     * Generated by route entry hit with action trap/log/copy
+     */
+    SAI_HOSTIF_USER_DEFINED_TRAP_TYPE_ROUTE,
```
j-bos: This case is already covered by the existing _ROUTER, so this case should be removed.

j-bos (changes requested):

inc/saihostif.h
nipunjoy: done. addressed in the next diff.

tjchadaga:
@JaiOCP - please help review as well

tjchadaga:
@nipunjoy  - please squash your commits and force push

nipunjoy:
@tjchadaga done.

inc/saifdb.h
```diff
@@ -126,6 +126,7 @@ typedef enum _sai_fdb_entry_attr_t
      * @objects SAI_OBJECT_TYPE_HOSTIF_USER_DEFINED_TRAP
      * @allownull true
      * @default SAI_NULL_OBJECT_ID
+     * @validonly SAI_FDB_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_TRAP or SAI_FDB_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_COPY or SAI_FDB_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_LOG
```
j-bos, line 129: Slight concern on the validonly here. If there is a sequence where you want to atomically switch to trap under this UDT, you might have a problem, since you would first have to change the packet action, then assign the trap?

j-bos: Same for other additions of validonly for UDT below. User may prefer to be able to assign the UDT, even if the action is currently just to forward and not using it.

It's not an issue for creation where both attributes can be passed in at once, but for later modification.

nipunjoy: @j-bos  Good Point, and I reviewed similar attributes that exist today. example:SAI_IPMC_ENTRY_ATTR_OUTPUT_GROUP_ID, which is validonly SAI_IPMC_ENTRY_ATTR_PACKET_ACTION == SAI_PACKET_ACTION_FORWARD. The same issue of assigning group id and then changing the packet action exists. Please advise if we can leave it for vendor implementations to decide whether this configuration should be allowed or not?

PR #2346, opened 2026-09-02
[doc]: Clarify HFT enterprise number layout

Pterosaur:
## What I did

Clarified the HFT IPFIX enterprise-number wire layout and corrected the example value:

- the first 16 bits on the wire encode the SAI object type;
- the following 16 bits encode the SAI stats ID;
- multi-byte values use network byte order.

For queue type `0x0015` and stat ID `0x0022`, this gives `0x00150022`, or wire bytes `00 15 00 22`.

## Why I did it

The previous formula placed the stat ID in the high 16 bits and contradicted both the adjacent bit diagram and the current SONiC countersyncd HFT IPFIX decoder.

## How I verified it

This is a one-line documentation correction. `git diff --check` passes and the old example value no longer appears in the document.

The matching sonic-net/SONiC HLD change is submitted separately.

Matching SONiC HLD PR: https://github.com/sonic-net/SONiC/pull/2537


DavidZagury (approved):

kcudnik (approved):

bandaru-viswanath:
Please understand that this breaks backward compatibility. Why were the placements of OBJ-TYPE and STA_ID awapped ? 
@Pterosaur @tjchadaga @JaiOCP

Pterosaur:
> Please understand that this breaks backward compatibility. Why were the placements of OBJ-TYPE and STA_ID awapped ? @Pterosaur @tjchadaga @JaiOCP

Hi @bandaru-viswanath 

Sorry for the confusion.

This PR does not introduce a new wire layout. It corrects an inconsistency in the original documentation.

The existing bit diagram places the object type first on the wire (the upper 16 bits of the enterprise-number value), followed by the stat ID (the lower 16 bits), with an EF bit in each half. The subsequent port example already follows that layout: type 0x0001 and stat ID 0x0004 produce 0x00010004. Only the formula in this paragraph had the two fields reversed.

The existing SONiC decoder also follows the diagram and the port example. The purpose of this PR is to make the text consistent with them and prevent future SAI implementations from being misled by the contradictory formula. Neither the diagrams nor any implementation code were changed by this PR.

PR #2348, opened 2026-09-09
HW Protection Group amendments and clarifications

j-bos:
- Amendments to SAI HW protection group config model
    - Support for non-revertive behavior
    - Option to use different distribution methods for backup groups (hint any kind of ECMP group as a backup group)
- Clarifications and fixes to documentation and examples.


prsunny:
@tjchadaga , @rlhui for viz

inc/sainexthopgroup.h
```diff
@@ -198,10 +209,13 @@ typedef enum _sai_next_hop_group_attr_t
...
      *
+     * deprecated in favor of SAI_NEXT_HOP_GROUP_ATTR_ADMIN_ROLE
+     *
      * @type bool
      * @flags CREATE_AND_SET
      * @default false
      * @validonly SAI_NEXT_HOP_GROUP_ATTR_TYPE == SAI_NEXT_HOP_GROUP_TYPE_PROTECTION
+     * @deprecated true
      */
     SAI_NEXT_HOP_GROUP_ATTR_SET_SWITCHOVER,
```
JaiOCP, line 220: You can not deprecate this attribute.

Premise of the attribute and use model assumed is incorrect.

JaiOCP (changes requested):

inc/sainexthopgroup.h
j-bos: Hi, it seems this can be covered by the admin mode enum. I.e. true here is equivalent to standby while false is equivalent to primary. Do you see a sequence where it could be useful to keep this besides backwards compatibility?

inc/sainexthopgroup.h
```diff
@@ -398,6 +412,39 @@ typedef enum _sai_next_hop_group_attr_t
...
+     *
+     * When false the hardware switchover is one way: hardware still switches
+     * from primary to standby on failure of the monitored object, but never
+     * switches back once the monitored object recovers, so a recovering or
+     * flapping object does not move traffic. The control plane moves it back
+     * through #SAI_NEXT_HOP_GROUP_ATTR_ADMIN_ROLE.
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default true
```
j-bos: To check on compatibility option and best default.

PR #2349, opened 2026-09-09
TRIM to Sender and TC-DSCP Map

JaiOCP:
This PR address the usecase of setting different DSCP for TRIM packets going out on different port as mentioned in
https://github.com/JaiOCP/SAI/blob/master/doc/SAI-Proposal-Packet-Trimming.md

Current usage of SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP creates an issue where when TRIM TC is enabled for DSCP marking any other traffic flowing out on the port need to be correctly marked for TC to DSCP even if there is no change in outgoing DSCP. This is because once a map is enabled for a port, SAI spec mandates a default behavior of mapping to TC 0. This is not acceptable if there is requirement to map TC to DSCP only for TRIM packets.

New attribute SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP is introduced to control only TRIM related TC to DSCP map and not touch the other data traffic maps.

New attribute SAI_BUFFER_PROFILE_PACKET_ADMISSION_FAIL_ACTION_DROP_AND_TRIM_TO_SENDER is introduced where TRIM packets can be forwarded in the reverse direction to the sender. All the other attributes for TRIM as applicable including stats remains same.

### DIFF
--- a/doc/SAI-Proposal-Packet-Trimming.md
+++ b/doc/SAI-Proposal-Packet-Trimming.md
@@ -224,3 +224,47 @@ typedef enum _sai_packet_trim_dscp_resolution_mode_t
 |saiqueue.h |SAI_QUEUE_STAT_TRIM_PACKETS|Per-queue counter of packets trimmed on trimming-eligible queue due to failed shared buffer admission|
 |           |SAI_QUEUE_STAT_DROPPED_TRIM_PACKETS|Per-queue counter of packets trimmed on trimming-eligible queue but dropped due to failed shared buffer admission on a trim queue|
 |           |SAI_QUEUE_STAT_TX_TRIM_PACKETS|Per-queue counter of packets trimmed on trimming-eligible queue and successfully sent via a trim queue|
+
+## TRIM or Data Packet DSCP Marking
+The current SAI specification defines a single port attribute, SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP, for specifying TRIM and/or Data Packet TC to DSCP marking.
+
+This creates an issue when data packet remarking is not needed and there is no corresponding entry in the TC map. According to the SAI specification, the default value of TC 0 becomes effective and maps to the default DSCP value, which inadvertently changes the DSCP value of the data packet.
+
+To address this unintended behavior, we are introducing a new port-based attribute: SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP. This attribute handles TC to DSCP mapping exclusively for TRIM packets.
+
+### Capability Query
+NOS can do a cpability query for SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP attribure to check an implementation's support.
+
+If both SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP and SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP are configured then TRIM packets will use SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP and data packets will use SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP.
+
+If only SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP is confiured then NOS can configure both data packet TC and TRIM TC as long as the desired behavior of DSCP value of data packets is maintained.
+
+If only SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP is configured then data packets do not undergo any DSCP remarking and only TRIM packets will use the configured map for DSCP marking.
+
+### TRIM to Sender TC and DSCP Value
+Trim to Sender may use a different TC value if in DSCP resolution mode or may use a different DSCP value from TRIM to receiver.
+
+Two new switch attributes are introduced specific to TRIM to sender. Note that the SAI_SWITCH_ATTR_PACKET_TRIM_DSCP_RESOLUTION_MODE is same for TRIM to sender and TRIM to reciever. 
+Also note that a queue can only be eligible for TRIM or TRIM_TO_SENDER but not for both.
+
+```
+    /**
+     * @brief New packet trim to sender DSCP value
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
+     * @validonly SAI_SWITCH_ATTR_PACKET_TRIM_DSCP_RESOLUTION_MODE == SAI_PACKET_TRIM_DSCP_RESOLUTION_MODE_DSCP_VALUE
+     */
+    SAI_SWITCH_ATTR_PACKET_TRIM_TO_SENDER_DSCP_VALUE,
+
+    /**
+     * @brief New packet trim to sender TC value
+     *
+     * @type sai_uint8_t
+     * @flags CREATE_AND_SET
+     * @default 0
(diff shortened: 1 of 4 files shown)


kcudnik (approved):

tjchadaga:
+ @vmittal-msft, @kperumalbfn

mholankar (approved):

inc/saiport.h
```diff
@@ -3493,6 +3493,20 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_LINK_DOWN_DEBOUNCE_TIMEOUT,
 
+    /**
+     * @brief Enable TRIM TC AND COLOR -> DSCP MAP
+     *
+     * Map id = #SAI_NULL_OBJECT_ID to disable trim map on port.
```
rck-innovium, line 3499: Please clarify that 1) SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP this map applies to trimmed packets when this attribute is non-NULL, 2) SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP does NOT apply to non-trimmed packets, 3) When SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP is NULL, the existing SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP applies to trimmed packets as well

eddyk-nvidia:
@JaiOCP  - please put the explanation from this PR into https://github.com/JaiOCP/SAI/blob/master/doc/SAI-Proposal-Packet-Trimming.md document

inc/saiport.h
```diff
@@ -3493,6 +3493,20 @@ typedef enum _sai_port_attr_t
      */
     SAI_PORT_ATTR_LINK_DOWN_DEBOUNCE_TIMEOUT,
 
+    /**
+     * @brief Enable TRIM TC AND COLOR -> DSCP MAP
```
rck-innovium, line 3497: Does this attribute apply for Trim-to-sender packets as well

inc/saibuffer.h
```diff
@@ -645,6 +645,19 @@ typedef enum _sai_buffer_profile_packet_admission_fail_action_t
...
+     *
+     * Try sending a shortened packet over a different
+     * queue. Original packet will be dropped and trimmed copy of the packet will be send to the sender.
+     * The IP length and checksum fields will be updated in a trimmed copy.
+     * SAI_QUEUE_STAT_DROPPED_PACKETS as well as SAI_QUEUE_STAT_DROPPED_BYTES
+     * will count the original discarded frames even if they will be trimmed afterwards.
+     * Interface statistics must show dropped packets.
+     * Interface statistics may show sent trimmed packets.
+     */
+    SAI_BUFFER_PROFILE_PACKET_ADMISSION_FAIL_ACTION_DROP_AND_TRIM_TO_SENDER,
```
rbudhia-upscaleai, line 660: A separate trim_tc_value field needs to be added to saiswitch.h for back-to-sender (BTS) packets, enabling the sender to distinguish BTS-trimmed packets from normal trimmed packets for which it is the receiver

rck-innovium:
@JaiOCP 
On which port do the below stats increment for Trim-to-Sender packets? the port towards the receiver or the port towards the sender?

    /** Packets trimmed but dropped due to failed shared buffer admission on a trim queue */
    SAI_PORT_STAT_DROPPED_TRIM_PACKETS,

    /** Packets trimmed and successfully transmitted on port */
    SAI_PORT_STAT_TX_TRIM_PACKETS,

inc/saibuffer.h
rck-innovium: Agree. An endpoint must be able to tell whether the trimmed packet was a back-to-sender or to-receiver case as it has to use that to identify congestion control context (in the case of BTS, we swap the DIP and SIP etc) as well as to know whether it needs to generate (in the case of to-receiver) a NACK  or (in case of BTS) selective retransmit of the original packet.

inc/saiport.h
JaiOCP: Added details in the TRIM spec

JaiOCP:
> @JaiOCP - please put the explanation from this PR into https://github.com/JaiOCP/SAI/blob/master/doc/SAI-Proposal-Packet-Trimming.md document

Updated the spec

inc/saibuffer.h
JaiOCP: Agreed. Introduced a new TRIM to sender specific TC and DSCP value as a switch attribute

JaiOCP:
> @JaiOCP On which port do the below stats increment for Trim-to-Sender packets? the port towards the receiver or the port towards the sender?
> 
> ```
> /** Packets trimmed but dropped due to failed shared buffer admission on a trim queue */
> SAI_PORT_STAT_DROPPED_TRIM_PACKETS,
> 
> /** Packets trimmed and successfully transmitted on port */
> SAI_PORT_STAT_TX_TRIM_PACKETS,
> ```

I went over the details of this stats again and here is my interpretation. 
SAI_PORT_STAT_TRIM_PACKETS: These are TRIM packets because of the congestion on trim eligible queue and has nothing to do with direction these packets are send. Many queues may experience congestion and may be trim eligible. All these trim packets across queues are accounted under this stat.

SAI_PORT_STAT_DROPPED_TRIM_PACKETS: The are dropped TRIM packets because the queue on which they were supposed to go out got congested. Again I would say that there is nothing specific about direction here.

SAI_PORT_STAT_TX_TRIM_PACKETS: This one the TRIM packets that were forwarded on the port and is an important metrics to know how of a port bandwidth is consumed by this traffic. If we look carefully if a TRIM is sent to receiver or sender we want to know the BW consumed on the port. I would suggest that these stats are also inclusive of TRIM in both the directions.

Let me know what you and others think about it.

inc/saiport.h
JaiOCP: Added details in the TRIM spec. Please take a look.

JaiOCP:
@rbudhia-upscaleai @rck-innovium Please take a look. Review comments are addressed

doc/SAI-Proposal-Packet-Trimming.md
```diff
@@ -224,3 +224,47 @@ typedef enum _sai_packet_trim_dscp_resolution_mode_t
...
+To address this unintended behavior, we are introducing a new port-based attribute: SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP. This attribute handles TC to DSCP mapping exclusively for TRIM packets.
+
+### Capability Query
+NOS can do a cpability query for SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP attribure to check an implementation's support.
+
+If both SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP and SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP are configured then TRIM packets will use SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP and data packets will use SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP.
+
+If only SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP is confiured then NOS can configure both data packet TC and TRIM TC as long as the desired behavior of DSCP value of data packets is maintained.
+
+If only SAI_PORT_ATTR_QOS_TRIM_TC_AND_COLOR_TO_DSCP_MAP is configured then data packets do not undergo any DSCP remarking and only TRIM packets will use the configured map for DSCP marking.
```
kperumalbfn, line 242: @JaiOCP If map attached to port doesn't have any mapping for non-trim TCs, then DSCP remarking for regular packets is bypassed. For default TC=0, if there is no entry in the map attached to egress port, then packet remarking should be bypassed. Could you please clarify this?

PR #2350, opened 2026-09-09
add Passive Optical Network (PON) to experimental

dpelton-ciena:
This PR adds API definitions to control Passive Optical Network (PON) hardware components.  These are added to the experimental SAI headers.

The changes cover the objects necessary to implement a data center based PON deployment as described in these SONiC PRs:
[PON HLD](https://github.com/sonic-net/SONiC/pull/2377)
[Dual Homed PON gateway HLD](https://github.com/sonic-net/SONiC/pull/2376)

As the SONiC features are planned for the SONiC 202611 release, we are proposing these changes for inclusion in the 1.19 SAI release.


experimental/saiexperimentalpon.h
```diff
@@ -0,0 +1,712 @@
...
+ *
+ * @{
+ */
+
+/**
+ * @brief PON action types for one-shot RPC operations
+ */
+typedef enum _sai_pon_action_type_t
+{
+    SAI_PON_ACTION_TYPE_OLT_RESET                   = 0,
```
kcudnik, line 48: no need for expicit numeration if numbers are continous

experimental/saiexperimentalpon.h
```diff
@@ -0,0 +1,712 @@
...
+    sai_bulk_get_pon_onu_statistics_accumulating_olt_pon_attribute_fn                                                                 get_pon_onu_statistics_accumulating_olt_pons_attribute;
+    sai_bulk_get_pon_onu_statistics_accumulating_olt_pon_omcc_attribute_fn                                                            get_pon_onu_statistics_accumulating_olt_pon_omccs_attribute;
+    sai_bulk_get_pon_onu_statistics_accumulating_olt_pon_service_attribute_fn                                                         get_pon_onu_statistics_accumulating_olt_pon_services_attribute;
+    sai_bulk_get_pon_onu_statistics_accumulating_onu_enhanced_tc_pm_attribute_fn                                                      get_pon_onu_statistics_accumulating_onu_enhanced_tc_pms_attribute;
+    sai_bulk_get_pon_onu_statistics_accumulating_onu_ethernet_frame_extended_pm_attribute_fn                                          get_pon_onu_statistics_accumulating_onu_ethernet_frame_extended_pms_attribute;
```
kcudnik, line 401: why those naming must be so long ?

experimental/saiexperimentalpon.h
```diff
@@ -0,0 +1,712 @@
...
+    sai_create_pon_service_config_profile_extended_vlan_tagging_operation_config_data_received_frame_vlan_tagging_operation_fn        create_pon_service_config_profile_extended_vlan_tagging_operation_config_data_received_frame_vlan_tagging_operation;
+    sai_remove_pon_service_config_profile_extended_vlan_tagging_operation_config_data_received_frame_vlan_tagging_operation_fn        remove_pon_service_config_profile_extended_vlan_tagging_operation_config_data_received_frame_vlan_tagging_operation;
+    sai_set_pon_service_config_profile_extended_vlan_tagging_operation_config_data_received_frame_vlan_tagging_operation_attribute_fn set_pon_service_config_profile_extended_vlan_tagging_operation_config_data_received_frame_vlan_tagging_operation_attribute;
+    sai_get_pon_service_config_profile_extended_vlan_tagging_operation_config_data_received_frame_vlan_tagging_operation_attribute_fn get_pon_service_config_profile_extended_vlan_tagging_operation_config_data_received_frame_vlan_tagging_operation_attribute;
```
kcudnik, line 516: this naming is insane !

experimental/saiexperimentalpon.h
kcudnik: and add empty line before each enum

experimental/saiexperimentalpononu.h
```diff
@@ -0,0 +1,13556 @@
...
+     */
+    SAI_PON_ONU_STATISTICS_ACCUMULATING_ONU_ETHERNET_FRAME_EXTENDED_PM_64BIT_ATTR_CRC_ERRORED_FRAMES,
+
+    /**
+     * @brief Drop Events
+     *
+     * @type sai_uint64_t
+     * @flags READ_ONLY
+     */
+    SAI_PON_ONU_STATISTICS_ACCUMULATING_ONU_ETHERNET_FRAME_EXTENDED_PM_64BIT_ATTR_
```
kcudnik, line 13465: reconsider all those long names to short them

meta/parse.pl
```diff
@@ -1458,6 +1466,12 @@ sub ProcessType
 {
     my ($attr, $type) = @_;
 
+    if ((!defined $type or $type eq "") and $attr eq "SAI_ACL_COUNTER_ATTR_TABLE_ID")
+    {
+        LogWarning "missing type for $attr, defaulting to sai_object_id_t";
+        return "SAI_ATTR_VALUE_TYPE_OBJECT_ID";
```
kcudnik, line 1472: why this condition is added?

meta/parse.pl
```diff
@@ -1569,6 +1583,12 @@ sub ProcessFlags
 
     if (not defined $flags)
     {
+        if ($value eq "SAI_ACL_COUNTER_ATTR_TABLE_ID")
```
kcudnik, line 1586: why this condition is added?

meta/parse.pl
```diff
@@ -1804,7 +1824,7 @@ sub ProcessStoreDefaultValue
         return "true";
     }
 
-    my @flags = @{ $flags };
+    my @flags = (ref($flags) eq "ARRAY") ? @{ $flags } : ();
```
kcudnik, line 1827: why this condition is added?

meta/parse.pl
```diff
@@ -2782,6 +2802,8 @@ sub ProcessStructObjects
 
     my $objects = $struct->{objects};
 
+    return "NULL" if not defined $objects;
```
kcudnik, line 2805: why this condition is added?

meta/parse.pl
```diff
@@ -2802,6 +2824,8 @@ sub ProcessStructObjectLen
 
     return 0 if not $type eq "sai_object_id_t" and not $type eq "sai_object_list_t" and not $type eq "sai_attribute_t*";
 
+    return 0 if not defined $struct->{objects};
```
kcudnik, line 2827: why this condition is added?

meta/parse.pl
```diff
@@ -3097,24 +3121,26 @@ sub ProcessCreate
...
         else
         {
-            WriteSource "return sai_metadata_sai_${api}_api->create_$small(&meta_key->objectkey.key.object_id, switch_id, attr_count, attr_list);";
+            WriteSource "return sai_metadata_sai_${api}_api->${method}(&meta_key->objectkey.key.object_id, switch_id, attr_count, attr_list);";
         }
     }
     else
     {
-        WriteSource "return sai_metadata_sai_${api}_api->create_$small(&meta_key->objectkey.key.$small, attr_count, attr_list);";
+        WriteSource "return sai_metadata_sai_${api}_api->${method}(&meta_key->objectkey.key.$small, attr_count, attr_list);";
```
kcudnik, line 3143: why this condition is added?

meta/parse.pl
```diff
@@ -3200,25 +3230,27 @@ sub ProcessGet
     WriteSource "_Inout_ sai_attribute_t *attr_list)";
     WriteSource "{";
 
-    if (IsSpecialObject($ot))
+    my $method = "get_${small}_attribute";
+
+    if (IsSpecialObject($ot) or not defined $API_METHODS{$api} or not defined $API_METHODS{$api}{$method})
     {
-        WriteSource "return SAI_STATUS_NOT_IMPLEMENTED;";
```
kcudnik, line 3205: why this is changed ?

meta/parse.pl
```diff
@@ -4357,6 +4402,10 @@ sub CheckApiStructNames
     {
         next if not $name =~ /^sai_(\w+)_api_t$/;
 
+        # Split experimental PON uses a single SAI_API_PON runtime API while
+        # exposing component-level API structs in separate headers.
+        next if $name =~ /^sai_pon_(controller|olt_intf|olt_plug|onu|onu_template|service_config_profile)_api_t$/;
```
kcudnik, line 4407: why this condition is added? each header should have its own api

meta/parse.pl
```diff
@@ -4687,7 +4747,33 @@ sub ExtractApiToObjectMap
...
+        if (defined $API_METHODS{$api})
+        {
+            for my $name (keys %methods)
+            {
+                $API_METHODS{$api}{$name} = 1;
+            }
+        }
+        else
+        {
+            $API_METHODS{$api} = \%methods;
```
kcudnik, line 4775: each header should have its own apid and this logic would not be needed, you are breaking header design here !

kcudnik: please follow entire SAI design

meta/style.pm
```diff
@@ -208,9 +208,9 @@ sub CheckStatsFunction
     return if $fname eq "sai_bulk_object_get_stats_fn"; # exception
     return if $fname eq "sai_bulk_object_clear_stats_fn"; # exception
 
-    if (not $fname =~ /^sai_((get|clear)_(\w+)_stats|get_\w+_stats_ext)_fn$/)
+    if (not $fname =~ /^sai_((get|clear)_(\w+)_stat(s)?|get_\w+_stat(s)?_ext)_fn$/)
```
kcudnik, line 211: why this condition is added?

meta/style.pm
```diff
@@ -550,9 +552,10 @@ sub CheckQuadApi
     $order =~ s/012/s/g;        # order should be: get_stats,get_stats_ext,clear_stats
     $order =~ s/CR/E/g;         # order should be: bulk_create,bulk_remove
     $order =~ s/SG/T/g;         # order should be: bulk_set,bulk_get
+    $order =~ s/g+/g/g;         # order can include runs of read-only get attribute APIs
```
kcudnik, line 555: why such api exists only ?

meta/style.pm
```diff
@@ -626,7 +629,7 @@ sub CheckStructAlignment
         {
             my $itemname = $2;
 
-            if ($1 ne $spaces or (length($2) != length($inside) and $struct =~ /_api_t/))
+            if ($1 ne $spaces or (length($2) != length($inside) and $struct =~ /_api_t/ and $struct !~ /sai_pon_api_t/))
```
kcudnik, line 632: you are breaking design here

meta/style.pm
```diff
@@ -1268,6 +1271,7 @@ sub CheckHeadersStyle
             next if $line =~ m![^\\]\\$!;           # macro multiline
             next if $line =~ /^ {4}(\w+);$/;        # union entries
             next if $line =~ /^union _sai_\w+ \{/;  # union entries
+            next if $line =~ /^ {20,}\w+/;          # api struct member lines
```
kcudnik, line 1274: why this condition is added?

kcudnik (changes requested):

meta/parse.pl
dpelton-ciena: Thanks for the feedback.  A number of these changes were made such that the header file could be broken up into smaller pieces.  We will revert these changes.

experimental/saiexperimentalpononu.h
dpelton-ciena: The long names are due to the SAI being derived from the proposed SONiC PON YANG models.  As described in the HLD, the current design pushes config data from YANG through SONiC to the SAI layer.  We had a number of internal discussions about the effect this has on the SAI headers prior to raising this PR, and while some of the field names could be shortened using abbreviations, this would make understanding the mapping more difficult.

kcudnik: ok

meta/parse.pl
kcudnik: you can still break this to smaller modules, just eachmodule should have it's own API
and not all put in single SAI_API_PON

dpelton-ciena:
@kcudnik - I've reworked the changes to the meta directory so that they should represent the minimal set required to introduce this change.  I have been running the checks manually, but I would appreciate if you could kick off the automated Azure job to ensure that I haven't made mistakes.

This change increases the number of extension objects, and explicitly disables the check for EXTENSION_OBJECT_TYPE_COUNT in saisanitycheck.c.  Please let me know how this should be addressed, as the counts are now:
`INFO: SAI_OBJECT_TYPE_MAX = 121, EXTENSION_OBJECT_TYPE_COUNT = 213, CUSTOM_OBJECT_TYPE_COUNT = 0`

meta/style.pm
dpelton-ciena: This change has been removed

meta/style.pm
dpelton-ciena: This change has been removed

meta/style.pm
dpelton-ciena: This change has been removed

meta/style.pm
dpelton-ciena: This change has been removed

meta/parse.pl
dpelton-ciena: This logic has been removed

experimental/saiexperimentalpon.h
dpelton-ciena: This has been reworked as per suggestion

experimental/saiexperimentalpon.h
dpelton-ciena: As with the other comments on long names, these are due to the SAI being derived from the proposed SONiC PON YANG models.

experimental/saiexperimentalpon.h
dpelton-ciena: As with the other comments on long names, these are due to the SAI being derived from the proposed SONiC PON YANG models.

meta/parse.pl
dpelton-ciena: The multiple headers have been merged, and this logic has been removed.

meta/parse.pl
dpelton-ciena: The revised code has reworked this change to clarify the intent.  In cases where there is no corresponding function, defined, the check code can not be used.

meta/parse.pl
dpelton-ciena: This change has been removed and the changes have been reworked to minimize the diff footprint.

meta/parse.pl
dpelton-ciena: This change has been removed.

meta/parse.pl
dpelton-ciena: This change has been removed.

meta/parse.pl
dpelton-ciena: This change has been removed.

meta/parse.pl
dpelton-ciena: This change has been removed.

meta/parse.pl
dpelton-ciena: This change has been removed - I'm not sure how these ACL checks made it into our initial change set.

experimental/saitypesextensions.h
```diff
@@ -113,6 +113,187 @@ typedef enum _sai_object_type_extensions_t
...
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_ENHANCED_TC_PM,
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_ETHERNET_FRAME_PM_DOWNSTREAM,
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_ETHERNET_FRAME_PM_UPSTREAM,
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_ETHERNET_PM,
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_FEC_PM,
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_GAL_ETHERNET_PM,
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_GEM_PORT_NETWORK_CTP_PM,
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_RS232_RS485_PERF_MON_HIST_DATA,
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_XG_PON_DOWNSTREAM_MGMT_PM,
+    SAI_OBJECT_TYPE_PON_ONU_STATISTICS_STREAMING_ONU_XG_PON_UPSTREAM_MGMT_PM,
```
kcudnik, line 292: i think it's less than 256, since currentliu yin sonic extension object types are encoded in 1 byte, so it should be fine at this moment but in future sairedis ODI/VID/RID encoding would need to be updated to accomodate this

meta/style.pm
```diff
@@ -507,6 +509,9 @@ sub CheckQuadApi
 
     my $apis = $2;
 
+    # this giant experimental API is in sections that do not follow the order checks below
```
kcudnik: why not follow? maybe make them follow order check?

experimental/saiextensions.h
```diff
@@ -109,6 +112,8 @@ typedef enum _sai_api_extensions_t
 
     SAI_API_OCS,
 
+    SAI_API_PON,
```
kcudnik, line 115: since now everything is in one big file, would it be bette to split this to some more convinient modules and smaller files like it was before but adding SAI_API_PON_XXX modules ?

meta/saisanitycheck.c
```diff
@@ -445,7 +445,7 @@ void check_attr_by_object_type()
      * Extensions object types for now should be minimum, since it could be
      * encoded on 1 byte in OID, but this could be later on relaxed.
      */
-    META_ASSERT_TRUE(EXTENSION_OBJECT_TYPE_COUNT < 64, "too many experimental object types");
+    META_ASSERT_TRUE(EXTENSION_OBJECT_TYPE_COUNT < 256, "too many experimental object types");
```
kcudnik, line 449: add also comment here that if this number will be >255 then sonic/sairedis VID/RID object encoding must be addressed

meta/saisanitycheck.c
```diff
@@ -6291,9 +6272,11 @@ void check_object_type_extension_max_value()
     /*
      * This check may be removed, but it will need to be brought into attention on SAI meeting.
      */
-    META_ASSERT_TRUE(SAI_OBJECT_TYPE_MAX < 192 && EXTENSION_OBJECT_TYPE_COUNT < 64, "exceeding this range will not allow to encode object types to single byte");
+    /* XXX need to determine what checks should be here now that the byte width limits are exceeded.
```
kcudnik: don't disable those checks, on some of those checks sonic/sairedis relay!

meta/saiserialize.c
```diff
@@ -36,7 +36,7 @@
 #include "saimetadata.h"
 #include "saiserialize.h"
 
-#define PRIMITIVE_BUFFER_SIZE 128
+#define PRIMITIVE_BUFFER_SIZE 171  /* Must match the value in saiserializetest.c */
```
kcudnik: make this 256

meta/saiserialize.c
```diff
@@ -679,6 +679,13 @@ int sai_serialize_enum(
     {
         if (meta->values[i] == value)
         {
+            size_t enum_len = strlen(meta->valuesnames[i]);
+            if (enum_len >= PRIMITIVE_BUFFER_SIZE)
+            {
+                SAI_META_LOG_WARN("buffer too small to serialize enum value %s", meta->valuesnames[i]);
+                sprintf(buffer, "PRIMITIVE_BUFFER_SIZE too small for enum (%zd >= %u)", enum_len, PRIMITIVE_BUFFER_SIZE);
+                return SAI_SERIALIZE_ERROR;
+            }
```
kcudnik, line 688: why this is introduced in the first place ?

meta/saiserializetest.c
```diff
@@ -52,7 +52,7 @@
                 __func__, __LINE__, r, strlen(a));                          \
         exit(1);}
 
-#define PRIMITIVE_BUFFER_SIZE 128
+#define PRIMITIVE_BUFFER_SIZE 171  /* Must match the value in saiserialize.c */
```
kcudnik: make 256

meta/saiserialize.c
kcudnik: there should be serialize test generated automaticaly for each enum in  the metadata and then size checke whether buffer is enough, so this value will never be exceeded

kcudnik (changes requested):

meta/saiserialize.c
dpelton-ciena: I was seeing core dumps prior to adding this change (when the names were longer than PRIMITIVE_BUFFER_SIZE).  Adding snprintf instead of sprintf violated the -ansi compiler flag, so I added this logic with strlen.

meta/saisanitycheck.c
dpelton-ciena: So this check should be for:
`SAI_OBJECT_TYPE_MAX < 192 && EXTENSION_OBJECT_TYPE_COUNT < 256`
?

meta/saiserialize.c
dpelton-ciena: Done.

meta/saiserializetest.c
dpelton-ciena: Done

meta/saisanitycheck.c
dpelton-ciena: Done

meta/saisanitycheck.c
kcudnik: yes can be

meta/saiserialize.c
kcudnik: i remember adding check for taht in saisanitycheck to check maximum enum length, and that serialization should always succeed, so buffer should be long enough to serialize all possible enums in metadata

kcudnik: so i would suggest to add check for buffer in saisanity to make sure that buffer is long enough, and remove this one

meta/style.pm
dpelton-ciena: The latest commit removes this exception but needs to add a new ordering pattern that allows:
```
sai_get_*
sai_get_bulk_*
```
These are objects that do not have corresponding create/remove/set.  One example is:
```
    sai_get_pon_olt_intf_onu_operational_state_attribute_fn
    sai_bulk_get_pon_olt_intf_onu_operational_state_attribute_fn    
```
The first form gets data for a specific ONU on an OLT interface, and the second form gets data for all ONUs on an OLT interface.  The state cannot be "set" and the ONUs are not created or removed by the SAI.

experimental/saiextensions.h
dpelton-ciena: I'm not sure what impact this will have on the folks we have working on the corresponding syncd code.  The HLD proposes a single syncd instance for all PON operations.

meta/saisanitycheck.c
dpelton-ciena: Latest submission has the updated checks as noted above.

meta/saiserialize.c
dpelton-ciena: Thanks - I will investigate.

dpelton-ciena: The code in meta/saisanitycheck.c doesn't appear to check field name length, and that code doesn't run until after saiserializetest.  Where should I be looking for the buffer length check that you're referring to?

kcudnik: hmm, i remembe radding something like that but it was long time ago, actually easier way would be to take the longest line in each headers and use it as max buffer length check this script:
```bash
SAI $ cat inc/*h experimental/*h|perl -ne 'print length$_," -> $_\n" if /^\s+SAI_/ and not /=/'|sort -n|tail -n 1
```

dpelton-ciena: I've incorporated the core of this logic into a perl script that checks the headers, and verifies that PRIMITIVE_BUFFER_SIZE matches in the two source files that define it.  The define could be moved into a C header file instead of having this check.  I didn't do that as I wasn't sure if there was an explicit reason for keeping the define directly in each of the C files.

meta/check_primitive_buffer_size.pl
```diff
@@ -0,0 +1,139 @@
...
+# This script checks that saiserialize.c and saiserializetest.c define the same
+# PRIMITIVE_BUFFER_SIZE value. It also scans the SAI header enum-style values in
+# the repository root and verifies that the longest name fits in a buffer size
+# defined by PRIMITIVE_BUFFER_SIZE.
+
+use strict;
+use warnings;
+use FindBin qw($Bin);
+
+use constant MACRO_NAME => 'PRIMITIVE_BUFFER_SIZE';
```
kcudnik: i don't think you need this script at all, since parse.pl reads all headers and all enums and you can extract maximum length just from that iterating all enums, and i would go that way, and then you can generate max buffer size constant in saimetadata.h programatically for future use

meta/saiserialize.c
kcudnik: check my comment: https://github.com/opencomputeproject/SAI/pull/2350#pullrequestreview-5269040943

meta/check_primitive_buffer_size.pl
dpelton-ciena: Thanks for the pointer to parse.pl - the latest commit incorporates your suggested change and removes this script.

meta/parse.pl
```diff
@@ -5306,9 +5334,18 @@ sub WriteHeaderHeader
     WriteSectionComment "AUTOGENERATED FILE! DO NOT EDIT";
 
     my @ch = GetCustomHeaderFiles();
+    my $longest_enum;
```
kcudnik: you can set this to 1 or 128, and then you don't need to check "if not defined" every time in line 5341, and just assign lenght($enum) directly, or if you want to use this as string then initialize it with ""

kcudnik: otherwise this looks good

dpelton-ciena: Thanks - I've updated the logic to track the longest string _length_ found rather than keeping track of the longest string found.

kcudnik (approved):
I would still split PON to multiple APIs not keep it in one large file

meta/parse.pl
```diff
@@ -5306,9 +5334,18 @@ sub WriteHeaderHeader
...
+    for my $enum (@ALL_ENUMS)
+    {
+        $longest_enum_len = length($enum) if length($enum) > $longest_enum_len;
+    }
+
+    my $primitive_buffer_size = $longest_enum_len + 1;
 
     WriteHeader "#ifndef __SAI_METADATA_H__";
     WriteHeader "#define __SAI_METADATA_H__";
+    WriteHeader "#define PRIMITIVE_BUFFER_SIZE $primitive_buffer_size";
```
kcudnik: one more thing, this should be defined later in the file after ALL includes, and should be prefixed SAI_METADATA_XXX and have extra descritpion, check how SAI_METADATA_HAVE_ATTR_VERSION is created, and make this definition of primitive buffer size as separate function, and call it after 
ProcessAttrVersion at the end of the file. you can call it CreateEnumPrimitiveBufferSize

kcudnik (changes requested):

meta/parse.pl
kcudnik: all definition in saimetadata.h shouold be always prefixed by SAI_METADATA_xxx

dpelton-ciena: Updated in the latest commit.

kcudnik (approved):

meta/parse.pl
kcudnik: thanks, approved

PR #2353, opened 2026-09-10
Add port MRU attribute

AnandhiDhanabalan:
**Summary**
Adds two port attributes that allow the maximum receive frame size (MRU) to be configured independently of `SAI_PORT_ATTR_MTU`. Today `SAI_PORT_ATTR_MTU` governs the egress frame-size limit, and on many implementations the same value is applied as the ingress admission threshold. This change lets an implementation that supports separate RX/TX enforcement expose an independent receive limit, while leaving single-threshold implementations unaffected.

**Backward compatibility**
`SAI_PORT_ATTR_MRU` takes effect only when `SAI_PORT_ATTR_MRU_ENABLED` is `true`. The default is `false`, under which receive-side behavior is unchanged: implementations that apply `SAI_PORT_ATTR_MTU` to both directions continue to do so. Existing applications require no change.

**New Attributes**
- `SAI_PORT_ATTR_MRU_ENABLED` 
- `SAI_PORT_ATTR_MRU`

### DIFF
--- a/inc/saiport.h
+++ b/inc/saiport.h
@@ -3494,6 +3494,32 @@ typedef enum _sai_port_attr_t
     SAI_PORT_ATTR_LINK_DOWN_DEBOUNCE_TIMEOUT,
 
     /**
+     * @brief Enable independent Maximum Receive Unit (MRU) enforcement
+     *
+     * When true, SAI_PORT_ATTR_MRU governs the ingress frame-size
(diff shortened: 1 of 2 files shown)


kcudnik (approved):

tjchadaga (approved):

tjchadaga:
@JaiOCP, @prgeor, @j-bos - could you please help sign off on this?

PR #2354, opened 2026-09-14
[debug_counter] Add support for SAI_IN_DROP_REASON_LPM_ANY

tsmail1441:
### Summary
Introduce `SAI_IN_DROP_REASON_LPM_ANY` to support ingress drop counting for both IPv4 and IPv6 L3 route lookup drops across various forwarding implementations and routing modes.

### Motivation
Depending on hardware architecture and routing table configurations, unrouted or dropped L3 packets may not always trigger a native hardware lookup miss; they may instead be dropped by explicit route discard entries. Consequently, relying solely on `SAI_IN_DROP_REASON_LPM_MISS` can result in undercounting unrouted packets, while tracking broader L3 drop categories can conflate route resolution drops with unrelated forwarding drops.
`SAI_IN_DROP_REASON_LPM_ANY` resolves this by providing a unified drop reason that captures both IPv4 and IPv6 packets dropped due to LPM table misses or explicit route discard actions, providing consistent drop accounting across both address families.

### Changes
- Add `SAI_IN_DROP_REASON_LPM_ANY` enum to `inc/saidebugcounter.h`.

Signed-off-by: Tommy Smail <tommysmail@google.com>


inc/saidebugcounter.h
```diff
@@ -364,6 +364,13 @@ typedef enum _sai_in_drop_reason_t
      */
     SAI_IN_DROP_REASON_EGRESS_LINK_DOWN,
 
+    /**
+     * @brief Packet is dropped during IPv4 or IPv6 routing table (LPM) lookup.
+     * Generic counter across all routing lookup modes when a packet cannot be forwarded.
+     * e.g.: LPM lookup miss, explicit blackhole / discard route
+     */
+    SAI_IN_DROP_REASON_LPM_ANY,
```
j-bos: Can this be achieved by user creating a debug counter that includes the following reasons?

SAI_IN_DROP_REASON_BLACKHOLE_ROUTE SAI_IN_DROP_REASON_LPM_MISS

skannan-CLS: Hi,
Generally we use separate Table or ACL for Policy Based Routing along with LPM will this counter cover for that too ? Since ANY should ideally count those drop packets ?

Thanks,
Kannan.S

mobinmohan (approved):

kcudnik:
you will need rebase to master and force push to your branch again to resolve build

